PR #49, opened 2022-01-12
TB-3197 Create market_language_config.dart

pro100svitlo:
Added FeedMarket to the `configuration.dart`

- create model
- add it to the Configuration constructor
- add it to the ChangeConfiguration event
- update old tests
- add a few new tests to cover new code


discovery_engine/lib/src/domain/models/configuration.dart
```diff
@@ -13,19 +13,22 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 import 'package:freezed_annotation/freezed_annotation.dart';
+import 'package:xayn_discovery_engine/discovery_engine.dart';
 
 part 'configuration.freezed.dart';
+
```
acrrd, line 19: Why this line?

acrrd (approved):

discovery_engine/lib/src/domain/models/market/feed_market.dart
```diff
@@ -0,0 +1,28 @@
+import 'package:equatable/equatable.dart';
```
biernacki, line 1: could we move this file into parent folder => `discovery_engine/lib/src/domain/models/feed_market.dart`? all the other files in `domain/models` are at the same level, would be good to have consistency

biernacki (approved):
LGTM! just one minor optional request

discovery_engine/lib/src/domain/models/market/feed_market.dart
biernacki: also please add licence notice
```
// Copyright 2022 Xayn AG
//
// This program is free software: you can redistribute it and/or modify
// it under the terms of the GNU Affero General Public License as
// published by the Free Software Foundation, version 3.
//
// This program is distributed in the hope that it will be useful,
// but WITHOUT ANY WARRANTY; without even the implied warranty of
// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
// GNU Affero General Public License for more details.
//
```

acrrd: Thanks!

discovery_engine/lib/src/domain/models/configuration.dart
pro100svitlo: I guess cos of the auto-formatter.

discovery_engine/lib/src/domain/models/market/feed_market.dart
pro100svitlo: Sure, if youd like - I will.
But in my personal opinion, `models` become a messy place, and it's better to split those files into separate folders (so all `feed_market` related in one folder and all `configuration` related - in another.)

But sure, you are right, this is not a part of this pr, so I will put it back

discovery_engine/lib/src/domain/models/configuration.dart
acrrd: the formatter seems fine with both

PR #50, opened 2022-01-12
TY 2250 integrate async library

rustonaut:
**As this pulls in a async runtime 1159 of the additions and 27 deletions are from the `Cargo.lock` file.**

This temporary introduces a local bash script, which will be replaced with a job-runner in a follow up PR.

The CI is only as far updated  as necessary, updates to the release process are not yet done.


discovery_engine/lib/discovery_engine.dart
```diff
@@ -19,4 +19,6 @@ library discovery_engine;
 
 export 'package:xayn_discovery_engine/src/api/api.dart';
 export 'package:xayn_discovery_engine/src/discovery_engine_base.dart';
+//FIXME: remove once unneded
```
acrrd, line 22:
```suggestion
//FIXME: remove once unneeded
```
When it will be unneeded?

discovery_engine/lib/src/ffi/load_lib.dart
```diff
@@ -0,0 +1,39 @@
+// Copyright 2021 Xayn AG
```
acrrd, line 1:
```suggestion
// Copyright 2022 Xayn AG
```

discovery_engine_core/bindings/src/async_bindings.rs
```diff
@@ -0,0 +1,9 @@
+// Copyright 2022 Xayn AG
+
```
acrrd, line 2: Rest of the license is missing.

discovery_engine_flutter/.gitignore
```diff
@@ -6,6 +6,7 @@
 .packages
 build/
 
+ios/Classes/XaynDiscoveryEngineBindings.h
```
acrrd, line 9: Do we really need this? We don't need to call these functions from swift anymore.

tmp_script.sh
```diff
@@ -0,0 +1,161 @@
+#!/bin/bash
```
acrrd, line 1: It should have a better name even if is going to be removed in the next pr

tmp_script.sh
```diff
@@ -0,0 +1,161 @@
...
+    cargo install \
+        --git https://github.com/xaynetwork/xayn_async_bindgen.git \
+        async-bindgen-gen-dart \
+        "$@"
+}
+
+function gen_dart_genesis_extension() {
+    cd discovery_engine_core
+    # Doesn't respect cargo config option, but good enough for us for now
+    "${CARGO_INSTALL_ROOT:-${CARGO_HOME:-$HOME/.cargo}}/bin/async-bindgen-gen-dart" \
```
acrrd, line 21: @janpetschexain @Robert-Steiner @biernacki can you test that this works on mac?

tmp_script.sh
```diff
@@ -0,0 +1,161 @@
...
+        --ffi-class XaynDiscoveryEngineBindingsFfi \
+        --genesis ../discovery_engine/lib/src/ffi/genesis.ffigen.dart
+    cd ..
+}
+
+function get_dart_deps() {
+    if [ "${_DART_PUB_GET_DONE:-0}" = 0 ]; then
```
acrrd, line 28: I guess this will not be needed when we port it to just because the task will be already run only once.

tmp_script.sh
```diff
@@ -0,0 +1,161 @@
...
+        _CODEGEN_WORKAROUND_DONE=1
+        cd discovery_engine_core
+        cargo check --quiet 2>/dev/null || :
+        cd ..
+    fi
+}
+
+function check_fmt_rust() {
+    cd discovery_engine_core
+    cargo +nightly fmt --all -- --check
```
acrrd, line 67: Would be more useful we can just run them to have the code as we want.

tmp_script.sh
```diff
@@ -0,0 +1,161 @@
...
+function test_rust_full() {
+    check_fmt_rust
+    check_rust
+    test_rust
+}
+
+function build_bindgen_glue() {
+    codegen_workaround
+    # make sure cbindgen did run properly, but we don't really
+    # care if there is a problem with rust at this point in time
```
acrrd, line 126: After we execute `codegen_workaround` shouldn't we care about the errors?

discovery_engine/lib/discovery_engine.dart
rustonaut: When we have any functionality which actually uses async bindgen;

discovery_engine_flutter/.gitignore
rustonaut: I'm not sure, but didn't we need to provide the headers to iOS so that it gets included
into the binary and not stripped out?

tmp_script.sh
rustonaut: Maybe we should use a custom install root in the workspace (git-ignored)?

Then we also don't have a problem with any situation where the install root is 
changed in a cargo config file (`.cargo/config.toml`).

tmp_script.sh
rustonaut: yes, this is one of the reasons why I still went with `just` instead of `tusk` even through `just` is a big ... strange.

tmp_script.sh
rustonaut: I'm not sure what you want to say with that.

tmp_script.sh
rustonaut: Not necessary, some random errors in rust should prevent generating cbindings for dart.

But I did change it in the justfile.

The optimal solution would not to run `cbindgen` implicitly on build. This would decouple `cbindgen` from `cargo check` which fits better with CI and local job runners and would like speed things slightly up.

At the cost that installing cbindgen now needs to be handle alongside to installing cargo-sort and the async_bindgen CLI utility.

rustonaut:
Generally tmp_script.sh and all changes to the CI are only there to make it mergeable/locally testable.

If I could I would not have commited either of them and only commited the local scripts in #61 and #54.

As the tmp_script will be fully replaced and any change to CI will be also replaced/touched again in the follow up PR I would prefer to to spend time on discussion their finer design aspects in this PR, but the follow up PR's instead.

tmp_script.sh
rustonaut: If you mean we should run cargo fmt, I would disagree. It's a check command it should not modify your code, we use a nightly i.e. unstable version of rustfmt, it might brake your code with formatting.

But that's a discussion for the follow up PR not this PR.

rustonaut:
I removed tmp_script.sh (after initially renaming it to job_runner.sh).

I inlined all commands used in CI (some deduplication, .ci/ composite actions will be added with other CI changes in #61).

Part of the reason I did so was not to conflict with work from Team Sliver to enable borse and merge dart_ci/rust_ci.

discovery_engine_flutter/.gitignore
rustonaut: If we need the header, then just copying it there is probably not enough.

As I can't really test that we probably should put it into a follow up PR.

tmp_script.sh
acrrd: Seems a good idea

tmp_script.sh
acrrd: I mean that locally we can just run it and have the files in the correct format.

tmp_script.sh
rustonaut: Ok I reposted the comment on #61 so that's not lost.

acrrd (approved):

PR #55, opened 2022-01-17
TY-2287 Check that copyright is there in the CI

acrrd:


acrrd:
this is a test where I removed some copyright:
https://github.com/xaynetwork/xayn_discovery_engine/runs/4843191215?check_suite_focus=true

rustonaut (approved):
lgtm

I was wondering what made you use `set -p` (using it is fine, I just think knowing the reason could be useful).

acrrd:
> I was wondering what made you use `set -p` (using it is fine, I just think knowing the reason could be useful).

It'll remove it. I'm not sure what I wanted to put there but not `-p` for sure. Thanks

PR #61, opened 2022-01-18
TY 2273 use job runner for local scripts

rustonaut:
CI will be updated in follow up PR (dedup, etc.)

### DIFF
--- a/justfile
+++ b/justfile
@@ -1,0 +1,94 @@
+
+export RUST_WORKSPACE := env_var_or_default("RUST_WORKSPACE", "discovery_engine_core")
+export DART_WORKSPACE := env_var_or_default("DART_WORKSPACE", "discovery_engine")
+
+default:
+    @{{just_executable()}} --list
+
+_codegen_workaround:
+    cd "$RUST_WORKSPACE"; \
+    cargo check --quiet 2>/dev/null || :
+
+rust_check_fmt:
+    cd "$RUST_WORKSPACE"; \
+    cargo +nightly fmt --all -- --check ;\
+    cargo sort --grouped --workspace --check
+
+_rust_check_only: _codegen_workaround
+    cd "$RUST_WORKSPACE"; \
+    cargo clippy --all-targets --locked
+
+_rust_build_only: _codegen_workaround
+    cd "$RUST_WORKSPACE"; \
+    cargo build --locked
+
+_rust_test_only: _codegen_workaround
+    cd "$RUST_WORKSPACE"; \
+    cargo test --locked
+
+
+rust_check: rust_check_fmt _rust_check_only
+
+rust_test: rust_check _rust_test_only
+
+copy_ios_header:
+    cp ./discovery_engine_core/bindings/include/*.h ./discovery_engine_flutter/ios/Classes/
+
+rust_gen_ffi: _rust_check_only copy_ios_header
+
+dart_get_deps:
+    cd "$DART_WORKSPACE"; \
+    dart pub get
+
+dart_check_fmt:
+    cd "$DART_WORKSPACE"; \
+    dart format --output=none --set-exit-if-changed .
+
+_dart_check_only: dart_get_deps
+    cd "$DART_WORKSPACE"; \
+    dart analyze --fatal-infos
+
+_dart_test_only:
+    cd "$DART_WORKSPACE"; \
+    dart test
+
+_dart_gen_build_runner: dart_get_deps
+    cd "$DART_WORKSPACE"; \
+    dart run build_runner build --delete-conflicting-outputs
+
+_dart_gen_genesis_only: dart_get_deps
+    cd "$DART_WORKSPACE"; \
+    dart run ffigen --config ffigen.yaml
+
+_dart_gen_genesis_ext_only:
+    cd "$RUST_WORKSPACE"; \
+    "${CARGO_INSTALL_ROOT:-${CARGO_HOME:-$HOME/.cargo}}/bin/async-bindgen-gen-dart" \
+        --ffi-class XaynDiscoveryEngineBindingsFfi \
+        --genesis ../discovery_engine/lib/src/ffi/genesis.ffigen.dart
+
+_dart_gen_ffi: rust_gen_ffi _dart_gen_genesis_only _dart_gen_genesis_ext_only
+
+dart_gen_code: _dart_gen_ffi _dart_gen_build_runner
+
+dart_check: dart_gen_code _dart_check_only
+
+dart_test: dart_gen_code dart_check _rust_build_only _dart_test_only
+
+test: rust_test dart_test
+
+check: rust_check dart_check
+
+codegen_clean:
+    find . \( -name '*.g.dart' \
+        -or -name '*.freezed.dart' \
+        -or -name '*.ffigen.dart' \
+        -or -name '*.ext.dart' \
+    \) -exec rm '{}' \;
+    -rm "$RUST_WORKSPACE"/bindings/include/*
+    -rm "$RUST_WORKSPACE"/bindings/src/async_bindings/* || :
+
+install_async_bindgen:
+    cargo install \
+        --git https://github.com/xaynetwork/xayn_async_bindgen.git \
+        async-bindgen-gen-dart \
+        "$@"
--- a/justfile
+++ b/justfile
@@ -3,6 +3,7 @@
 #FIXME: Consider using .env for this, but only if used also on CI
 export RUST_WORKSPACE := env_var_or_default("RUST_WORKSPACE", "discovery_engine_core")
 export DART_WORKSPACE := env_var_or_default("DART_WORKSPACE", "discovery_engine")
+export CARGO_INSTALL_ROOT := env_var_or_default("CARGO_INSTALL_ROOT", "cargo-installs")
 
 # Run just --list
 default:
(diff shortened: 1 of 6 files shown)


justfile
```diff
@@ -0,0 +1,118 @@
...
+    cd "$DART_WORKSPACE"; \
+    dart run build_runner build --delete-conflicting-outputs
+
+_dart_gen_genesis_only: dart_get_deps
+    cd "$DART_WORKSPACE"; \
+    dart run ffigen --config ffigen.yaml
+
+_dart_gen_genesis_ext_only:
+    cd "$RUST_WORKSPACE"; \
+    "${CARGO_INSTALL_ROOT:-${CARGO_HOME:-$HOME/.cargo}}/bin/async-bindgen-gen-dart" \
```
rustonaut, line 82: > @janpetschexain @Robert-Steiner @biernacki can you test that this works on mac?


> Maybe we should use a custom install root in the workspace (git-ignored)?

> Then we also don't have a problem with any situation where the install root is
> changed in a cargo config file (.cargo/config.toml).

Comments moved from previous PR where this code was in the end not included.

rustonaut: acrrd:

> Seems a good idea

justfile
```diff
@@ -0,0 +1,156 @@
...
+
+# Format all code (checks only on CI)
+fmt: rust_fmt dart_fmt
+
+# Check dart code, fails on info on CI
+dart_check: dart_build
+    cd "$DART_WORKSPACE"; \
+    dart analyze {{ if env_var_or_default("CI", "false") == "true" { "--fatal-infos" } else { "" } }}
+
+_codegen_order_workaround:
```
x3ro, line 52: To me it's not obvious what this does or why it's needed. Since it's labeled workaround, could you add a comment explaining why this is needed?

justfile
```diff
@@ -0,0 +1,156 @@
...
+    cd "$DART_WORKSPACE"; \
+    dart pub global run dartdoc:dartdoc --no-generate-docs --no-quiet
+
+# Check if dart documentation can be build without issues
+rust_check_doc: _codegen_order_workaround
+    cd "$RUST_WORKSPACE"; \
+    cargo doc --all-features --no-deps --document-private-items --locked
+
+# Check if documentation can be build without issues
+check_doc: dart_check_doc rust_check_doc
```
x3ro, line 75: How is `check_doc` different from just generating the documentation and seeing if it fails (it looks like it's doing that). In that case it could just be called `doc`, right?

In other words, would I run this locally to generate the docs? If so, `check_doc` would be an odd name.

justfile
```diff
@@ -0,0 +1,156 @@
...
+export RUST_WORKSPACE := env_var_or_default("RUST_WORKSPACE", "discovery_engine_core")
+export DART_WORKSPACE := env_var_or_default("DART_WORKSPACE", "discovery_engine")
+export CARGO_INSTALL_ROOT := env_var_or_default("CARGO_INSTALL_ROOT", "cargo-installs")
+
+# Run just --list
+default:
+    @{{just_executable()}} --list
+
+# Get/update dart deps
+dart_deps:
```
x3ro, line 13: Personal note: it appears that justfiles support dashes instead of underscores in names. I find that it's easier to type `just dart-fmt` than `just dart_fmt`, and it also seems more "idiomatic" (I'm using this term very loosely here) when it comes to command line arguments, which in my experience rarely use underscores.

x3ro (approved):
Looks good overall 🚀

justfile
rustonaut: I fully agree, I somehow had the misconception you can't do that. I will change it.

justfile
rustonaut: We use cbindgen to generate C headers for ffi functions.

We use async_bindgen to generate some of the C functions.

The problem is that cbindgen doesn't the code generated by derives and pro-macros, normal macros etc. (because it doesn't tightly integrate into rust, in turn you can use it without running cargo).

So it doesn't see our generate C functions.

As a ad-hoc fix for this we wrote the generated functions into a file.

The problem is that the file is generate after cbindgen runs, also after rust parsed all files.

Hence we must run a (likely failing) cargo check before running what we actually want. A bit like in latex.

This obvious is not the best solution to the problem, just the fastest to get working.

But replacing it with a more proper solution has currently no-priority.

Ironically I missed this problem for a while because vscode being so nice as to doing that "pre-run" for me, sometimes even after I closed it to test exactly that case (there is/was some bug where in some edge cases rust-analyzer continues to run for a while even after you close the workspace).

x3ro:
One more thing: I like looking at the generated docs, and for convenience it'd be nice to have something that builds them and then spins up a server in the right directory (e.g. `python3 -m http.server`). I realize that's a nice-to-have though :D

justfile
x3ro: Thanks for the in-depth explanation :) Could you also put a condensed version in the justfile as a comment? :>

rustonaut:
I will make a jira ticket for some additional improvements like:

- the python dart doc server thingy
- installing `just pre-push` as pre-push hook, maybe also some pre-commit hooks

justfile
rustonaut: It's not quite the same, dart-doc as called here doesn't generate documentation (but it will tell you it did :shrug: ).

Similar for rust we omit documentation for dependencies, but in my experience having documentation for dependencies appearing in the public interface is quite important.

I added a separate `doc` step instead.

justfile
```diff
@@ -0,0 +1,180 @@
+
+# Make sure that some env variables are set for all jobs.
+#FIXME: Consider using .env for this, but only if used also on CI
```
acrrd, line 3: I thought we decided to use it. Could be a problem with the CI?

justfile
```diff
@@ -0,0 +1,180 @@
...
+    cd "$RUST_WORKSPACE"; \
+    cargo fetch {{ if env_var_or_default("CI", "false") == "true" { "--locked" } else { "" } }}
+
+# Installs the async-bindgen CLI tool
+install-async-bindgen *args:
+    cargo install \
+        --git https://github.com/xaynetwork/xayn_async_bindgen.git \
+        {{args}} \
+        async-bindgen-gen-dart \
+        "$@" ;
```
acrrd, line 28: What are the expected args here?

justfile
```diff
@@ -0,0 +1,180 @@
...
+
+# Remvoes all local cargo isntalls
+remove-local-cargo-installs:
+    rm -r "$CARGO_INSTALL_ROOT"
+
+# Removes all local dependency artifacts
+clean-deps: rust-clean-deps dart-clean-deps remove-local-cargo-installs
+
+# Removes all local cached dependencies and generated files
+clean-fully: clean-files clean-deps
```
acrrd, line 168: Just `clean`?

justfile
```diff
@@ -0,0 +1,180 @@
...
+    find . \( -name '*.g.dart' \
+        -or -name '*.freezed.dart' \
+        -or -name '*.ffigen.dart' \
+        -or -name '*.ext.dart' \
+    \) -exec rm '{}' \;
+    -rm "$RUST_WORKSPACE"/bindings/include/*
+    -rm "$RUST_WORKSPACE"/bindings/src/async_bindings/*
+
+# Cleans up rusts build cache
+rust-clean-deps:
```
acrrd, line 152: Just `rust-clean`? Same for `dart`. Why did you add `deps`?

justfile
```diff
@@ -0,0 +1,180 @@
...
+# Tests rust
+rust-test:
+    cd "$RUST_WORKSPACE"; \
+    cargo test --locked
+
+# Tests dart and rust
+test: rust-test dart-test
+
+# Cleans up all generated files
+clean-files:
```
acrrd, line 142: Maybe `clean-gen-files`?

justfile
```diff
@@ -0,0 +1,180 @@
...
+    -rm "$RUST_WORKSPACE"/bindings/src/async_bindings/*
+
+# Cleans up rusts build cache
+rust-clean-deps:
+    cd "$RUST_WORKSPACE"; \
+    cargo clean
+
+# Cleans up darts build cache
+dart-clean-deps:
+    find "$DART_WORKSPACE" -type d -name .dart_tool -prune -exec rm -r '{}' \;
```
acrrd, line 158: There a `flutter clean` we can run it in `discovery_engine_flutter` . But I think it only does something similar to this .

justfile
rustonaut: just not in this PR, the comment is older then the decision

justfile
rustonaut: None, you could put `--force` in there.

justfile
rustonaut: ok, Through that cleans much more then you normally would ever want.

justfile
rustonaut: It doesn't clean the package cache, or generated files. I will rename it.

justfile
rustonaut: There is no `dart clean` so we need to do it by hand for the dart repo.

acrrd: Yep, I know, not sure why is in flutter and not in dart

justfile
rustonaut: updated the fixme

justfile
acrrd: Is probably a matter of opinions, When I want to clean I'm usually looking for something like this.
If we realize we want less we can always redirect this to something else.

acrrd (approved):
Great job this will help us a lot!

PR #63, opened 2022-01-19
DEVOPS-671 Adding a new CI merging rust and dart ci

jorge-cr-13:
Creating a new CI with the merger of the rust and dart CI's for bors.


jorge-cr-13:
I'm merging the dart CI and the rust CI, there is a new ci-ok job that will be used by bors.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,352 @@
...
+        id: filter
+        with:
+          filters: |
+            dart:
+              - '.github/workflows/dart_ci.yml'
+              - 'discovery_engine/**'
+              - 'discovery_engine_core/**'
+              - 'job_runner.sh'
+            rust:
+              - '.github/workflows/rust_ci.yml'
```
rustonaut, line 48: As I understand this replaces `dart_ci.yml` and `rust_ci.yml` so they can be removed from the files which trigger job runs (and deleted, I guess).

.github/workflows/ci.yml
```diff
@@ -0,0 +1,352 @@
...
+              - 'discovery_engine_core/**'
+
+  # DART CI 
+  dart-format:
+    name: dart-format
+    needs: selection
+    if: ${{ needs.changes.outputs.dart == 'true' }}
```
rustonaut, line 55:
```suggestion
    if: ${{ needs.selection.outputs.dart == 'true' }}
```
or do I overlooked something?

Same for the other ifs.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,352 @@
...
+    runs-on: ubuntu-20.04
+    timeout-minutes: 5
+    steps:
+      - name: CI OK Testing
+        if: ${{ needs.selection.outputs.dart == 'true' }} || ${{ needs.selection.outputs.rust == 'true' }}
+        run: |
+          if ( ${{ needs.selection.outputs.dart == 'true' }} && ${{ needs.selection.outputs.rust == 'true' }} && ${{ needs.dart-doc.result == 'success'}} && ${{ needs.cargo-ok.result == 'success'}} )
+          then
+            exit 0
+          elif ( ${{ needs.selection.outputs.dart == 'true' }} && ${{ needs.dart-doc.result == 'success'}} && ${{ needs.changes.outputs.test2 == 'false' }} )
```
rustonaut, line 343: What is the `test2` output for?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,352 @@
...
+      - cargo-test
+      - cargo-doc
+    runs-on: ubuntu-20.04
+    steps:
+      - run: exit 0
+  
+  ci-ok:
+    name: ci-ok
+    needs: [selection, dart-doc, cargo-ok]
+    if: ${{always()}}
```
rustonaut, line 333: For better understanding why is this needed, instead of just not having an if clause.

rustonaut, line 333: For better understanding why is this needed, instead of just not having an if clause.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,352 @@
...
+      - name: cargo doc
+        working-directory: ${{ env.RUST_WORKSPACE }}
+        run: |
+          # workaround for some code being created after rust wants to parse it
+          cargo check --quiet 2>/dev/null || :
+          cargo doc --all-features --no-deps --document-private-items
+
+    # this is an helper that needs all the real leafs of the workflow.
+    # It makes easier notify_main_failure because we only need to check
+    # for this job
```
rustonaut, line 317: This can be deleted, or updated there is also no `notify_main_failure` around anymore in this script.

rustonaut:
you might have forgotten to push given that the last commit is from 6 hours ago?

.github/workflows/ci.yml
acrrd: It force it to run even if some needed job was skipped.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,352 @@
...
+    name: ci-ok
+    needs: [selection, dart-doc, cargo-ok]
+    if: ${{always()}}
+    runs-on: ubuntu-20.04
+    timeout-minutes: 5
+    steps:
+      - name: CI OK Testing
+        if: ${{ needs.selection.outputs.dart == 'true' }} || ${{ needs.selection.outputs.rust == 'true' }}
+        run: |
+          if ( ${{ needs.selection.outputs.dart == 'true' }} && ${{ needs.selection.outputs.rust == 'true' }} && ${{ needs.dart-doc.result == 'success'}} && ${{ needs.cargo-ok.result == 'success'}} )
```
acrrd, line 340: Can't we write all of this as:
`(outputs.dart implies dart.result == success) and (outputs.rust implies rust.result == success)`
?
instead of having multiple cases?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,352 @@
...
+      - name: Install dart
+        uses: dart-lang/setup-dart@6a218f2413a3e78e9087f638a238f6b40893203d # v1.3
+        with:
+          sdk: ${{ env.DART_VERSION }}
+
+      - name: Check formatting
+        working-directory: ${{ env.DART_WORKSPACE }}
+        run: dart format --output=none --set-exit-if-changed .
+
+  dart-analyze:
```
acrrd, line 71: as in rust we can have all dart-* run in parallel. Don't know how easy it is but would be nice if `dart-test` could depend on `rust-test` if it must run.

.github/workflows/ci.yml
jorge-cr-13: Yes, good point I forgot to delete them from the repository and from this path.

.github/workflows/ci.yml
jorge-cr-13: Yes, Andrea told me that given that we were going to use `bors` the `notifyy_main_failure` was not needed anymore.

.github/workflows/ci.yml
jorge-cr-13: Yes, that is correct. The `changes` and `test` are from my testing, but I updated them to `rust` and `dart` respectively.

.github/workflows/ci.yml
jorge-cr-13: The changes and test are from my testing, but I updated them to rust and dart respectively.

.github/workflows/ci.yml
jorge-cr-13: Because it needs to run after `selection`, `dart-doc` and `cargo-ok` have run, but if any has been skipped then anyways run.

.github/workflows/ci.yml
jorge-cr-13: Ok, I will try.

jorge-cr-13: I made some changes, let me know if that was what you were referring to.

.github/workflows/ci.yml
jorge-cr-13: I'm going to create a new task for this.

.github/workflows/ci.yml
acrrd: How can it work if we are looking for a key in `needs` that we don't specify?

.github/workflows/ci.yml
```diff
@@ -41,7 +65,8 @@ jobs:
 
   dart-analyze:
     name: dart-analyze
-    needs: dart-format
+    needs: [dart-format, selection]
```
acrrd, line 68: We discussed about running them in parallel. Will be implemented in another PR?

.github/workflows/ci.yml
```diff
@@ -189,40 +216,123 @@ jobs:
...
+  ci-ok:
+    name: ci-ok
+    needs: [selection, dart-doc, cargo-ok]
+    if: ${{always()}}
+    runs-on: ubuntu-20.04
+    timeout-minutes: 5
+    steps:
+      - name: CI OK Testing
+        run: |
+          if [[ ${{ needs.selection.outputs.dart == 'true' }} && ${{ needs.selection.outputs.rust == 'true' }} && ${{ needs.dart-doc.result == 'success'}} && ${{ needs.cargo-ok.result == 'success'}} ||
```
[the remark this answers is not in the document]
jorge-cr-13: There is the case where only the `dart` section changes and the `rust` jobs should not run, so we just need:
```
outputs.dart == 'true' => dart-doc.result == 'success'
```
My understanding is that the `dart` jobs have to run all the time, right?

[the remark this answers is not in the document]
acrrd: dart-test will run in both cases yes. but then could be worth to move it out from `dart-ok`. 
We discuss this when we have the PR that run the dart process in parallel then and merge this in the mean while.

.github/workflows/ci.yml
```diff
@@ -19,11 +17,37 @@ env:
...
+      rust: ${{ steps.filter.outputs.rust }}
+    steps:
+      # For pull requests it's not necessary to checkout the code
+      - uses: dorny/paths-filter@v2.10.2
+        id: filter
+        with:
+          filters: |
+            dart:
+              - 'discovery_engine/**'
+              - 'discovery_engine_core/**'
```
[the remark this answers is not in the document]
jorge-cr-13: Spoke to @rustonaut  and he explained me that: 
`discovery_engine_core/` generates a C header which is then used by `discovery_engine` to generate dart code using a external tool (which happend to have been written in rust, but it's  in another repo). Through for testing dart we need to fully build rust.
`discovery_engine_flutter` will only have dart code? Yes (ignoring the kotlin/swift/etc. glue code for compiling it as app, in which case you also need fully build rust code, testing it also needs to have rust build as it might call functions implemented in rust during tests).

.github/workflows/ci.yml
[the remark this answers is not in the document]
jorge-cr-13: I made the changes, I think we have covered all the cases. I'll open a PR with the parallel runs.

.github/workflows/ci.yml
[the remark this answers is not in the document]
acrrd: Ok, we can see if we can do something better later with a cache for generated file bu is also a bit risky.

acrrd (approved):

PR #65, opened 2022-01-20
TY 2276 define client api

acrrd:
It adds a mock of the `Client` struct and api.

It adds the data structure to let the user filter the documents he is interested in (`Filter`).

### DIFF
--- a/discovery_engine_core/providers/src/client.rs
+++ b/discovery_engine_core/providers/src/client.rs
@@ -1,0 +1,48 @@
+// Copyright 2022 Xayn AG
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, version 3.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+//! Client do get new documents.
+
+use xayn_discovery_engine_core::document::Document;
+
+use crate::{filter::Filter, query::Query};
+
+// TODO: replace with error form the http library + Filter error
+pub enum Error {}
+
+/// Client that can provide documents.
+pub struct Client {
+    token: String,
+    url: String,
+}
+
+impl Client {
+    /// Create a client.
+    pub fn new(token: String, url: String) -> Self {
+        Self { token, url }
+    }
+
+    /// Get document from a provider,
+    ///
+    /// `filter` can be used to query only documents of interests,
+    /// `size` optionally limit the number of item we want.
+    pub async fn get(&self, filter: Filter, size: Option<u8>) -> Result<Vec<Document>, Error> {
+        let _query = Query::new(self.token.clone(), filter, size);
+
+        // just to use it until we really use it
+        let _url = self.url.clone();
+
+        Ok(vec![])
+    }
+}
--- a/discovery_engine_core/providers/src/client.rs
+++ b/discovery_engine_core/providers/src/client.rs
@@ -37,7 +37,7 @@ impl Client {
     ///
     /// `filter` can be used to query only documents of interests,
     /// `size` optionally limit the number of item we want.
-    pub async fn get(&self, filter: Filter, size: Option<u8>) -> Result<Vec<Document>, Error> {
+    pub async fn get(&self, filter: &Filter, size: Option<u8>) -> Result<Vec<Document>, Error> {
         let _query = Query::new(self.token.clone(), filter, size);
 
         // just to use it until we really use it
--- a/discovery_engine_core/providers/src/expression.rs
+++ b/discovery_engine_core/providers/src/expression.rs
@@ -1,0 +1,376 @@
+// Copyright 2022 Xayn AG
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, version 3.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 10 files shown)


discovery_engine_core/providers/src/client.rs
```diff
@@ -0,0 +1,48 @@
...
+    /// Create a client.
+    pub fn new(token: String, url: String) -> Self {
+        Self { token, url }
+    }
+
+    /// Get document from a provider,
+    ///
+    /// `filter` can be used to query only documents of interests,
+    /// `size` optionally limit the number of item we want.
+    pub async fn get(&self, filter: Filter, size: Option<u8>) -> Result<Vec<Document>, Error> {
```
x3ro, line 40: Maybe `Option<usize>`? Won't be slower on a 64-bit machine afaik.

discovery_engine_core/providers/src/client.rs
```diff
@@ -0,0 +1,48 @@
...
+        let _query = Query::new(self.token.clone(), filter, size);
```
x3ro, line 41: I was wondering if we'd want to pass by reference into `get` and `Query::new`. Are you worried about that leading to unforseen compilation issues in `async` context?

discovery_engine_core/providers/src/expression.rs
```diff
@@ -0,0 +1,376 @@
...
+    fn test_build_and_or() {
+        let a = Expr::Value("a".to_string());
+        let b = Expr::Value("b".to_string());
+        let c = Expr::Value("c".to_string());
+        let d = Expr::Value("d".to_string());
+
+        // (a /\ b) \/ c
+        assert_eq!(
+            "(a AND b) OR c",
+            a.clone().and(b.clone()).or(c.clone()).build()
```
x3ro, line 358: Are these expressions that we would directly pass to WebZ? Or who will be the consumer for these strings?

discovery_engine_core/providers/src/expression.rs
```diff
@@ -0,0 +1,376 @@
...
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub(crate) enum Expr {
+    // contained `Expr` must not be an `And`
+    And(BTreeSet<Expr>),
+    // contained `Expr` must not be an `Or`
+    Or(BTreeSet<Expr>),
+    Value(String),
+}
+
+impl Expr {
```
x3ro, line 31: It may be worth also implementing the following, to avoid all those explicit `.clone()` calls :D 

```rust
impl Into<Expr> for &Expr {
    fn into(self) -> Expr {
        self.clone()
    }
}
```

x3ro (commented):
Looks good overall, let me know your thoughts on my comments :)

discovery_engine_core/providers/src/client.rs
acrrd: but the maximum that we can request from this specific endpoint is 100 and in general asking for 255 news seems more then enough.

discovery_engine_core/providers/src/expression.rs
acrrd: The endpoint what the string in this format.

discovery_engine_core/providers/src/client.rs
acrrd: It was because it seems easier to consume it when we convert it to the query string. I'll try taking a ref in the build to see how much we need to change.
It was also because if the markets are changing we need to throw the filter away anyway, and for personalized query the filter will be always different so we don't gain much in not taking the ownership. We can implement clone on it so at least the user can have a choice.

discovery_engine_core/providers/src/expression.rs
acrrd: Unfortunately does not work, with this we can only remove a `clone` when it is used only on right hand side.
`a.clone().and(b.clone().and(c.clone()))` can simplified only to `a.clone().and(b.clone().and(&c))`.

`and` and `or` are taking `self` so we can reuse the btree without cloning.

x3ro: Mmh, for me the following works fine with that extra `impl`:

```rust
&a.and(&b.or(&c)).build()
```

What am I missing?

discovery_engine_core/providers/src/client.rs
acrrd: Done, the changes were contained to the `Into<Expr>` implementations and since internally we crate new string we didn't really need to consume it.

x3ro (approved):

discovery_engine_core/providers/src/expression.rs
acrrd: I'll try it again, but I added it and modified a test and it wasn't working

x3ro: If you have sequences that all start with `&a`, you do need to clone the first one. Maybe that's it?

```rust
&a.clone().and(&b.or(&c)).build()
```

All the inner ones should definitely work, though. At least in my scratchpad :D

acrrd: Did you try to have two?
```rust
&a.and(&b.or(&c)).build()
&a.and(&b.or(&c)).build()
```
on the second I get the error that `a` and `b` were moved on the previous line.

acrrd: The inner has the same problem because rust is implicitly defer the reference to allow `or` to take ownership of `b`, there is no `Into` on the left hand side so it should not even call the new impl

x3ro: Ah, I see the issue. It works in this case:

```rust
#[test]
    fn test_build_and_or() {
        let a = Expr::Value("a".to_string());
        let b = Expr::Value("b".to_string());
        let c = Expr::Value("c".to_string());
        let d = Expr::Value("d".to_string());

        // (a /\ b) \/ c
        assert_eq!(
            "(a AND b) OR c",
            a.clone().and(&b).or(&c).build()
        );

        // a /\ (b \/ c)
        assert_eq!(
            "(b OR c) AND a",
            a.clone().and(&b.or(&c)).build()
        );
    }
```

But not for e.g. `a.clone().or(&b.and(&c))`. Didn't test far enough. Fair enough, in that case one would likely need something more complex. Not worth the effort.

PR #53, opened 2022-01-17
TY-2142 feed manager: feedDocumentsClosed, nextFeedBatchRequested

finiteprods:
**Summary**

adds a `FeedManager` class, with handlers for the feed events:

- _feed documents closed_
- _next feed batch requested_

(_feed requested_ event will be handled in the next pr)

PR #66, opened 2022-01-20
TY-2267 feed manager: restore feed

finiteprods:
**Summary**

Previously,

- feed manager was added in #53 handling _docs closed_ and _next batch_.
- time stamps were added to documents in #64.

this pr adds handling of _restore feed_ to the feed manager.
the returned feed is ordered by `timestamp`, then `personalizedRank`.

### DIFF
--- a/discovery_engine/lib/src/domain/feed_manager.dart
+++ b/discovery_engine/lib/src/domain/feed_manager.dart
@@ -18,6 +18,8 @@ import 'package:xayn_discovery_engine/src/domain/document_manager.dart'
     show DocumentManager;
 import 'package:xayn_discovery_engine/src/domain/engine/engine.dart'
     show Engine;
+import 'package:xayn_discovery_engine/src/domain/models/document.dart'
+    show Document;
 import 'package:xayn_discovery_engine/src/domain/repository/active_document_repo.dart'
     show ActiveDocumentDataRepository;
 import 'package:xayn_discovery_engine/src/domain/repository/document_repo.dart'
@@ -38,7 +40,7 @@ class FeedManager {
   /// Handle the given feed client event.
   ///
   /// Fails if [event] does not have a handler implemented.
-  Future<void> handleFeedClientEvent(FeedClientEvent event) async {
+  Future<List<Document>?> handleFeedClientEvent(FeedClientEvent event) async {
     await event.maybeWhen(
       feedRequested: () => restoreFeed(),
       nextFeedBatchRequested: () => nextFeedBatch(),
@@ -47,7 +49,14 @@ class FeedManager {
     );
   }
 
-  Future<void> restoreFeed() async {} // TODO once timestamps addded to Document
+  /// Restore (or initialise, if called for the first time) the feed.
+  Future<List<Document>?> restoreFeed() async {
+    await _docRepo.fetchAll().then((docs) {
+      docs.retainWhere((doc) => doc.isActive);
+      docs.sort((doc1, doc2) => doc1.timestamp.compareTo(doc2.timestamp));
+      return docs;
+    });
+  }
 
   /// Obtain the next batch of feed documents and persist to repositories.
   Future<void> nextFeedBatch() async {
--- a/discovery_engine/lib/src/domain/feed_manager.dart
+++ b/discovery_engine/lib/src/domain/feed_manager.dart
@@ -13,7 +13,7 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 import 'package:xayn_discovery_engine/src/api/events/client_events.dart'
-    show FeedClientEvent;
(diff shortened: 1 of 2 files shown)


discovery_engine/lib/src/domain/feed_manager.dart
```diff
@@ -38,16 +40,31 @@ class FeedManager {
   /// Handle the given feed client event.
   ///
   /// Fails if [event] does not have a handler implemented.
-  Future<void> handleFeedClientEvent(FeedClientEvent event) async {
-    await event.maybeWhen(
+  /// If [event] is [FeedRequested], returns the feed documents.
+  /// Returns an empty list in all other cases.
+  Future<List<Document>> handleFeedClientEvent(FeedClientEvent event) {
```
acrrd, line 45: I guess I misinterpreted you message when we talked about this.
Will be difficult for the handler to then return the correct message with this type. 
`feedRequested` return a list of documents like `nextFeedBatchRequested` but `feedDocumentsClosed` return only `ClientRequestSuccesful`.
@biernacki do you have any good idea on how to handle this? I guess one option will be to have two different methods or return a more descriptive type or return the event itself.

biernacki: Well, at this time we should just use `EngineEvent` as the return type from the `FeedManager` cause it contains all the possible responses (when we add later more events, like search related ones, that will change).

We could also think about a different structure for Engine events where we have for each client event freezed union with possible responses. But maybe we could revisit that option later. We could combine it then with another option, which would be to have usecases instead of managers, one usecase per event.

finiteprods: thanks :+1: will try `EngineEvent` as the return type for now then and see how that goes. btw I notice it doesn't have a response for `feedDocumentsClosed` so will add a `FeedDocumentsClosedSuccess` (or similar)..

biernacki: the intended response for `FeedDocumentsClosed` is atm `ClientEventSucceeded`, but ofc. we can change that. Though I feel we should be consistent and maybe add success/failure response events to all client events (maybe with freezed union classes)

finiteprods: thanks, since `ClientEventSucceeded` is the intended return value I've gone for that.

discovery_engine/lib/src/domain/models/document.dart
```diff
@@ -52,6 +53,15 @@ class Document {
     this.isActive = true,
   })  : documentId = DocumentId(),
         timestamp = DateTime.now().toUtc();
+
+  api.Document toApiDocument() => api.Document(
```
biernacki, line 57: it's not a concern for this PR, but maybe we could think about renaming one of the `Document` classes to avoid further confusion? I was thinking that the `Document` exposed to the app could be named `DocumentDto`, which is closer to the truth, but it would be a nicer api if we kept it and renamed the domain `Document` instead, maybe to `DocumentEntity`? wdyt? cc @acrrd

discovery_engine/lib/src/domain/feed_manager.dart
```diff
@@ -38,25 +40,46 @@ class FeedManager {
...
-      orElse: throw UnimplementedError('handler not implemented for $event'),
-    );
-  }
+  Future<EngineEvent> handleFeedClientEvent(FeedClientEvent event) =>
+      event.maybeWhen(
+        feedRequested: () => restoreFeed(),
+        nextFeedBatchRequested: () => nextFeedBatch(),
+        feedDocumentsClosed: (ids) => _docMgr
+            .deactivateDocuments(ids)
+            .then((_) => const EngineEvent.clientEventSucceeded()),
```
biernacki, line 49: also not for this PR, just to open up a discussion, but I feel like it's a bit confusing that we have this indirection and calling one manager from the other. This looks like a good reason to have dedicated usecases for every event (by "usecase" I mean something like a manager that handles only 1 event). On the other hand, that would be more boilerplate'y to initialize. Wdyt?

biernacki (approved):
looks good 💪

I left some questions up for general purpose discussion, not concerning this particular PR.

discovery_engine/lib/src/domain/models/document.dart
finiteprods: I think either of those suggestions would be a little less confusing, yes.

discovery_engine/lib/src/domain/feed_manager.dart
finiteprods: yes, the dependency on the document manager here also makes testing a little more awkward.
@acrrd is it necessary to have it this way, or could the `deactivateDocuments` method be part of the feed manager, say?

PR #69, opened 2022-01-24
DEVOPS-671 Creating the reusable wf file

jorge-cr-13:
Before using reusable workflows we need to have the workflow on the `main` branch.

### DIFF
--- a/.github/workflows/ci_reusable_wf.yml
+++ b/.github/workflows/ci_reusable_wf.yml
@@ -1,0 +1,349 @@
+name: Reusable Dev CI
+
+on:
+  workflow_call:
+    outputs:
+      dart_output:
+        description: Output if dart has to be run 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci_reusable_wf.yml
```diff
@@ -0,0 +1,349 @@
+name: Reusable Dev CI
```
acrrd, line 1: Should't we call this from `ci.yml`?

acrrd (approved):

.github/workflows/ci_reusable_wf.yml
jorge-cr-13: Yes, but in order for this to work, it has to be first on the main branch. So I have to create a PR for creating the file and merging it into main and then I can call it on `ci.yml`.

acrrd (approved):

PR #71, opened 2022-01-24
TY 2295 Fix release ci

acrrd:
With the current release workflow the flutter plugin always depends on the `main` branch.
This PR change that to the previous commit so that it always depends on the same branch and at a fixed location.

### DIFF
--- a/discovery_engine_flutter/pubspec.yaml
+++ b/discovery_engine_flutter/pubspec.yaml
@@ -14,7 +14,7 @@ dependencies:
   xayn_discovery_engine:
     git:
       url: git@github.com:xaynetwork/xayn_discovery_engine_release.git
-      ref: main
+      ref: change_me_to_commit_ref
       path: discovery_engine
 
 dev_dependencies:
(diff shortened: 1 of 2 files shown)


discovery_engine_flutter/pubspec.yaml
```diff
@@ -14,7 +14,7 @@ dependencies:
   xayn_discovery_engine:
     git:
       url: git@github.com:xaynetwork/xayn_discovery_engine_release.git
-      ref: main
+      ref: change_me_to_commit_ref
```
rustonaut, line 17: How does that interact with local flutter builds/testing, non-release CI flutter testing?

acrrd:
https://github.com/xaynetwork/xayn_discovery_engine_release/tree/TY-2295-fix_release_ci

discovery_engine_flutter/pubspec.yaml
acrrd: It doesn't interact because there is the override to the current directory

rustonaut (approved):
lgtm

PR #59, opened 2022-01-18
TY-2257 Download assets + TY-2275 retry logic

biernacki:
[Jira ref: Download assets](https://xainag.atlassian.net/browse/TY-2257)
[Jira ref: Retry logic](https://xainag.atlassian.net/browse/TY-2275)

This PR adds `DataProvider` abstract class with native and web implementations, and `AssetFetcher` with file and http implementations, with the ability to retry when initial requests were unsuccessful.

It also adds tests for `NativeDataProvider` and `HttpAssetFetcher` using a local http server.

PR #74, opened 2022-01-25
TY-2327 Added `assetsUrl` to `Configuration`

biernacki:
[Jira ref](https://xainag.atlassian.net/browse/TY-2327)

Added `assetsUrl` to `Configuration` object for Discovery Engine.

### DIFF
--- a/discovery_engine_flutter/example/pubspec.lock
+++ b/discovery_engine_flutter/example/pubspec.lock
@@ -104,6 +104,27 @@ packages:
       url: "https://pub.dartlang.org"
     source: hosted
     version: "2.0.4"
+  http:
+    dependency: transitive
+    description:
+      name: http
+      url: "https://pub.dartlang.org"
+    source: hosted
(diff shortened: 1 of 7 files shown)


discovery_engine_flutter/example/pubspec.lock
```diff
@@ -104,6 +104,27 @@ packages:
       url: "https://pub.dartlang.org"
     source: hosted
     version: "2.0.4"
+  http:
```
acrrd, line 107: Why the lock is updated when no new deps are adde or modifed in this pr?

acrrd (approved):

discovery_engine_flutter/example/pubspec.lock
biernacki: I think it was supposed to be updated in with one of the previous PRs, namely #59, but not sure why `flutterp pub get` didn't pick it up before

PR #80, opened 2022-01-26
[Merged by Bors] - TY-2279 Implement news client

x3ro:
[TY-2279]
[TY-2278]

Notes to the reviewer: There are some open questions regarding how to initialize `Document`s from Newscatcher results. Please take a look at the `TODO`s in `newscatcher.rs`.

[TY-2279]: https://xainag.atlassian.net/browse/TY-2279?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[TY-2278]: https://xainag.atlassian.net/browse/TY-2278?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


discovery_engine_core/core/src/document.rs
```diff
@@ -99,8 +99,7 @@ pub enum UserReaction {
 }
 
 /// A d-dimensional sequence embedding.
-#[derive(Clone, Debug, Deref, Serialize, Deserialize)]
-#[cfg_attr(test, derive(Default))]
+#[derive(Clone, Debug, Deref, Serialize, Deserialize, Default)]
```
acrrd, line 102: It wasn't properly document, but the reason that we have the default only for test is because it does make sense to have a default implementation for this types outside tests.

discovery_engine_core/providers/src/client.rs
```diff
@@ -12,37 +12,256 @@
...
-    /// Get document from a provider,
-    ///
-    /// `filter` can be used to query only documents of interests,
-    /// `size` optionally limit the number of item we want.
-    pub async fn get(&self, filter: &Filter, size: Option<u8>) -> Result<Vec<Document>, Error> {
-        let _query = Query::new(self.token.clone(), filter, size);
+    /// Retrieve news from the remote API
+    pub async fn news(&self, params: &NewsQuery) -> Result<Vec<Document>, Error> {
+        let mut query: HashMap<String, String> = HashMap::new();
+        query.insert("sort_by".into(), "relevancy".into());
```
acrrd, line 73: Doesn't seems to be an option anymore from the documentation.

discovery_engine_core/providers/src/client.rs
```diff
@@ -12,37 +12,256 @@
...
+            .header("x-api-key", &self.token)
+            .query(&query)
+            .send()
+            .await
+            .map_err(Error::RequestExecution)?
+            .error_for_status()
+            .map_err(Error::StatusCode)?;
+
+        let news: NewscatcherResponse = response.json().await.map_err(Error::Decoding)?;
+        let result: Vec<Document> = news.articles.iter().map(|article| article.into()).collect();
```
acrrd, line 88: We cannot return a `Document` from here then. To create a `Document` we need the smbert. I think we can then return the `Vec<Article>`?
This could also make sense because then we can optimize a bit better on which document we need to run it.
We can a set of filter that does not depend on it that we can run and the compute the smbert only on the remaining.
@janpetschexain

discovery_engine_core/providers/src/expression.rs
```diff
@@ -94,8 +96,8 @@ impl Expr {
 
     pub(crate) fn build(self) -> String {
         match self {
-            Expr::And(exprs) => intersperse_exprs(exprs.into_iter(), " AND "),
-            Expr::Or(exprs) => intersperse_exprs(exprs.into_iter(), " OR "),
+            Expr::And(exprs) => intersperse_exprs(exprs.into_iter(), " && "),
+            Expr::Or(exprs) => intersperse_exprs(exprs.into_iter(), " || "),
```
acrrd, line 100: Is there any reason? They seems to support also the string variant

discovery_engine_core/providers/src/newscatcher.rs
```diff
@@ -0,0 +1,71 @@
...
+}
+
+/// The Newscatcher API does not always return an article `_score`. For example,
+/// articles returned through the `latest_headlines` API are currently not scored
+/// at all. As a result, we need to specify a default value.
+const DEFAULT_SCORE: f32 = 1.;
+
+impl From<&Article> for Document {
+    fn from(article: &Article) -> Self {
+        // TODO (lj): Is this the correct value to use for `rank`? From the Newscatcher API:
```
acrrd, line 38: We can drop `score` and  `rank`

discovery_engine_core/providers/src/filter.rs
```diff
@@ -35,182 +30,39 @@ impl Filter {
...
-
-        keywords.and(markets).and(site_types).build()
+        keywords.build()
     }
 }
 
 /// Define area and language of interests.
 #[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
-pub struct Market {
+pub(crate) struct Market {
```
acrrd, line 42: If we are going to keep then why not using it in the QueryParam instead of two distinct strings?

acrrd (commented):
Why not keeping `test-fixtures` under `provider`?

About the fixuters themself, I'm not confortable in having real domain, links, names, and handles in here. We can also shorten the summary and reduce the page size to 2.

discovery_engine_core/core/src/document.rs
x3ro: I expected as much, I just needed a way to write the conversion and wasn't clear on how to instantiate the types. Since we'll be dropping them, this is no longer an issue anyway.

discovery_engine_core/providers/src/expression.rs
x3ro: You're right, from some of their examples it seemed like they only support the one and I didn't investigate further. Will revert that part.

discovery_engine_core/providers/src/filter.rs
x3ro: Mmh, I'm actually not quite sure what happened here. I wanted to delete this struct as well, either me or IntelliJ made a mistake with the deletion refactor :D

x3ro:
> Why not keeping `test-fixtures` under `provider`?

I was expecting that this (and other) test data might also later be used to perform more intricate integration testing, including the machine learning parts. I assume then that that is not the plan?

> About the fixuters themself, I'm not confortable in having real domain, links, names, and handles in here. We can also shorten the summary and reduce the page size to 2.

I see the potential advantage of anonymizing the data. Shortening it would make it unusable for the more intricate testcases I mentioned above though, I assume.

x3ro:
Made the requested changes. Can you re-review @acrrd?

discovery_engine_core/providers/src/client.rs
acrrd: if I go that link i don't see a table with that data :/

x3ro: Ah yes, it's a bit hidden.
Screenshot 2022-01-28 at 11 25 21

acrrd: thanks

discovery_engine_core/providers/src/expression.rs
```diff
@@ -20,6 +20,7 @@ use itertools::{intersperse, Either, Itertools};
 use maplit::btreeset;
 
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+#[allow(dead_code)]
```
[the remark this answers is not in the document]
acrrd: For some reason if we move the `allow[dead_code]` to the `and` function it is not needed here.

acrrd:
> I see the potential advantage of anonymizing the data. Shortening it would make it unusable for the more intricate testcases I mentioned above though, I assume.

At the moment is not planned. For the e2e test with ai we have data coming from them, but the length will not affect the result if we both use the same data.

discovery_engine_core/providers/src/client.rs
```diff
@@ -12,37 +12,256 @@
...
-        let _url = self.url.clone();
+        let docs = client.headlines(&params).await.unwrap();
+        assert_eq!(docs.len(), 2);
 
-        Ok(vec![])
+        let doc = docs.get(1).unwrap();
+        assert_eq!(
+            doc.title,
+            "Jerusalem blanketed in white after rare snowfall"
+        );
```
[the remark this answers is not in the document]
x3ro: We could, but that would only make sense if we were actually talking to their API. Here we're comparing against a previously stored response. What would the benefit be of testing against that?

acrrd (approved):

x3ro:
bors merge

PR #81, opened 2022-01-26
[Merged by Bors] - Add WebResource

acrrd:
Add the `WebResource` struct as a counterpart of the class that we have in dart.


acrrd:
> I guess the test only impl. will be used in follow up PRs?

The one for `WebResource` is already used in the current test when we create a document.
For `WebResourceProvider` in the future when it will make sense to have it in the tests.

discovery_engine_core/core/src/document.rs
```diff
@@ -63,23 +79,86 @@ pub struct Document {
...
+
+    /// Resource this document refers to.
+    pub web_resource: WebResource,
+}
 
-    /// Text title of the document.
+/// Represents different kinds of resources like web, image, video, news,
+/// that are delivered by an external content API.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct WebResource {
```
x3ro, line 92: Can you elaborate how this structure would be extended to encompass e.g. video-related information such as its length, likes, etc? Is this something that would be added to the `WebResourceProvider`?

discovery_engine_core/core/src/document.rs
```diff
@@ -141,3 +220,165 @@ pub struct UserReacted {
...
+
+        let provider = web_resource.provider.unwrap();
+
+        assert_eq!(String::new(), provider.name);
+        assert_eq!(media, provider.thumbnail.unwrap().to_string());
+    }
+
+    #[test]
+    fn test_web_resource_from_article_invalid_links() {
+        let article = Article {
```
x3ro, line 326: Instead of writing all of these out by hand and adding a lot of boilerplate for the tests, I think we should load these from the provider fixtures instead I used for the newscatcher tests instead.

discovery_engine/lib/src/domain/engine/mock_engine.dart
```diff
@@ -43,6 +43,11 @@ class MockEngine implements Engine {
         'name': 'domain',
         'thumbnail': 'http://thumbnail.domain.com',
       },
+      'rank': 10,
+      'score': 0.1,
+      'country': 'en',
```
x3ro, line 48: `country` would always be in all-caps (ISO 3166-1 alpha-2).

discovery_engine/test/document_manager_test.dart
```diff
@@ -59,17 +59,22 @@ Future<void> main() async {
...
+      url: Uri.parse('https://domain.com'),
+      displayUrl: Uri.parse('domain.com'),
+      datePublished: DateTime.utc(2022, 01, 01),
+      provider: WebResourceProvider(
+        name: 'domain',
+        thumbnail: Uri.parse('http://thumbnail.domain.com'),
+      ),
+      rank: 10,
+      score: 0.1,
+      country: 'en',
```
x3ro, line 74: `country` would always be in all-caps (ISO 3166-1 alpha-2).

discovery_engine/test/feed_manager_test.dart
```diff
@@ -66,17 +66,22 @@ Future<void> main() async {
...
+        url: Uri.parse('https://domain.com'),
+        displayUrl: Uri.parse('domain.com'),
+        datePublished: DateTime.utc(2022, 01, 01),
+        provider: WebResourceProvider(
+          name: 'domain',
+          thumbnail: Uri.parse('http://thumbnail.domain.com'),
+        ),
+        rank: 10,
+        score: 0.1,
+        country: 'en',
```
x3ro, line 81: `country` would always be in all-caps (ISO 3166-1 alpha-2).

discovery_engine/test/repository/hive_document_repo_test.dart
```diff
@@ -32,17 +32,22 @@ Future<void> main() async {
...
+      url: Uri.parse('https://domain.com'),
+      displayUrl: Uri.parse('domain.com'),
+      datePublished: DateTime.utc(2022, 01, 01),
+      provider: WebResourceProvider(
+        name: 'domain',
+        thumbnail: Uri.parse('http://thumbnail.domain.com'),
+      ),
+      rank: 10,
+      score: 0.1,
+      country: 'en',
```
x3ro, line 47: `country` would always be in all-caps (ISO 3166-1 alpha-2).

x3ro (commented):
I think I haven't full grasped the sub-division between WebResource and WebResourceProvider yet. It'd be great if you could elaborate on my question, then I can take another look :)

discovery_engine_core/core/src/document.rs
```diff
@@ -63,23 +79,86 @@ pub struct Document {
...
+                    .then(|| Url::parse(&media))
+                    .transpose()?,
+            }),
+        })
+    }
+}
+
+/// Represents the provider of a [`WebResource`].
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct WebResourceProvider {
```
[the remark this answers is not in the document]
acrrd: It keeps the name of the provider and the link to the thumbnail of the logo of the provider. Is something we just copied from what was done in the app. Some change will be needed here because we don't have a name anymore with the current provider, but we need to sync with blue.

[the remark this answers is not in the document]
frankpepermans: If we get no official "name", then we can just parse the host name from the url.
We need to display the source for legal reasons, Bing used to provide this, cause they often hosted external providers via msn.com.

Ps the data structure may be altered, webResource and Provider are basically matching the old Bing api really.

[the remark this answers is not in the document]
frankpepermans: ...if we do parse the host from the url, then it's also slightly tricky, for example:
washingtonpost.com
Or telegraph.co.uk
Or worse, media.telegraph.co.uk

That's hard to pour into a deterministic name of course, but imo it's fine to just display the full host name, whatever it is.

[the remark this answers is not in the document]
acrrd: We don't have the info from Article for this so I'm going to remove it.

discovery_engine/lib/src/domain/engine/mock_engine.dart
acrrd: done. Do you why it should be all capital? Internally we can probably relax this.

discovery_engine_core/core/src/document.rs
acrrd: It was supposed to be more general but we support only news. I renamed this to `NewsResource`.

discovery_engine_core/core/src/document.rs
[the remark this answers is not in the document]
acrrd: The api provide `clean_url`, that represent the url of the source (usually the base domain). We can export that as `source_url`.

discovery_engine_core/core/src/document.rs
acrrd: I refactored this to reduce boilerplate. I don't want to import it from a json because, for example, we add new fields then we will never notice. If we build the instance instead we will have an error and hopefully we will add the relevant tests.
Also we have an error at compile time if the struct changes.
In dart in most places it was loaded from a json and was a bit annoying because you only notice at runtime.

acrrd:
Another change: in `NewsResource` the field `topic` is now a `String` instead of the enum. At the moment we don't use it in the app and exporting it as `String` will be less work when preparing the ffi.

discovery_engine_core/core/src/document.rs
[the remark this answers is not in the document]
acrrd: `source_url` is there!

discovery_engine/lib/src/domain/engine/mock_engine.dart
x3ro: My guess is they're specified all-capital because the country codes have different semantics from the language codes. For example, the ISO 639-1 language code `cy` means Welsh, whereas the ISO 3166-1 alpha-2 country code `CY` means Cyprus. Later on we may want to use an `enum` internally, to avoid any confusion on the matter ^^

x3ro (approved):
LGTM  🚀

discovery_engine/lib/src/domain/engine/mock_engine.dart
acrrd: I'll prefer a new type with a check in the constructor instead of listing all the codes. Hopefully we will find a library already support this format.

acrrd:
bors merge

discovery_engine/lib/src/domain/engine/mock_engine.dart
x3ro: I don't think I understand. What's the difference between having a library that uses an `enum` with a `from_string` vs. your proposed solution? The crates that I've skimmed earlier all internally represented their states as enums. Is the aversion to enums related to the FFI?

acrrd: That the library has already all the definitions of the languages and code to check if is a valid one and we don't need to write it. If the lib use an enum is even better.

acrrd:
bors merge

PR #84, opened 2022-01-28
[Merged by Bors] - TY-2310 add user reaction to time spent struct

Robert-Steiner:
Ticket:

- [TY-2310]

part of [TY-2282]

Summary:

Adds the `UserReaction` field to the `TimeSpent` struct. The field will be used later in the `log_document_view_time` method in xayn_ai (https://github.com/xaynetwork/xayn_ai/pull/401)

Question:
Do we need to do the same in dart?

[TY-2310]: https://xainag.atlassian.net/browse/TY-2310?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[TY-2282]: https://xainag.atlassian.net/browse/TY-2282?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine/lib/src/domain/document_manager.dart
+++ b/discovery_engine/lib/src/domain/document_manager.dart
@@ -90,7 +90,7 @@ class DocumentManager {
       throw ArgumentError('id $id does not identify an active document');
     }
     final doc = await _documentRepo.fetchById(id);
-    if (doc == null) {
+    if (doc == null || !doc.isActive) {
       throw ArgumentError('id $id does not identify an active document');
     }
 
--- a/discovery_engine/lib/src/domain/document_manager.dart
+++ b/discovery_engine/lib/src/domain/document_manager.dart
@@ -89,6 +89,11 @@ class DocumentManager {
     if (activeData == null) {
       throw ArgumentError('id $id does not identify an active document');
     }
+    final doc = await _documentRepo.fetchById(id);
+    if (doc == null) {
+      throw ArgumentError('id $id does not identify an active document');
+    }
+
     activeData.addViewTime(mode, Duration(seconds: sec));
     await _activeRepo.update(id, activeData);
 
@@ -101,6 +106,7 @@ class DocumentManager {
       id,
       smbertEmbedding: activeData.smbertEmbedding,
       seconds: sumDuration,
+      reaction: doc.feedback,
     );
   }
 }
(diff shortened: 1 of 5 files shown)


janpetschexain (approved):

Robert-Steiner:
@biernacki, @acrrd do we also need to add this field to the dart class?

acrrd:
> @biernacki, @acrrd do we also need to add this field to the dart class?

We should already have it.

biernacki:
> @biernacki, @acrrd do we also need to add this field to the dart class?

@Robert-Steiner it's called `DocumentFeedback` on the Dart side.

Robert-Steiner:
> `DocumentFeedback`

Thanks 👍 and this needs to be added to the `DocumentTimeSpent` event right?

Robert-Steiner:
@biernacki I added the `DocumentFeedback` on the dart side. Can you please have a look?

discovery_engine/lib/src/domain/document_manager.dart
```diff
@@ -81,6 +81,7 @@ class DocumentManager {
     DocumentId id,
     DocumentViewMode mode,
     int sec,
+    DocumentFeedback feedback,
```
biernacki, line 84: I don't think we should add `DocumentFeedback` to the list of args here or to the `DocumentTimeSpent` event. Insteas we should use the `DocumentId` to get a `Document` from the database, and get the feedback from that document.

```dart
final doc = await _documentRepo.fetchById(id);
/*...*/
_engine.timeLogged(
  id,
  smbertEmbedding: activeData.smbertEmbedding,
  seconds: sumDuration,
  reaction: doc.feedback, // access the feedback from the document
);
```

Robert-Steiner: That makes sense 👍 I will implement it like this
Is it possible that a document does not exist?

Robert-Steiner: Done also I added two tests

biernacki (approved):

PR #86, opened 2022-01-28
[Merged by Bors] - TY-2336 update all stacks

janpetschexain:
**References**

- [TY-2336]

**Summary**

- wrap the stacks of the engine in an async lock:
  - make the engine functions async
  - the mocking of the `Ops` trait requires some fiddly setup
- refactor `pull_arms`, there are fewer error variants now, but the outcome of the selection iter is identical, also the mutating logic is moved out of `pull_arms` into the calling selector
- implement update stacks function

[TY-2336]: https://xainag.atlassian.net/browse/TY-2336?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


discovery_engine_core/core/src/mab.rs
```diff
@@ -46,113 +57,100 @@ impl BetaSample for BetaSampler {
...
-            (first_sample, first_bucket),
-            |max, bucket| -> Result<_, Error> {
-                let sample = sample_from_bucket(bucket)?;
-                if let Ordering::Greater = nan_safe_f32_cmp(&sample, &max.0) {
-                    Ok((sample, bucket))
+/// Samples the next element from the buckets.
+#[allow(clippy::future_not_send)]
+async fn pull_arms<T>(
+    beta_sampler: &impl BetaSample,
+    buckets: &[&RwLock<impl Bucket<T>>],
```
acrrd, line 75: I don't think is logically correct to allow modification to the stacks while pulling. This module should not know anything about async on how the data is stored. I think that the user of the mab need to acquire the writing lock for every stack before calling it. Maybe we need to generalize the signature to `&Deref<impl Bucket<T>` or something similar.

janpetschexain: this is just the async implementation of the previous sync code (as it took `buckets: &mut [&mut impl Bucket<T>]` before and mutated them). i agree that it would be better to split it in some way, should we change it in this pr or in another?

acrrd: But making it async in this way we allow that different pull can act on data that has been updated in the mean while. Nothing in here should be async. I would revert this somehow already here.

discovery_engine_core/core/src/stack.rs
```diff
@@ -87,12 +87,11 @@ impl Stack {
...
     /// Updates the internal documents with the new one and returns an updated [`Stack`].
-    #[allow(dead_code)]
-    pub(crate) fn update<R: Ranker>(
-        mut self,
+    pub(crate) fn update(
+        &mut self,
         new_documents: &[Document],
-        ranker: &R,
-    ) -> Result<Self, Error> {
+        ranker: &impl Ranker,
```
acrrd, line 93: With the PR of the client we are returning `Article` from the api and we will use that to construct `Documents`.
I was thinking that we can do that in the `update` function instead after we already filter out some of them so we have run smbert on less documents.

discovery_engine_core/core/src/engine.rs
```diff
@@ -92,60 +97,118 @@ where
...
+            .collect::<FuturesUnordered<_>>()
+            .try_for_each(|(stack, documents)| async move {
+                stack
+                    .write()
+                    .await
+                    .update(&documents, ranker)
+                    .map_err(GenericError::from)
+            })
+            .await
+            .map_err(Into::into)
```
rustonaut, line 190: It might not matter for us (does it?) but this code can have race conditions.

Mainly two tasks might calculated a update for the same stack at the same time
and then write it. Whoever writes first will be overwritten by whoever writes second.

discovery_engine_core/core/src/stack.rs
janpetschexain: makes sense, although i'm not sure how this relates to this pr? i didn't touch the logic of `update()` here (just changed the receiver from `mut self` to `&mut self` as discussed beforehand).

discovery_engine_core/core/src/engine.rs
```diff
@@ -92,60 +97,118 @@ where
...
-        rank_stacks(self.stacks.values_mut(), &self.ranker)
+        rank_stacks(self.stacks.values(), &self.ranker).await
+    }
+
+    /// Updates the stacks with data related to the top key phrases of the current data.
+    #[allow(dead_code)]
+    #[allow(clippy::future_not_send)]
+    async fn update_stacks(&mut self, top: usize) -> Result<(), Error> {
+        let key_phrases = self.ranker.select_top_key_phrases(top);
+        let key_phrases = &key_phrases;
```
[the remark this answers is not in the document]
janpetschexain: in this case the compiler said, that it referenced a temporary value and that i should introduce a binding before referencing it.

discovery_engine_core/core/src/engine.rs
janpetschexain: within the method all stacks should be distinct (as they come from the hashmap) and they are updated individually. from the outside only a single task should be able to call the method at a time (as the receiver is `&mut self`). so i'm not sure if i can see where the race condition could come from?

in any case, i agree that we have to very carefully review the async code to avoid race conditions. for now i was told that we shouldn't spend too much time on this as it is more important to have a first running version.

discovery_engine_core/core/src/engine.rs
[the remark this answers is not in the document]
janpetschexain: nevermind, now it works, probably was some intermediate issue.

discovery_engine_core/core/src/engine.rs
rustonaut: With "it might not  matter for us" I meant exactly things like "the way we use this this situation can't happen".

discovery_engine_core/core/src/mab.rs
```diff
@@ -46,113 +58,115 @@ impl BetaSample for BetaSampler {
...
+            .collect::<Vec<_>>();
+        pin_mut!(buckets);
+
+        match buckets
+            .poll(cx)
+            .map(|buckets| pull_arms(&self.beta_sampler, buckets))
+        {
+            Poll::Ready(Some(Ok(index))) => {
+                let bucket = self.buckets[index].write();
+                pin_mut!(bucket);
```
rustonaut, line 156: Same as the other case it should be fine the way we use it.

But if this could be run concurrent there could be minor "off" results if the second run starts between releasing the read lock and gaining the write lock.

rustonaut (approved):
As far as I can tell all issues where addressed, and while new comments where added no changes are needed as far as I can tell (still check the new comment).

discovery_engine_core/core/src/engine.rs
```diff
@@ -72,85 +74,134 @@ where
...
 
-    if errors.is_empty() {
-        Ok(())
-    } else {
-        Err(Error::Errors(errors))
+    /// Updates the stacks with data related to the top key phrases of the current data.
+    #[allow(dead_code)]
+    async fn update_stacks(&mut self, top: usize) -> Result<(), Error> {
+        let key_phrases = &self.ranker.select_top_key_phrases(top);
+        let ranker = &self.ranker;
```
Robert-Steiner, line 182: since we now require `&mut` we could use `while let`

```rust
        let key_phrases = &self.ranker.select_top_key_phrases(top);
        let ranker = &mut self.ranker;

        while let Some(stack) = self.stacks.values().into_iter().next() {
            let (stack, documents) = stack
                .read()
                .await
                .ops
                .new_items(key_phrases, ranker)
                .await
                .map(|documents| (stack, documents))?;

            stack
                .write()
                .await
                .update(&documents, ranker)
                .map_err(GenericError::from)?
        }

        Ok(())
```

discovery_engine_core/core/src/mab.rs
```diff
@@ -46,113 +58,115 @@ impl BetaSample for BetaSampler {
...
         Self {
             beta_sampler,
-            buckets,
+            buckets: buckets.into_iter().collect(),
             bucket_type: PhantomData,
         }
     }
+
+    /// Selects up to n elements.
+    #[must_use = "futures do nothing unless they are awaited/polled"]
```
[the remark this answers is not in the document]
janpetschexain: ah ok, that's good to know that the lint is automatically aplied to a `impl Future`, i'll remove it.

discovery_engine_core/core/src/engine.rs
janpetschexain: this would make the loop synchronous again. currently we collect the "new items future" into the `FuturesUnordered` and run the async try_collect loop on those.

janpetschexain: my bad, now i get what you mean, we can't pass in `ranker` anyways anymore due to the owned reference. i'll check if we can work around this asynchronously in some way, otherwise i'll follow your suggestion.

Robert-Steiner: I'm not so sure if we're going to benefit that much from the async stuff either. Since almost every engine method requires `&mut`, we will only be able to process the engine calls sequentially and not concurrently. expect we are wrapping the engine in a mutex. But this is something that we can tackle later.

discovery_engine_core/core/src/engine.rs
acrrd: "it might not matter for us" and  "the way we use this this situation can't happen" have two really different meaning.
The method take a `&mut` so this code cannot have a race condition.

discovery_engine_core/core/src/mab.rs
```diff
@@ -46,113 +58,115 @@ impl BetaSample for BetaSampler {
...
 }
 
-impl<'bs, 'b, BS, B, T> SelectionIter<'bs, 'b, BS, B, T>
-where
-    B: Bucket<T>,
-{
-    fn new(beta_sampler: &'bs BS, buckets: Vec<&'b mut B>) -> Self {
+impl<'b, BS, B, T> Selection<'b, BS, B, T> {
+    /// Creates a selective steam.
+    pub(crate) fn new(beta_sampler: BS, buckets: impl IntoIterator<Item = &'b RwLock<B>>) -> Self {
```
acrrd, line 111: Can't we remove `RwLock` from this module completely and leave to the user to job to make the data in a format that the mab module can use?

discovery_engine_core/core/src/mab.rs
acrrd: This does not solve the original problem. When ask for k elements we want to extract them from the current state we no update in between.
Seems that we cannot download new items from the stack and get items from them at the same time with this approach, we might need to separate the operations from the data.

discovery_engine_core/core/src/mab.rs
acrrd: Having:
```rust
 pub async fn get_feed_documents(&self, max_documents: usize) -> Result<Vec<Document>, Error> {
        let mut stacks = Vec::with_capacity(self.stacks.len());
        for stack in self.stacks.values() {
            let stack = stack.write().await;
            stacks.push(stack);
        }

        Selection::new(BetaSampler)
            .select(stacks, max_documents)
            .map_err(|e| e.into())
}
```
and w.r.t the mab.rs that we have in main, changing:
```rust
fn pull_arms<B, T>(
    beta_sampler: &impl BetaSample,
    buckets: &mut [impl DerefMut<Target = B>],
) -> Result<T, Error>
```

```rust
struct SelectionIter<'bs, BS, D, B, T>
where
    D: DerefMut<Target = B>,
    B: Bucket<T>,
{
    beta_sampler: &'bs BS,
    buckets: Vec<D>,
    bucket_type: PhantomData<T>,
}
```

```rust
struct SelectionIter<'bs, BS, D, B, T>
where
    D: DerefMut<Target = B>,
    B: Bucket<T>,
{
    beta_sampler: &'bs BS,
    buckets: Vec<D>,
    bucket_type: PhantomData<T>,
}
```
and some other adjustment that derives from this we can have the mab work without async code and minimal changes.

discovery_engine_core/core/src/engine.rs
```diff
@@ -72,85 +74,134 @@ where
...
+                stack
+                    .read()
+                    .await
+                    .ops
+                    .new_items(key_phrases, ranker)
+                    .await
+                    .map(|documents| (stack, documents))
+            })
+            .collect::<FuturesUnordered<_>>()
+            .try_for_each(|(stack, documents)| async move {
```
acrrd, line 196: Here we want to do as in `rank_stacks` and collect all the errors and return them.

discovery_engine_core/core/src/engine.rs
```diff
@@ -72,85 +74,134 @@ where
...
-            .select(self.stacks.values_mut().collect(), max_documents)
+    pub async fn get_feed_documents(&self, max_documents: usize) -> Result<Vec<Document>, Error> {
+        Selection::new(BetaSampler, self.stacks.values())
+            .select(max_documents)
+            .await
             .map_err(|e| e.into())
     }
 
+    /// The ranker could rank the documents in a different order so we update the stacks with it.
+    async fn rank_stacks(&self) -> Result<(), Error> {
```
acrrd, line 134: Can't we leave the body of this in another function as it was? So to test it we don't have to create an engine.

acrrd:
In some places we could use `unfold` instead of collecting into a `FutureUnordered`
```rust
    stream::unfold(self.stacks.values(), |mut vals| async {
        let stack = vals.next()?;
        let stack = stack.write().await;
        Some((stack, vals))
    })
```

acrrd:
@janpetschexain @rustonaut I dismissed the approval because mab is still broken. I wanted to tag you there but failed.

discovery_engine_core/core/src/mab.rs
janpetschexain: i'll rewrite the code in a way that we acquire all write locks for all stacks (instead of all read locks and afterwards one write lock) and pop from the sampled stack before releasing any lock. this way we guarantee that the state hasn't changed, but on the other hand it is more restrictive of course (in the current context this should be ok).

discovery_engine_core/core/src/mab.rs
janpetschexain: my original idea was to have `stacks: Hashmap<StackId, RwLock<Stack>>` to allow concurrent operations on each stack, for example having a selection stream instead of a selection iter. i guess it makes sense to change it to `stacks: RwLock<HashMap<StackId, Stack>>` now, since we need to acquire all locks in many cases anyways. this will also make the code simpler/changes less, if we need more fine grained locks, we can introduce them again later on.

acrrd: What we would at the end is to be able to download the data from the api and run (part of?) the filters without locking the selection on the stack. For the mab we want to get all the `n` elements that are requested with the state we have when we start, so for this we need to lock all stacks first and get extract elements from that. I'm also fine with having an `RwLock` around the map at the moment and we can optimize it later. Also when we have the database this can also be removed and run the mab inside a transaction or something like that.

discovery_engine_core/core/src/engine.rs
janpetschexain: do we want to collect the errors from `new_items()` or from `update()` or from both?

acrrd: let's get everything. at the moment is our only way to tell outside that something happened

discovery_engine_core/core/src/mab.rs
```diff
@@ -46,108 +46,102 @@ impl BetaSample for BetaSampler {
...
-                } else {
-                    Ok(max)
-                }
-            },
-        )?
-        .1;
-
-    bucket.pop().ok_or(Error::EmptyBucket)
+    fn alpha(&self) -> f32 {
+        (&**self).alpha()
```
Robert-Steiner, line 65: would also work. same for `beta` and `is_emtpy`
```suggestion
        (**self).alpha()
```

Robert-Steiner (approved):
LGTM 👍

acrrd (approved):

janpetschexain:
bors merge

PR #87, opened 2022-01-31
[Merged by Bors] - TY-2320 basic ffi components. [1]

rustonaut:
**References**

**Summary**

Create some basic tools used by the follow up PRs: 

- UUID [TY-2322](https://xainag.atlassian.net/browse/TY-2320)
- Embedding1 [TY-2340](https://xainag.atlassian.net/browse/TY-2340)
- Duration [TY-2341](https://xainag.atlassian.net/browse/TY-2341)
- Document [TY-2342](https://xainag.atlassian.net/browse/TY-2342)
- TimeSpent [TY-2343](https://xainag.atlassian.net/browse/TY-2343)
- UserReacted [TY-2344](https://xainag.atlassian.net/browse/TY-2344)

In previous code we did translate dart types to C types and then the C types to rust types.

Due to limitations of dart (strong preference for heap allocated values and limitations in how you can abstract over alloc and free) this wasn't always a pleasant task. Changes took often long and even if the relatively simple abstractions like `CBoxedSlice` can easily introduce subtle soundness issues if not implemented very carefully.

This time I tried to avoid this problems by using a pattern which can sometimes be found in C of opaque pointers (to rust types) and C-FFI accessor function.

This allows us:
- to directly allocate in rust from dart and then (if wanted) move values over, avoiding problems with how dart implements alloc/free and generally making the handling of such data simple. It also means we can always use rust Drop and never have to manually free any nested structs (just call drop on the outermost type).
- have very small independent FFI functions on both sites, making reviews easier
- composing larger functions by composing already reviewed very small FFI functions, in dart compose the dart wrapper (e.g. on the `StringFfi` extension) not the rust FFI functions.
- have a common repeating pattern across (nearly all) types, making changes and additions easier
  - if it wouldn't be for limitations  of cbindgen we could probably generate half of all ffi functions easily, including most functions we likely will add in the future.
  

To show where this PR are going here is how document will later be implemneted (is already in a follow up PR):

```dart
factory Document.readNative(final Pointer<RustDocument> place) {
    return Document(
      id: DocumentIdFfi.readNative(ffi.get_document_id_ptr(place)),
      stackId: StackIdFfi.readNative(ffi.get_document_stack_id_ptr(place)),
      rank: ffi.get_document_rank_ptr(place).value,
      title: StringFfi.readNative(ffi.get_document_title_ptr(place)),
      snipped: StringFfi.readNative(ffi.get_document_snipped_ptr(place)),
      url: StringFfi.readNative(ffi.get_document_url_ptr(place)),
      domain: StringFfi.readNative(ffi.get_document_domain_ptr(place)),
      smbertEmbedding: Embedding1Ffi.readNative(
        ffi.get_document_smbert_embedding_ptr(place),
      ),
    );
  }

  void writeNative(final Pointer<RustDocument> place) {
    id.writeNative(ffi.get_document_id_ptr(place));
    stackId.writeNative(ffi.get_document_stack_id_ptr(place));
    ffi.get_document_rank_ptr(place).value = rank;
    title.writeNative(ffi.get_document_title_ptr(place));
    snipped.writeNative(ffi.get_document_snipped_ptr(place));
    url.writeNative(ffi.get_document_url_ptr(place));
    domain.writeNative(ffi.get_document_domain_ptr(place));
    smbertEmbedding.writeNative(ffi.get_document_smbert_embedding_ptr(place));
  }
```

Some notes:

- the `init_<type>_at` C-FFI functions will initialize a type at a given pointer position, e.g. initi the `title` field of the `Document``
- I use the word `place` to refer to a valid memory object (i.e. points to some allocated memory, with the right alignment etc.). Depending on context I might or might not expect it to be initialized.
- In later PR I for each new type 2 entries get added to `cbindgen.toml`
  - one line to the header like `typedef struct RustDuration RustDuration;` this will make `ffigen` treat rust types as opaque
  - one line for renaming e.g. `Duration` to `RustDuration` to avoid naming conflicts and confusion in the dart code
- at no place do I use `alloc` directly, hence I do at no place need to handle zero size specialization of alloc failing. Instead I use `Box`/`Vec` in combination with  `MaybeUninit<T>` (where necessary).
- only the rust functions are supposed to generate the various for Dart opaque pointers, we never generate null pointers!
- The `TY-2320-0-save-non-split-pr-upstream` branch contains the full, but by now likely somewhat out of date parts including all follow up PRs.


discovery_engine_core/bindings/cbindgen.toml
```diff
@@ -5,3 +5,27 @@ autogen_warning = "/* Warning, this file is autogenerated by cbindgen. Don't mod
 include_version = true
 sys_includes = ["stdint.h"]
 no_includes = true
+
+header = """
```
acrrd, line 9: We don't really need this on generated files.

rustonaut: The `header` is mostly there to add `typedef struct A A;` later one (which forces `ffigen` to treat it as an opaque type).

I can remove the license.

acrrd (commented):
Copyright year should be 2022.

discovery_engine/test/ffi/types/string_test.dart
```diff
@@ -0,0 +1,37 @@
...
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+import 'package:test/test.dart';
+import 'package:xayn_discovery_engine/src/ffi/load_lib.dart' show ffi;
+import 'package:xayn_discovery_engine/src/ffi/types/string.dart' show StringFfi;
+
+void main() {
+  test('parsing written empty string works', () {
```
acrrd, line 20: We are not parsing.

discovery_engine_core/bindings/cbindgen.toml
rustonaut: No.

Alias will generate an alias i.e. `typedef A B;` but we need something which acts as an opaque type definition. (note the missing `struct` part).

discovery_engine_core/bindings/src/types/string.rs
```diff
@@ -0,0 +1,132 @@
...
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+//! FFI functions for handling `String`
+
+use std::ptr;
+
+/// Creates a rust `String` with given capacity at given memory address.
+///
+/// The `String` is created at given address, not it's content.
```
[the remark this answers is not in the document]
rustonaut: For me yes, but I'm not always good at judging what is obvious for other people, e.g. for people unfamiliar with rust FFI.

discovery_engine_core/bindings/src/types/string.rs
```diff
@@ -0,0 +1,132 @@
...
+        ptr::write(place, s);
+    }
+    data_ptr
+}
+
+/// Sets the length of the rust `String` at given memory address.
+///
+/// Use this after you write to the strings data buffer behind
+/// `len` (but before it's end indicated by `capacity`) to make
+/// the newly written string data available to rust.
```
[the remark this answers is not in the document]
rustonaut: It kinda does, you can't set `len` to a value grater then capacity in rust. Through we probably should explicitly abort in that case

discovery_engine_core/bindings/src/types/slice.rs
```diff
@@ -0,0 +1,72 @@
...
+}
+
+/// Creates a `Box<[T]>` from a pointer to the first element and the slice len.
+pub(super) unsafe fn boxed_slice<T>(ptr: *mut T, len: usize) -> Box<[T]> {
+    unsafe { Box::from_raw(slice::from_raw_parts_mut(ptr, len)) }
+}
+
+/// Get length of an `Box<Vec<T>>`.
+#[allow(dead_code)]
+pub(super) unsafe fn get_boxed_vec_len<T>(ptr: *mut Vec<T>) -> usize {
```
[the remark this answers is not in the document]
rustonaut: Yes

discovery_engine_core/bindings/src/types/slice.rs
```diff
@@ -0,0 +1,72 @@
...
+    //SAFE: MaybeUninit doesn't need initialization
+    unsafe {
+        vec.set_len(len);
+    }
+    let boxed_slice = vec.into_boxed_slice();
+    Box::into_raw(boxed_slice).cast()
+}
+
+/// Creates a `Box<[T]>` from a pointer to the first element and the slice len.
+pub(super) unsafe fn boxed_slice<T>(ptr: *mut T, len: usize) -> Box<[T]> {
```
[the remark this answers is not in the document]
rustonaut: I think boxed slice the the more common terminology there. 

Std also uses it like e.g. `into_boxed_str`.

discovery_engine_core/bindings/src/types/boxed.rs
```diff
@@ -0,0 +1,38 @@
...
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+//! FFI functions for handling `Box`
+
+use std::mem::MaybeUninit;
+
+/// Allocates a box of a uninitialized `T` and returns the pointer to it.
+///
+/// Mostly used for testing `init_T_at` ffi functions
+#[cfg(feature = "additional-ffi-methods")]
```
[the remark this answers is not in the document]
rustonaut: The idea was to alter one disable the feature in release builds.

I can remove it.

discovery_engine_core/bindings/src/types/slice.rs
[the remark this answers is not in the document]
acrrd: ok

discovery_engine_core/bindings/src/types/slice.rs
```diff
@@ -0,0 +1,72 @@
...
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+//! Modules containing FFI glue for `Box<[u8]>`.
+
+use std::{mem::MaybeUninit, slice};
+
+/// Allocates a slice of (uninitialized) bytes of given length.
+pub(super) fn alloc_uninitialized_slice<T>(len: usize) -> *mut T {
+    let mut vec = Vec::<MaybeUninit<T>>::with_capacity(len);
```
acrrd, line 21: Coming back here from my question on the slice on f32 since it applies to all slices.
So seems that the main difference is that here we don't have a type to name for `place` because we cannot write `* mut &[T]` that is what we want. 
What about creating our own struct 
```rust
struct Slice<T> {
  ptr: *mut T,
  len: usize,
}
```
To represent a slice a make the interface uniform with String and Vec?

discovery_engine_core/bindings/src/types/string.rs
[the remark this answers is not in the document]
acrrd: we can we `set_len`, the doc says: "This is a low-level operation that maintains none of the normal invariants of the type."
And does not seems to panic: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=81977167f66edce0222cb87d6c894abe

discovery_engine_core/bindings/src/types/boxed.rs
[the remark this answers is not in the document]
acrrd: So it was to expose more functions to the dart tests but not to final build. 
That's a good idea but I tried to build the libs for android (merging all the TY-2320-* PR) and it save ~1k, so I'm not sure it is really worth it.

discovery_engine_core/bindings/src/types/string.rs
```diff
@@ -0,0 +1,132 @@
...
+    unsafe { &mut *string }.as_mut_ptr()
+}
+
+/// Alloc an uninitialized `Box<String>`, mainly used for testing.
+#[cfg(feature = "additional-ffi-methods")]
+#[no_mangle]
+pub extern "C" fn alloc_uninitialized_string_box() -> *mut String {
+    use super::boxed::alloc_uninitialized_box;
+
+    alloc_uninitialized_box()
```
janpetschexain, line 83: are there limitations which force us to import this locally? otherwise we could just import it at the top of the module as usual or we could call it here with the full path. same for `drop_string_box()`.
```suggestion
    super::boxed::alloc_uninitialized_box()
```

discovery_engine_core/bindings/src/types/slice.rs
```diff
@@ -0,0 +1,72 @@
...
+    let boxed_slice = vec.into_boxed_slice();
+    Box::into_raw(boxed_slice).cast()
+}
+
+/// Creates a `Box<[T]>` from a pointer to the first element and the slice len.
+pub(super) unsafe fn boxed_slice<T>(ptr: *mut T, len: usize) -> Box<[T]> {
+    unsafe { Box::from_raw(slice::from_raw_parts_mut(ptr, len)) }
+}
+
+/// Get length of an `Box<Vec<T>>`.
```
janpetschexain, line 35:
```suggestion
/// Get length of a `Box<Vec<T>>`.
```

discovery_engine_core/bindings/src/types/slice.rs
```diff
@@ -0,0 +1,72 @@
...
+#[allow(dead_code)]
+pub(super) unsafe fn get_boxed_vec_len<T>(ptr: *mut Vec<T>) -> usize {
+    unsafe { &*ptr }.len()
+}
+
+/// Get a pointer to the beginning of an `Box<Vec<T>>`
```
janpetschexain, line 41:
```suggestion
/// Get a pointer to the beginning of a `Box<Vec<T>>`'s buffer.
```

discovery_engine_core/bindings/src/types/slice.rs
```diff
@@ -0,0 +1,72 @@
...
+#[allow(dead_code)]
+pub(super) unsafe fn get_boxed_vec_buffer<T>(ptr: *mut Vec<T>) -> *mut T {
+    unsafe { &mut *ptr }.as_mut_ptr()
+}
+
+/// Increments the pointer by one element
```
janpetschexain, line 47:
```suggestion
/// Increments the pointer by one element.
```

discovery_engine_core/bindings/src/types/slice.rs
```diff
@@ -0,0 +1,72 @@
...
+/// # Safety
+///
+/// This is basically an alias for `ptr.offset(1)` and
+/// all safety constraints from `offset` apply.
+#[allow(dead_code)]
+pub(super) unsafe fn next_slice_element<T>(ptr: *mut T) -> *mut T {
+    unsafe { ptr.offset(1) }
+}
+
+/// Allocates an uninitialized slice of bytes
```
janpetschexain, line 58:
```suggestion
/// Allocates an uninitialized slice of bytes.
```

discovery_engine_core/bindings/src/types/slice.rs
```diff
@@ -0,0 +1,72 @@
...
+#[no_mangle]
+pub extern "C" fn alloc_uninitialized_f32_slice(len: usize) -> *mut f32 {
+    alloc_uninitialized_slice(len)
+}
+
+/// Drops a `Box<[u8]>`
```
janpetschexain, line 64: wrong type
```suggestion
/// Drops a `Box<[f32]>`.
```

discovery_engine_core/bindings/src/types/string.rs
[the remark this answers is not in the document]
rustonaut: Right the bounds check is a debug assertion only.

discovery_engine_core/bindings/src/types/slice.rs
rustonaut: This is not a `* mut &[T]` it's a `Box<[T]>`.

It's also meant to be directly called by the type specific ffi-exposed methods, in which case we can't return a `Slice` but only `Box<Slice>` which would just add additional complexity with no value.

In difference to other constructors which create a type at a specific place, this is meant to create a place we pass to rust.

discovery_engine_core/bindings/src/types/string.rs
rustonaut: It's only imported once and only if the feature which feature gates the function is enabled.

Hence rust-analyzer auto places it in the function.

I will switch it to prefix.

discovery_engine_core/bindings/src/types/boxed.rs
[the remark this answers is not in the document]
rustonaut: I removed it for now

discovery_engine_core/bindings/src/types/string.rs
```diff
@@ -0,0 +1,132 @@
...
+///
+/// Use this after you write to the strings data buffer behind
+/// `len` (but before it's end indicated by `capacity`) to make
+/// the newly written string data available to rust.
+///
+/// # Safety
+///
+/// The pointer must point to a valid `String` instance.
+#[no_mangle]
+pub unsafe extern "C" fn set_string_len(string_place: *mut String, len: usize) {
```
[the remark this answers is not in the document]
rustonaut: I will apply it to all pending and future PRs, too.

rustonaut:
Some review things I did/will apply to all future PRs:

- for now remove  the feature gate
- `alloc_`/`drop_` implies `Box` so e.g. `drop_string` insted of `drop_boxed_string`/`drop_string_box`, or `drop_f32_slice`. Methods which diverge will be explicit about this in their naming.
- use `place` only where we expect uninitialized memory objects

janpetschexain (approved):
lgtm

rustonaut:
borse merge

rustonaut:
bors merge

discovery_engine_core/bindings/src/types/slice.rs
acrrd: But instead of having pointer and len separate we can have a struct that holds them together like a String and Vec will do. We can also implement this using the Vec directly without having our own struct.

PR #88, opened 2022-01-31
[Merged by Bors] - TY-2341  duration [2.1]

rustonaut:
**References**

- [TY-2341](https://xainag.atlassian.net/browse/TY-2341)

**Preceded by**

**Parallel too**

- [TY-2322](https://xainag.atlassian.net/browse/TY-2322)
- [TY-2340](https://xainag.atlassian.net/browse/TY-2340)

**Followed by**

- [TY-2342](https://xainag.atlassian.net/browse/TY-2342)
- [TY-2343](https://xainag.atlassian.net/browse/TY-2343)
- [TY-2344](https://xainag.atlassian.net/browse/TY-2344)

**summary**

adds bindings for `Duration`

### DIFF
--- a/discovery_engine/lib/src/ffi/types/duration.dart
+++ b/discovery_engine/lib/src/ffi/types/duration.dart
@@ -1,0 +1,50 @@
+// Copyright 2021 Xayn AG
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, version 3.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+import 'dart:ffi' show Pointer;
+
+import 'package:xayn_discovery_engine/src/ffi/genesis.ffigen.dart'
+    show RustDuration;
+import 'package:xayn_discovery_engine/src/ffi/load_lib.dart' show ffi;
+
+extension DurationFfi on Duration {
+  void writeNative(final Pointer<RustDuration> durationPlace) {
+    final nanos = (inMicroseconds % Duration.microsecondsPerSecond) * 1000;
+    ffi.init_duration_at(durationPlace, inSeconds, nanos);
+  }
+
+  /// Reads a dart [Duration] from a rust duration.
+  ///
+  /// Be aware that darts [Duration] has both less precision and  is more
+  /// limited wrt. the max duration as it stores the duration as a `int`
+  /// of microseconds.
+  ///
+  /// Sub microseconds precision will be ignored.
+  ///
+  /// In case of to large durations an exception will be throw.
+  ///
+  static Duration readNative(final Pointer<RustDuration> durationPlace) {
+    const maxDurationSeconds = 0x8637bd05af6;
+    final seconds = ffi.get_duration_seconds(durationPlace);
+    if (seconds > maxDurationSeconds) {
+      throw ArgumentError.value(
+        seconds,
+        'seconds',
+        'duration is bigger then what dart Duration supports',
+      );
+    }
+    final microseconds = ffi.get_duration_nanos(durationPlace) ~/ 1000;
+    return Duration(seconds: seconds, microseconds: microseconds);
+  }
+}
--- a/discovery_engine/lib/src/ffi/types/duration.dart
+++ b/discovery_engine/lib/src/ffi/types/duration.dart
@@ -1,4 +1,4 @@
-// Copyright 2021 Xayn AG
+// Copyright 2022 Xayn AG
 //
 // This program is free software: you can redistribute it and/or modify
 // it under the terms of the GNU Affero General Public License as
@@ -27,15 +27,15 @@ extension DurationFfi on Duration {
   /// Reads a dart [Duration] from a rust duration.
   ///
   /// Be aware that darts [Duration] has both less precision and  is more
-  /// limited wrt. the max duration as it stores the duration as a `int`
+  /// limited wrt. the max duration as it stores the duration as an `int`
   /// of microseconds.
   ///
   /// Sub microseconds precision will be ignored.
   ///
-  /// In case of to large durations an exception will be throw.
-  ///
+  /// In case of a too large durations an exception will be throw.
   static Duration readNative(final Pointer<RustDuration> durationPlace) {
-    const maxDurationSeconds = 0x8637bd05af6;
+    const i64Max = 0x7fffffffffffffff;
+    const maxDurationSeconds = i64Max ~/ Duration.microsecondsPerSecond;
     final seconds = ffi.get_duration_seconds(durationPlace);
     if (seconds > maxDurationSeconds) {
       throw ArgumentError.value(
--- a/discovery_engine/test/ffi/types/duration_test.dart
+++ b/discovery_engine/test/ffi/types/duration_test.dart
@@ -1,4 +1,4 @@
-// Copyright 2022 Xayn AG
+// Copyright 2021 Xayn AG
 //
 // This program is free software: you can redistribute it and/or modify
 // it under the terms of the GNU Affero General Public License as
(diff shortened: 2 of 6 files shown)


acrrd:
Can't we encoded duration as a number since we are interested only in seconds and avoid pointers?

rustonaut:
> Can't we encoded duration as a number since we are interested only in seconds and avoid pointers?

We can't avoid pointer as we write to a field an an struct. Even with just a `u32` there would be an pointer.

We also decided to use `Duration` in rust, not `u32`/`u64`.

acrrd:
I changed the base from `main` the the branch of the base PR

discovery_engine/lib/src/ffi/types/duration.dart
```diff
@@ -0,0 +1,50 @@
...
+
+  /// Reads a dart [Duration] from a rust duration.
+  ///
+  /// Be aware that darts [Duration] has both less precision and  is more
+  /// limited wrt. the max duration as it stores the duration as a `int`
+  /// of microseconds.
+  ///
+  /// Sub microseconds precision will be ignored.
+  ///
+  /// In case of to large durations an exception will be throw.
```
acrrd, line 35:
```suggestion
  /// In case of a too large durations an exception will be throw.
```

discovery_engine/lib/src/ffi/types/duration.dart
```diff
@@ -0,0 +1,50 @@
...
+  ///
+  static Duration readNative(final Pointer<RustDuration> durationPlace) {
+    const maxDurationSeconds = 0x8637bd05af6;
```
acrrd, line 38: How this was calculated? 
I did `(2^53) - 1 / 1000000` and I get `9007199254` while yours is 3 magnitude larger then this (`9223372036854`).
Instead of this opaque number is better to put the calculation as an expression.

discovery_engine_core/bindings/src/types/duration.rs
```diff
@@ -0,0 +1,99 @@
...
+
+#[cfg(test)]
+mod tests {
+    use rand::random;
+
+    use super::*;
+
+    #[test]
+    fn test_reading_duration() {
+        let place = &mut Duration::new(random(), random());
```
acrrd, line 81: There is no need to use random here. If we want this kind of tests we can use to appropriate tools and libraries.

discovery_engine_core/bindings/src/types/duration.rs
```diff
@@ -0,0 +1,99 @@
...
+    }
+}
+
+/// Gets the seconds of a duration at given place.
+///
+/// # Safety
+///
+/// The pointer must point to a valid [`Duration`] instance.
+#[no_mangle]
+pub unsafe extern "C" fn get_duration_seconds(place: *mut Duration) -> u64 {
```
acrrd, line 37: Here and in following we can use `duration` and not `place`.

discovery_engine_core/bindings/src/types/duration.rs
```diff
@@ -0,0 +1,99 @@
...
+/// The pointer must point to a valid [`Duration`] instance.
+#[no_mangle]
+pub unsafe extern "C" fn get_duration_nanos(place: *mut Duration) -> u32 {
+    unsafe { &*place }.subsec_nanos()
+}
+
+/// Alloc an uninitialized `Box<Duration>`, mainly used for testing.
+#[cfg(feature = "additional-ffi-methods")]
+#[no_mangle]
+pub extern "C" fn alloc_uninitialized_duration_box() -> *mut Duration {
```
acrrd, line 54: We can drop the `box` suffix.

discovery_engine_core/bindings/src/types/duration.rs
```diff
@@ -0,0 +1,99 @@
...
+}
+
+/// Drops a `Box<Duration>`, mainly used for testing.
+///
+/// # Safety
+///
+/// The pointer must represent a valid `Box<Duration>` instance.
+#[cfg(feature = "additional-ffi-methods")]
+#[no_mangle]
+pub unsafe extern "C" fn drop_duration_box(boxed: *mut Duration) {
```
acrrd, line 67: `boxed` -> `duration` and as before we can drop `box` in the function of the name.

acrrd (commented):
Copyright year should be 2022

discovery_engine/test/ffi/types/duration_test.dart
```diff
@@ -0,0 +1,29 @@
...
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+import 'package:test/test.dart';
+import 'package:xayn_discovery_engine/src/ffi/load_lib.dart' show ffi;
+import 'package:xayn_discovery_engine/src/ffi/types/duration.dart'
+    show DurationFfi;
+
+void main() {
+  test('parsing written duration yields same result', () {
```
acrrd, line 21: We are not parsing.

discovery_engine_core/bindings/src/types/duration.rs
rustonaut: No we can't drop box here as it must be a box and no other form.

discovery_engine_core/bindings/src/types/duration.rs
rustonaut: but we alloc a box of an `Duration`, not something else.

discovery_engine/lib/src/ffi/types/duration.dart
rustonaut: It's native code the the ints are `i64`s.

acrrd: Ah true, this code cannot be used on the browser!

discovery_engine_core/bindings/src/types/duration.rs
acrrd: But the user cannot see it. They get a pointer to `Duration` and they don't care how it was allocated.

discovery_engine/lib/src/ffi/types/duration.dart
janpetschexain: unfortunately dart doesn't have an equivalent to `i64::MAX`, but we could write it like
```suggestion
    const i64Max = 0x7fffffffffffffff;
    const maxDurationSeconds = i64Max ~/ Duration.microsecondsPerSecond;
```

discovery_engine_core/bindings/src/types/duration.rs
```diff
@@ -0,0 +1,99 @@
...
+
+/// Initializes a `Duration` field at given place.
+///
+/// # Safety
+///
+/// It must be valid to write a [`Duration`] to given pointer.
+#[no_mangle]
+pub unsafe extern "C" fn init_duration_at(place: *mut Duration, seconds: u64, nanos: u32) {
+    unsafe {
+        place.write(Duration::new(seconds, nanos));
```
janpetschexain, line 27: related to

> > Can't we encoded duration as a number since we are interested only in seconds and avoid pointers?
>
> We can't avoid pointer as we write to a field of a struct. Even with just a u32 there would be an pointer.
>
> We also decided to use Duration in rust, not u32/u64.

if we only expect seconds granularity from dart, we could simplify the ffi api a bit and still get the benefits of the `Duration` api internally in rust

```suggestion
pub unsafe extern "C" fn init_duration_at(place: *mut Duration, seconds: u64) {
    unsafe {
        place.write(Duration::from_secs(seconds));
```

then we could also remove `get_duration_nanos()`.

discovery_engine_core/bindings/src/types/duration.rs
rustonaut: They do have to care, because if this is returned to dart (by a rust ffi function) dart still needs to call the right drop method (because we returned it to dart).

And who says that we don't maybe at some point also return some `Arc<T>`'s in some future API in which case for some types there would be two drop methods both taking a `*mut T`.

discovery_engine_core/bindings/src/types/duration.rs
rustonaut: yes, I considered that, but I preferred to have a dart Duration to rust Duration type we write once and then can use anywhere we need a duration without touching it again.

It also means that we will have an leaky abstraction which seem to do one thing (api for passing a dart/rust Duration through the ffi) but does another thing (API for truncating duration to seconds precision and passing them from/to dart).

discovery_engine_core/bindings/src/types/duration.rs
rustonaut: I know it's not a prop/fuzz test. But I also don't want to bother choosing values.

discovery_engine_core/bindings/src/types/duration.rs
rustonaut: Now `alloc_`/`drop_` imply box, so the prefix is removed.

discovery_engine_core/bindings/src/types/duration.rs
rustonaut: Now `alloc_`/`drop_` imply box, so the suffix is removed.

janpetschexain (approved):
lgtm

rustonaut:
bors merge

PR #90, opened 2022-01-31
[Merged by Bors] - TY-2340  embedding bindings [2.3]

rustonaut:
**References**

- [TY-2340](https://xainag.atlassian.net/browse/TY-2340)

**Preceded by**

**Parallel too**

- [TY-2322](https://xainag.atlassian.net/browse/TY-2322)
- [TY-2341](https://xainag.atlassian.net/browse/TY-2341)

**Followed by**

- [TY-2342](https://xainag.atlassian.net/browse/TY-2342)
- [TY-2343](https://xainag.atlassian.net/browse/TY-2343)
- [TY-2344](https://xainag.atlassian.net/browse/TY-2344)

**summary**

adds bindings for `Embedding1`

### DIFF
--- a/discovery_engine/lib/src/ffi/types/embedding.dart
+++ b/discovery_engine/lib/src/ffi/types/embedding.dart
@@ -16,24 +16,24 @@ import 'dart:ffi' show Pointer, FloatPointer;
 import 'dart:typed_data' show Float32List;
 
 import 'package:xayn_discovery_engine/src/ffi/genesis.ffigen.dart'
-    show RustEmbedding1;
+    show RustEmbedding;
 import 'package:xayn_discovery_engine/src/ffi/load_lib.dart' show ffi;
 
-extension Embedding1Ffi on Float32List {
+extension EmbeddingFfi on Float32List {
   void writeNative(
-    final Pointer<RustEmbedding1> place,
+    final Pointer<RustEmbedding> place,
   ) {
     final len = length;
     final buffer = ffi.alloc_uninitialized_f32_slice(len);
     buffer.asTypedList(len).setAll(0, this);
-    ffi.init_embedding1_at(place, buffer, len);
+    ffi.init_embedding_at(place, buffer, len);
   }
 
   static Float32List readNative(
-    final Pointer<RustEmbedding1> place,
+    final Pointer<RustEmbedding> place,
   ) {
-    final len = ffi.get_embedding1_buffer_len(place);
-    final data = ffi.get_embedding1_buffer(place).asTypedList(len);
+    final len = ffi.get_embedding_buffer_len(place);
+    final data = ffi.get_embedding_buffer(place).asTypedList(len);
     return Float32List.fromList(data);
   }
 }
--- a/discovery_engine/lib/src/ffi/types/embedding.dart
+++ b/discovery_engine/lib/src/ffi/types/embedding.dart
@@ -1,4 +1,4 @@
-// Copyright 2021 Xayn AG
+// Copyright 2022 Xayn AG
 //
 // This program is free software: you can redistribute it and/or modify
 // it under the terms of the GNU Affero General Public License as
--- a/discovery_engine/lib/src/ffi/types/embedding.dart
+++ b/discovery_engine/lib/src/ffi/types/embedding.dart
@@ -1,0 +1,39 @@
+// Copyright 2021 Xayn AG
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, version 3.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+import 'dart:ffi' show Pointer, FloatPointer;
+import 'dart:typed_data' show Float32List;
+
+import 'package:xayn_discovery_engine/src/ffi/genesis.ffigen.dart'
+    show RustEmbedding1;
+import 'package:xayn_discovery_engine/src/ffi/load_lib.dart' show ffi;
+
+extension Embedding1Ffi on Float32List {
+  void writeNative(
+    final Pointer<RustEmbedding1> place,
+  ) {
+    final len = length;
+    final buffer = ffi.alloc_uninitialized_f32_slice(len);
+    buffer.asTypedList(len).setAll(0, this);
+    ffi.init_embedding1_at(place, buffer, len);
+  }
+
+  static Float32List readNative(
+    final Pointer<RustEmbedding1> place,
+  ) {
+    final len = ffi.get_embedding1_buffer_len(place);
+    final data = ffi.get_embedding1_buffer(place).asTypedList(len);
+    return Float32List.fromList(data);
+  }
+}
--- a/discovery_engine/test/ffi/types/embedding_test.dart
+++ b/discovery_engine/test/ffi/types/embedding_test.dart
@@ -1,4 +1,4 @@
-// Copyright 2021 Xayn AG
+// Copyright 2022 Xayn AG
 //
 // This program is free software: you can redistribute it and/or modify
 // it under the terms of the GNU Affero General Public License as
@@ -20,22 +20,22 @@ import 'package:xayn_discovery_engine/src/ffi/types/embedding.dart'
     show Embedding1Ffi;
 
 void main() {
-  test('parsing written empty embeddings works', () {
+  test('reading written empty embeddings works', () {
     final embedding = Float32List(0);
-    final place = ffi.alloc_uninitialized_embedding1_box();
+    final place = ffi.alloc_uninitialized_embedding1();
     embedding.writeNative(place);
     final res = Embedding1Ffi.readNative(place);
-    ffi.drop_embedding1_box(place);
+    ffi.drop_embedding1(place);
     expect(res, equals(embedding));
   });
 
-  test('parsing written embeddings yields same result', () {
+  test('reading written embeddings yields same result', () {
     final embedding =
         Float32List.fromList([18.4, 6.9, 13.2, 7.8945, 8.2, 0.3, 7.8, 9.479]);
-    final place = ffi.alloc_uninitialized_embedding1_box();
+    final place = ffi.alloc_uninitialized_embedding1();
     embedding.writeNative(place);
     final res = Embedding1Ffi.readNative(place);
-    ffi.drop_embedding1_box(place);
+    ffi.drop_embedding1(place);
     expect(res, equals(embedding));
   });
 }
--- a/discovery_engine/test/ffi/types/embedding_test.dart
+++ b/discovery_engine/test/ffi/types/embedding_test.dart
@@ -1,0 +1,41 @@
+// Copyright 2021 Xayn AG
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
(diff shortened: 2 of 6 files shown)


acrrd:
Changed the base to  `TY-2320-1-basic-utilities`

discovery_engine/lib/src/ffi/types/embedding.dart
```diff
@@ -0,0 +1,39 @@
...
+import 'package:xayn_discovery_engine/src/ffi/genesis.ffigen.dart'
+    show RustEmbedding1;
+import 'package:xayn_discovery_engine/src/ffi/load_lib.dart' show ffi;
+
+extension Embedding1Ffi on Float32List {
+  void writeNative(
+    final Pointer<RustEmbedding1> place,
+  ) {
+    final len = length;
+    final buffer = ffi.alloc_uninitialized_f32_slice(len);
```
acrrd, line 27: Why we can't use the same approach we used for strings?

discovery_engine/test/ffi/types/embedding_test.dart
```diff
@@ -0,0 +1,41 @@
...
+
+import 'dart:typed_data';
+
+import 'package:test/test.dart';
+import 'package:xayn_discovery_engine/src/ffi/load_lib.dart' show ffi;
+import 'package:xayn_discovery_engine/src/ffi/types/embedding.dart'
+    show Embedding1Ffi;
+
+void main() {
+  test('parsing written empty embeddings works', () {
```
acrrd, line 23: not parsing.

discovery_engine_core/bindings/src/types/embedding.rs
```diff
@@ -0,0 +1,135 @@
...
+
+use core::document::{Embedding, Embedding1};
+
+use ndarray::Array;
+
+use crate::types::slice::boxed_slice;
+
+/// Creates a rust `Embedding1` with given capacity at given memory address.
+///
+/// The `Embedding1` is created at given address, not it's content.
```
acrrd, line 25: Isn't this obvious by the type?

discovery_engine_core/bindings/src/types/embedding.rs
```diff
@@ -0,0 +1,135 @@
...
+///
+/// Returns a pointer to the begin of the buffer of the embedding.
+///
```
acrrd, line 28:
```suggestion
```

discovery_engine_core/bindings/src/types/embedding.rs
```diff
@@ -0,0 +1,135 @@
...
+/// # Safety
+///
+/// - It must be valid to write an `Embedding1` instance to given pointer.
+/// - The passed in slice must represent a `Box<[f32]>` and transfers ownership,
+///   it must be fully initialized.
+#[no_mangle]
+pub unsafe extern "C" fn init_embedding1_at(
+    place: *mut Embedding1,
+    float_slice: *mut f32,
```
acrrd, line 37: This is not a slice, a slice is pointer plus len.

discovery_engine_core/bindings/src/types/embedding.rs
```diff
@@ -0,0 +1,135 @@
...
+    }
+}
+
+/// Returns a pointer to the begin of the `[f32]` backing the `Embedding1`
+///
+/// # Safety
+///
+/// The pointer must point to a valid [`Embedding1`] instance.
+#[no_mangle]
+pub unsafe extern "C" fn get_embedding1_buffer(place: *mut Embedding1) -> *mut f32 {
```
acrrd, line 53:
```suggestion
pub unsafe extern "C" fn get_embedding1_buffer(embedding: *mut Embedding1) -> *mut f32 {
```

discovery_engine_core/bindings/src/types/embedding.rs
```diff
@@ -0,0 +1,135 @@
...
+    let embedding = unsafe { &mut *place };
+    embedding.0.as_mut_ptr()
+}
+
+/// Returns len of given embeddings buffer.
```
acrrd, line 58:
```suggestion
/// Returns len of the given embedding.
```

discovery_engine_core/bindings/src/types/embedding.rs
```diff
@@ -0,0 +1,135 @@
...
+///
+/// # Safety
+///
+/// The pointer must point to a valid [`Embedding1`] instance.
+#[no_mangle]
+pub unsafe extern "C" fn get_embedding1_buffer_len(place: *mut Embedding1) -> usize {
+    //WARNING: This holds for `Embedding1` but not for all possible `ndarray::Array<...>`.
```
acrrd, line 65: What this means?

discovery_engine_core/bindings/src/types/embedding.rs
```diff
@@ -0,0 +1,135 @@
...
+    unsafe { &*place }.len()
+}
+
+/// Alloc an uninitialized `Box<String>`, mainly used for testing.
```
acrrd, line 69:
```suggestion
/// Alloc an uninitialized `Box<Embedding1>`, mainly used for testing.
```

discovery_engine_core/bindings/src/types/embedding.rs
```diff
@@ -0,0 +1,135 @@
...
+    embedding.0.as_mut_ptr()
+}
+
+/// Returns len of given embeddings buffer.
+///
+/// # Safety
+///
+/// The pointer must point to a valid [`Embedding1`] instance.
+#[no_mangle]
+pub unsafe extern "C" fn get_embedding1_buffer_len(place: *mut Embedding1) -> usize {
```
acrrd, line 64:
```suggestion
pub unsafe extern "C" fn get_embedding1_buffer_len(embedding: *mut Embedding1) -> usize {
```

discovery_engine_core/bindings/src/types/embedding.rs
```diff
@@ -0,0 +1,135 @@
...
+    //WARNING: This holds for `Embedding1` but not for all possible `ndarray::Array<...>`.
+    unsafe { &*place }.len()
+}
+
+/// Alloc an uninitialized `Box<String>`, mainly used for testing.
+#[cfg(feature = "additional-ffi-methods")]
+#[no_mangle]
+pub extern "C" fn alloc_uninitialized_embedding1_box() -> *mut Embedding1 {
```
acrrd, line 72: We can drop box

discovery_engine_core/bindings/src/types/embedding.rs
```diff
@@ -0,0 +1,135 @@
...
+}
+
+/// Drops a `Box<String>`, mainly used for testing.
+///
+/// # Safety
+///
+/// The pointer must represent a valid `Box<Embedding1>` instance.
+#[cfg(feature = "additional-ffi-methods")]
+#[no_mangle]
+pub unsafe extern "C" fn drop_embedding1_box(boxed: *mut Embedding1) {
```
acrrd, line 85:
```suggestion
pub unsafe extern "C" fn drop_embedding1_box(embedding: *mut Embedding1) {
```

discovery_engine_core/bindings/src/types/embedding.rs
```diff
@@ -0,0 +1,135 @@
...
+
+    unsafe { drop_box(boxed) };
+}
+
+#[cfg(test)]
+mod tests {
+    use std::{ptr, slice};
+
+    use ndarray::arr1;
+    use rand::{distributions::Standard, prelude::Distribution, thread_rng, Rng};
```
acrrd, line 96: We don't need to this tests. If we want to proper random testing we need the proper library.

acrrd (commented):
Copyright year must be 2022

discovery_engine/lib/src/ffi/types/embedding.dart
rustonaut: because `ndarray::Array` has I think no `set_len` method (I think, it would conflict with it's Dimension design).

but we could add methods for `Box<[u8]>` and then use this approach for `String`.

discovery_engine_core/bindings/src/types/embedding.rs
rustonaut: Changed it, as a side note for  `ndarray::Array` the array len and arrays buffer len might not be the same (through in this case they are guaranteed to be the same).

discovery_engine_core/bindings/src/types/embedding.rs
rustonaut: For ndarray::Array the array len and arrays buffer len might not be the same (through in this case they are guaranteed to be the same).

The dimension parameter in an `Array` can have a strides which can both skip and overlap the buffer.

Like you can(1) have a buffer of one element but a huge multidimensional array. Just by having a stride of 0 all elements refer to the same element.

(1): Actually I'm not 100% sure if that exact example works with the current `ndarray` interface/library, though it does work with the technology they use and is the reason why some methods returns options where you wouldn't expect it.

discovery_engine_core/bindings/src/types/embedding.rs
rustonaut: yes, but also choosing arbitrary values is just annoying IMHO.

discovery_engine_core/bindings/src/types/embedding.rs
```diff
@@ -0,0 +1,119 @@
...
+        let place: &mut Embedding = &mut Embedding::from(arr1(&[]));
+        unsafe {
+            let data_ptr = alloc_uninitialized_f32_slice(len);
+            ptr::copy(embedding.as_ptr(), data_ptr, len);
+            init_embedding_at(place, data_ptr, len);
+        }
+        assert_eq!(*place, embedding);
+    }
+
+    fn arbitrary_embeddig() -> Embedding {
```
janpetschexain, line 116:
```suggestion
    fn arbitrary_embedding() -> Embedding {
```

discovery_engine_core/bindings/src/types/embedding.rs
```diff
@@ -0,0 +1,119 @@
...
+}
+
+/// Returns len of the given embedding.
+///
+/// # Safety
+///
+/// The pointer must point to a valid [`Embedding`] instance.
+#[no_mangle]
+pub unsafe extern "C" fn get_embedding_buffer_len(embedding: *mut Embedding) -> usize {
+    //WARNING: This holds for `Embedding` but not for all possible `ndarray::Array<...>`.
```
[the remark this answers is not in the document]
rustonaut: In our case `Embedding` only can have cases where `len` == "buffer len".

But it's a good point that we should handle that more explicitly/put in some fail safe.

(I had the misconception that this constraints is implicitly statically encoded in `Embedding`, and hence this isn't needed, but it is not.)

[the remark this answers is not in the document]
rustonaut: I changed it to:

`unsafe { &*embedding }.as_slice().unwrap_or_else(|| abort()).len()`

with following note:

```rust
/// # Aborts
///
/// Aborts if the embedding is not "contiguous and in standard order".
///
//Note: Wether this is or isn't the case should (for our use case) be always
// the same independent of input data. Hence this should be caught by
// tests. If that isn't the case anymore it should be considered to
// change this interface, e.g. to support reading a buffer with strides.
```

Through this note is based on some older understanding of our algorithms and they have changed since then.
Maybe we should just teach dart to parse strides?

discovery_engine_core/bindings/src/types/embedding.rs
```diff
@@ -0,0 +1,130 @@
...
+/// Returns len of the given embedding.
+///
+/// # Safety
+///
+/// The pointer must point to a valid [`Embedding`] instance.
+///
+/// # Aborts
+///
+/// Aborts if the embedding is not "contiguous and in standard order".
+//Note: Wether this is or isn't the case should (for our use case) be always
```
janpetschexain, line 63:
```suggestion
//Note: Whether this is or isn't the case should (for our use case) be always
```

janpetschexain (approved):
👌

rustonaut:
bors merge

PR #95, opened 2022-02-01
[Merged by Bors] - TY-2304 create engine from config

finiteprods:
**Summary**

adds a `Config` struct and an `Engine` constructor method `from_config`, which initializes the stacks.
it takes a `Ranker` for now, but this will be constructed as well later.

### DIFF
--- a/discovery_engine_core/core/src/engine.rs
+++ b/discovery_engine_core/core/src/engine.rs
@@ -67,8 +67,12 @@ pub struct Config {
     api_key: String,
     api_base_url: String,
     markets: Vec<Market>,
-    smbert_vocab: Vec<u8>,
-    smbert_model: Vec<u8>,
+    smbert_vocab: String,
+    smbert_model: String,
+    kpe_vocab: String,
+    kpe_model: String,
+    kpe_cnn: String,
+    kpe_classifier: String,
 }
 
 /// Discovery Engine.
--- a/discovery_engine_core/core/src/engine.rs
+++ b/discovery_engine_core/core/src/engine.rs
@@ -84,12 +84,21 @@ where
     R: Ranker,
 {
     /// Creates a new `Engine` from configuration.
-    pub fn from_config(config: Config, ranker: R) -> Self {
-        Self {
+    pub fn from_config(config: Config, ranker: R, stack_ops: Vec<BoxedOps>) -> Result<Self, Error> {
+        let stacks = stack_ops
+            .into_iter()
+            .map(|ops| {
+                let id = ops.id();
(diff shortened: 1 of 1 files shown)


discovery_engine_core/core/src/engine.rs
```diff
@@ -63,11 +82,25 @@ impl<R> Engine<R>
...
+    }
+
     /// Creates a new `Engine` from serialized state and stack operations.
     ///
     /// The `Engine` only keeps in its state data related to the current [`BoxedOps`].
     /// Data related to missing operations will be dropped.
-    pub fn new(state: &[u8], ranker: R, stacks_ops: Vec<BoxedOps>) -> Result<Self, Error> {
+    pub fn new(
+        state: &[u8],
+        config: Config, // FIXME part of state?
```
acrrd, line 100: no

discovery_engine_core/core/src/engine.rs
```diff
@@ -63,11 +82,25 @@ impl<R> Engine<R>
 where
     R: Ranker,
 {
+    /// Creates a new `Engine` from configuration.
+    pub fn from_config(config: Config, ranker: R) -> Self {
+        Self {
+            config,
+            stacks: HashMap::new(), // FIXME check
```
acrrd, line 89: We always need stack_ops otherwise we have no stack.

discovery_engine_core/core/src/engine.rs
finiteprods: ok, so the `serialize` method stays exactly as it is.

discovery_engine_core/core/src/engine.rs
```diff
@@ -53,8 +53,27 @@ pub enum Error {
...
+}
+
+#[allow(dead_code)]
+/// Discovery Engine configuration settings.
+pub struct Config {
+    api_key: String,
+    api_base_url: String,
+    markets: Vec<Market>,
+    smbert_vocab: Vec<u8>,
+    smbert_model: Vec<u8>,
```
Robert-Steiner, line 70: here we also need the kpe assets (vocab, model, cnn, classifer)
https://github.com/xaynetwork/xayn_discovery_engine/pull/93#issuecomment-1026717249

Robert-Steiner, line 70: The type for all smbert/kpe assets can be something that implements `AsRef<Path>` for example `String` or `PathBuf`

Robert-Steiner (approved):
for me it looks good 👍

acrrd (approved):

finiteprods:
bors merge

PR #104, opened 2022-02-03
[Merged by Bors] - TY-2351 ffi user reaction [3.1]

rustonaut:
Based on #90

Needed for [TY-2344] and [TY-2343].

Adds ffi support for `UserRection` which seems to be called `DocumentFeedback` in dart.

[TY-2344]: https://xainag.atlassian.net/browse/TY-2344?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[TY-2343]: https://xainag.atlassian.net/browse/TY-2343?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine/lib/src/domain/models/document.dart
+++ b/discovery_engine/lib/src/domain/models/document.dart
@@ -15,7 +15,7 @@
 import 'package:hive/hive.dart'
     show HiveType, HiveField, TypeAdapter, BinaryReader, BinaryWriter;
 import 'package:json_annotation/json_annotation.dart'
-    show $enumDecode, JsonEnum, JsonSerializable, JsonValue;
+    show $enumDecode, JsonEnum, JsonValue;
 import 'package:xayn_discovery_engine/src/api/models/document.dart' as api;
 import 'package:xayn_discovery_engine/src/domain/models/news_resource.dart'
     show NewsResource;
--- a/discovery_engine/lib/src/domain/models/document.dart
+++ b/discovery_engine/lib/src/domain/models/document.dart
@@ -14,7 +14,8 @@
 
 import 'package:hive/hive.dart'
     show HiveType, HiveField, TypeAdapter, BinaryReader, BinaryWriter;
-import 'package:json_annotation/json_annotation.dart' show JsonValue;
+import 'package:json_annotation/json_annotation.dart'
+    show JsonSerializable, JsonValue, $enumDecode;
 import 'package:xayn_discovery_engine/src/api/models/document.dart' as api;
 import 'package:xayn_discovery_engine/src/domain/models/news_resource.dart'
     show NewsResource;
@@ -22,6 +23,8 @@ import 'package:xayn_discovery_engine/src/domain/models/unique_id.dart'
     show DocumentId, StackId;
 import 'package:xayn_discovery_engine/src/domain/repository/type_id.dart'
     show documentTypeId, userReactionTypeId;
+import 'package:xayn_discovery_engine/src/ffi/genesis.ffigen.dart'
+    show RustUserReaction;
 
 part 'document.g.dart';
 
@@ -71,13 +74,27 @@ class Document {
 /// essentially if the user "liked" or "disliked" the document.
 @HiveType(typeId: userReactionTypeId)
 enum UserReaction {
-  @JsonValue(0)
-  @HiveField(0)
+  @JsonValue(RustUserReaction.Neutral)
+  @HiveField(RustUserReaction.Neutral)
   neutral,
-  @JsonValue(1)
-  @HiveField(1)
+  @JsonValue(RustUserReaction.Positive)
+  @HiveField(RustUserReaction.Positive)
   positive,
-  @JsonValue(2)
-  @HiveField(2)
+  @JsonValue(RustUserReaction.Negative)
+  @HiveField(RustUserReaction.Negative)
   negative,
+}
+
+extension UserReactionIntConversion on UserReaction {
+  int toIntRepr() => _$UserReactionEnumMap[this]!;
+  static UserReaction fromIntRepr(int intRepr) =>
+      $enumDecode(_$UserReactionEnumMap, intRepr);
+}
+
+// Without this `_$DocumentFeedbackEnumMap` is not generated.
+@JsonSerializable()
+class _ForceGenerationOfUserReactionMapping {
+  final DocumentFeedback feedback;
+
+  _ForceGenerationOfUserReactionMapping(this.feedback);
 }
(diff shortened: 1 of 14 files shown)


discovery_engine_core/bindings/src/types/document/user_reaction.rs
```diff
@@ -0,0 +1,113 @@
...
+
+/// Gets the int representation of an [`UserReaction`].
+///
+/// # Safety
+///
+/// The pointer must point to a valid [`UserReaction`] instance.
+#[no_mangle]
+pub unsafe extern "C" fn get_user_reaction(reaction: *mut UserReaction) -> IntRepr {
+    let reaction = unsafe { &*reaction };
+    match reaction {
```
janpetschexain, line 52: this could be inlined without the readability being reduced
```suggestion
    match unsafe { &*reaction } {
```

discovery_engine/lib/src/ffi/types/document/feedback.dart
```diff
@@ -0,0 +1,65 @@
...
+    final reaction = ffi.get_user_reaction(place);
+    switch (reaction) {
+      case 0:
+        return DocumentFeedback.neutral;
+      case 1:
+        return DocumentFeedback.positive;
+      case 2:
+        return DocumentFeedback.negative;
+    }
+    throw ArgumentError.value(
```
janpetschexain, line 59: since we have to do the mapping on the dart side anyways, why don't we simply put a `#[repr(u8)]` (or a `#[repr(IntRepr)]` if we want to stick with the alias) at the rust `UserReaction` (which only has fieldless variants anyways) and export it in the ffi? this would give us a bunch of named constants which we could use in the mappings, and the ffi functions could take a `UserReaction` value directly instead of a `*mut UserReaction` pointer (which would make the alloc/init/drop ffi unnecessary).

rustonaut: You still need a `*mut UserReaction` as outside of testing you are basically always write the enum to a field.

Though you could merge the field addr + getter/setter, but that is less reusable/composable and less consistent with he other fields.

The `alloc_`/`drop_` tests are used by dart tests only, so we are not really saving anything there.

rustonaut: I didn't make `UserReaction` non-opaque, as then cbindgen has to spider additional crates.

rustonaut: I originally had it as `repr(u8)` but changed it away from it. I don't remember why. But one reason could have been that I didn't want to add ffi-only int representations to the enum :thinking: . Not sure.

janpetschexain: i see, thx for the explanation. i'm still not super happy with the solution, but at least for the sake of consistency it's ok for me.

discovery_engine/lib/src/ffi/types/document/feedback.dart
```diff
@@ -0,0 +1,65 @@
...
+import 'package:xayn_discovery_engine/src/ffi/genesis.ffigen.dart'
+    show RustUserReaction;
+import 'package:xayn_discovery_engine/src/ffi/load_lib.dart' show ffi;
+
+extension DocumentFeedbackFfi on DocumentFeedback {
+  void writeNative(final Pointer<RustUserReaction> place) {
+    var asInt = 128;
+    switch (this) {
+      case DocumentFeedback.neutral:
+        asInt = 0;
```
acrrd, line 28: Can we use `serde_repr` like in xayn-ai so we don't have to keep the values in sync?

rustonaut: `serde_repr` is mainly used due to wasm-bindgen using `Serde`.

In dart we can try to abuse the json code genration again, but that is tricky due to:

- us relying on implementation details of the json library
- relying on the annotations in dart and rust being the same
- it not generating the int mapping if the enum is not used in another json annotated struct.

Alternatively we can use `cbindgen` to spider into `core`  and
then have `DocumentFeedbackFfi` constants and maybe?? we can put that into
the json libraries annotations?

In rust we could use `serde_repr` + `serde_json`, or `as` cast + bounds checked transmute.

rustonaut:
@janpetschexain @acrrd , I updated it to:

- generate C bindings for the enum itself
- use them for annotations in dart
- force dart to generate the enums json table and use that for conversion
- made the enum `repr(u8)` used `serde_repr` derives
- to directly ptr. write/read the enum (i.e. no valid variant checks from dart -> rust, but then
  we always ship them together so that should be fine)

discovery_engine/lib/src/domain/models/document.dart
```diff
@@ -71,13 +74,27 @@ class Document {
...
   negative,
 }
+
+extension DocumentFeedbackIntConversion on DocumentFeedback {
+  int toIntRepr() => _$DocumentFeedbackEnumMap[this]!;
+  static DocumentFeedback fromIntRepr(int intRepr) =>
+      $enumDecode(_$DocumentFeedbackEnumMap, intRepr);
+}
+
+// Without this `_$DocumentFeedbackEnumMap` is not generated.
```
acrrd, line 94: Can we use `@JsonEnum(alwaysCreate=true)` on the enum instead?

discovery_engine_core/core/src/document.rs
```diff
@@ -149,18 +150,25 @@ impl TryFrom<Article> for NewsResource {
...
+    Positive = 1,
 
     /// The user is not interested.
-    Negative,
+    Negative = 2,
+}
+
+#[derive(Debug, DisplayDoc, Error)]
+/// Received an unsupported user reaction, int repr: {reaction}
+pub struct UnsupportedUserReaction {
```
acrrd, line 170: This seems unused.

discovery_engine/lib/src/domain/models/document.dart
rustonaut: yes, nice that they added that :+1:

discovery_engine_core/core/src/document.rs
rustonaut: Right, I had a `TryFrom` implementation before I change to writing to the pointer directly.

janpetschexain (commented):
lgtm

when resolving the merge conflicts, wdyt about changing the `cbindgen.toml`'s `header` field to `after_includes`?  this way the auto-generated warning would be at the first line of the header file, currently it comes after the additional typedefs.

```
after_includes = """

// Typedefs to make sure this "unknown" types are treated as opaque by ffigen.
typedef struct RustDuration RustDuration;
typedef struct RustEmbedding RustEmbedding;
typedef struct RustUuid RustUuid;"""
```

janpetschexain (approved):
👍

rustonaut:
bors merge

PR #106, opened 2022-02-03
[Merged by Bors] - Ty 2344 ffi user reacted [4.2]

rustonaut:
Based on #104 

Adds ffi support for `UserReacted`.


discovery_engine_core/bindings/src/types/document/user_reacted.rs
```diff
@@ -0,0 +1,95 @@
...
+}
+
+/// Returns a pointer to the `snipped` field of an [`UserReacted`].
+///
+/// # Safety
+///
+/// The pointer must point to a valid [`UserReacted`] memory object, it
+/// might be uninitialized.
+#[no_mangle]
+pub unsafe extern "C" fn user_reacted_place_of_snipped(place: *mut UserReacted) -> *mut String {
```
janpetschexain, line 51:
```suggestion
pub unsafe extern "C" fn user_reacted_place_of_snippet(place: *mut UserReacted) -> *mut String {
```

discovery_engine_core/core/src/stack.rs
```diff
@@ -63,7 +63,7 @@ pub type BoxedOps = Box<dyn Ops + Send + Sync>;
 /// `Id` is used to connect [`Ops`](ops::Ops) with the corresponding data of a stack.
 #[derive(Serialize, Deserialize, Clone, Copy, Debug, PartialEq, Eq, Hash, From, Display)]
 #[cfg_attr(test, derive(Default))]
-pub struct Id(Uuid);
+pub struct Id(pub Uuid);
```
janpetschexain, line 66: we could put a `#[repr(transparent)]` on the `Id`, then we wouldn't have to make the wrapped field public. similarly for `document::Id`.

discovery_engine/lib/src/ffi/types/document/user_reacted.dart
```diff
@@ -0,0 +1,83 @@
...
+import 'package:xayn_discovery_engine/src/ffi/types/string.dart' show StringFfi;
+import 'package:xayn_discovery_engine/src/ffi/types/uuid.dart'
+    show DocumentIdFfi, StackIdFfi;
+
+//FIXME user reacted isn't yet in models/ so we define it here for now
+// instead of having an extension here
+class UserReacted with EquatableMixin {
+  final DocumentId id;
+  final StackId stackId;
+  final String snipped;
```
janpetschexain, line 39:
```suggestion
  final String snippet;
```

discovery_engine_core/core/src/document.rs
```diff
@@ -129,7 +129,7 @@ pub struct UserReacted {
     pub snippet: String,
 
     /// Precomputed S-mBert of the document.
-    pub smbert: Embedding,
+    pub smbert_embedding: Embedding,
```
janpetschexain, line 132: while we are at it we could also rename the `id` field in l.123 to `document_id` to better differentiate it from the `stack_id` field and to align the namings in the ffi.

discovery_engine/test/ffi/types/document/user_reacted_test.dart
```diff
@@ -0,0 +1,40 @@
...
+import 'package:xayn_discovery_engine/src/ffi/load_lib.dart' show ffi;
+import 'package:xayn_discovery_engine/src/ffi/types/document/user_reacted.dart'
+    show UserReacted;
+
+void main() {
+  test('reading written user reacted instance yields same result', () {
+    final document = UserReacted(
+      id: DocumentId(),
+      stackId: StackId(),
+      snipped: 'Cloning bought back the dodo.',
```
janpetschexain, line 30: doesn't matter much for the test, but probably this was meant?
```suggestion
      snipped: 'Cloning brought back the dodo.',
```

discovery_engine_core/core/src/document.rs
acrrd: The type is already `Embedding` why we need to repeat it?

acrrd: field's named `id` are the id of the entity they are in. I don't think will be necessary to additionally add `document_` to avoid stuff like `document.document_id`, `document.id` seems already clear.

discovery_engine/lib/src/ffi/types/document/user_reacted.dart
```diff
@@ -0,0 +1,83 @@
...
+import 'package:xayn_discovery_engine/src/ffi/load_lib.dart' show ffi;
+import 'package:xayn_discovery_engine/src/ffi/types/document/feedback.dart'
+    show DocumentFeedbackFfi;
+import 'package:xayn_discovery_engine/src/ffi/types/embedding.dart'
+    show EmbeddingFfi;
+import 'package:xayn_discovery_engine/src/ffi/types/string.dart' show StringFfi;
+import 'package:xayn_discovery_engine/src/ffi/types/uuid.dart'
+    show DocumentIdFfi, StackIdFfi;
+
+//FIXME user reacted isn't yet in models/ so we define it here for now
```
acrrd, line 34: This is ok because there corresponding event is different.

discovery_engine_core/core/src/document.rs
janpetschexain: > field's named `id` are the id of the entity they are in. I don't think will be necessary to additionally add `document_` to avoid stuff like `document.document_id`, `document.id` seems already clear.

true, i haven't looked at it this way.

discovery_engine_core/core/src/stack.rs
rustonaut: I'm not sure if a `#[repr(transparent)]` which you are supposed to use to write the inner type without making the inner type public is an anit-pattern or not tbh. :thinking: 

It kinda makes the field pseudo public, which is why I think it might be an anit-pattern.

But then only for unsafe code, mainly ffi code, which makes it a reasonable pattern.

rustonaut: I will for nor change it like you said.

discovery_engine_core/core/src/document.rs
rustonaut: I called it `smbert_embedding` because it is named that way in other places. And it's not a `smbert`. Sure field names don't have to repeat the type. But IMHO the parts you can cut out are the non-stem parts. For S-mBert Embeddings the stem is Embedding not S-mBert. (And yes that linguistically seen probably not a fully correct usage of the word stem, I hope it still conveys what I mean.)

janpetschexain (approved):
lgtm

janpetschexain:
it looks like the previous pr caused a name conflict during the `ffigen` call. the error is
```
[WARNING]: Resolved name conflict: Declaration 'RustUserReaction' and has been renamed to 'RustUserReaction1'.
```
it seems that the enum is correctly generated as `RustUserReaction`, but wherever the ffi uses a pointer to this (ie. alloc/drop), `ffigen` renames the "conflicting" name to `RustUserReaction1`. we have to check if `ffigen` can be configured correctly or if we can remove the alloc/drop test ffi utils.

acrrd:
> [WARNING]: Resolved name conflict: Declaration 'RustUserReaction' and has been renamed to 'RustUserReaction1'.
>

It seems that is cbindgen that generates:
```c
enum UserReaction {
 ...
};
typedef uint8_t UserReaction;
```
And then ffigen complain that the type is defined to times.

rustonaut:
> And then ffigen complain that the type is defined to times.

yes

the funny thing is:

- it needs to declare the type twice, once as a class with constants and once as a uint8_t alias
- it just happens to not expect  it and as such runs into a name collision
- I think the C declaration is correct due to the strange way C works, like you have the type `enum RustUserReaction` and the type `RustUserReaction`. `enum RustUserReaction` provides the values for the enum, but `RustUserReaction` is the representation used in practice

acrrd:
We also have the same C declarations in xayn_ai  but no warning there. The configuration seems the same and the version of ffigen is the same

janpetschexain:
> We also have the same C declarations in xayn_ai but no warning there. The configuration seems the same and the version of ffigen is the same

the main difference is, that in xayn_ai the enums are only ever used by value, but never by pointer. on the other hand, in the DE the enum is used by value and pointer, i guess this is where the genererator gets confused.

rustonaut:
Can I still merge it?

acrrd:
Yes. It does not seems something that we can act quickly or at all. I gave a quicklook at generating c++ headers but ffigen does not support it.

rustonaut:
bors merge

PR #107, opened 2022-02-03
[Merged by Bors] - TY-2333 - Include flutter plugin in the CI

x3ro:
[TY-2333]

[TY-2333]: https://xainag.atlassian.net/browse/TY-2333?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/actions/setup-job/action.yml
+++ b/.github/actions/setup-job/action.yml
@@ -4,6 +4,9 @@ inputs:
   dart:
     description: 'Setup dart'
     default: false
+  flutter:
+    description: 'Setup flutter'
+    default: false
   rust:
     description: 'Setup rust'
     default: false
@@ -19,13 +22,27 @@ runs:
     - name: Install just
       uses: extractions/setup-just@aa5d15c144db4585980a44ebfdd2cf337c4f14cb #v1.4.0
       with:
-        just-version: '0.10.5'
+        just-version: ${{ env.JUST_VERSION }}
 
     - name: Install dart
       uses: dart-lang/setup-dart@6a218f2413a3e78e9087f638a238f6b40893203d # v1.3
-      if: inputs.dart
+      # Flutter already installs a dart SDK, so we never need to install both
+      if: ${{ inputs.dart != 'false' && inputs.flutter != 'true' }}
       with:
         sdk: ${{ env.DART_VERSION }}
+
+    - name: Install flutter
+      uses: subosito/flutter-action@5e1529bc12adae5b26f30f0a4046ac743d6f0c5d # v2.3.0
+      if: inputs.flutter
+      with:
+        flutter-version: ${{ env.FLUTTER_VERSION }}
+        cache: true
+
+    - name: Setup flutter dependencies
+      shell: bash
+      if: inputs.flutter
+      run: |
+        just flutter-deps
 
     - name: Install rust toolchain
       working-directory: ${{ env.RUST_WORKSPACE }}
@@ -52,12 +69,12 @@ runs:
       shell: bash
       if: inputs.deps
       run: |
-        if [[ "${{ inputs.dart }}" == "true" && "${{ inputs.rust }}" != "false" ]]; then
+        if [[ "${{ inputs.dart }}" == "true" || "${{ inputs.flutter }}" == "true" ]] && [[ "${{ inputs.rust }}" == "true" ]]; then
           # ffigen dep, not handled by `just deps` as it's OS specific.
           sudo apt-get install libclang-10-dev
           just deps
-        elif [ "${{ inputs.dart }}" = "true" ]; then
+        elif [ "${{ inputs.dart }}" == "true" ]; then
           just dart-deps
-        elif [ "${{ inputs.rust }}" != "false" ]; then
+        elif [ "${{ inputs.rust }}" == "true" ]; then
           just rust-deps
         fi
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ permissions:
 
 jobs:
   dev-ci:
-    uses: xaynetwork/xayn_discovery_engine/.github/workflows/ci_reusable_wf.yml@main
+    uses: ./.github/workflows/ci_reusable_wf.yml
 
   ci-ok:
     name: ci-ok
--- a/justfile
+++ b/justfile
@@ -1,5 +1,6 @@
 # We import environment variables from .env
 set dotenv-load := true
+set shell := ["bash", "-uc"]
 
 # If CI is set and add rust flags
 export RUSTFLAGS := if env_var_or_default("CI", "false") == "true" {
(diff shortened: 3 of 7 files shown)


.github/actions/setup-job/action.yml
```diff
@@ -19,14 +22,43 @@ runs:
     - name: Install just
       uses: extractions/setup-just@aa5d15c144db4585980a44ebfdd2cf337c4f14cb #v1.4.0
       with:
-        just-version: '0.10.5'
+        just-version: ${{ env.JUST_VERSION }}
+
+    - name: Restore Rust toolchain cache
```
acrrd, line 27: This is not restoring the toolchain cache it cannot be called before the toolchain is there.

.github/actions/setup-job/action.yml
```diff
@@ -19,14 +22,43 @@ runs:
...
+      if: inputs.rust != 'false'
+      with:
+        working-directory: ${{ env.RUST_WORKSPACE }}
+
+    - name: Restore Rust build cache
+      uses: actions/cache@v2
+      if: inputs.rust != 'false'
+      with:
+        path: |
+          target/
```
acrrd, line 38: This is already cached by `rust-cache`.
We can move this call at the end together with the other rust parts.

.github/workflows/ci.yml
```diff
@@ -21,7 +21,7 @@ permissions:
 
 jobs:
   dev-ci:
-    uses: xaynetwork/xayn_discovery_engine/.github/workflows/ci_reusable_wf.yml@main
+    uses: ./.github/workflows/ci_reusable_wf.yml
```
acrrd, line 24: Does this really work now? When we started doing this it wasn't possible. Would be very nice!

justfile
```diff
@@ -45,6 +57,12 @@ deps: dart-deps rust-deps install-async-bindgen
 dart-fmt:
     cd "$DART_WORKSPACE"; \
     dart format {{ if env_var_or_default("CI", "false") == "true" { "--output=none --set-exit-if-changed" } else { "" } }} .
+    cd "$FLUTTER_WORKSPACE"; \
+    dart format {{ if env_var_or_default("CI", "false") == "true" { "--output=none --set-exit-if-changed" } else { "" } }} .
```
acrrd, line 61: Can we put
`{{ if env_var_or_default("CI", "false") == "true" { "--output=none --set-exit-if-changed" } else { "" } }}` in a variable and pass that to `dart format`?

justfile
```diff
@@ -1,5 +1,6 @@
 # We import environment variables from .env
 set dotenv-load := true
+set shell := ["bash", "-uc"]
```
acrrd, line 3: Why we set this? If we set this do we still have to specify `#!/usr/bin/env sh` in other jobs?

justfile
```diff
@@ -45,6 +57,12 @@ deps: dart-deps rust-deps install-async-bindgen
 dart-fmt:
     cd "$DART_WORKSPACE"; \
     dart format {{ if env_var_or_default("CI", "false") == "true" { "--output=none --set-exit-if-changed" } else { "" } }} .
+    cd "$FLUTTER_WORKSPACE"; \
+    dart format {{ if env_var_or_default("CI", "false") == "true" { "--output=none --set-exit-if-changed" } else { "" } }} .
+
+# Installs the required rust toolchain version
+install-rust-toolchain:
```
acrrd, line 64: This is misleading because the nightly toolchain is only needed to format while 1.55 is needed for everything else.
Also I'm not too incline to add stuff like this here since we don't have the same for dart or flutter toolchains and is something that we need to do once on our machine

.github/actions/setup-job/action.yml
x3ro: Mmh, I'm going to double-check then. I saw some build times cut in half with this caching added, but maybe that was just a fluke.

justfile
x3ro: We could, but I felt like it's not worth it for just one copy of the thing. If you think this is a must-have I can implement it though.

justfile
x3ro: This is the shell that all commands are run with.  I assume you're referring to this:

```
dart-deps:
    #!/usr/bin/env sh
    ....
```

This is what Just calls a "shebang recipe", and it's always preceded by the shell you want to run it with. 

> Shebang recipe bodies are extracted and run as scripts [...].

So long story short: I set `bash` so I can use `=~` and `[[`, but this does not affect the recipes you're referring to, I think.

.github/actions/setup-job/action.yml
acrrd: `rust-cache` does not keep our own build code: https://github.com/Swatinem/rust-cache/issues/37 
The code is not versioned for the cache so this can create problem on the long run.

justfile
acrrd: not must-have

justfile
x3ro: Mmh, I don't see why not having feature parity is a reason to exclude it. I added this because I was confused that we're referencing the `RUST_NIGHTLY` variable in the scripts, but don't have a way to install it. As a result, I myself was using a much more recent version of Rust nightly to format my code, which could've led to odd problems. 

In any case, I'll remove it for now, as it is probably not the right place to put it.

.github/workflows/ci.yml
acrrd: Can you please also do this change in the bors ci?

x3ro: At least for me, on my branch, this has been working fine :)

.github/actions/setup-job/action.yml
x3ro: Sure, but the cache here is bound to the Cargo.lock, which changes quite frequently for us. So for branches I definitely caching `target/` is desirable, though one could think about disabling it in bors / main.

x3ro: In any case, I'm going to revert that part to the state of before for now. No need to look into this in detail at the moment.

justfile
acrrd: I was wondering if with this definition all commands are implitly shebang recipe for every task or if we still have to it to avoid to have to put \ at the end of a line to have all the commands run in the same shell and not a new shell per line.

acrrd (approved):
Can you please do the same change you did in `ci.yml` to `bors_ci.yml` so we don't have to do a one line pr for that.

justfile
x3ro: No, that is unfortunately not how it works, I think 🤔 But I will double-check.

x3ro: From the docs:

> Recipes without an initial shebang are evaluated and run line-by-line [...]

So yeah, if we'd want to remove the backslashes we'd have to include the shebang for every recipe.

x3ro:
bors merge

PR #108, opened 2022-02-07
[Merged by Bors] - TY-2349 Reading assets manifest before engine initialization

biernacki:
[Jira ref](https://xainag.atlassian.net/browse/TY-2349)

This PR fixes the flutter plugin & example app, by moving reading of the assets `Manifest` file outside of the engine.

`PlatformAssetBundle` which is used used to read the manifest file from the flutter assets, can't be used in an isolate other than main isolate. We also need to use `PlatformAssetBundle` to read the file, cause Flutter doesn't support the `Isolate.Isolate.resolvePackageUri` method. So easiest solution is to read the file first and pass it with `Configuration` object during engine initialisation.


janpetschexain (approved):
runs on my machine 👍 (needs a rebase though)

since running the dart example will put some real assets in `xayn_discovery_engine/discovery_engine/example/engine/` now, can we add this folder to the respective `.gitignore` such that they are not flagged as new files once the example is run?

biernacki:
> runs on my machine 👍 (needs a rebase though)
> 
> since running the dart example will put some real assets in `xayn_discovery_engine/discovery_engine/example/engine/` now, can we add this folder to the respective `.gitignore` such that they are not flagged as new files once the example is run?

Thanks, I've rebased and added folder for fetched assets to `.gitignore` as suggested!

biernacki:
bors merge

PR #110, opened 2022-02-07
TY-2384 update config

janpetschexain:
**References**

- [TY-2384]
- related to https://github.com/xaynetwork/xayn_ai/pull/413

**Summary**

- add `kpe_top` config value to the ranker

### DIFF
--- a/discovery_engine_core/Cargo.lock
+++ b/discovery_engine_core/Cargo.lock
@@ -1034,7 +1034,7 @@ dependencies = [
 [[package]]
 name = "kpe"
 version = "0.1.0"
-source = "git+https://github.com/xaynetwork/xayn_ai?branch=master#61060cee615034e433f050d096094dffa880f5e7"
+source = "git+https://github.com/xaynetwork/xayn_ai?branch=TY-2384-update_config#178ddf8cab60d1300c1910e512f54bff85c0a0ca"
 dependencies = [
  "derive_more",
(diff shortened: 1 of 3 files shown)


Robert-Steiner:
I think we will also need add those fields to the [dart part](https://github.com/xaynetwork/xayn_discovery_engine/blob/main/discovery_engine/lib/src/domain/engine/engine_config.dart). btw @rustonaut do we also need a ffi struct for the config and if so is it a problem that we don't have a `SetupData` sruct in rust?

acrrd:
I think this approach will create problem with the ffi part because it means that we still to set the values from dart or the struct cannot be correctly initialized. We can a different kind of Config that stay only in rust and from which we can move the fields to this one when we want to expose them to dart.

@Robert-Steiner why we should have the SetupData, we are already taking the paths.

janpetschexain:
> I think this approach will create problem with the ffi part because it means that we still to set the values from dart or the struct cannot be correctly initialized. We can a different kind of Config that stay only in rust and from which we can move the fields to this one when we want to expose them to dart.

i'll split the config into two sets.

acrrd:
Thanks

Robert-Steiner:
> @Robert-Steiner why we should have the `SetupData`, we are already taking the paths.

I'm was not saying that we should have it i was only asking if this is a problem if they are different.

PR #111, opened 2022-02-07
[Merged by Bors] - TY-2386 filter top documents

janpetschexain:
**References**

- [TY-2386]

**Summary**

- adds a filter to retain the top documents per stack/bucket
- call the filter at the end of `update_stacks()`

[TY-2386]: https://xainag.atlassian.net/browse/TY-2386?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine_core/core/src/engine.rs
+++ b/discovery_engine_core/core/src/engine.rs
@@ -229,11 +229,12 @@ where
                 Ok(documents) => {
                     if let Err(error) = stack.update(&documents, &mut self.ranker) {
                         errors.push(Error::StackOpFailed(error));
+                    } else {
+                        stack.data.retain_top(self.core_config.keep_top);
                     }
                 }
                 Err(error) => errors.push(error.into()),
             }
-            stack.data.retain_top(self.core_config.keep_top);
         }
 
         if errors.is_empty() {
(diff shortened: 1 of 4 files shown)


discovery_engine_core/core/src/engine.rs
```diff
@@ -234,6 +233,7 @@ where
                 }
                 Err(error) => errors.push(error.into()),
             }
+            stack.data.retain_top(self.core_config.keep_top);
```
acrrd, line 236: We want to run this only if `stack.update` succeed.

discovery_engine_core/core/src/stack.rs
```diff
@@ -162,6 +162,10 @@ impl Bucket<Document> for Stack {
     fn pop(&mut self) -> Option<Document> {
         self.data.documents.pop()
     }
+
+    fn retain_top(&mut self, keep: usize) {
```
acrrd, line 166: We use directly the implementation from `self.data` so we don't need to expose this.

discovery_engine_core/core/src/mab.rs
```diff
@@ -55,6 +55,9 @@ pub(crate) trait Bucket<T> {
 
     /// Removes the next best element from this bucket and returns it, or `None` if it is empty.
     fn pop(&mut self) -> Option<T>;
+
```
acrrd, line 58: Why we need to modify the bucket? This method has nothing to do with it.

janpetschexain: it's just an additional impl which could be useful, but we can impl it later on if we need it, i'll remove it.

acrrd (approved):

janpetschexain:
bors merge

PR #113, opened 2022-02-07
[Merged by Bors] - TY-2385 TY-2387 call update stacks

janpetschexain:
**References**

- [TY-2385]
- [TY-2387]
- based on #111
- based on https://github.com/xaynetwork/xayn_ai/pull/415

**Summary**

- call `update_stacks()` in constructor independent of stack size
- call `update_stacks()` in `get_feed_documents()` wrt stack size after popping

[TY-2385]: https://xainag.atlassian.net/browse/TY-2385?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[TY-2387]: https://xainag.atlassian.net/browse/TY-2387?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine_core/core/src/engine.rs
+++ b/discovery_engine_core/core/src/engine.rs
@@ -12,7 +12,7 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-use std::{collections::HashMap, iter::repeat};
+use std::collections::HashMap;
 
 use displaydoc::Display;
 use serde::{Deserialize, Serialize};
@@ -188,9 +188,7 @@ where
             stacks,
             ranker,
         };
-        engine
-            .update_stacks(repeat(engine.core_config.keep_top))
-            .await?;
+        engine.update_stacks(0).await?;
 
         Ok(engine)
     }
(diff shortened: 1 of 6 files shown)


discovery_engine_core/core/src/engine.rs
```diff
@@ -158,12 +161,15 @@ where
...
 
-        Ok(Self {
+        let mut engine = Self {
             config,
             core_config,
             stacks,
             ranker,
-        })
+        };
+        engine.update_stacks(0).await?;
```
acrrd, line 170: This will update the stacks only if they are empty but we want to update them whatever is their size.

discovery_engine_core/core/src/mab.rs
```diff
@@ -50,6 +50,9 @@ pub(crate) trait Bucket<T> {
     /// Returns the beta parameter of the beta distribution.
     fn beta(&self) -> f32;
 
+    /// Gets the number of elements in the bucket.
```
acrrd, line 53: We don't need to modify mab.

discovery_engine_core/core/src/engine.rs
janpetschexain: you are absolutely right, i'll change it to `usize::MAX`, i shouldn't work that late oO

discovery_engine_core/core/src/mab.rs
janpetschexain: similarly to the previous pr, this might be useful, but we can always add it later on, i'll remove it for now.

acrrd (approved):

PR #114, opened 2022-02-08
[Merged by Bors] - TY-2305 updateable markets config

finiteprods:
previously: #95 added a configuration struct to initialize the engine

**Summary**

- puts the markets part of the configuration behind a lock
- adds a method `set_markets` to update it
- config struct now called `InitConfig` with the "endpoint-specific" parts in a new `EndpointConfig` struct

### DIFF
--- a/discovery_engine_core/core/src/engine.rs
+++ b/discovery_engine_core/core/src/engine.rs
@@ -68,18 +68,36 @@ pub struct Market {
     lang_code: String,
 }
 
-/// Discovery Engine configuration settings.
-#[allow(dead_code)]
-pub struct Config {
+/// Configuration settings to initialize Discovery Engine with a [`xayn_ai::ranker::Ranker`].
+pub struct InitConfig {
     api_key: String,
     api_base_url: String,
-    markets: RwLock<Vec<Market>>,
+    markets: Vec<Market>,
     smbert_vocab: String,
     smbert_model: String,
     kpe_vocab: String,
     kpe_model: String,
     kpe_cnn: String,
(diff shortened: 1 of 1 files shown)


discovery_engine_core/core/src/engine.rs
```diff
@@ -185,6 +185,12 @@ where
         bincode::serialize(&state_data).map_err(|err| Error::Serialization(err.into()))
     }
 
+    /// Updates the markets configuration.
+    pub async fn set_markets(&mut self, markets: Vec<Market>) {
+        let mut mkts = self.config.markets.write().await;
+        *mkts = markets;
```
janpetschexain, line 191: this could be done without the intermediate binding, but both ways are ok i guess.
```suggestion
        *self.config.markets.write().await = markets;
```

janpetschexain (approved):
lgtm

discovery_engine_core/core/src/engine.rs
```diff
@@ -68,7 +68,7 @@ struct Market {
 pub struct Config {
     api_key: String,
     api_base_url: String,
-    markets: Vec<Market>,
+    markets: RwLock<Vec<Market>>,
```
acrrd, line 71: We cannot do this because it will complicate how the ffi can work with this struct. 
We can keep the relevant data in another struct and use this only as a way of passing messages between dart and rust.
In this new struct we can drop the paths to the models.

acrrd (approved):

finiteprods:
bors merge

PR #118, opened 2022-02-08
[Merged by Bors] - TY-2358 add Just task for compiling for ios

Robert-Steiner:
Ticket: 

- [TY-2358]

e.g.

`just compile-ios aarch64-apple-ios`

wait for #125 

[TY-2358]: https://xainag.atlassian.net/browse/TY-2358?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/justfile
+++ b/justfile
@@ -195,7 +195,7 @@ rust-clean:
 dart-clean:
     find "$DART_WORKSPACE" -type d -name .dart_tool -prune -exec rm -r '{}' \;
 
-# Remvoes all local cargo isntalls
+# Removes all local cargo installs
 clean-tools:
     -rm -r "$CARGO_INSTALL_ROOT"
 
@@ -230,6 +230,11 @@ compile-android-local:
     for TARGET in $ANDROID_TARGETS; do
         {{just_executable()}} compile-android $TARGET
     done
+
+# Compiles the core lib for the given iOS target
+compile-ios target *args:
+    cd "$RUST_WORKSPACE"; \
+    cargo build --target {{target}} -p xayn-discovery-engine-core {{args}}
 
 alias d := dart-test
 alias r := rust-test
--- a/justfile
+++ b/justfile
@@ -231,10 +231,10 @@ compile-android-local:
         {{just_executable()}} compile-android $TARGET
     done
 
-# Compiles the core lib for the given iOS target
-compile-ios target *args:
+# Compiles the bindings for the given iOS target
+compile-ios target:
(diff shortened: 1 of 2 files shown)


justfile
```diff
@@ -197,6 +197,11 @@ _dry-run-release: clean deps dart-build
 dry-run-release:
      @CI=true {{just_executable()}} _dry-run-release
 
+# Compiles the core lib for the given iOS target
+compile-ios target *args:
+    cd "$RUST_WORKSPACE"; \
+    cargo build --target {{target}} -p xayn-discovery-engine-core {{args}}
```
acrrd, line 203: This should target `bindings` not the core.

acrrd: > `just compile-ios aarch64-apple-ios` `just compile-ios aarch64-apple-ios --release`

In xayn_ai we are always compiling with `--release` because otherwise it is too slow to run. I think we should do the same here.

acrrd:
Did we had a broken master that we have an error in rust?

Robert-Steiner:
> Did we had a broken master that we have an error in rust?

Good question. I'm nor sure why the ci is failing

Robert-Steiner:
it is because i added
```
[lib]
crate-type = ["staticlib"]
```
in core

justfile
Robert-Steiner: done

acrrd (approved):

x3ro:
One question: since the `compile-ios` recipe does not currently contain anything iOS specific, couldn't we call it `compile-release` or something like that, and also use it for Android? 🤔

Robert-Steiner:
> One question: since the `compile-ios` recipe does not currently contain anything iOS specific, couldn't we call it `compile-release` or something like that, and also use it for Android? 🤔
> 
> Edit: Never mind, since this is already approved, lets get it merged. We can always change it later if it makes sense.

For android we use cargo-ndk. We could also do it with a simple cargo build command but then we would have to set a bunch of android ndk related environment variable ourselves. 

here is an example with a plain cargo build command:
https://xainag.atlassian.net/wiki/spaces/M2D/pages/2192211979/Run+tract+cli+on+mobile

Robert-Steiner:
bors merge

Robert-Steiner:
bors cancel

acrrd:
cargo ndk is also already able to strip down the libraries and put them in the correct folder layouts

PR #119, opened 2022-02-08
[Merged by Bors] - TY-2379, TY-2380 Added EngineStateRepository and Hive implementation

biernacki:
[Jira ref TY-2379](https://xainag.atlassian.net/browse/TY-2379)
[Jira ref TY-2380](https://xainag.atlassian.net/browse/TY-2380)

This PR adds an interface for `EngineStateRepository` and it's Hive implementation, together with some tests and updated initialisation steps of the event handler.

`EngineStateRepository` will hold engine state serialized into a byte array (`Uint8List`).

### DIFF
--- a/discovery_engine/lib/src/domain/repository/engine_state_repo.dart
+++ b/discovery_engine/lib/src/domain/repository/engine_state_repo.dart
@@ -21,7 +21,4 @@ abstract class EngineStateRepository {
 
   /// Persist the state of the engine.
   Future<void> save(Uint8List bytes);
-
-  /// Clear the repository.
-  Future<void> reset();
 }
--- a/discovery_engine/lib/src/domain/repository/engine_state_repo.dart
+++ b/discovery_engine/lib/src/domain/repository/engine_state_repo.dart
@@ -1,4 +1,4 @@
-// Copyright 2021 Xayn AG
+// Copyright 2022 Xayn AG
 //
 // This program is free software: you can redistribute it and/or modify
 // it under the terms of the GNU Affero General Public License as
@@ -12,7 +12,16 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-const documentBox = 'document';
-const changedDocumentIdBox = 'changed_document';
-const activeDocumentDataBox = 'active_document';
-const engineStateBox = 'engine_state';
+import 'dart:typed_data' show Uint8List;
+
+/// Repository interface for serialized engine state.
+abstract class EngineStateRepository {
+  /// Fetch the current state of the engine.
+  Future<Uint8List?> load();
+
+  /// Persist the state of the engine.
+  Future<void> save(Uint8List bytes);
+
+  /// Clear the repository.
+  Future<void> reset();
+}
--- a/discovery_engine/lib/src/infrastructure/repository/hive_engine_state_repo.dart
+++ b/discovery_engine/lib/src/infrastructure/repository/hive_engine_state_repo.dart
@@ -31,7 +31,4 @@ class HiveEngineStateRepository implements EngineStateRepository {
 
   @override
   Future<void> save(Uint8List bytes) => box.put(stateKey, bytes);
-
-  @override
-  Future<void> reset() => box.clear();
 }
--- a/discovery_engine/lib/src/infrastructure/repository/hive_engine_state_repo.dart
+++ b/discovery_engine/lib/src/infrastructure/repository/hive_engine_state_repo.dart
@@ -1,0 +1,37 @@
+// Copyright 2022 Xayn AG
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, version 3.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+import 'dart:typed_data' show Uint8List;
+
+import 'package:hive/hive.dart' show Hive, Box;
+import 'package:xayn_discovery_engine/src/domain/repository/engine_state_repo.dart'
+    show EngineStateRepository;
+import 'package:xayn_discovery_engine/src/infrastructure/box_name.dart'
+    show engineStateBox;
+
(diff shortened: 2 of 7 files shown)


discovery_engine/test/repository/hive_engine_state_repo_test.dart
```diff
@@ -0,0 +1,100 @@
...
+    });
+
+    group('"save" method', () {
+      test('when we save some data it will be present in the box', () async {
+        await _repo.save(Uint8List.fromList([5, 6, 7, 8]));
+
+        expect(_box.values.first, equals(Uint8List.fromList([5, 6, 7, 8])));
+      });
+
+      test('when we override data it will be updated in the box', () async {
```
acrrd, line 64: Isn't this a special case of next test?

discovery_engine/lib/src/infrastructure/repository/hive_engine_state_repo.dart
```diff
@@ -0,0 +1,37 @@
...
+import 'package:xayn_discovery_engine/src/domain/repository/engine_state_repo.dart'
+    show EngineStateRepository;
+import 'package:xayn_discovery_engine/src/infrastructure/box_name.dart'
+    show engineStateBox;
+
+/// Hive repository implementation of [EngineStateRepository].
+class HiveEngineStateRepository implements EngineStateRepository {
+  static const stateKey = 0;
+
+  Box<Uint8List> get box => Hive.box<Uint8List>(engineStateBox);
```
acrrd, line 27: This is the same approach that we use in other repo, but why we cannot open it only once in the constructor?

discovery_engine/test/repository/hive_engine_state_repo_test.dart
```diff
@@ -0,0 +1,100 @@
...
+
+import 'package:xayn_discovery_engine/src/infrastructure/box_name.dart'
+    show engineStateBox;
+import 'package:xayn_discovery_engine/src/infrastructure/repository/hive_engine_state_repo.dart'
+    show HiveEngineStateRepository;
+
+Future<void> main() async {
+  group('HiveEngineStateRepository', () {
+    late Box<Uint8List> _box;
+    final _repo = HiveEngineStateRepository();
```
acrrd, line 28: Shouldn't we create it fresh in each test?

discovery_engine/lib/src/domain/repository/engine_state_repo.dart
```diff
@@ -0,0 +1,27 @@
...
+/// Repository interface for serialized engine state.
+abstract class EngineStateRepository {
+  /// Fetch the current state of the engine.
+  Future<Uint8List?> load();
+
+  /// Persist the state of the engine.
+  Future<void> save(Uint8List bytes);
+
+  /// Clear the repository.
+  Future<void> reset();
```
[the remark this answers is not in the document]
biernacki: I added that because of the `ResetEngine` event, but now I see it was removed. I'll drop it from the PR.

discovery_engine/test/repository/hive_engine_state_repo_test.dart
biernacki: yup, we can just remove it, as the next test also covers that case

discovery_engine/lib/src/infrastructure/repository/hive_engine_state_repo.dart
biernacki: constructors can't be async, so it would need to ba a static method + private default constructor. I think we could make this change for all repos, just not sure if this PR would be the best one to do it. Should we do that in a separate PR, wdyt? I can add a task for that.

discovery_engine/test/repository/hive_engine_state_repo_test.dart
biernacki: the repo itself in it's current form is stateless, so there is no need, but if we change it so that we open a box when creating an instance, then yes, we will need to update tests accordingly.

discovery_engine/lib/src/infrastructure/repository/hive_engine_state_repo.dart
```diff
@@ -0,0 +1,37 @@
...
+
+import 'package:hive/hive.dart' show Hive, Box;
+import 'package:xayn_discovery_engine/src/domain/repository/engine_state_repo.dart'
+    show EngineStateRepository;
+import 'package:xayn_discovery_engine/src/infrastructure/box_name.dart'
+    show engineStateBox;
+
+/// Hive repository implementation of [EngineStateRepository].
+class HiveEngineStateRepository implements EngineStateRepository {
+  static const stateKey = 0;
```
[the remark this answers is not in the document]
biernacki: I'm using it in tests, so it would be easier to leave it as is.

discovery_engine/lib/src/infrastructure/repository/hive_engine_state_repo.dart
acrrd: Yes we do this later. I thought there was some problem with hive in doing that.

discovery_engine/test/repository/hive_engine_state_repo_test.dart
acrrd: Since is not a requirement that it must be state less we can do it so the tests don't have to assume that is stateless or stateful.

acrrd: If tests on repository are similar we can do this n the PR to open the box only once.

biernacki: I've adjusted other repo tests, other changes might be better to do in a separate PR

discovery_engine/lib/src/infrastructure/repository/hive_engine_state_repo.dart
acrrd: I saw that is not really opening anyway but only returning the handle to the open one so we can leave it like this.

discovery_engine/test/repository/hive_active_document_repo_test.dart
```diff
@@ -32,17 +32,24 @@ import '../logging.dart' show setupLogging;
...
+    late Box<ActiveDocumentData> box;
+    late HiveActiveDocumentDataRepository repo;
     final data = ActiveDocumentData(Uint8List(0));
     final id1 = DocumentId();
     final id2 = DocumentId();
 
+    setUpAll(() async {
+      box = await Hive.openBox<ActiveDocumentData>(
+        activeDocumentDataBox,
+        bytes: Uint8List(0),
```
acrrd, line 45: `bytes` will be taken in consideration only the first time. This can be a bit misleading.

acrrd: My bad, it run only once.

biernacki:
bors merge

PR #124, opened 2022-02-09
[Merged by Bors] - Don't expose the isActive field

acrrd:
We don't have cases in which we return the non active document so it does not make sense to expose it.


biernacki (approved):
seems legit!

acrrd:
bors merge

PR #125, opened 2022-02-09
[Merged by Bors] - Small fixes CI and Justfile

acrrd:
* Do not always install flutter
* Remove unused env var in the release ci
* set github token do install just
* add `flutter-deps` to the `deps` target
* update the pubspec lock for flutter
* remove path filters in the ci.

### DIFF
--- a/.github/actions/setup-job/action.yml
+++ b/.github/actions/setup-job/action.yml
@@ -22,7 +22,7 @@ runs:
     - name: Install just
       uses: extractions/setup-just@aa5d15c144db4585980a44ebfdd2cf337c4f14cb #v1.4.0
       env:
-        GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+        GITHUB_TOKEN: ${{ env.GITHUB_TOKEN }}
       with:
         just-version: ${{ env.JUST_VERSION }}
 
--- a/.github/actions/setup-job/action.yml
+++ b/.github/actions/setup-job/action.yml
@@ -71,12 +71,12 @@ runs:
       shell: bash
       if: inputs.deps
       run: |
-        if [[ "${{ inputs.dart }}" == "true" || "${{ inputs.flutter }}" == "true" ]] && [[ "${{ inputs.rust }}" != "false" ]]; then
+        if [[ "${{ inputs.dart }}" != "false" || "${{ inputs.flutter }}" != "false" ]] && [[ "${{ inputs.rust }}" != "false" ]]; then
           # ffigen dep, not handled by `just` as it's OS specific.
           sudo apt-get install libclang-10-dev
         fi
 
-        if [ "${{ inputs.rust }}" == "true" ]; then
+        if [ "${{ inputs.rust }}" != "false" ]; then
           just rust-deps
         fi
 
--- a/.github/actions/setup-job/action.yml
+++ b/.github/actions/setup-job/action.yml
@@ -71,7 +71,7 @@ runs:
       shell: bash
       if: inputs.deps
       run: |
-        if [[ "${{ inputs.dart }}" == "true" || "${{ inputs.flutter }}" == "true" ]] && [[ "${{ inputs.rust }}" == "true" ]]; then
+        if [[ "${{ inputs.dart }}" == "true" || "${{ inputs.flutter }}" == "true" ]] && [[ "${{ inputs.rust }}" != "false" ]]; then
           # ffigen dep, not handled by `just` as it's OS specific.
           sudo apt-get install libclang-10-dev
         fi
--- a/.github/actions/setup-job/action.yml
+++ b/.github/actions/setup-job/action.yml
@@ -21,7 +21,9 @@ runs:
 
     - name: Install just
       uses: extractions/setup-just@aa5d15c144db4585980a44ebfdd2cf337c4f14cb #v1.4.0
-      with:
+     env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+     with:
         just-version: ${{ env.JUST_VERSION }}
 
     - name: Install dart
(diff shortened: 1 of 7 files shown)


.github/actions/setup-job/action.yml
```diff
@@ -21,6 +21,8 @@ runs:
 
     - name: Install just
       uses: extractions/setup-just@aa5d15c144db4585980a44ebfdd2cf337c4f14cb #v1.4.0
+      env:
+        GITHUB_TOKEN: ${{ env.GITHUB_TOKEN }}
```
rustonaut, line 25: Its still ab bit strange:

> this is because this workflow has to make requests to GitHub API in order to list available releases.

We explicitly define the exact version we want to use, it shouldn't make the requests. In the long run for improved supply chain security we should consider doing the same thing we do for cargo-sort, fetch the binary by known URL and known Hash.

.github/actions/setup-job/action.yml
```diff
@@ -70,11 +72,19 @@ runs:
       if: inputs.deps
       run: |
         if [[ "${{ inputs.dart }}" == "true" || "${{ inputs.flutter }}" == "true" ]] && [[ "${{ inputs.rust }}" == "true" ]]; then
-          # ffigen dep, not handled by `just deps` as it's OS specific.
+          # ffigen dep, not handled by `just` as it's OS specific.
           sudo apt-get install libclang-10-dev
-          just deps
```
rustonaut, line 75: `just deps` is more then just dart and flutter deps, but also `install-async-bindgen`

.github/workflows/release.yml
```diff
@@ -10,7 +10,6 @@ permissions:
   contents: read
 
 env:
-  FLUTTER_VERSION: '2.8.1'
   DART_WORKSPACE: ${{ github.workspace }}/discovery_engine
   FLUTTER_WORKSPACE: ${{ github.workspace }}/discovery_engine_flutter
```
rustonaut, line 14: I just realized we can also delete the workspace env variables as they also should be in `.env`.

justfile
```diff
@@ -37,7 +37,7 @@ flutter-deps:
     flutter pub get
 
 # Fetches rust dependencies
-rust-deps:
+rust-deps: install-async-bindgen
```
rustonaut, line 40: It's more  a dart dependency, tbh.

Like it's used only by dart not by rust, it just happen to be written in rust.

But one option worth considering when (if) improving the story around async bindgen is to port the CLI tool to dart, making it a `build_runner` based tool.

.github/actions/setup-job/action.yml
```diff
@@ -33,14 +35,14 @@ runs:
 
     - name: Install flutter
       uses: subosito/flutter-action@5e1529bc12adae5b26f30f0a4046ac743d6f0c5d # v2.3.0
-      if: inputs.flutter
+      if: inputs.flutter != 'false'
```
rustonaut, line 38: As long as `inputs.flutter` can only be `true`/`false` this is not necessary.

Or to avoid future problems we could use `!= 'false'` everywhere `inputs.flutter/dart/rust` is used.

.github/actions/setup-job/action.yml
rustonaut: I saw it's handled different in the `justfile` now, through then `sudo apt-get install libclang-10-dev` probably should move into the `[ "${{ inputs.dart }}" == "true" ]` branch.

Also when flutter was added in a previous commit, it messed up the rust checks it should be `"${{ inputs.rust }}" != "false"` as rust can be `true`,`false` or `nightly`. It just happens to not cause problems currently as we only use `nightly` with `deps == false`.

.github/actions/setup-job/action.yml
acrrd: with out this it was always installing the flutter stuff, not sure why.

Robert-Steiner: i think the problem is that `default` is always a string 
https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#inputsinput_iddefault

justfile
acrrd: I didn't put for dart because sometimes we install only dart stuff on some job without rust. Otherwise we have that dart = true also imply rust = true.
 
Let's see about rewriting it because if we move the message handler directly in rust we might not need it.

.github/actions/setup-job/action.yml
acrrd: I gave a quick look, it does make the request and gets the download link from the response.

.github/actions/setup-job/action.yml
rustonaut: Probably, before for the "false" case I simply had no default specified.

Githubs type handling is so strange, it's like JSON typed, but then requires you to specify thing stringly typed and sometimes it's basically stringly typed. :pouting_cat: 

Like:

>  If the types do not match, GitHub coerces the type to a number. GitHub casts data types to a number using these conversions:

except somewhere below this is hidden:

> Some functions cast values to a string to perform comparisons.

and boolean conversion is not even mentioned...

acrrd:
I also remove duplicate listing from the path filters

rustonaut:
`dart-ok` fails because it doesn't handle the case where some of the dart actions are skipped.

.github/workflows/ci_reusable_wf.yml
```diff
@@ -136,6 +139,7 @@ jobs:
   dart-ok:
     name: dart-ok
```
Robert-Steiner, line 140: not related to this pr but we could also remove the names for the jobs where the id and name are the same

Robert-Steiner (commented):
looks good

rustonaut (approved):
lgtm

Robert-Steiner:
bors try

PR #129, opened 2022-02-10
[Merged by Bors] - Ty-2381 Save serialized Engine state

biernacki:
[Jira ref](https://xainag.atlassian.net/browse/TY-2381)

This PR adds `serialize` method to the `Engine` interface and updates the managers so that after each call to the engine that changes its state, the state is serialized and stored in a database via `EngineStateRepository`.


janpetschexain (approved):
lgtm (although i'm not familiar with the whole repository topic, so i guess it makes sense if somebody else also has a quick look)

acrrd (commented):
looks good

biernacki:
bors merge

PR #131, opened 2022-02-10
TY-2382 Pass serialized state when initializing the engine

biernacki:
[Jira ref](https://xainag.atlassian.net/browse/TY-2382)

### DIFF
--- a/discovery_engine/lib/src/domain/engine/engine.dart
+++ b/discovery_engine/lib/src/domain/engine/engine.dart
@@ -14,6 +14,8 @@
 
 import 'dart:typed_data' show Uint8List;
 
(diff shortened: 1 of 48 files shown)


discovery_engine/lib/src/domain/engine/engine.dart
```diff
@@ -23,6 +25,9 @@ import 'package:xayn_discovery_engine/src/domain/models/unique_id.dart'
 
 /// Interface to Discovery Engine core.
 abstract class Engine {
+  /// NOTE: this is to enforce subclasses to have these parameters during instantiation
```
acrrd, line 28: We don't need to enforce this. If the mock does not need to work is ok.
The real implementation will have it.

acrrd:
I think we can close this

PR #132, opened 2022-02-10
[Merged by Bors] - TY-2362 Build flutter example for android and iOS on bors CI

x3ro:
[TY-2362]

[TY-2362]: https://xainag.atlassian.net/browse/TY-2362?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/ci_bors.yml
+++ b/.github/workflows/ci_bors.yml
@@ -22,8 +22,14 @@ jobs:
   build-android-libs:
     uses: ./.github/workflows/ci_build_android.yml
 
+  build-flutter-example:
+    needs: build-ios-libs
+    uses: ./.github/workflows/ci_flutter_example.yml
+    with:
+      artifact-dir-base-ios: ${{ needs.build-ios-libs.outputs.artifact-dir-base }}
+
   ci-ok:
-    needs: [dev-ci, build-ios-libs, build-android-libs]
+    needs: [dev-ci, build-ios-libs, build-android-libs, build-flutter-example]
     runs-on: ubuntu-20.04
     timeout-minutes: 5
     if: always()
@@ -33,7 +39,8 @@ jobs:
           if [[
             "${{ needs.dev-ci.outputs.wf-result }}" == "success" &&
             "${{ needs.build-ios-libs.result }}" == "success" &&
-            "${{ needs.build-android-libs.result }}" == "success"
+            "${{ needs.build-android-libs.result }}" == "success" &&
+            "${{ needs.build-flutter-example.result }}" == "success"
           ]]; then
             exit 0
           else
--- a/.github/workflows/ci_flutter_example.yml
+++ b/.github/workflows/ci_flutter_example.yml
@@ -1,0 +1,54 @@
+name: Build flutter example
+
+on:
+  workflow_call:
+    inputs:
+      artifact-dir-base-ios:
+        description: The base name of artifact directory (without the target suffix)
+        required: true
+        type: string
+
+permissions:
+  contents: read
+
+jobs:
+  flutter-build-example:
+    # The reason we don't need the Android libraries here is that, for Android, we have
+    # a dynamic library that is loaded at runtime, while for iOS we have a static library
+    # that needs to exist during the linking process.
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 30
+    strategy:
+      matrix:
+        target: [android, ios]
+        include:
+          - target: android
+            os: ubuntu-20.04
+            cmd: just flutter-build apk
+          - target: ios
+            os: macos-11
+            cmd: just flutter-build ios --no-codesign
+    steps:
+      - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # v2.4.0
+
+      - name: Setup the CI.
+        uses: ./.github/actions/setup-job
+        with:
+          rust: true
+          flutter: true
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Download artifacts
+        if: matrix.target == 'ios'
+        id: artifacts
+        uses: actions/download-artifact@f023be2c48cc18debc3bacd34cb396e0295e2869 # v2.1.0
+        with:
+          path: ${{ runner.temp }}/artifacts/
+
+      - name: Copy artifacts
+        if: matrix.target == 'ios'
+        working-directory: ${{ steps.artifacts.outputs.download-path }}
+        run: cp -R ${{ inputs.artifact-dir-base-ios }}-*/* ${{ github.workspace }}/${{ env.FLUTTER_WORKSPACE }}/ios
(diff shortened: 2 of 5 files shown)


justfile
```diff
@@ -219,6 +223,26 @@ _dry-run-release: clean deps dart-build
...
+        cargo ndk -t $(echo "{{target}}" | sed 's/[^ ]* */&/g') -p $ANDROID_PLATFORM_VERSION -o "{{justfile_directory()}}/$FLUTTER_WORKSPACE/android/src/main/jniLibs" build --release
+
+compile-android-local:
+    #!/usr/bin/env sh
+    set -eu
+    for TARGET in $ANDROID_TARGETS; do
+        {{just_executable()}} compile-android $TARGET
+    done
+
+flutter-run: dart-build
```
x3ro, line 238: Just moving a [previous discussion](https://github.com/xaynetwork/xayn_discovery_engine/pull/126#discussion_r803738386) into new PR
> @Robert-Steiner: In general we try to keep the PRs small and self-contained but you can also keep it in the pr.
But in this case at least flutter-build should also call compile-android-local if the target is apk. For flutter-run we could build both ios and android or also add a target there. wdyt?

x3ro:
bors try

x3ro:
bors try-

x3ro:
bors try-

x3ro:
bors try

x3ro:
bors try-

x3ro:
bors try

x3ro:
bors try-

x3ro:
bors try

justfile
x3ro: I'll double check how this currently works. I'm assuming it works in the CI because we build the libraries in the step before.

x3ro:
* The `ci_build_ios` workflow is currently the only one where we depend on the `macos-11` image. Building the flutter example as part of this workflow has the benefit that we're not spreading the use of this special case.
* Building the flutter example adds ~ 3 minutes to the workflow, which I assume makes no relevant difference for a release build
* Since the builds run in parallel, there's not going to be a significant difference in wall time, even though we build it for all targets
* We would be replacing what's currently a couple of lines with two completely new workflows (flutter iOS, flutter Android)

While I see the point that, strictly speaking, the two could definitely be separated, I don't think the cons outweigh the benefits here, so I would only go down this route if you would otherwise not sign off on the change^^

acrrd:
We need to be careful that building the example does not add files that we don't want to export on the release repo.

x3ro:
bors try

x3ro:
bors try-

x3ro:
bors try

x3ro:
bors try-

x3ro:
bors try

x3ro:
bors try-

x3ro:
bors try

.github/workflows/ci_flutter.yml
```diff
@@ -0,0 +1,53 @@
+name: Build flutter example
```
Robert-Steiner, line 1: what about `ci_build_flutter_example` or `ci_flutter_example` as file name?

.github/workflows/ci_flutter.yml
```diff
@@ -0,0 +1,53 @@
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  flutter-build-example:
+    # The Android dependencies are built automatically when running the flutter build,
```
Robert-Steiner, line 15: The reason we don't need the Android libraries here is that for Android we have a dynamic library that is loaded at runtime, while for iOS we have a static library that needs to exist during the linking process. The android example will build but fail when you try to run the example

justfile
```diff
@@ -267,6 +267,14 @@ compile-ios-ci target prod_flag="\"\"": _codegen-order-workaround
         {{just_executable()}} _compile-ios {{target}}
     fi
 
+flutter-run: dart-build
+    cd "$FLUTTER_WORKSPACE/example" && \
+        flutter run
+
+flutter-build target *args: dart-build
```
Robert-Steiner, line 274: For local development it would be really handy if we also automatically build the right rust libraries (depending on `target`). However, the recipe then no longer works on the CI :/ so we would also take that into account. 
same for flutter run
maybe we can create a ticket in jira to improve it later

.github/workflows/ci_flutter.yml
```diff
@@ -0,0 +1,53 @@
+name: Build flutter example
+
+on:
+  workflow_call:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
Robert-Steiner, line 8: I think we should remove this until they update the documentation that explains how this works in a reusable workflow. Also, we already define a concurrency group in the calling workflow, so that workflow will already be canceled when a new version is pushed

Robert-Steiner (commented):
Some minor tweaks on the names the rest looks very good 👍 
We can make the improvement with the justfile in another pr

x3ro:
bors try-

x3ro:
bors try

x3ro:
bors try-

x3ro:
bors try

x3ro:
bors try-

x3ro:
bors try

x3ro:
bors try

x3ro:
bors try

x3ro:
bors try-

x3ro:
bors try

.github/workflows/ci_flutter_example.yml
```diff
@@ -0,0 +1,55 @@
+name: Build flutter example
+
+on:
+  workflow_call:
+    inputs:
+      artifact-dir-base-ios:
+        description: The base name of artifact directory (without the target suffix)
+        default: true
```
Robert-Steiner, line 8: let's remove the default if we don't have a good value for that
```suggestion
```

.github/workflows/ci_bors.yml
```diff
@@ -33,7 +39,8 @@ jobs:
           if [[
             "${{ needs.dev-ci.outputs.wf-result }}" == "success" &&
             "${{ needs.build-ios-libs.result }}" == "success" &&
-            "${{ needs.build-android-libs.result }}" == "success"
+            "${{ needs.build-android-libs.result }}" == "success" &&
+            "${{ needs.build-flutter.result }}" == "success"
```
Robert-Steiner, line 43:
```suggestion
            "${{ needs.build-flutter-example.result }}" == "success"
```

Robert-Steiner (approved):
only two things left

x3ro:
bors try-

x3ro:
bors merge

PR #133, opened 2022-02-11
[Merged by Bors] - TY-2391 Take model assets from app configuration

biernacki:
[Jira ref](https://xainag.atlassian.net/browse/TY-2391)

In order to use the AI assets bundled with the application, we need to copy them first to a directory accessible by the worker isolate, because `rootBundle` can't be used in an isolate (other then the main one). This PR adds a `FlutterAiAssetCopier` to copy the AI assets from the app bundle to a directory used by the `DataProvider`.

The other option would be to move the code used to retrieve the AI assets entirely to the main isolate and just pass the `SetupData` with the `Configuration` during engine init. This would require a refactor (maybe not a big issue, depends also a bit on needed tests adjustments), but also then limiting ourselves to use the assets only from the bundle replacing the `HttpAssetFetcher`.

This way (implemented by this PR) we can start with assets bundled with the app, but then fetch updated assets on the fly over the network if needed, without the need to release a new version of the app.

### DIFF
--- a/discovery_engine_flutter/example/.gitignore
+++ b/discovery_engine_flutter/example/.gitignore
@@ -47,5 +47,11 @@ app.*.map.json
 
 # Assets related
 /assets/*
-# exception to the rule
-!/assets/.gitkeep
+/assets/kpe_v0000/*
+/assets/smbert_v0000/*
+
+# Keep the filder structure for assets
+!/assets/kpe_v0000
+!/assets/kpe_v0000/.gitkeep
+!/assets/smbert_v0000
+!/assets/smbert_v0000/.gitkeep
--- a/discovery_engine_flutter/example/.gitignore
+++ b/discovery_engine_flutter/example/.gitignore
@@ -44,3 +44,8 @@ app.*.map.json
 /android/app/debug
 /android/app/profile
 /android/app/release
+
+# Assets related
+/assets/*
+# exception to the rule
+!/assets/.gitkeep
--- a/discovery_engine_flutter/example/.gitignore
+++ b/discovery_engine_flutter/example/.gitignore
@@ -47,11 +47,11 @@ app.*.map.json
 
 # Assets related
 /assets/*
-/assets/kpe_v0000/*
+/assets/kpe_v0001/*
 /assets/smbert_v0000/*
 
 # Keep the filder structure for assets
-!/assets/kpe_v0000
-!/assets/kpe_v0000/.gitkeep
+!/assets/kpe_v0001
+!/assets/kpe_v0001/.gitkeep
 !/assets/smbert_v0000
 !/assets/smbert_v0000/.gitkeep
--- a/discovery_engine_flutter/example/.gitignore
+++ b/discovery_engine_flutter/example/.gitignore
@@ -46,11 +46,9 @@ app.*.map.json
 /android/app/release
 
 # Assets related
-/assets/*
-/assets/kpe_v0001/*
-/assets/smbert_v0000/*
+/assets/**/*
 
-# Keep the filder structure for assets
+# Keep the folder structure for assets
 !/assets/kpe_v0001
 !/assets/kpe_v0001/.gitkeep
 !/assets/smbert_v0000
--- a/discovery_engine_flutter/example/download_assets.sh
+++ b/discovery_engine_flutter/example/download_assets.sh
@@ -1,0 +1,38 @@
+#!/bin/sh
+
+set -eu
+
+# We can't use `pushd` or `readlink -f` so we
+# fall back to this.
+CALLING_BASE_DIR="$(pwd -L)"
+
+# path to the directory where this file is
+SELF_DIR_PATH="$(dirname "$0")"
+
+# in this way we can call the script from different directory
+# but the data should go in the correct destination
+DATA_DIR="$SELF_DIR_PATH/assets"
+
+CHECKSUM_FILE="sha256sums"
+
+download()
+{
+  cd "$CALLING_BASE_DIR"
+  NAME="$1"
+  VERSION="$2"
+  ARCHIVE_BASENAME="${NAME}_$VERSION"
+  ARCHIVE_NAME="$ARCHIVE_BASENAME.tgz"
+  URL="http://s3-de-central.profitbricks.com/xayn-yellow-bert/$NAME/$ARCHIVE_NAME"
+
+  curl "$URL" -o "$DATA_DIR/$ARCHIVE_NAME"
+
+  cd "$DATA_DIR"
+  tar -zxf "$ARCHIVE_NAME"
+
+  # check content
+  cd "$ARCHIVE_BASENAME"
+  shasum -c "$CHECKSUM_FILE"
+}
+
+download smbert v0000
+download kpe v0000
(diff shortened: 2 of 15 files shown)


discovery_engine_flutter/example/download_assets.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/sh
```
biernacki, line 1: script is copied from `xayn_ai` repo, with just small modifications

discovery_engine_flutter/example/download_assets.sh
```diff
@@ -0,0 +1,38 @@
...
+  cd "$DATA_DIR"
+  tar -zxf "$ARCHIVE_NAME"
+
+  # check content
+  cd "$ARCHIVE_BASENAME"
+  shasum -c "$CHECKSUM_FILE"
+}
+
+download smbert v0000
+download kpe v0000
```
biernacki, line 38: in `xayn_ai` kpe is at version v0001, but in discovery engine manifest, we fetch version v0000, so please confirm if this is correct

janpetschexain: `v0001` is the latest and correct version. i must have missed the changes in `discovery_engine` when i changed it for `xayn_ai`, sorry for that :/

discovery_engine_flutter/example/download_assets.sh
```diff
@@ -0,0 +1,38 @@
...
+CHECKSUM_FILE="sha256sums"
+
+download()
+{
+  cd "$CALLING_BASE_DIR"
+  NAME="$1"
+  VERSION="$2"
+  ARCHIVE_BASENAME="${NAME}_$VERSION"
+  ARCHIVE_NAME="$ARCHIVE_BASENAME.tgz"
+  URL="http://s3-de-central.profitbricks.com/xayn-yellow-bert/$NAME/$ARCHIVE_NAME"
```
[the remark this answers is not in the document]
Robert-Steiner: yes

discovery_engine_flutter/example/download_assets.sh
biernacki: no worries :) does it mean we should update the version also in our manifest file `asset_manifest.json`?

discovery_engine_flutter/example/download_assets.sh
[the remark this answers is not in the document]
Robert-Steiner: although do you want to download the zip files or the files that are in the manifest?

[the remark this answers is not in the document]
biernacki: for this script we want to fetch non-fragmented files

discovery_engine_flutter/example/download_assets.sh
biernacki: I've updated the kpe to v0001 everywhere

discovery_engine_flutter/example/.gitignore
```diff
@@ -44,3 +44,14 @@ app.*.map.json
 /android/app/debug
 /android/app/profile
 /android/app/release
+
+# Assets related
+/assets/*
+/assets/kpe_v0001/*
+/assets/smbert_v0000/*
+
+# Keep the filder structure for assets
```
janpetschexain, line 53:
```suggestion
# Keep the folder structure for assets
```

discovery_engine_flutter/example/.gitignore
```diff
@@ -44,3 +44,14 @@ app.*.map.json
 /android/app/debug
 /android/app/profile
 /android/app/release
+
+# Assets related
+/assets/*
+/assets/kpe_v0001/*
+/assets/smbert_v0000/*
```
janpetschexain, line 51: isn't this ignored anyways due to the glob ignore in l.49?

biernacki: it wasn't ignored, but I've changed it to a onliner now

discovery_engine_flutter/lib/src/assets/asset_copier.dart
```diff
@@ -0,0 +1,61 @@
...
+  final String appDir;
+  final AssetLoader _loadAsset;
+
+  FlutterBundleAssetCopier({
+    required this.appDir,
+    required this.bundleAssetsPath,
+    AssetLoader? loadAsset,
+  })  : assert(appDir.isNotEmpty),
+        assert(bundleAssetsPath.isNotEmpty),
+        _loadAsset = loadAsset ?? rootBundle.load;
```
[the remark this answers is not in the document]
acrrd: >  it's weird that dart doesn't allow for direct assignment of private fields in an optional named argument list oO so i'd say up to you if you want to change it.

It's not that weird, if it's private it should not affect the public interface.

discovery_engine_flutter/example/.gitignore
```diff
@@ -44,3 +44,12 @@ app.*.map.json
 /android/app/debug
 /android/app/profile
 /android/app/release
+
+# Assets related
+/assets/**/*
+
+# Keep the folder structure for assets
```
acrrd, line 51: Why we need this?

biernacki: the flutter analyzer is complaining if those directories don't exist, because they are specified in example's `pubspec.yaml` as bundled assets directories

discovery_engine_flutter/lib/src/assets/asset_copier.dart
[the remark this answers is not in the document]
janpetschexain: not true, for example in the following it's perfectly fine to have a private field in a public interface, the limitation only exists for optional private arguments and hence is quite inconsistent.
```dart
class MyClass {
  final int _myField;
  MyClass(this._myField);
}
```

discovery_engine_flutter/example/.gitignore
acrrd: Can we specify only `assets/models/` in the `pubspec.yaml` and put the models in there or it will not be possible to get them out of the bundle without the explicit name? 
Having to keep this is really prone to errors when we change model

discovery_engine_flutter/lib/src/assets/asset_copier.dart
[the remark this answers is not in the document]
acrrd: Because that will not break the public interface. In this case you can rename `_myField` and all the code will continue to work, this does not holds for named parameters.

discovery_engine_flutter/example/.gitignore
biernacki: we could put everything in `assets/models/` but it would need to be a flat structure. To bundle files from subdirectories we need to specify each directory in `pubspec.yaml` like so:

```yaml
flutter:
  assets:
    - directory/ # this includes all the files in the directory, but not in the subdirectory
    - directory/subdirectory/ # we need to additionally add this to include all the files from subdirectory
```

acrrd: Let's keep it like this then :/

Robert-Steiner (approved):
looks good 🚀

janpetschexain (approved):

biernacki:
bors merge

PR #134, opened 2022-02-11
[Merged by Bors] - TY-2360 Build android libs on bors CI

x3ro:
[TY-2360]

[TY-2360]: https://xainag.atlassian.net/browse/TY-2360?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.env
+++ b/.env
@@ -11,6 +11,7 @@ FLUTTER_VERSION=2.5.3
 
 ANDROID_TARGETS="arm64-v8a x86_64 x86"
 ANDROID_PLATFORM_VERSION="21"
+
 IOS_TARGETS="aarch64-apple-ios x86_64-apple-ios"
 IOS_LIB_BASE="libxayn_discovery_engine_bindings"
 
--- a/.github/actions/setup-job/action.yml
+++ b/.github/actions/setup-job/action.yml
@@ -16,6 +16,10 @@ inputs:
   rust-target:
     description: 'Install an additional rust target'
     required: false
+  android:
+    description: 'Install tooling for cross-compiling for Android'
+    required: false
+
   deps:
     description: 'Fetch/Install dependencies'
     default: "true"
@@ -49,6 +53,13 @@ runs:
       with:
         flutter-version: ${{ env.FLUTTER_VERSION }}
         cache: true
+
+    - name: Install cargo-ndk
+      shell: bash
+      if: inputs.android != 'false'
+      run: |
+        echo "${{ github.workspace }}/cargo-installs/bin" >> $GITHUB_PATH
+        cargo install --git https://github.com/xaynetwork/cargo-ndk.git cargo-ndk
 
     - name: Setup flutter dependencies
       shell: bash
--- a/.github/workflows/ci_bors.yml
+++ b/.github/workflows/ci_bors.yml
@@ -19,15 +19,22 @@ jobs:
   build-ios-libs:
     uses: ./.github/workflows/ci_build_ios.yml
 
+  build-android-libs:
+    uses: ./.github/workflows/ci_build_android.yml
+
   ci-ok:
-    needs: [dev-ci, build-ios-libs]
+    needs: [dev-ci, build-ios-libs, build-android-libs]
     runs-on: ubuntu-20.04
     timeout-minutes: 5
     if: always()
     steps:
       - name: CI OK Testing
         run: |
-          if [[ ${{ needs.dev-ci.outputs.wf-result }} == "success" && "${{ needs.build-ios-libs.result }}" == "success" ]]; then
+          if [[
+            "${{ needs.dev-ci.outputs.wf-result }}" == "success" &&
+            "${{ needs.build-ios-libs.result }}" == "success" &&
+            "${{ needs.build-android-libs.result }}" == "success"
+          ]]; then
             exit 0
           else
             exit 1
--- a/.github/workflows/ci_build_android.yml
+++ b/.github/workflows/ci_build_android.yml
@@ -1,0 +1,47 @@
+name: Build Android library
+
+on:
+  workflow_call:
+    outputs:
+      artifact-dir-base:
+        description: The base name of artifact directory (without the target suffix)
+        value: ${{ jobs.build-android-libs.outputs.artifact-dir-base }}
+
+jobs:
+  build-android-libs:
+    runs-on: ubuntu-20.04
+    timeout-minutes: 35
+    strategy:
+      matrix:
+        target: [
+          "aarch64-linux-android",
+          "armv7-linux-androideabi",
+          "x86_64-linux-android",
+          "i686-linux-android",
+        ]
+    env:
+      ARTIFACT_DIR_BASE: ${{ github.job }}
+    outputs:
+      artifact-dir-base: ${{ env.ARTIFACT_DIR_BASE }}
+    steps:
+      - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # v2.4.0
+
+      - name: Setup the rust CI.
+        uses: ./.github/actions/setup-job
+        with:
+          rust: true
+          android: true
+          rust-target: ${{ matrix.target }}
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build Android library
+        run: |
+          just compile-android "${{ matrix.target }}"
+
+      - name: Upload library artifacts
+        uses: actions/upload-artifact@82c141cc518b40d92cc801eee768e7aafc9c2fa2 # v2.3.1
+        with:
+          name: ${{ env.ARTIFACT_DIR_BASE }}-${{ matrix.target }}-${{ github.sha }}
+          retention-days: 1
+          if-no-files-found: error
+          path: ${{ env.FLUTTER_WORKSPACE }}/android/src/main/jniLibs
--- a/justfile
+++ b/justfile
@@ -219,7 +219,7 @@ _dry-run-release: clean deps dart-build
 dry-run-release:
      @CI=true {{just_executable()}} _dry-run-release
 
-compile-android target:
+compile-android target: _codegen-order-workaround
     # See also: https://developer.android.com/studio/projects/gradle-external-native-builds#jniLibs
     cd "$RUST_WORKSPACE"; \
         cargo ndk -t $(echo "{{target}}" | sed 's/[^ ]* */&/g') -p $ANDROID_PLATFORM_VERSION -o "{{justfile_directory()}}/$FLUTTER_WORKSPACE/android/src/main/jniLibs" build --release


.github/workflows/ci.yml
```diff
@@ -13,19 +13,81 @@ permissions:
   contents: read
```
acrrd, line 13: This is not where we want this but we want it in `ci_bors.yml`.

discovery_engine_core/rust-toolchain.toml
```diff
@@ -1,3 +1,4 @@
 [toolchain]
 channel = "1.55.0"
 profile = "default"
+targets = ["aarch64-apple-ios", "x86_64-apple-ios"]
```
acrrd, line 4: This will force to install the targets in evrey job but we want to have them only in few of them so we can have an action or step to do it.

.github/workflows/ci.yml
x3ro: This is only so I can do test runs while developing.

x3ro:
bors try

x3ro:
bors try

x3ro:
bors try-

x3ro:
bors try

x3ro:
bors try

.env
```diff
@@ -9,8 +9,13 @@ FLUTTER_WORKSPACE=discovery_engine_flutter
 DART_VERSION=2.14.4
 FLUTTER_VERSION=2.5.3
 
-ANDROID_TARGETS="arm64-v8a x86_64 x86"
```
x3ro, line 12: I've removed this target to improve build times for now, as I don't really know if we need a non 64-bit Intel target, considering that we're e.g. not supporting armv7. Let me know if that is not a good idea.

.github/actions/setup-job/action.yml
```diff
@@ -50,6 +54,18 @@ runs:
         flutter-version: ${{ env.FLUTTER_VERSION }}
         cache: true
 
+    - name: Install Android NDK
+      if: inputs.android != 'false'
+      uses: nttld/setup-ndk@4a541f506ff27d545f89e860b7cdc86efc3ca496 # HEAD 2022-02-15
+      with:
+        ndk-version: ${{ env.ANDROID_NDK_VERSION }}
```
Robert-Steiner, line 61: We shouldn't need that. The Ubuntu runner should have ndk installed [ref](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md#environment-variables-3)

.github/actions/setup-job/action.yml
```diff
@@ -50,6 +54,18 @@ runs:
...
+
+    - name: Install cargo-ndk
+      shell: bash
+      if: inputs.android != 'false'
+      run: |
+        cargo install --git https://github.com/xaynetwork/cargo-ndk.git cargo-ndk
```
Robert-Steiner, line 67:
```suggestion
      run: cargo install --git https://github.com/xaynetwork/cargo-ndk.git cargo-ndk
```

.github/workflows/ci_bors.yml
```diff
@@ -19,15 +19,22 @@ jobs:
...
     runs-on: ubuntu-20.04
     timeout-minutes: 5
     if: always()
     steps:
       - name: CI OK Testing
         run: |
-          if [[ ${{ needs.dev-ci.outputs.wf-result }} == "success" && "${{ needs.build-ios-libs.result }}" == "success" ]]; then
+          if [[ \
+            "${{ needs.dev-ci.outputs.wf-result }}" == "success" && \
+            "${{ needs.build-ios-libs.result }}" == "success" &&
```
Robert-Steiner, line 35: do we also need a `\` here or does it work without?
```suggestion
            "${{ needs.build-ios-libs.result }}" == "success" && \
```

.github/workflows/ci_build_android.yml
```diff
@@ -0,0 +1,50 @@
...
+        description: The base name of artifact directory (without the target suffix)
+        value: ${{ jobs.build-android-libs.outputs.artifact-dir-base }}
+
+jobs:
+  build-android-libs:
+    runs-on: ubuntu-20.04
+    timeout-minutes: 35
+    strategy:
+      matrix:
+        target: ["arm64-v8a", "x86_64"]
```
Robert-Steiner, line 16: In cargo ndk we can also use the [rust triple](https://github.com/bbqsrc/cargo-ndk/blob/f91b56d4114dd18fa7d2fa3b484fb214288e6450/src/meta.rs#L88). 
```suggestion
        target: ["aarch64-linux-android", "x86_64-linux-android"]
```

.github/workflows/ci_build_android.yml
```diff
@@ -0,0 +1,50 @@
...
+    steps:
+      - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # v2.4.0
+
+      - name: Setup the rust CI.
+        uses: ./.github/actions/setup-job
+        with:
+          rust: true
+          dart: true
+          android: true
+          rust-target: "aarch64-linux-android x86_64-linux-android"
```
Robert-Steiner, line 30: with the rust triple that can be then written as
```suggestion
          rust-target: ${{ matrix.target }}
```

.github/workflows/ci_build_android.yml
```diff
@@ -0,0 +1,50 @@
...
+    timeout-minutes: 35
+    strategy:
+      matrix:
+        target: ["arm64-v8a", "x86_64"]
+    env:
+      ARTIFACT_DIR_BASE: ${{ github.job }}
+    outputs:
+      artifact-dir-base: ${{ env.ARTIFACT_DIR_BASE }}
+    steps:
+      - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # v2.4.0
```
Robert-Steiner, line 22: @acrrd not part of this pr but i think we can also move the checkout to `setup-job`

.github/workflows/ci_build_android.yml
```diff
@@ -0,0 +1,50 @@
...
+        with:
+          rust: true
+          dart: true
+          android: true
+          rust-target: "aarch64-linux-android x86_64-linux-android"
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build Android library
+        run: |
+          export PATH="$(pwd)/cargo-installs/bin:${PATH}"
```
Robert-Steiner, line 35: why do we need that?

.github/workflows/ci_build_android.yml
```diff
@@ -0,0 +1,50 @@
...
+          just compile-android "${{ matrix.target }}"
+
+      - name: Prepare lib for upload
+        id: lib
+        run: |
+          FILE=${{ env.FLUTTER_WORKSPACE }}/android/src/main/jniLibs/${{ matrix.target }}/libxayn_discovery_engine_bindings.so
```
Robert-Steiner, line 41: we can also use 
```suggestion
          FILE=${{ env.FLUTTER_WORKSPACE }}/android/src/main/jniLibs/${{ matrix.target }}/${{ env.IOS_LIB_BASE }}.so
```
but then we should better remame it to `LIB_BASE` or something like that

justfile
```diff
@@ -219,7 +219,7 @@ _dry-run-release: clean deps dart-build
 dry-run-release:
      @CI=true {{just_executable()}} _dry-run-release
 
-compile-android target:
+compile-android target: dart-build
```
Robert-Steiner, line 222: That probably fails because we need to run cargo check before in order to run the build scripts. can you please try if this would also work
```suggestion
compile-android target: _codegen-order-workaround
```

.github/workflows/ci_build_android.yml
```diff
@@ -0,0 +1,50 @@
...
+          rust: true
+          dart: true
+          android: true
+          rust-target: "aarch64-linux-android x86_64-linux-android"
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build Android library
+        run: |
+          export PATH="$(pwd)/cargo-installs/bin:${PATH}"
+          just compile-android "${{ matrix.target }}"
```
Robert-Steiner, line 36:
```suggestion
          just compile-android ${{ matrix.target }}
```

.github/workflows/ci_build_android.yml
```diff
@@ -0,0 +1,50 @@
...
+    outputs:
+      artifact-dir-base: ${{ env.ARTIFACT_DIR_BASE }}
+    steps:
+      - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # v2.4.0
+
+      - name: Setup the rust CI.
+        uses: ./.github/actions/setup-job
+        with:
+          rust: true
+          dart: true
```
Robert-Steiner, line 28:
```suggestion
```

Robert-Steiner (commented):
looks good! added some small suggestions

.github/workflows/ci_build_android.yml
acrrd: On the CI we want build for all the android targets: aarch64-linux-android, armv7-linux-androideabi, i686-linux-android, x86_64-linux-android

x3ro: Mmh, I think I was told that we don't support armv7. I'm assuming that information is incorrect, then? Also what's the point of building Intel non-64-bit architectures?

Robert-Steiner: for ios we don't support armv7 because 32Bit is no longer supported since ios 11 i think

acrrd: Someone still have an emulator that wants the non 64 version.

.github/workflows/ci_build_android.yml
x3ro: Because otherwise it doesn't find `cargo-ndk`

.github/workflows/ci_build_android.yml
x3ro: I think the general advice that shell linters (e.g. shellcheck) give is to quote variables / dynamic values at all times, to avoid unpleasant surprises. While this may not be needed here, I mostly try to follow that pattern :)

x3ro:
bors try

.github/workflows/ci_build_android.yml
Robert-Steiner: that is weird because we never had to do this before 
also cargo-ndk shound't be install in `cargo-installs/bin` but in `~/.cargo/bin/`
`cargo-installs/bin` that is only for `async-bindgen-gen-dart`
can you try a run without that line please

.github/workflows/ci_build_android.yml
Robert-Steiner: you normally need quotes if the variable can be empty, contain spaces or wildcards. but I'm ok if you leave the quotes here

x3ro:
bors try

.github/workflows/ci_build_android.yml
```diff
@@ -0,0 +1,53 @@
...
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build Android library
+        run: |
+          just compile-android "${{ matrix.target.ndk }}"
+
+      - name: Prepare lib for upload
+        id: lib
+        run: |
+          FILE=${{ env.FLUTTER_WORKSPACE }}/android/src/main/jniLibs/${{ matrix.target.ndk }}/libxayn_discovery_engine_bindings.so
```
Robert-Steiner, line 44: i thing i forgot. to preserve the folder that cargo ndk creates for us we should change this to 
```suggestion
          FILE=${{ env.FLUTTER_WORKSPACE }}/android/src/main/jniLibs
```
so it will upload 
```
<arch_folder>/lib.so
```
instead of 
```
lib.so
```

we can then also remove this step and put this line directly to the upload action path

.github/workflows/ci_build_android.yml
x3ro: Breaks [here](https://github.com/xaynetwork/xayn_discovery_engine/runs/5214694142?check_suite_focus=true#step:4:25)

.github/workflows/ci_build_android.yml
x3ro: Outdated now because we'll upload the entire folder.

x3ro:
bors try

x3ro:
bors try-

x3ro:
bors try

.github/workflows/ci_build_android.yml
Robert-Steiner: ah i think it because of `CARGO_INSTALL_ROOT` in `.env`. that will install cargo-ndk in `cargo-installs/bin`. @acrrd, @rustonaut is `cargo-installs/bin` only there for `async-bindgen-gen-dart` or was there the plan to also use it also for cargo-ndk, cargo sort,... ?

.github/workflows/ci_build_android.yml
x3ro: Done

x3ro:
@acrrd @Robert-Steiner I think I've addressed everything, could you take another look?

x3ro:
bors try

.github/actions/setup-job/action.yml
```diff
@@ -50,6 +54,18 @@ runs:
         flutter-version: ${{ env.FLUTTER_VERSION }}
         cache: true
 
+    # - name: Install Android NDK
+    #   if: inputs.android != 'false'
+    #   uses: nttld/setup-ndk@4a541f506ff27d545f89e860b7cdc86efc3ca496 # HEAD 2022-02-15
+    #   with:
+    #     ndk-version: ${{ env.ANDROID_NDK_VERSION }}
```
Robert-Steiner, line 61:
```suggestion
```

.env
```diff
@@ -11,6 +11,11 @@ FLUTTER_VERSION=2.5.3
 
 ANDROID_TARGETS="arm64-v8a x86_64 x86"
 ANDROID_PLATFORM_VERSION="21"
+# Note that this is _not_ the same as the platform version, and
+# does not need to align with it necessarily.
+# Versions can be found here https://developer.android.com/ndk/downloads
+ANDROID_NDK_VERSION="r23b"
```
Robert-Steiner, line 17:
```suggestion
```

.github/workflows/ci_build_android.yml
```diff
@@ -0,0 +1,48 @@
...
+        value: ${{ jobs.build-android-libs.outputs.artifact-dir-base }}
+
+jobs:
+  build-android-libs:
+    runs-on: ubuntu-20.04
+    timeout-minutes: 35
+    strategy:
+      matrix:
+        target: [
+          { rust: "aarch64-linux-android", ndk: "arm64-v8a" },
```
Robert-Steiner, line 17: I'd rather just use the Rust target triple here to have less code

x3ro: Since this mapping is not exactly obvious, I think it's good to have here in the CI. Especially considering that the rust triplets don't really have meaning within the Android build process, and we'd be naming the zip artifacts based on them. Following your suggestion, we would build a `build-android-libs-aarch64-linux-android.zip`, which would then contain a directory `arm64-v8a`, leaving the mapping to the person who has to deal with the artifact. 

Is this a must-have from your perspective?

Robert-Steiner: > Following your suggestion, we would build a build-android-libs-aarch64-linux-android.zip, which would then contain a directory arm64-v8a, leaving the mapping to the person who has to deal with the artifact.

That's ok I think because these archives are just for sharing the artifacts between github jobs. It is not an artifact to be used by a person. The main purpose of the target name is to create a unique name for these artifacts within the matrix jobs

x3ro:
bors try

.github/workflows/ci_build_android.yml
Robert-Steiner: I think the `CARGO_INSTALL_ROOT` is not a blocker for now. We can do it in a follow-up pr if we find a better solution. 
to make it a bit cleaner we could move this line to the step where we install cargo-ndk. 
but the line would need to change to: [`echo "${{ github.workspace }}/cargo-installs/bin" >> $GITHUB_PATH`](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#adding-a-system-path)
wdyt?

Robert-Steiner (approved):

.github/workflows/ci_build_android.yml
rustonaut: How is that failing, I mean sure `CARGO_INSTALL_ROOT` does install it somewhere else, but it also makes look `cargo` at that place when running a `cargo ..` sub-command :thinking: 

And `just` always loads `.env` files so the env should match.

Am I missing something?

x3ro:
bors try

.github/workflows/ci_build_android.yml
Robert-Steiner: Seems that cargo ignores `CARGO_INSTALL_ROOT` when you try to run an executable :( They also point out adding the path:
```
  Installing cargo-installs/bin/async-bindgen-gen-dart
   Installed package `async-bindgen-gen-dart v0.1.0 (https://github.com/xaynetwork/xayn_async_bindgen.git#fe195d99)` (executable `async-bindgen-gen-dart`)
warning: be sure to add `cargo-installs/bin` to your PATH to be able to run the installed binaries
```

rustonaut: Yes, but it's to run them without going through cargo I thought.

I mean what point is there in being able to change the install root if it then is ignored?

x3ro: I think whatever the reasoning may be, `CARGO_INSTALL_ROOT` appears to not automagically also make the sub-command available to cargo, which is why we need to modify the `PATH` for `cargo-ndk` to work. This is necessary because we also can't invoke the `cargo-ndk` binary directly. I will go ahead and merge once the current state passes, then we can think about a different solution later on.

x3ro:
bors merge

PR #136, opened 2022-02-11
TY-2302 add ffi functions rust

janpetschexain:
**References**

- [TY-2321]
- [TY-2399]
- [TY-2401]
- [TY-2403]
- [TY-2405]
- [TY-2413]

**Summary**

- impl rust ffi functions:
  - some input arguments still need ffi constructors
  - the results still need ffi unwrappers/mappers, alternatively we could return options and pass a `&mut String` to transport the error message but those options would also need unwrappers/mappers so it shouldn't make much of a difference
- core engine:
  - adjust `Engine::from_config()` constructor to use the intended stack ops
  - reduce visibility of private constructors & expose required types
  - add `LockedEngine` wrapper to ensure the `&mut Engine` constraints across the ffi


discovery_engine_core/core/src/engine.rs
```diff
@@ -342,20 +352,24 @@ impl Engine<xayn_ai::ranker::Ranker> {
...
             let ranker = builder.build().map_err(|err| Error::Ranker(err.into()))?;
-            Engine::new(config.into(), ranker, stacks_ops).await
+            Engine::new(config.into(), ranker, stack_ops).await
         }
     }
 }
 
+/// Locked discovery engine.
+#[allow(clippy::module_name_repetitions)]
+pub struct LockedEngine(pub Mutex<Engine<xayn_ai::ranker::Ranker>>);
```
rustonaut, line 365: `Locked` is often seen as "has a the lock" so maybe `SharedEngine` or similar would be better.

discovery_engine_core/bindings/src/lib.rs
```diff
@@ -31,8 +31,83 @@ pub mod types;
 
 #[async_bindgen::api]
 impl AsyncCore {
-    /// Adds two bytes.
-    pub async fn add(x: u8, y: u8) -> u8 {
-        x + y
+    /// Initializes the engine.
+    pub async fn initialize(
+        config: Box<core::InitConfig>,
+        state: &Option<Vec<u8>>,
```
rustonaut, line 37: Iff we always drop the vec in the calling code then having a `Option<Vec<u8>>` would be easier as then rust handles the dropping of it.

rustonaut (commented):
I guess I can make a PR to have the right generic interfaces in ffi.
(like e.g. `Option<Vec<u8>>`, there is `Option`, there is `Vec<u8>` but not yet the combination of both.

discovery_engine_core/bindings/src/lib.rs
rustonaut: Also `Option<&Vec<u8>>` means we don't need to have `Option<Vec<u8>>` specific methods, as it's guaranteed nullptr optimized.

rustonaut: And with `Option<Box<[u8]>>` or ( `Option<&[u8]>`) we would already have all necessary FFI functions in place.

rustonaut: Or  `ptr: *mut u8, len: usize` with nullptr support for `None`.

janpetschexain: reagrding pointer arguments, i wasn't sure if we can use them at all. as we would have to do unsafe conversions in the function body, making the whole function unsafe, is this supported by `async-bindgen`? and what would be the way you intended `async-bindgen` to be used with, pointer or reference arguments?

janpetschexain: > Also Option<&Vec<u8>> means we don't need to have Option<Vec<u8>> specific methods, as it's guaranteed nullptr optimized.

i'll go with this for now.

> Iff we always drop the vec in the calling code then having a Box<Option<Vec<u8>>> would be easier as then rust handles the dropping of it.

i don't know how this will be handled on the dart side, might make sense to change to this later on.

rustonaut: The idea of async-bindgen is to add `async` to a normal C FFI, so sure `unsafe` is supported and meant to be used where necessary.

The only problem could be `Send`, as pointers are by default not send :thinking: .

So for simplicity lets go with `Option<&Vec<u8>>`.

rustonaut: Or `Option<Box<Vec<u8>>>`

discovery_engine_core/bindings/src/async_bindings.rs
```diff
@@ -17,5 +17,5 @@
 //! This exists as a workaround to cbindgen not being able to handle
 //! proc-macro generated code on stable.
 
-#[rustfmt::skip]
```
rustonaut, line 20: This still needs to skip fmt.

discovery_engine/lib/src/ffi/types/document/document_vec.dart
```diff
@@ -15,13 +15,13 @@
 import 'dart:ffi' show Pointer;
 
 import 'package:xayn_discovery_engine/src/ffi/genesis.ffigen.dart'
-    show RustDocument, RustDocumentVec;
+    show RustDocument, RustVecDocument;
```
acrrd, line 18: Why this rename is necessary? It is really common to have `Vec`,`List`,`Buffer`,`Stream`, `Future` at the end.
Also with this rename all file names are out of sync.

rustonaut: I would prefer the naming to change in that way too as it reads better for thinks like `Vec<u8>` (`VecU8` is readable `U8Vec` irritating, for me).

But I also think it's something we can do later.

The file names are out of sync because their are anyway sub-optimal. Like instead of `document_vec.rs` it probably should `document/vec.rs` or `vec/document.rs`.  Either way not something we should do now/this week.

acrrd: ok I see.For me `U8Vec` is better, like `Uint8List` but I prefer `OptionU8` over `U8Option`.

rustonaut:
Besides `#[rustfmt::skip]` still being needed on main and some smaller things like the naming which we can fix either way in one or two weeks this lgtm.

rustonaut (approved):
lgtm

discovery_engine_core/core/src/engine.rs
```diff
@@ -28,9 +29,18 @@ use crate::{
...
+        BoxedOps,
+        BreakingNews,
+        Data as StackData,
+        Id as StackId,
+        PersonalizedNews,
+        Stack,
+    },
 };
 
+/// Potential engine errors.
```
acrrd, line 43: This are real errors. Not always an error means that the operation was completely unsuccessful but only partially.

discovery_engine_core/core/src/engine.rs
```diff
@@ -360,20 +374,25 @@ impl Engine<xayn_ai::ranker::Ranker> {
...
-            Engine::new(config.into(), ranker, stacks_ops).await
+            Self::new(config.into(), ranker, stack_ops).await
         }
     }
 }
 
+/// A shared discovery engine with a lock.
+#[allow(clippy::module_name_repetitions)]
+#[derive(AsRef, From)]
+pub struct SharedEngine(Mutex<XaynAiEngine>);
```
acrrd, line 388: Is still true that this must stay here for some limitation we found in cbindgen?

discovery_engine/test/ffi/ffi_test.dart
```diff
@@ -13,15 +13,13 @@
...
 import '../logging.dart' show setupLogging;
 
 void main() {
   setupLogging();
 
-  test('calling async ffi functions works', () async {
-    final x = await asyncCore.add(10, 22);
-    expect(x, equals(32));
+  test('calling async ffi functions works', () {
+    // TODO
```
acrrd, line 23: What forbid us to start writing some test?

janpetschexain: we could indeed take the raw ffi from the genesis extension file for this test. but since we don't use primitives to call the ffi anymore, there is more setup required, which is handled in the dart wrapper of the follow up pr anyways.

discovery_engine_core/core/src/engine.rs
janpetschexain: this might be a missunderstanding in the wording, i used "potential" in the sense that these are the errors which can occur during runtime, but in a successful run they will of course not happen. not "potential" in a sense that they might or might not be real errors. i'll remove the word to avoid confusion.

discovery_engine_core/core/src/engine.rs
janpetschexain: i'm not entirely sure what the exact issue was, i'll check again if i can move it to the bindgen crate.

janpetschexain: i moved the `SharedEngine` into the `bindings` crate. to avoid the `unused_qualifications` lint, i placed it into its own submodule instead of the crate root, because we still have to call it with full path within the `async-bindgen` macro. i also renamed the `core` dependency to its original name `xayn-discovery-engine-core` as this otherwise causes name conflicts for example in other proc macros which expect the standard library's `core` module behind this name.

janpetschexain:
bors merge

PR #135, opened 2022-02-11
[Merged by Bors] - TY-2366 build ios prod

Robert-Steiner:
Ticket:

- [TY-2366]

**Update:** lto error was addressed in #140 

---

can't use `lto` because of:
https://github.com/rust-lang/rust/issues/51009

running:
`RUSTFLAGS="-Ccodegen-units=1 -Clto=on -Cembed-bitcode=yes" just compile-ios-local`

gives the error:
```
error: lto can only be run for executables, cdylibs and static library outputs

error: could not compile `hyper` due to previous error
```

but `-Ccodegen-units=1` alone still decreases the size from 221mb to 144mb

[TY-2366]: https://xainag.atlassian.net/browse/TY-2366?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #140, opened 2022-02-14
[Merged by Bors] - Point to a version of hyper that support lto

acrrd:
In #135 we have an error that `hyper` cannot be used together with lto.
We forked hyper and removed the the configuration that wasn't allowing it.
This pr points to that version.

Running
`RUSTFLAGS="-Ccodegen-units=1 -Clto=on -Cembed-bitcode=yes" just compile-android-local` works correctly.

### DIFF
--- a/discovery_engine_core/Cargo.lock
+++ b/discovery_engine_core/Cargo.lock
@@ -57,7 +57,7 @@ dependencies = [
 [[package]]
 name = "async-bindgen"
 version = "0.1.0"
-source = "git+https://github.com/xaynetwork/xayn_async_bindgen.git?branch=main#0b8b127b3c86f9c9d9bdc75fc3f6a6b7ff39efe2"
+source = "git+https://github.com/xaynetwork/xayn_async_bindgen.git?branch=main#bde325e484a762e9a3f4b3c2652aef3a5e83d8da"
 dependencies = [
  "async-bindgen-derive",
  "once_cell",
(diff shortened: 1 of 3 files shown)


janpetschexain (approved):
lgtm

a question out of interest, wouldn't it be sufficient to remove `"lib"` and leave `"staticlib", "cdylib"` in the `crate-type` of `hyper`? at least we can compile `xayn_ai` for android with this `[lib]` config.
also (maybe this was already answered somewhere else?) why isn't the iappropriate option just ignored by the compiler? for example like we see for compiling `xayn_ai` for ios targets where one of `"staticlib"` or `"cdylib"` is dropped (because unsupported).

acrrd:
> lgtm
> 
> a question out of interest, wouldn't it be sufficient to remove `"lib"` and leave `"staticlib", "cdylib"` in the `crate-type` of `hyper`? 

It should work but we are not intrested in compiling hyper as a standalone c/static library so I just removed everything.

> at least we can compile `xayn_ai` for android with this `[lib]` config. also (maybe this was already answered somewhere else?) > ? why isn't the iappropriate option just ignored by the compiler? 
The compiler should but is not doing it. https://github.com/rust-lang/rust/issues/51009

acrrd:
bors merge

Robert-Steiner:
> a question out of interest, wouldn't it be sufficient to remove "lib" and leave "staticlib", "cdylib" in the crate-type of hyper

I tired that first and this didn't work because other rust crates that depend on hyper (like hyper-rustls) will fail to compile (because hyper as a rust lib does not exist)

PR #146, opened 2022-02-16
[Merged by Bors] - Always fail dart-check on infos from dart analyze.

rustonaut:
Since check now also runs flutter it's really easy to overlook "infos" when running `just check`.

As we anyway treat them as errors later on and I already run multiple times into situation where 
I overlooked some I would argue it's best to always fail on `info` (they are also always easy to
fix).


janpetschexain:
the change makes sense. looks like we also need to increase timeouts for the ci jobs, the dart test job timed out in the last moment.

acrrd (approved):

rustonaut:
bors merge

PR #147, opened 2022-02-16
[Merged by Bors] - TY-2416 decouple ranker from ops

finiteprods:
**Summary**

- changes to `Ops::new_items` method:
  - it no longer takes a ranker
  - it returns `Article`s rather than `Document`s
- also introduces an additional method `filter_articles` to `Ops`.
- `Stack` convenience methods to expose the above.  
- `compute_smbert` exposed through to `Ranker` trait.
- `Engine::update_stacks` changed accordingly using the above to build documents from articles.

### DIFF
--- a/discovery_engine_core/core/src/engine.rs
+++ b/discovery_engine_core/core/src/engine.rs
@@ -25,7 +25,7 @@ use xayn_ai::{
 };
 
 use crate::{
-    document::{Document, TimeSpent, UserReacted},
+    document::{self, document_from_article, Document, TimeSpent, UserReacted},
     mab::{self, BetaSampler, SelectionIter},
     ranker::Ranker,
     stack::{
@@ -65,6 +65,9 @@ pub enum Error {
 
     /// Error while using the ranker.
     Ranker(#[from] GenericError),
+
+    /// Document-related error.
+    Document(#[from] document::Error),
 
     /// A list of errors that could occur during some operation.
     Errors(Vec<Error>),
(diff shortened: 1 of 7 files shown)


discovery_engine_core/core/src/engine.rs
```diff
@@ -48,15 +48,21 @@ pub enum Error {
     /// Invalid stack id: {0}.
     InvalidStackId(StackId),
 
-    /// An operation on a stack failed: {0}.
+    /// An operation on a stack failed with a stack error: {0}.
```
acrrd, line 51: Maybe we call it internal operation on stack failed? so we don't have to specify the kind of the error in the string.

discovery_engine_core/core/src/engine.rs
```diff
@@ -48,15 +48,21 @@ pub enum Error {
...
     StackOpFailed(#[source] stack::Error),
 
+    /// An operation on a stack failed with a generic error: {0}.
+    StackOpGen(#[source] GenericError),
+
     /// Error while selecting the documents to return: {0}.
     Selection(#[from] mab::Error),
 
     /// Error while using the ranker.
     Ranker(#[from] GenericError),
```
acrrd, line 61: We cannot use `from` here now that we have another error using that.

discovery_engine_core/core/src/engine.rs
```diff
@@ -283,15 +289,27 @@ where
         let mut errors = Vec::new();
         for stack in self.stacks.write().await.values_mut() {
             if stack.len() <= request_new {
-                match stack.ops.new_items(key_phrases, &self.ranker).await {
+                let articles = stack.ops.new_items(key_phrases).await;
+                match articles.map_err(Error::StackOpGen).and_then(|articles| {
+                    let id = stack.id();
+                    articles
```
acrrd, line 295: We want to run `filter_articles` on them before running smbert

finiteprods: ok. I've combined "`new_items` followed by `filter_articles`" into a single `Stack` method `filter_new_articles`, which is probably where it ought to belong. so now in `update_stacks`:
```rust
                let articles = stack.filter_new_articles(key_phrases).await;
                match articles.map_err(Error::StackOpFailed).and_then(|articles| {
                    // the rest same as before ...
```
which also eliminates the `Error::StackOpGen` variant I introduced so overall it's a bit tidier (see the other comments)

discovery_engine_core/core/src/engine.rs
finiteprods: I removed `StackOpGen` so the error variants here are back to the way they were

discovery_engine_core/core/src/engine.rs
finiteprods: I removed `StackOpGen` so this string has been reverted back

discovery_engine_core/core/src/engine.rs
acrrd: But they are two logical different things so I'm more to keep them separate, but we can keep the new error variants and use them to aovid `StackOpGen`.

acrrd (approved):

janpetschexain:
since the `Ops` trait doesn't need a lifetime for `new_items()` anymore, we can go back to automocking to reduce code duplication. this means we can remove the `mock! {}` impl from the tests and place a `#[cfg_attr(test, automock)]` on the `Ops` trait again (`#[automock]` must be placed before `#[async_trait]` to work properly).

finiteprods:
> since the `Ops` trait doesn't need an explicit lifetime for `new_items()` anymore, we can go back to automocking to reduce code duplication. this means we can remove the `mock! {}` impl from the tests and place a `#[cfg_attr(test, automock)]` on the `Ops` trait again (`#[automock]` must be placed before `#[async_trait]` to work properly).

thanks for pointing this out

finiteprods:
bors merge

PR #142, opened 2022-02-15
TY-2302 ffi engine dart

janpetschexain:
**References**

- [TY-2302]
  - [TY-2398]
  - [TY-2400]
  - [TY-2402]
  - [TY-2404]
  - [TY-2406]
  - [TY-2414]
- follow up from #136

**Summary**

- impl dart ffi engine functions
- fix dart ffi test
- impl engine result ffi adapter
- cleanup box ffi

PR #151, opened 2022-02-16
[Merged by Bors] - TY-2423 improve result ffi

janpetschexain:
**References**

- [TY-2423]
- follow up from https://github.com/xaynetwork/xayn_discovery_engine/pull/142#pullrequestreview-883679114

**Summary**

rust side:
- add ffi to move ok and err values out of results into new boxes
- add ffi to drop `SharedEngine`

dart side:
- move boxing of pointer logic into `consumeNative()`
- add `move` and `free` handles to the `ResultFfiAdapter` and impl `moveNative()`
- align generics names for `Boxed`

**Notes**

- ~we could remove the move of error values, i just included it for completeness, but i'd also be ok omitting it~

[TY-2302]: https://xainag.atlassian.net/browse/TY-2302?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[TY-2423]: https://xainag.atlassian.net/browse/TY-2423?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine/lib/src/ffi/types/result.dart
+++ b/discovery_engine/lib/src/ffi/types/result.dart
@@ -30,39 +30,47 @@ class ResultFfiAdapter<Ok, Err, RustResult extends NativeType,
   final Pointer<RustErr> Function(Pointer<RustResult>) getErr;
   final Ok Function(Pointer<RustOk>) readNativeOk;
   final Err Function(Pointer<RustErr>) readNativeErr;
+  final Never Function(Err) throwErr;
+  final void Function(Pointer<RustResult>) freeResult;
 
   ResultFfiAdapter({
     required this.getOk,
     required this.getErr,
     required this.readNativeOk,
     required this.readNativeErr,
+    required this.throwErr,
+    required this.freeResult,
   });
 
-  Ok readNative(
-    Pointer<RustResult> result, {
-    required Exception Function(Err) mapErr,
-  }) {
+  /// Reads the result and returns the success value or throws in case of an error value.
+  Ok readNative(Pointer<RustResult> result) {
     final ok = getOk(result);
     if (ok.address != 0) {
       return readNativeOk(ok);
     }
     final err = getErr(result);
     if (err.address != 0) {
-      throw mapErr(readNativeErr(err));
+      throwErr(readNativeErr(err));
     }
     throw AssertionError('result should be either Ok or Err');
   }
 
-  Ok consumeNative(
-    Boxed<RustResult> result, {
-    required Exception Function(Err) mapErr,
-  }) {
+  /// Consumes the result and returns the success value or throws in case of an error value.
+  ///
+  /// # Safety
+  /// Must only be called on owned pointers.
+  Ok consumeNative(Pointer<RustResult> result) {
+    final boxedResult = Boxed(result, freeResult);
     try {
-      return readNative(result.ref, mapErr: mapErr);
+      return readNative(boxedResult.ref);
     } finally {
-      result.free();
+      boxedResult.free();
     }
   }
+}
+
+Never _throwStringErr(final String error) {
+  throw Exception(error);
 }
 
 final resultVoidStringFfiAdapter = ResultFfiAdapter(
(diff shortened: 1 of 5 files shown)


discovery_engine/lib/src/ffi/types/result.dart
```diff
@@ -28,67 +28,128 @@ class ResultFfiAdapter<Ok, Err, RustResult extends NativeType,
...
   }
 
-  Ok consumeNative(
-    Boxed<RustResult> result, {
-    required Exception Function(Err) mapErr,
-  }) {
+  /// Consumes the result and returns the success value or throws in case of an error value.
+  ///
+  /// # Safety
+  /// Must only be called on owned pointers.
```
rustonaut, line 69:
```suggestion
  /// # Safety
  ///
  /// Must only be called on owned pointers.
```

discovery_engine/lib/src/ffi/types/result.dart
```diff
@@ -28,67 +28,128 @@ class ResultFfiAdapter<Ok, Err, RustResult extends NativeType,
...
     } finally {
-      result.free();
+      freeResult(result);
     }
   }
+
+  /// Consumes the result and moves the success value or throws in case of an error value.
+  ///
+  /// # Safety
+  /// Must only be called on owned pointers.
```
rustonaut, line 81:
```suggestion
  /// # Safety
  ///
  /// Must only be called on owned pointers.
```

rustonaut (approved):
lgtm,

Through  I think we should remove the `moveErr` until we actually use it/provide
a dart abstraction around it.

As a side note the reason I originally had `mapErr` in the `readNative` function instead of an field was because I wasn't sure if we might not have cases where we have the same stringly typed error for multiple ffi functions but want to throw different context dependent exceptions.

But the current design is fine, too. 

I also think using a subclass or composition for having move functionality for some but not all results would be a better design, but again I don't think it matters much so lets for now keep it in the way it is in the PR.

janpetschexain:
> Through I think we should remove the `moveErr` until we actually use it/provide a dart abstraction around it.

done.

> As a side note the reason I originally had `mapErr` in the `readNative` function instead of an field was because I wasn't sure if we might not have cases where we have the same stringly typed error for multiple ffi functions but want to throw different context dependent exceptions.

makes sense once we actually want to use it, currently we just had to repeat the same closure at the call sites.

> I also think using a subclass or composition for having move functionality for some but not all results would be a better design, but again I don't think it matters much so lets for now keep it in the way it is in the PR.

i agree, i've split the adapter into a base class and mixins.

discovery_engine/lib/src/ffi/types/result.dart
```diff
@@ -24,132 +24,161 @@ import 'package:xayn_discovery_engine/src/ffi/types/primitives.dart'
...
-    required this.freeOk,
-    required this.freeErr,
-    required this.freeResult,
-  });
+  late final Pointer<RustOk> Function(Pointer<RustResult>) _getOk;
+  late final Pointer<RustErr> Function(Pointer<RustResult>) _getErr;
+  late final Ok Function(Pointer<RustOk>) _readNativeOk;
+  late final Err Function(Pointer<RustErr>) _readNativeErr;
+  late final Never Function(Err) _throwErr;
+  late final void Function(Pointer<RustResult>) _freeResult;
```
rustonaut, line 34: Why are they lazy internalized?

janpetschexain: because the field isn't initialized in the abstract class/mixin, but only in the subclass later on. even though it gets immediately initialized in the subclass constructor, dart doesn't allow this field to not be lazy (i guess the superclass can in general not know what the subclass is going to do with the field).

rustonaut: yes, but you could just have a constructor in the super-class and then call that in the initializer list in the sub-class.

janpetschexain: that would only remove the `late` from the base class, the mixins still have to use `late`.

rustonaut (approved):
lgtm

PR #152, opened 2022-02-16
[Merged by Bors] - TY-2420 Add Embedding new type.

rustonaut:
**References:**

[TY-2420](https://xainag.atlassian.net/browse/TY-2420)

### DIFF
--- a/discovery_engine/lib/src/domain/models/embedding.dart
+++ b/discovery_engine/lib/src/domain/models/embedding.dart
@@ -1,0 +1,46 @@
+// Copyright 2021 Xayn AG
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, version 3.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+import 'dart:typed_data' show Float32List;
+
+import 'package:equatable/equatable.dart' show EquatableMixin;
+import 'package:hive/hive.dart' show BinaryReader, BinaryWriter, TypeAdapter;
+import 'package:xayn_discovery_engine/src/domain/repository/type_id.dart'
+    show embeddingAdapter;
+
+/// 1-Dimensional Embedding
+///
+/// Values are stored in native byte order in hive.
+class Embedding with EquatableMixin {
+  final Float32List values;
+
+  Embedding(this.values);
+
+  @override
+  List<Object?> get props => [values];
+}
+
+class EmbeddingAdapter extends TypeAdapter<Embedding> {
+  @override
+  int get typeId => embeddingAdapter;
+
+  @override
+  Embedding read(BinaryReader reader) =>
+      Embedding(reader.readByteList().buffer.asFloat32List());
+
+  @override
+  void write(BinaryWriter writer, Embedding obj) {
+    writer.writeByteList(obj.values.buffer.asUint8List());
+  }
+}
--- a/discovery_engine/lib/src/domain/repository/type_id.dart
+++ b/discovery_engine/lib/src/domain/repository/type_id.dart
@@ -21,3 +21,4 @@ const newsResourceTypeId = 5;
 const uriTypeId = 6;
 const durationTypeId = 7;
 const stackIdTypeId = 8;
+const embeddingAdapter = 9;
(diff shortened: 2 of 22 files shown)


discovery_engine/lib/src/domain/models/embedding.dart
```diff
@@ -0,0 +1,46 @@
+// Copyright 2021 Xayn AG
```
biernacki, line 1:
```suggestion
// Copyright 2022 Xayn AG
```
shouldn't matter too much I think

discovery_engine/lib/src/domain/models/embedding.dart
```diff
@@ -0,0 +1,46 @@
...
+
+  @override
+  Embedding read(BinaryReader reader) =>
+      Embedding(reader.readByteList().buffer.asFloat32List());
+
+  @override
+  void write(BinaryWriter writer, Embedding obj) {
+    writer.writeByteList(obj.values.buffer.asUint8List());
+  }
+}
```
biernacki, line 46: we keep code related to Hive in `lib/src/infrastructure/*` so this class should go in a separate file, possibly in `lib/src/infrastructure/type_adapters/hive_embedding_adapter.dart`

biernacki (commented):
Code looks good, I would only suggest to move type adapter from `domain` to `infrastructure`

discovery_engine/lib/src/domain/models/embedding.dart
janpetschexain: @acrrd is it possible with github to only run the copyright job on modified files of the current pr? then we could change the copyright check to allow only for the current year, which will catch such copy-paste-issues.

discovery_engine/lib/src/domain/models/embedding.dart
```diff
@@ -0,0 +1,46 @@
...
+/// 1-Dimensional Embedding
+///
+/// Values are stored in native byte order in hive.
+class Embedding with EquatableMixin {
+  final Float32List values;
+
+  Embedding(this.values);
+
+  @override
+  List<Object?> get props => [values];
```
janpetschexain, line 31:
```suggestion
  List<Object> get props => [values];
```

discovery_engine/lib/src/domain/models/embedding.dart
acrrd: Not sure that will solve the problem. We can modify a file that was introduced last year so checking for current here will be wrong. We need to check for newly added files.
For the script that added the copyright to every file I used the year in which the file was introduced in git but it is more complex.

discovery_engine/lib/src/domain/repository/type_id.dart
```diff
@@ -21,3 +21,4 @@ const newsResourceTypeId = 5;
 const uriTypeId = 6;
 const durationTypeId = 7;
 const stackIdTypeId = 8;
+const embeddingAdapter = 9;
```
janpetschexain, line 24: to align it with the other type ids
```suggestion
const embeddingTypeId = 9;
```

discovery_engine/lib/src/domain/models/embedding.dart
```diff
@@ -0,0 +1,46 @@
...
+import 'package:xayn_discovery_engine/src/domain/repository/type_id.dart'
+    show embeddingAdapter;
+
+/// 1-Dimensional Embedding
+///
+/// Values are stored in native byte order in hive.
+class Embedding with EquatableMixin {
+  final Float32List values;
+
+  Embedding(this.values);
```
janpetschexain, line 28: we could add a convenience helper and use it in the test setups
```dart
  @visibleForTesting
  Embedding.fromList(List<double> values) => Embedding(Float32List.fromList(values));

discovery_engine/lib/src/domain/models/embedding.dart
janpetschexain: mhh i see :/ i guess it's easier to keep it like it is and be careful in copy-pasting stuff around.

discovery_engine/lib/src/domain/models/embedding.dart
rustonaut: The signature of `props` is `List<Object?>` not `List<Object>` you can change it to `List<Object>` as it's a sub-type, but why should we do so in this case?

janpetschexain: to help readability/analysis by having less type ambiguity/the most constrained type. i'd even argue that we should write `List<Float32List>`.

janpetschexain (approved):

discovery_engine/lib/src/domain/models/embedding.dart
rustonaut: Not in this case, the whole point of `probs` is to be as generic as possible i.e. `List<Object?>`,

Any place calling it will always treat it as `List<Object?>`.

There is no benefit gained and it will just obfuscate what this is doing.

rustonaut:
bors merge

PR #153, opened 2022-02-17
[Merged by Bors] - TY-2409 ChangedDocumentsReporter

biernacki:
[Jira ref](https://xainag.atlassian.net/browse/TY-2409)

This PR adds a `ChangedDocumentsReporter` class to manage a stream of `DocumentsUpdated` engine event, which is sent when `Document` entities that are exposed to the app change their internal state. This event acts as a notification that something has updated, so that the app can replace changed documents in the state that it manages.

For now this notyfication will happen only in response to `UserReactionChanged` event, and will be extended if needed for other events.

### DIFF
--- a/discovery_engine/test/document_manager_test.dart
+++ b/discovery_engine/test/document_manager_test.dart
@@ -133,7 +133,7 @@ Future<void> main() async {
         () => mgr.updateUserReaction(id3, UserReaction.positive),
         throwsArgumentError,
       );
-      expect(changedDocsReporter.changedDocuments, emitsDone);
+      expect(changedDocsReporter.changedDocuments, neverEmits(anything));
       await changedDocsReporter.close();
     });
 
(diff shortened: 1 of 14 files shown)


discovery_engine/test/document_manager_test.dart
```diff
@@ -116,13 +131,17 @@ Future<void> main() async {
         () => mgr.updateUserReaction(id3, UserReaction.positive),
         throwsArgumentError,
       );
+      expect(changedDocsReporter.changedDocuments, emitsDone);
+      await changedDocsReporter.close();
```
rustonaut, line 135: Is that necessary given that we do that anyway in `tearDown`.

biernacki: I'll take another look, but it might be, closing the underlying stream controller is necessary to finish this test, and if we only do it the `tearDown` the test will timeout

biernacki: ok, I've checked and we need to close the stream, otherwise the test will timeout.

rustonaut (approved):
lgtm

biernacki:
bors merge

PR #155, opened 2022-02-18
[Merged by Bors] - Implement Engine for DiscoveryEngineFfi so that it can be loaded.

rustonaut:


discovery_engine/lib/src/domain/event_handler.dart
```diff
@@ -201,6 +197,23 @@ class EventHandler {
     return engine;
   }
 
+  Future<Engine> initializeEngine(
+    Configuration config,
+    SetupData setupData,
+    Uint8List? engineState,
+  ) async {
+    if (config.apiKey == 'use-mock-engine' &&
+        config.apiBaseUrl == 'https://use-mock-engine.test') {
```
rustonaut, line 206: Alternatives:

- we add a `useMockEngine` field to the `Configuration` 
- we do not use a mock engine anymore (in which case we also can not not use a dummy vocabulary anymore)

biernacki: I would move the `MockEngine` to test folder, so it can be used as a mock for managers tests, but otherwise I think we don't need it in the `lib`.

Where do we use a dummy vacab?

rustonaut: In `discovery_engine_flutter/test/xayn_discovery_engine_flutter_test.dart`.

We also can't really move it to the test folder as we still need to explicitly initialize the `MockEngine` from this file.

We could use some form of DI but it's likely not worth the complexity it will introduce.

biernacki (approved):
As mentioned in the meeting, we can merge it as is, and then improve upon it later, when we decide on a better testing setup.

rustonaut:
bors merge

PR #157, opened 2022-02-18
[Merged by Bors] - TY 2477 update dart deps

acrrd:
Update dart deps.
Freezed 1.0 change the type tag name from 'type' to 'runtimeType'.
Relaxed the deps version to avoid conflicts with app.

### DIFF
--- a/discovery_engine/pubspec.yaml
+++ b/discovery_engine/pubspec.yaml
@@ -23,7 +23,7 @@ dependencies:
     git:
       url: https://github.com/xaynetwork/xayn_async_bindgen.git
       path: async_bindgen_dart_utils
-      ref: fe195d9
+      ref: fe195d99954c691c5e606ebbb3c77d371ba0e65b
 
 dev_dependencies:
   build_runner: ^2.1.7
--- a/discovery_engine/pubspec.yaml
+++ b/discovery_engine/pubspec.yaml
@@ -10,11 +10,11 @@ environment:
 dependencies:
   crypto: 3.0.1
   equatable: 2.0.3
-  freezed_annotation: 0.15.0
-  hive: 2.0.4
+  freezed_annotation: ^1.1.0
+  hive: ^2.0.5
   http: 0.13.4
   http_retry: 0.2.0
-  json_annotation: 4.3.0
+  json_annotation: ^4.4.0
   logger: 1.1.0
   meta: 1.7.0
   universal_platform: 1.0.0+1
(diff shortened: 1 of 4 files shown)


discovery_engine/pubspec.yaml
```diff
@@ -8,27 +8,27 @@ environment:
...
+  equatable: ^2.0.3
+  freezed_annotation: ^1.1.0
+  hive: ^2.0.5
+  http: ^0.13.4
+  http_retry: ^0.2.0
+  json_annotation: ^4.4.0
+  logger: ^1.1.0
+  meta: ^1.7.0
+  universal_platform: ^1.0.0+1
+  uuid: ^3.0.5
```
biernacki, line 21: We wanted to use exact version, as we don't track lock file in git. What is the reason for starting with using caret?

discovery_engine/test/codecs/json_codecs_test.dart
```diff
@@ -61,7 +61,7 @@ void main() {
       final message_1 = converter.convert(request_1) as Map;
 
       expect(message_1[kSenderKey], isA<SendPort>());
-      expect(message_1[kPayloadKey], equals({'type': 'feedRequested'}));
+      expect(message_1[kPayloadKey], equals({'runtimeType': 'feedRequested'}));
```
biernacki, line 64: there is an option to still use the key `type` if we specify it in `build.yaml` file like so:
```yaml
targets:
  $default:
    builders:
      freezed:
        options:
          union_key: type
```

But I guess `runtimeType` is a better name, so let's go with that.

discovery_engine/pubspec.yaml
```diff
@@ -8,27 +8,27 @@ environment:
...
   async_bindgen_dart_utils:
     git:
       url: https://github.com/xaynetwork/xayn_async_bindgen.git
       path: async_bindgen_dart_utils
```
biernacki, line 25: maybe we should specify `ref` as well, so we have more controll over the version used, wdyt?

discovery_engine_flutter/example/pubspec.lock
```diff
@@ -13,7 +13,7 @@ packages:
     description:
       path: async_bindgen_dart_utils
       ref: HEAD
-      resolved-ref: bde325e484a762e9a3f4b3c2652aef3a5e83d8da
+      resolved-ref: fe195d99954c691c5e606ebbb3c77d371ba0e65b
```
biernacki, line 16: as above, it's probably not so good if it can change on it's own

discovery_engine/pubspec.yaml
biernacki: ah, ok.. I've just read the PR description. Maybe we should then start tracking the lockfile? Not sure what is the best solution here..

acrrd: To give more freedom to blue to use different versions. We can start using the lock and not exporting it on the release repo if we want.

biernacki (approved):
Lokks good! If you feel like it a full version of the hash would be a better option imho, but it's also fine as is

acrrd:
bors merge

PR #166, opened 2022-02-21
[Merged by Bors] - TY-2469 init engine with models

Robert-Steiner:
Ticket:

- [TY-2469]

[TY-2469]: https://xainag.atlassian.net/browse/TY-2469?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/justfile
+++ b/justfile
@@ -199,11 +199,15 @@ dart-clean:
 clean-tools:
     -rm -r "$CARGO_INSTALL_ROOT"
 
+# Removes all asset data
+remove-assets:
+    find $FLUTTER_EXAMPLE_WORKSPACE/assets/*_v* -type f ! -name .gitkeep -exec rm '{}' \;
+
 # Removes all local cached dependencies and generated files
 clean: clean-gen-files rust-clean dart-clean
 
 # Runs clean and removes local installed tools
-clean-fully: clean clean-tools
+clean-fully: clean clean-tools remove-assets
 
 # Workaround to set env variable CI for all job dependencies
 _pre-push: deps clean-gen-files fmt check test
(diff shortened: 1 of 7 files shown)


justfile
```diff
@@ -161,7 +161,7 @@ rust-build: _codegen-order-workaround
 build: rust-build dart-build
 
 # Tests dart (builds all necessary parts)
-dart-test: rust-build dart-build
+dart-test: rust-build dart-build download_assets
```
acrrd, line 164: We don't want to download the assets every time we run the tests.

acrrd (approved):

justfile
Robert-Steiner: With the curl flag `-C -` we don't download the assets again if they have already been downlaoded. 
> Use "-C -" to tell curl to automatically find out where/how to resume the transfer. It then uses the given
              output/input files to figure that out.

Is that ok? otherwise we should at least put it in the readme

acrrd: This also avoid the unpack of the archive?

acrrd: another point, should we add a command to delete them in `clean-fully`?

Robert-Steiner: unfortunately no. We can add the flag `--keep-old-files` but it doesn't make it noticeably faster. verifying the checksum takes the longest. we can also just add an if and check if the archive exist if not then we fetch, extract and verify. wdyt?

Robert-Steiner: > another point, should we add a command to delete them in clean-fully?

done

Robert-Steiner:
bors merge

acrrd (approved):

Robert-Steiner:
bors merge

PR #172, opened 2022-02-22
[Merged by Bors] - TY-2261 implement querying of new items

finiteprods:
previously, #147 refactored `Ops` trait.

**Summary**

implements `Ops::new_items` for `BreakingNews` and `PersonalizedNews` structs.

### DIFF
--- a/discovery_engine_core/core/src/stack/ops/breaking.rs
+++ b/discovery_engine_core/core/src/stack/ops/breaking.rs
@@ -50,17 +50,26 @@ impl Ops for BreakingNews {
     #[allow(clippy::cast_precision_loss)]
     #[allow(clippy::cast_possible_truncation)]
     async fn new_items(&self, _key_phrases: &[KeyPhrase]) -> Result<Vec<Article>, GenericError> {
-        Ok(if let Some(markets) = self.markets.as_ref() {
+        if let Some(markets) = self.markets.as_ref() {
             let mut articles = Vec::new();
+            let mut errors = Vec::new();
             let page_size = Some(20); // TODO pass through config later
+
             for market in markets.read().await.clone() {
                 let query = HeadlinesQuery { market, page_size };
-                articles.extend(self.client.headlines(&query).await?);
+                match self.client.headlines(&query).await {
+                    Ok(batch) => articles.extend(batch),
+                    Err(err) => errors.push(err),
+                }
             }
-            articles
+            if articles.is_empty() && !errors.is_empty() {
+                Err(errors.pop().unwrap(/* nonempty errors */).into())
+            } else {
+                Ok(articles)
+            }
         } else {
-            vec![]
-        })
+            Ok(vec![])
+        }
     }
 
     fn filter_articles(
--- a/discovery_engine_core/core/src/stack/ops/breaking.rs
+++ b/discovery_engine_core/core/src/stack/ops/breaking.rs
@@ -19,8 +19,7 @@ use chrono::NaiveDate;
 use tokio::sync::RwLock;
 use uuid::Uuid;
 use xayn_ai::ranker::KeyPhrase;
-
-use xayn_discovery_engine_providers::{Article, Market, Topic};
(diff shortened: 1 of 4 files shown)


discovery_engine_core/providers/src/client.rs
```diff
@@ -35,24 +35,28 @@ pub enum Error {
...
 pub struct NewsQuery {
-    market: Market,
-    filter: Filter,
+    /// Market of news.
+    pub market: Market,
+    /// News filter.
+    pub filter: Filter,
     /// How many articles to return (per page).
-    page_size: Option<usize>,
+    pub page_size: Option<usize>,
```
rustonaut, line 51: Is that really optional for us?

(same for the other query)

acrrd: It was a some point. Probably now not anymore.

acrrd: But we can change it later.

discovery_engine_core/core/src/stack/ops/breaking.rs
```diff
@@ -43,11 +42,22 @@ impl Ops for BreakingNews {
...
     }
 
     async fn new_items(&self, _key_phrases: &[KeyPhrase]) -> Result<Vec<Article>, GenericError> {
-        Ok(vec![])
+        Ok(if let Some(markets) = self.markets.as_ref() {
+            let mut articles = Vec::new();
+            let page_size = Some(20); // TODO pass through config later
+            for market in markets.read().await.clone() {
+                let query = HeadlinesQuery { market, page_size };
+                articles.extend(self.client.headlines(&query).await?);
```
acrrd, line 55: This will terminate with an error even if only one request returned an error. 
We don't have a way at the moment to signal only one error, but is better for to return an error only when all of them return an error

discovery_engine_core/core/src/stack/ops/breaking.rs
```diff
@@ -43,11 +42,22 @@ impl Ops for BreakingNews {
...
+            }
+            articles
+        } else {
+            vec![]
+        })
```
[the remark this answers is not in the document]
finiteprods: sure :+1:  it's changed a bit now after addressing the other comment

discovery_engine_core/core/src/stack/ops/breaking.rs
finiteprods: ok, adjusted so that errors are now recorded in a vec. in the case where no articles were retrieved, only errors, (for now) the last error is returned.

rustonaut (approved):
lgtm

discovery_engine_core/core/src/stack/ops/breaking.rs
```diff
@@ -44,34 +42,31 @@ impl Ops for BreakingNews {
...
 
-        Ok(articles)
+            for market in markets.read().await.clone() {
+                let query = HeadlinesQuery { market, page_size };
+                match self.client.headlines(&query).await {
+                    Ok(batch) => articles.extend(batch),
+                    Err(err) => errors.push(err),
+                }
+            }
+            if articles.is_empty() && !errors.is_empty() {
```
acrrd, line 62: I'm not sure about `articles.is_empty()` that is not really an error, it could happen that the api returns 0 documents.
Will be weird for the headlines and kinda possible for the personalized news if the keyword don't have a match.

I think we can merge and return on this in another pr if we want to remove it.

finiteprods: >I'm not sure about `articles.is_empty()` that is not really an error, it could happen that the api returns 0 documents.

yeah. if the api returns no articles and there were no errors throughout, that's definitely an ok. my thinking was if there are no articles but there _are_ errors, returning an error would be more informative. but I see your point if returning 0 articles is a very possible / frequent occurrence e.g. for personalized news.

ok will merge as is but if you prefer it changed I will follow it up.

finiteprods:
bors merge

PR #173, opened 2022-02-22
[Merged by Bors] - Add mock implementation instead of panic for stack ops

acrrd:


discovery_engine_flutter/example/pubspec.lock
```diff
@@ -12,7 +12,7 @@ packages:
     dependency: transitive
     description:
       path: async_bindgen_dart_utils
-      ref: HEAD
+      ref: fe195d99954c691c5e606ebbb3c77d371ba0e65b
```
Robert-Steiner, line 15: is this always changing because we use different flutter versions locally?

Robert-Steiner (approved):

discovery_engine_flutter/example/pubspec.lock
Robert-Steiner: was discussed in the daily

acrrd: I was on a branch where this wasn't already there. Now is on main so I don't expect it to change anymore.

acrrd:
bors merge

PR #170, opened 2022-02-21
[Merged by Bors] - TY-2465 historic document vec

rustonaut:
Added support for passing a `Vec<HistoricDocument>` to dart.

*References:*

- [TY-2465](https://xainag.atlassian.net/browse/TY-2465)

*Based on:*

- #170

PR #169, opened 2022-02-21
[Merged by Bors] - TY-2425 historic document

rustonaut:
Adds a `HistoricDocument` type and FFI bindings for it.

*References:*

- [TY-2425](https://xainag.atlassian.net/browse/TY-2425)

PR #174, opened 2022-02-22
[Merged by Bors] - TY-2318 pass history to rust

rustonaut:
*References:*

- [TY-2318](https://xainag.atlassian.net/browse/TY-2318)

*Based on:*

- #169 
- #170


rustonaut:
I removed the `Extended MockEngine to track more` commit as it's only relevant for code not in this PR.

discovery_engine_core/bindings/src/lib.rs
```diff
@@ -29,18 +29,22 @@
 pub mod async_bindings;
 pub mod types;
 
-#[async_bindgen::api]
+#[async_bindgen::api(
+    use xayn_discovery_engine_core::document::HistoricDocument;
+)]
```
janpetschexain, line 34: that's nice, i didn't know this was possible 👍  i'll open a pr to shorten the other full paths used in this module.

discovery_engine/lib/src/domain/engine/engine.dart
```diff
@@ -26,12 +35,39 @@ abstract class Engine {
...
+  );
 
   /// Process the feedback about the user spending some time on a document.
   Future<void> timeSpent(TimeSpent timeSpent);
 
   /// Process the user's reaction to a document.
   Future<void> userReacted(UserReacted userReacted);
 }
+
+/// Passed to constructors/initializers of `Engine` implementing classe.
```
janpetschexain, line 57:
```suggestion
/// Passed to constructors/initializers of `Engine` implementing classes.
```

discovery_engine/lib/src/domain/engine/engine.dart
```diff
@@ -26,12 +35,39 @@ abstract class Engine {
...
+class EngineInitializer with EquatableMixin {
```
janpetschexain, line 58: what's the purpose of this wrapper? the `Engine` effectively takes the same arguments and just unwraps this, rendering the wrapper somewhat pointless. and even though the `MockEngine` can take this optionally and store it in a field, this is never used there. is this wrapper important for a following pr?

rustonaut: Mainly I don't want to repeat the same number of keyword arguments across various call places, and update them manually every time we add/change a field (like the new `aiConfig`).

Like its used by:

- `initializeEngine` in event_handler.dart
- `initialize` in `engine.dart` (ffi)
- the constructor of `MockEngine` (in the future) so that we can better test if the right data is passed in

Also maybe at some point make it a FFI type and pass this as a single boxed struct.

janpetschexain (approved):
lgtm

rustonaut:
bors merge

PR #175, opened 2022-02-22
Return mock data for stacks

acrrd:
Returns mock data from stack.
This will be removed by #172.

Modify the handling of `source_url` since `curl_url` is only a domain. This will be changed in another PR.


Robert-Steiner (approved):

acrrd:
bors merge

PR #177, opened 2022-02-22
[Merged by Bors] - TY-2449 optional ai config

janpetschexain:
**References**

- [TY-2449]
- requires #164
- followed by #178

**Summary**

- add `Option<String>` handling to ffi
- add `ai_config: Option<String>` field to `InitConfig` & update ffi

[TY-2449]: https://xainag.atlassian.net/browse/TY-2449?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine/test/ffi/types/init_config_test.dart
+++ b/discovery_engine/test/ffi/types/init_config_test.dart
@@ -24,8 +24,8 @@ void main() {
       apiKey: 'hjlsdfhjfdhjk',
       apiBaseUrl: 'https://foo.example/api/v1',
       feedMarkets: [
-        const FeedMarket(countryCode: 'DE', langCode: 'DE'),
-        const FeedMarket(countryCode: 'US', langCode: 'EN'),
+        const FeedMarket(countryCode: 'DE', langCode: 'de'),
+        const FeedMarket(countryCode: 'US', langCode: 'en'),
       ],
       smbertVocab: 'foo/bar',
       smbertModel: 'bar/foot',
@@ -45,8 +45,8 @@ void main() {
       apiKey: 'hjlsdfhjfdhjk',
       apiBaseUrl: 'https://foo.example/api/v1',
       feedMarkets: [
-        const FeedMarket(countryCode: 'DE', langCode: 'DE'),
-        const FeedMarket(countryCode: 'US', langCode: 'EN'),
+        const FeedMarket(countryCode: 'DE', langCode: 'de'),
+        const FeedMarket(countryCode: 'US', langCode: 'en'),
       ],
       smbertVocab: 'foo/bar',
       smbertModel: 'bar/foot',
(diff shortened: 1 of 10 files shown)


discovery_engine/test/ffi/types/string_test.dart
```diff
@@ -34,4 +35,23 @@ void main() {
     ffi.drop_string(place);
     expect(res, equals(string));
   });
+
+  test('reading written some string yields same result', () {
+    // ignore: unnecessary_cast
+    const String? string = 'a&+KLA)&+fjw)&+f' as String?;
```
janpetschexain, line 41: unfortunately the type declaration `const String?` is not enough, but the type cast `as String?` is required, because dart otherwise infers that this is a non-null string indeed and tries to call `StringFfi.writeNative()` instead of `OptionStringFfi.writeNative()` in l.42.

declaring this `final` instead of `const` triggers another lint, but avoids the type cast. anybody got an idea if this can be fixed? it's just an issue in this test after all due to the const nature of the variable.

acrrd: It's only horrible with the cast or is not working even with the cast?
Are `const`/`final` changing the semantic of the type inference (a part from constness)?

janpetschexain: it's only working with the cast, the issue is that we need the additional `// ignore: unnecessary_cast` as this lint is bugged in this case.

`final String? string` will actually convince the compiler that this type is `String?` without an additional cast needed. with `const String? string` the compiler thinks that this type is `String` without the additional cast. but declaring it final doesn't solve the linter issue, because we then have to `// ignore: prefer_const_declarations`.

discovery_engine/test/ffi/types/init_config_test.dart
```diff
@@ -39,4 +39,26 @@ void main() {
     boxed.free();
     expect(res, equals(config));
   });
+
+  test('reading written init config ffi with ai config works', () {
+    final config = InitConfigFfi.fromParts(
+      apiKey: 'hjlsdfhjfdhjk',
+      apiBaseUrl: 'https://foo.example/api/v1',
+      feedMarkets: [
+        const FeedMarket(countryCode: 'DE', langCode: 'DE'),
```
acrrd, line 48: lang code should be lower case

discovery_engine/test/ffi/types/string_test.dart
acrrd: I don't think we can do much then.

acrrd (approved):

discovery_engine/test/ffi/types/init_config_test.dart
janpetschexain: i just copy-pasted this from other tests, so i will also fix those.

janpetschexain:
bors merge

PR #178, opened 2022-02-22
[Merged by Bors] - TY-2450 pass optional ai config

janpetschexain:
**References**

- [TY-2450]
- requires #177

**Summary**

- pass an optional JSON string for the AI config in the engine constructors & ffi

[TY-2450]: https://xainag.atlassian.net/browse/TY-2450?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine/lib/src/ffi/types/engine.dart
+++ b/discovery_engine/lib/src/ffi/types/engine.dart
@@ -61,10 +61,12 @@ class DiscoveryEngineFfi implements Engine {
   /// Initializes the engine.
   static Future<DiscoveryEngineFfi> initialize(
     final Configuration config,
-    final NativeSetupData setupData, [
+    final NativeSetupData setupData, {
+    final String? aiConfig,
     final Uint8List? state,
-  ]) async {
-    final boxedConfig = InitConfigFfi(config, setupData).allocNative();
+  }) async {
+    final boxedConfig =
+        InitConfigFfi(config, setupData, aiConfig: aiConfig).allocNative();
     final boxedState = state?.allocNative();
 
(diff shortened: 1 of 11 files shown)


discovery_engine/lib/src/ffi/types/engine.dart
```diff
@@ -61,10 +61,12 @@ class DiscoveryEngineFfi implements Engine {
   /// Initializes the engine.
   static Future<DiscoveryEngineFfi> initialize(
     final Configuration config,
-    final NativeSetupData setupData, [
+    final NativeSetupData setupData, {
+    final String? aiConfig,
     final Uint8List? state,
```
acrrd, line 66: we uniform the name and called it always `state` or `engineState` now we have a mix

acrrd (approved):
We have some ```const String? aiConfig = null;` around the code and I don't think is adding much value. When we want to pass a config we will add it.

discovery_engine/lib/src/ffi/types/engine.dart
janpetschexain: i'll open a follow up pr to clean this up across the repo. in rust and the rust/dart ffi we mostly (but not always) use `state`, in dart we have a mix of `state`/`engineState`. i'd prefer to just name it as `state` since it always is an argument to the engine and hence it should be clear that this is the engine state. wdyt?

janpetschexain:
bors merge

PR #180, opened 2022-02-22
[Merged by Bors] - TY 2489 Exclude arm64 for ios simulator

acrrd:


Robert-Steiner:
I saw that blue already merged the fix where they exclude the arm64 simulator on their side. just wanted to note that your fix would also work (at least it worked for our example)

acrrd:
We are going to add this exclusion and we will remove it when we can compile for aarch64-apple-ios-sim.

Robert-Steiner (approved):
can you please also remove those lines in `discovery_engine_flutter/example/ios/Podfile`
```
    target.build_configurations.each do |config|
      config.build_settings['EXCLUDED_ARCHS[sdk=iphonesimulator*]'] = 'i386 arm64'
    end
```

Robert-Steiner (approved):

acrrd:
bors merge

PR #182, opened 2022-02-22
[Merged by Bors] - TY 2487 Export message to main thread

acrrd:
Only the reason will not give us enough information to debug the error when on user phone. We export the message till the main thread so the app can log more information for us.

@biernacki Do we need to explicitly attach the message somewhere or the `EngineExceptionRaised` event is always attached to an exception?

### DIFF
--- a/discovery_engine/lib/src/api/events/engine_events.dart
+++ b/discovery_engine/lib/src/api/events/engine_events.dart
@@ -131,9 +131,10 @@ class EngineEvent with _$EngineEvent {
   /// UserReactionChanged.
   @Implements<SystemEngineEvent>()
   const factory EngineEvent.engineExceptionRaised(
-    EngineExceptionReason reason,
+    EngineExceptionReason reason, {
     String? message,
-  ) = EngineExceptionRaised;
+    String? stackTrace,
+  }) = EngineExceptionRaised;
 
   /// Event created as a successful response to some client events which are
   /// updating a [Document] (currently only "UserReactionChanged").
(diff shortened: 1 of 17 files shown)


discovery_engine/lib/src/api/events/engine_events.dart
```diff
@@ -132,6 +132,7 @@ class EngineEvent with _$EngineEvent {
   @Implements<SystemEngineEvent>()
   const factory EngineEvent.engineExceptionRaised(
     EngineExceptionReason reason,
+    String? message,
```
biernacki, line 135: what about adding stack trace? might be useful for the app

discovery_engine/lib/src/discovery_engine_base.dart
```diff
@@ -291,7 +291,7 @@ class DiscoveryEngine {
       logger.e(message, error, stackTrace);
 
       // into [EngineExceptionRaised] event with a specific reason
-      return EngineEvent.engineExceptionRaised(reason);
+      return EngineEvent.engineExceptionRaised(reason, error.toString());
```
biernacki, line 294: this will do the string conversion
```suggestion
      return EngineEvent.engineExceptionRaised(reason, '$error');
```

biernacki (approved):
thanks for that! left some small comments, non-blocking

biernacki:
> @biernacki Do we need to explicitly attach the message somewhere or the EngineExceptionRaised event is always attached to an exception?

We try to catch every exeption on our side and return `EngineExceptionRaised` event when it happens. Then the message field you've added needs to be explicitly attached to every exception event.

discovery_engine/lib/src/api/events/engine_events.dart
acrrd: I wasn't sure it can pass the isolate. I'll check and add it or change it to a string if possible.

acrrd:
Added stack trace. Made 'message' and 'stackTrace' optional named parameters.

biernacki (approved):
looks good!

acrrd:
bors merge

acrrd:
bors merge

PR #184, opened 2022-02-23
[Merged by Bors] - TY-2314 - Implement basic image and duplicate filtering

x3ro:
* [TY-2418] (parent)
    * [TY-2314]
    * [TY-2316]

---

* Filter out articles we're already aware of as Documents
* Filter out articles without media URLs

[TY-2418]: https://xainag.atlassian.net/browse/TY-2418?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[TY-2314]: https://xainag.atlassian.net/browse/TY-2314?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[TY-2316]: https://xainag.atlassian.net/browse/TY-2316?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine_core/core/Cargo.toml
+++ b/discovery_engine_core/core/Cargo.toml
@@ -27,4 +27,5 @@ xayn-discovery-engine-providers = { path = "../providers" }
 claim = "0.5.0"
 float-cmp = "0.9.0"
 mockall = "0.11.0"
+serde_json = "1.0.79"
 tokio = { version = "1.17.0", default-features = false, features = ["macros", "rt", "sync"] }
--- a/discovery_engine_core/core/src/stack/filters.rs
+++ b/discovery_engine_core/core/src/stack/filters.rs
@@ -1,0 +1,119 @@
+// Copyright 2022 Xayn AG
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, version 3.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+use crate::{document::Document, engine::GenericError};
+use std::collections::HashSet;
+use url::Url;
+use xayn_discovery_engine_providers::Article;
+
+pub(crate) trait ArticleFilter {
+    fn apply(current: &[Document], articles: Vec<Article>) -> Result<Vec<Article>, GenericError>;
+}
+
+struct DuplicateFilter;
+
+impl ArticleFilter for DuplicateFilter {
+    fn apply(
+        current: &[Document],
+        mut articles: Vec<Article>,
+    ) -> Result<Vec<Article>, GenericError> {
+        let mut urls = HashSet::new();
+        for doc in current {
+            urls.insert(doc.resource.url.as_str());
+        }
+
+        articles.retain(|article| !urls.contains(&article.link.as_str()));
+        Ok(articles)
+    }
+}
+
+struct MalformedImageUrlFilter;
+
+impl ArticleFilter for MalformedImageUrlFilter {
+    fn apply(
+        _current: &[Document],
+        mut articles: Vec<Article>,
+    ) -> Result<Vec<Article>, GenericError> {
+        articles.retain(|article| Url::parse(&article.media).is_ok());
+        Ok(articles)
+    }
+}
+
+pub(crate) struct CommonFilter;
+impl ArticleFilter for CommonFilter {
+    fn apply(current: &[Document], articles: Vec<Article>) -> Result<Vec<Article>, GenericError> {
+        DuplicateFilter::apply(current, articles)
+            .and_then(|articles| MalformedImageUrlFilter::apply(current, articles))
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    use crate::{
+        document::{document_from_article, Document},
+        stack::filters::CommonFilter,
+    };
+    use xayn_discovery_engine_providers::Article;
+
+    #[test]
+    fn test_duplicate_filter() {
+        let valid_articles: Vec<Article> =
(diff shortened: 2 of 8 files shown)


discovery_engine_core/core/src/stack/filters.rs
```diff
@@ -0,0 +1,121 @@
...
+    ) -> Result<Vec<Article>, GenericError> {
+        let urls = current
+            .iter()
+            .map(|a| a.resource.url.as_str())
+            .collect::<Vec<_>>();
+        let filtered = articles
+            .drain(..)
+            .filter(|article| !urls.contains(&article.link.as_str()))
+            .collect();
+        Ok(filtered)
```
janpetschexain, line 37: instead of draining & filtering & collecting we could just retain the desired elements. similarly for the `MalformedImageUrlFilter`.
```suggestion
        articles.retain(|article| !urls.contains(&article.link.as_str()));

        Ok(articles)
```

discovery_engine_core/core/src/stack/filters.rs
```diff
@@ -0,0 +1,121 @@
...
+            })
+            .collect::<Vec<_>>();
+
+        let result = CommonFilter::apply(documents.as_slice(), valid_articles).unwrap();
+        let titles = result.iter().map(|a| &a.title).collect::<Vec<_>>();
+
+        assert_eq!(titles.as_slice(), [
+            "Porsche entwickelt Antrieb, der E-Mobilit\u{00e4}t teilweise \u{00fc}berlegen ist",
+            "Mensch mit d\u{00fc}sterer Prognose: \"Kollektiv versagt!\" N\u{00e4}chste Pandemie wird schlimmer als Covid-19",
+        ]);
```
janpetschexain, line 92: instead of checking for the literal titles, it's probably more robust to test against the remaining titles of the `valid_articles`. maybe we could also check that the `valid_articles` actually hold more than 2 elements, just to catch issues if the test fixtures are changed. similarly for the other test.
```suggestion
        let result = CommonFilter::apply(documents.as_slice(), valid_articles.clone()).unwrap();
        let titles = result.iter().map(|a| &a.title).collect::<Vec<_>>();
        let expected = valid_articles
            .iter()
            .skip(2)
            .map(|a| &a.title)
            .collect::<Vec<_>>();

        assert_eq!(titles, expected);
```

discovery_engine_core/core/src/stack/filters.rs
```diff
@@ -0,0 +1,121 @@
...
+            .map(|article| {
+                let doc = Document::default();
+                document_from_article(article.clone(), doc.stack_id, doc.smbert_embedding).unwrap()
+            })
+            .collect::<Vec<_>>();
+
+        let result = CommonFilter::apply(documents.as_slice(), valid_articles).unwrap();
+        let titles = result.iter().map(|a| &a.title).collect::<Vec<_>>();
+
+        assert_eq!(titles.as_slice(), [
```
janpetschexain, line 89: no need for the additional referencing, as vectors can be compared with arrays via the macro directly.
```suggestion
        assert_eq!(titles, [
```

discovery_engine_core/core/test-fixtures/articles-some-malformed-media-urls.json
```diff
@@ -0,0 +1,48 @@
...
+    "authors": [
+      "Anonymous"
+    ],
+    "media": "a88260009773cd_hero-phone.png",
+    "is_opinion": false,
+    "twitter_account": "@XaynHQ",
+    "_score": 17.118832,
+    "_id": "b7ecc49a15bc38bf224ae01556e52641"
+  }
+]
```
janpetschexain, line 48: github wants to have a newline
```suggestion
]

```

discovery_engine_core/core/Cargo.toml
```diff
@@ -26,4 +26,6 @@ xayn-discovery-engine-providers = { path = "../providers" }
 claim = "0.5.0"
 float-cmp = "0.9.0"
 mockall = "0.11.0"
+serde = { version = "1.0.136", features = ["derive"] }
```
janpetschexain, line 29: `serde` is not needed as an additional dev-dependency because it is already a dependency with the same features.
```suggestion
```

discovery_engine_core/core/src/stack/filters.rs
```diff
@@ -0,0 +1,121 @@
...
+struct DuplicateFilter;
+impl ArticleFilter for DuplicateFilter {
+    fn apply(
+        current: &[Document],
+        mut articles: Vec<Article>,
+    ) -> Result<Vec<Article>, GenericError> {
+        let urls = current
+            .iter()
+            .map(|a| a.resource.url.as_str())
+            .collect::<Vec<_>>();
```
acrrd, line 32: We can collect in an HashSet to avoid quadratic complexity

discovery_engine_core/core/src/stack/filters.rs
x3ro: I like the idea of verifying that we're dealing with the expected number of articles, but I think the `expected` logic that you introduced just makes the test harder to read, so I would not introduce that overhead.

discovery_engine_core/core/src/stack/filters.rs
```diff
@@ -0,0 +1,119 @@
...
+
+impl ArticleFilter for DuplicateFilter {
+    fn apply(
+        current: &[Document],
+        mut articles: Vec<Article>,
+    ) -> Result<Vec<Article>, GenericError> {
+        let mut urls = HashSet::new();
+        for doc in current {
+            urls.insert(doc.resource.url.as_str());
+        }
```
janpetschexain, line 34: the hashset could be collected directly from an iterator to avoid potential reallocations when it grows in the loop:
```suggestion
        let urls = current
            .iter()
            .map(|doc| doc.resource.url.as_str())
            .collect::<HashSet<_>>();
```

janpetschexain (approved):
lgtm, the hashmap could be optimized a bit more

x3ro:
bors merge

x3ro:
bors merge

PR #188, opened 2022-02-23
[Merged by Bors] - TY 2490 Do not do a query if there no keywords

acrrd:
Do not do a query if there no keywords.
Use BTreeMap instead of HashMap to always have the same order of the parameters and be more cache friendly.
Use `bearer_token` instead of setting it manually.
Set timeout for requests.

### DIFF
--- a/discovery_engine_core/core/src/stack/ops/personalized.rs
+++ b/discovery_engine_core/core/src/stack/ops/personalized.rs
@@ -47,6 +47,9 @@ impl Ops for PersonalizedNews {
     }
 
     async fn new_items(&self, key_phrases: &[KeyPhrase]) -> Result<Vec<Article>, GenericError> {
+        if key_phrases.is_empty() {
+            return Ok(vec![]);
+        }
         if let Some(markets) = self.markets.as_ref() {
             let mut articles = Vec::new();
             let mut errors = Vec::new();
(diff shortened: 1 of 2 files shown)


rustonaut (approved):
lgtm

discovery_engine_core/core/src/stack/ops/personalized.rs
```diff
@@ -47,6 +47,9 @@ impl Ops for PersonalizedNews {
     }
 
     async fn new_items(&self, key_phrases: &[KeyPhrase]) -> Result<Vec<Article>, GenericError> {
+        if key_phrases.is_empty() {
```
x3ro, line 50: Would this state occur when a new user starts using the app? Or is there another scenario? If it's the case of a new user, do we then fill the key phrases through the breaking news? 🤔

x3ro (approved):

discovery_engine_core/core/src/stack/ops/personalized.rs
acrrd: It occurs for new users. A user interacts with the breaking news and we can start learning keywords. This can later happen also with active search.
I don't think we can go back to a state where we don't have keywords, even if we remove coi that are too old we want to keep a minimum in.

acrrd:
bors merge

acrrd:
bors merge

PR #189, opened 2022-02-23
[Merged by Bors] - TY 2493 discard net error on init

acrrd:


rustonaut (approved):
lgtm

I wonder if there is a standard style for ignoring results in rust.

Thinks I have seen include:

- `drop`
- `let _ = ...;`
- `.ok()` 

I guess the last one is a bit more explicit and doesn't "remove"
potential `must_use` of the `Ok` type :thinking:

acrrd:
> lgtm
> 
> I wonder if there is a standard style for ignoring results in rust.
> 
> Thinks I have seen include:
> 
>     * `drop`
> 
>     * `let _ = ...;`
> 
>     * `.ok()`
> 
> 
> I guess the last one is a bit more explicit and doesn't "remove" potential `must_use` of the `Ok` type thinking

clippy suggest to use `drop` because `let _ = ` does the same and is more explicit.
Didn't thought about about `ok` but it remove the must_use on the inner type because it will be hidden in a option so
`Result<Result<_, _>, _>::ok()` does not trigger the lint.

Allowing the linter also does not work. I think `drop` is a bit more explicit then converting to an option.

Robert-Steiner:
i have added a test for that in #166

acrrd:
bors merge

acrrd:
bors merge

PR #194, opened 2022-02-24
[Merged by Bors] - TY-2497 - Fix errors when deserializing articles with null fields

x3ro:
[TY-2497]

We discovered that, contrary to their documentation, Newscatcher responses
sometimes contain articles which are missing non-optional fields, causing
our deserializing to fail. As a remedy, we're introducing defaults for all
fields which may potentially exhibit this behaviour. In a follow-up, we may
need to introduce additional filtering to discard deserialized articles which
are missing crucial information (such as the excerpt).

[TY-2497]: https://xainag.atlassian.net/browse/TY-2497?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine_core/providers/src/newscatcher.rs
+++ b/discovery_engine_core/providers/src/newscatcher.rs
@@ -45,57 +45,106 @@ pub enum Topic {
 #[derive(Clone, Deserialize, Debug)]
 pub struct Article {
     /// Newscatcher API's unique identifier for each news article.
-    #[serde(rename(deserialize = "_id"))]
+    #[serde(
+        rename(deserialize = "_id"),
+        deserialize_with = "deserialize_null_default"
+    )]
     pub id: String,
 
     /// The title of the article.
+    #[serde(deserialize_with = "deserialize_null_default")]
     pub title: String,
 
     /// How well the article is matching your search criteria.
-    #[serde(rename(deserialize = "_score"))]
+    #[serde(
+        rename(deserialize = "_score"),
+        deserialize_with = "deserialize_null_default"
+    )]
     pub score: Option<f32>,
 
     /// The page rank of the source website.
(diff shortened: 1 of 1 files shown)


discovery_engine_core/providers/src/newscatcher.rs
```diff
@@ -98,6 +104,20 @@ where
     NaiveDateTime::parse_from_str(&s, "%Y-%m-%d %H:%M:%S").map_err(de::Error::custom)
 }
 
+fn default_topic() -> Topic {
```
acrrd, line 107: I think we can use derivative (https://mcarton.github.io/rust-derivative/latest/Default.html) to set `Unrecognized` as the default value and use `#[serde(default)]` instead on the field.

discovery_engine_core/providers/src/newscatcher.rs
```diff
@@ -49,31 +49,37 @@ pub struct Article {
...
     /// The title of the article.
+    #[serde(deserialize_with = "deserialize_null_default")]
     pub title: String,
 
     /// How well the article is matching your search criteria.
     #[serde(rename(deserialize = "_score"))]
     pub score: Option<f32>,
 
     /// The page rank of the source website.
+    #[serde(deserialize_with = "deserialize_null_default")]
```
acrrd, line 60: Since when we rank without cois we use rank for the headlines I think here would be better to have `usize::MAX` instead of 0 to penalize documents without infos.

acrrd (approved):

acrrd (approved):
We can merge this and address the comments in another PR.

acrrd (approved):

PR #195, opened 2022-02-24
[Merged by Bors] - TY-2492 Write tool to fetch lots of articles to verify our deserialization logic works

x3ro:
[TY-2492] 

The addition of `serde_path_to_error` seems worthwile, as it significantly improves the error messages in case something goes wrong when deserializing. With this change we'll get something like:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value:
DecodingAtPath("articles[93].excerpt", Error { path: Path { segments: [Map { key: "articles" },
Seq { index: 93 }, Map { key: "excerpt" }] }, original: Error("invalid type: null, expected a string",
line: 1, column: 64906) })', tooling/src/newscatcher.rs:45:66
```

Indicating the place where the error occurred (`articles[93].excerpt` in this case).

[TY-2492]: https://xainag.atlassian.net/browse/TY-2492?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine_core/providers/src/client.rs
+++ b/discovery_engine_core/providers/src/client.rs
@@ -33,8 +33,15 @@ pub enum Error {
     RequestExecution(#[source] reqwest::Error),
     /// Server returned a non-successful status code: {0}
     StatusCode(#[source] reqwest::Error),
+    /// Failed to fetch from the server: {0}
+    Fetching(#[source] reqwest::Error),
     /// Failed to decode the server's response: {0}
-    Decoding(#[source] reqwest::Error),
+    Decoding(#[source] serde_json::Error),
+    /// Failed to decode the server's response at JSON path {1}: {0}
+    DecodingAtPath(
+        String,
+        #[source] serde_path_to_error::Error<serde_json::Error>,
+    ),
 }
 
 /// Client that can provide documents.
@@ -62,6 +69,8 @@ pub struct HeadlinesQuery<'a> {
     pub market: &'a Market,
     /// How many articles to return (per page).
     pub page_size: usize,
+    /// Which page of the results to return.
+    pub page: usize,
 }
 
 impl Client {
@@ -88,7 +97,7 @@ impl Client {
             .error_for_status()
             .map_err(Error::StatusCode)?;
 
-        let news: NewscatcherResponse = response.json().await.map_err(Error::Decoding)?;
+        let news: NewscatcherResponse = response.json().await.map_err(Error::Fetching)?;
         let result: Vec<Article> = news.articles.into_iter().collect();
         Ok(result)
     }
@@ -114,6 +123,16 @@ impl Client {
 
     /// Retrieve headlines from the remote API
     pub async fn headlines(&self, params: &HeadlinesQuery<'_>) -> Result<Vec<Article>, Error> {
+        let news = self.headlines_query(params).await?;
+        let result: Vec<Article> = news.articles.into_iter().collect();
+        Ok(result)
+    }
+
+    /// Performs a query to retrieve headlines against the Newscatcher API
+    pub async fn headlines_query(
+        &self,
+        params: &HeadlinesQuery<'_>,
+    ) -> Result<NewscatcherResponse, Error> {
         let mut url = Url::parse(&self.url).map_err(|e| Error::InvalidUrlBase(Some(e)))?;
         Self::build_headlines_query(&mut url, params)?;
 
@@ -128,9 +147,10 @@ impl Client {
             .error_for_status()
             .map_err(Error::StatusCode)?;
 
-        let news: NewscatcherResponse = response.json().await.map_err(Error::Decoding)?;
-        let result: Vec<Article> = news.articles.into_iter().collect();
-        Ok(result)
+        let raw_response = response.text().await.map_err(Error::Fetching)?;
+        let deserializer = &mut serde_json::Deserializer::from_str(&raw_response);
+        serde_path_to_error::deserialize(deserializer)
+            .map_err(|error| Error::DecodingAtPath(error.path().to_string(), error))
     }
 
     fn build_headlines_query(url: &mut Url, params: &HeadlinesQuery<'_>) -> Result<(), Error> {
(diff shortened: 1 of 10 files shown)


discovery_engine_core/providers/src/client.rs
```diff
@@ -122,9 +141,10 @@ impl Client {
             .error_for_status()
             .map_err(Error::StatusCode)?;
 
-        let news: NewscatcherResponse = response.json().await.map_err(Error::Decoding)?;
-        let result: Vec<Article> = news.articles.into_iter().collect();
-        Ok(result)
+        let raw_response = response.text().await.map_err(Error::Fetching)?;
+        let deserializer = &mut serde_json::Deserializer::from_str(raw_response.as_str());
```
janpetschexain, line 145: this could be coerced automatically
```suggestion
        let deserializer = &mut serde_json::Deserializer::from_str(&raw_response);
```

discovery_engine_core/providers/src/client.rs
```diff
@@ -122,9 +141,10 @@ impl Client {
...
+        serde_path_to_error::deserialize(deserializer)
+            .map_err(|x| Error::DecodingAtPath(x.path().to_string(), x))
```
janpetschexain, line 147: `error` probably is more telling than `x`
```suggestion
            .map_err(|error| Error::DecodingAtPath(error.path().to_string(), error))
```

discovery_engine_core/providers/src/lib.rs
```diff
@@ -33,4 +33,4 @@ mod newscatcher;
 
 pub use client::{Client, HeadlinesQuery, NewsQuery};
 pub use filter::{Filter, Market};
-pub use newscatcher::{Article, Topic};
+pub use newscatcher::{Article, Response as NewscatcherResponse, Topic};
```
janpetschexain, line 36: can `Response` be confused with anything else that the crate exports? otherwise we don't have to rename it. if we have to rename it, we should do so consistently also for the other exported types of `newscatcher::`.

discovery_engine_core/tooling/src/newscatcher.rs
```diff
@@ -0,0 +1,61 @@
+// Copyright 2022 Xayn AG
```
janpetschexain, line 1: since this file is a bin as per the cargo declaration, should we put it under `bin/newscatcher.rs` instead of `src/newscatcher.rs`?

discovery_engine_core/tooling/src/newscatcher.rs
```diff
@@ -0,0 +1,61 @@
...
+use anyhow::{Context, Result};
+use xayn_discovery_engine_providers::{Client, HeadlinesQuery, Market};
+
+#[tokio::main]
+async fn main() -> Result<()> {
+    let url = "https://api-gw.xaynet.dev".to_string();
+    let token = std::env::var("GATEWAY_TOKEN").context(
+        "Please provide the GATEWAY_TOKEN environment variable for the dev environment. \
+                  The token can be found in 1Password",
+    )?;
```
janpetschexain, line 24: in 1password this is called bearer token, and there are different tokens, also without a prefix the env name is likely to have conflicts with other env. what about the following name instead?
```suggestion
    let token = std::env::var("NEWSCATCHER_DEV_BEARER_AUTH_TOKEN").context(
        "Please provide the NEWSCATCHER_DEV_BEARER_AUTH_TOKEN environment variable for the dev environment. \
                  The token can be found in 1Password",
    )?;
```

janpetschexain (commented):
works on my machine 👍

discovery_engine_core/tooling/src/newscatcher.rs
```diff
@@ -0,0 +1,61 @@
...
+    let market = Market {
+        lang_code: "en".to_string(),
+        country_code: "US".to_string(),
+    };
+
+    // This is updated every iteration, based on the response from Newscatcher. So in reality,
+    // we'll be fetching more than two pages.
+    let mut total_pages = 2;
+    let mut page = 1;
+    while page < total_pages {
```
janpetschexain, line 40: the counting is off by one, the final file and log i get is `Fetching page 99 of 100`. if the `page = total_pages` exist(?), then we should fetch it as well.
```suggestion
    // This is updated every iteration, based on the response from Newscatcher. So in reality,
    // we'll be fetching more than one page.
    let mut total_pages = 1;
    let mut page = 1;
    while page <= total_pages {
```

discovery_engine_core/tooling/src/newscatcher.rs
x3ro: Running `cargo new --bin` creates a `src/main.rs`, and I thought that was a bit generic, so I renamed it. I've also seen `src/bin/foo.rs` in other projects, but haven't come across just `bin/foo.rs`. No strong feelings in any direction though. The only thing I'm not sure of is if IntelliJ / VSCode plugins pick up `crate/bin/foo.rs` properly. 

Long story short: I don't particulary care, so if you do I'm happy to move it around :)

janpetschexain: i fully agree on the `newscatcher.rs` naming, any telling name is better than a generic `main.rs`.

regading the directory name, the [cargo book says the following](https://doc.rust-lang.org/cargo/reference/cargo-targets.html#binaries):
> Binary targets are executable programs that can be run after being compiled. The default binary filename is `src/main.rs`, which defaults to the name of the package. Additional binaries are stored in the `src/bin/` directory.

since the `newscatcher.rs` can be one of many binaries in the `tooling` crate, i'd says it's better to put it in a directory for additional binaries. i'm fine with either `src/bin/newscatcher.rs` or `bin/newscatcher` (we just have to add a `path = bin/newscatcher.rs` to the `[[bin]]` section in this case), though i have a slight preference for the latter as this would align the directory structure with the `benches/` and `examples/` directories we sometimes have.

janpetschexain (approved):
lgtm

x3ro:
bors merge

PR #196, opened 2022-02-24
[Merged by Bors] - TY-2433 Extended the `Document` model for active search

biernacki:
[Prepare repo for searched `Documents`](https://xainag.atlassian.net/browse/TY-2433)
[Adapt feed manager](https://xainag.atlassian.net/browse/TY-2442)

This PR adds a new attribute to the `Document` model to differentiate between documents that belong to a feed, and those which belong to active search.

- In feed manager we filter out the documents that belong to an active search
- `removeByIds` method was added to `DocumentRepository` to prepare for removing unneeded documents from the repo when a search is closed
- some smaller changes were made to the document manager, to clean it up a bit

### DIFF
--- a/discovery_engine/lib/src/domain/feed_manager.dart
+++ b/discovery_engine/lib/src/domain/feed_manager.dart
@@ -80,11 +80,6 @@ class FeedManager {
             });
 
           final feed = sortedActives.map((doc) => doc.toApiDocument()).toList();
-
-          if (feed.isEmpty) {
-            const reason = FeedFailureReason.noNewsForMarket;
-            return const EngineEvent.feedRequestFailed(reason);
-          }
           return EngineEvent.feedRequestSucceeded(feed);
         },
       );
--- a/discovery_engine/lib/src/domain/feed_manager.dart
+++ b/discovery_engine/lib/src/domain/feed_manager.dart
@@ -15,7 +15,7 @@
 import 'package:xayn_discovery_engine/src/api/events/client_events.dart'
     show FeedClientEvent;
 import 'package:xayn_discovery_engine/src/api/events/engine_events.dart'
-    show EngineEvent;
+    show EngineEvent, FeedFailureReason;
 import 'package:xayn_discovery_engine/src/domain/engine/engine.dart'
     show Engine;
 import 'package:xayn_discovery_engine/src/domain/models/unique_id.dart'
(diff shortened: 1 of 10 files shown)


discovery_engine/lib/src/domain/models/document.dart
```diff
@@ -47,19 +47,32 @@ class Document {
...
-    DocumentId? documentId,
+    required this.documentId,
     this.userReaction = UserReaction.neutral,
     this.isActive = true,
-  })  : documentId = documentId ?? DocumentId(),
-        timestamp = DateTime.now().toUtc();
+    this.isSearched = false,
+  }) : timestamp = DateTime.now().toUtc();
+
+  /// Register a [UserReaction]
```
janpetschexain, line 72:
```suggestion
  /// Registers a [UserReaction].
```

discovery_engine/lib/src/domain/models/document.dart
```diff
@@ -47,19 +47,32 @@ class Document {
...
+  void registerReaction(UserReaction reaction) {
```
janpetschexain, line 73: the `userReaction` field is public anyways and therefore has a default setter, it might cause more confusion than benefits to have another setter with a different name but same functionality.

i get that you want to call it in the ellipsis
```dart
doc..registerReaction(userReaction)
```
but this can already be achieved by
```dart
doc..userReaction = userReaction
```

or we could call it like the following (imo a bit easier to read)
```dart
doc.userReaction = userReaction;
doc
```

discovery_engine/lib/src/domain/feed_manager.dart
```diff
@@ -74,6 +80,11 @@ class FeedManager {
             });
 
           final feed = sortedActives.map((doc) => doc.toApiDocument()).toList();
+
+          if (feed.isEmpty) {
```
acrrd, line 84: When the user start the app the feed is empty and is not error. Also this is a breaking change.

discovery_engine/lib/src/domain/models/document.dart
acrrd: also 'register' is a bit weird I would expect to save it somewhere not only assign it to the field.

discovery_engine/lib/src/domain/feed_manager.dart
acrrd: The user can also decide to close all the documents in the feed.

discovery_engine/lib/src/domain/models/document.dart
biernacki: you're right, I removed the method and reverted back to a default setter for user reaction.

discovery_engine/lib/src/domain/feed_manager.dart
biernacki: I reverted this back to previous state.

If the empty case is not meant to be represented by `FeedRequestFailed` event, then maybe we don't need it at all?

janpetschexain (approved):
lgtm

biernacki:
bors merge

PR #198, opened 2022-02-24
[Merged by Bors] - TY-2353 handle config changed event

janpetschexain:
**References**

- [TY-2353]

**Summary**

- impl the handler for the `ConfigurationChanged` event in the `SystemManager`
- return error if `Init` is called on an already initialized engine

[TY-2353]: https://xainag.atlassian.net/browse/TY-2353?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine/lib/src/domain/engine/mock_engine.dart
+++ b/discovery_engine/lib/src/domain/engine/mock_engine.dart
@@ -103,8 +103,8 @@ class MockEngine implements Engine {
 
   @override
   Future<void> setMarkets(
-    final List<HistoricDocument> history,
-    final FeedMarkets markets,
+    List<HistoricDocument> history,
+    FeedMarkets markets,
   ) async {
     _incrementCount('setMarket');
   }
--- a/discovery_engine/lib/src/domain/engine/mock_engine.dart
+++ b/discovery_engine/lib/src/domain/engine/mock_engine.dart
@@ -23,7 +23,7 @@ import 'package:xayn_discovery_engine/src/domain/models/document.dart'
 import 'package:xayn_discovery_engine/src/domain/models/embedding.dart'
     show Embedding;
 import 'package:xayn_discovery_engine/src/domain/models/feed_market.dart'
-    show FeedMarket;
+    show FeedMarkets;
 import 'package:xayn_discovery_engine/src/domain/models/history.dart'
     show HistoricDocument;
 import 'package:xayn_discovery_engine/src/domain/models/news_resource.dart'
@@ -104,7 +104,7 @@ class MockEngine implements Engine {
   @override
   Future<void> setMarkets(
     List<HistoricDocument> history,
-    List<FeedMarket> markets,
+    FeedMarkets markets,
   ) async {
     _incrementCount('setMarket');
   }
--- a/discovery_engine/lib/src/domain/engine/mock_engine.dart
+++ b/discovery_engine/lib/src/domain/engine/mock_engine.dart
@@ -103,8 +103,8 @@ class MockEngine implements Engine {
 
   @override
   Future<void> setMarkets(
-    List<HistoricDocument> history,
-    FeedMarkets markets,
+    final List<HistoricDocument> history,
+    final FeedMarkets markets,
   ) async {
     _incrementCount('setMarket');
   }
--- a/discovery_engine/lib/src/domain/event_handler.dart
+++ b/discovery_engine/lib/src/domain/event_handler.dart
@@ -51,14 +51,6 @@ import 'package:xayn_discovery_engine/src/domain/models/news_resource.dart'
     show NewsResourceAdapter;
 import 'package:xayn_discovery_engine/src/domain/models/view_mode.dart'
     show DocumentViewModeAdapter;
-import 'package:xayn_discovery_engine/src/domain/repository/active_document_repo.dart'
-    show ActiveDocumentDataRepository;
-import 'package:xayn_discovery_engine/src/domain/repository/changed_document_repo.dart'
-    show ChangedDocumentRepository;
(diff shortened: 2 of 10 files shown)


discovery_engine/lib/src/domain/engine/mock_engine.dart
```diff
@@ -103,8 +103,8 @@ class MockEngine implements Engine {
 
   @override
   Future<void> setMarkets(
-    List<HistoricDocument> history,
-    List<FeedMarket> markets,
+    final List<HistoricDocument> history,
```
acrrd, line 106: We decided to not enable the linter to enforce final parameters so they should stay as they are.

discovery_engine/lib/src/domain/feed_manager.dart
```diff
@@ -47,18 +50,47 @@ class FeedManager {
...
+      } catch (e, st) {
+        return EngineEvent.engineExceptionRaised(
+          EngineExceptionReason.genericError,
+          message: '$e',
+          stackTrace: '$st',
+        );
+      }
+    }
+    if (maxItemsPerFeedBatch != null) {
+      _maxDocs = maxItemsPerFeedBatch;
```
acrrd, line 88: We need to check that is > 0

discovery_engine/lib/src/domain/event_handler.dart
```diff
@@ -167,8 +166,6 @@ class EventHandler {
         response = await _feedManager.handleFeedClientEvent(clientEvent);
       } else if (clientEvent is DocumentClientEvent) {
         await _documentManager.handleDocumentClientEvent(clientEvent);
-      } else if (clientEvent is SystemClientEvent) {
-        // TODO: we need to handle other system events (config changed)
```
acrrd, line 171: We should handle the event here. We can keep an `Engine?` in the fields instead of `Future<Engine>?`.
For the `maxDoc` that `FeedManager` needs we can have an internal configuration object that we update here and that the `FeedManager` can use to read. This will be also used by the `ActiveSearchManager`.

discovery_engine/lib/src/domain/engine/mock_engine.dart
janpetschexain: actually they are already `final` in the `Engine` (or more precisely in the `DiscoveryEngineFfi` implementation of the `Engine`) and this just aligns the `MockEngine`.

since we have a mixture of `final` and non-`final` arguments all over the code base, how should we proceed?

rustonaut: I would say:

- not add `final` to new parameters
- when touching a function with final parameters, remove the final
- do not remove final on functions in files which else-wise would not have been touched

discovery_engine/lib/src/domain/event_handler.dart
rustonaut: > await engine.requestFeed();

Or assign the engine returned from awaiting the future to a variable on line 151/152.

acrrd: but we have a variable around that we don't need.

discovery_engine/lib/src/domain/engine/mock_engine.dart
acrrd: We decided to not force it, not to not used it. If we never want final on parameters we should check if there is a linter for that.

discovery_engine/lib/src/domain/event_handler.dart
rustonaut: :thinking: Are we speaking about the same thing?

Like what fields do you mean?

acrrd: `_engineFuture`

rustonaut: But we want to make `_engineFuture` a future to avoid any potential race conditions between init started and making a request before initialization completes.

rustonaut: And that future will return the `Engine` instance, so if we need to use that anywhere we can just store it at l.151/152 in a variable.

rustonaut: As a related but slightly off-topic side question: How are we meant to handle engine failure (after returning an error). Will we retry the initialization?

acrrd: > But we want to make `_engineFuture` a future to avoid any potential race conditions between init started and making a request before initialization completes.

We can do the same.
```dart
Engine? _engine;

... handleEvent(...) {
  if(event is Init) {
     _engine = await _initEngine(...);
  }
  
  if(_engine == null) return Error(EngineNotReady);
}
```
We should also throw an error if the engine is already initialized and the user try to init it again.

acrrd: The only difference that I see now is that now a message sent after `Init` will wait, while with this it will fail.
I don't see a reason to wait. Is up to the user to send us messages after he knows that the engine has been initialized correctly.

acrrd: > As a related, but slightly off-topic side question: How are we meant to handle engine failure (after returning an error). Will we retry the initialization?

Why we should reinit after an error? We don't really have anything that can be fixed by a reinit.

rustonaut: We have nothing which automatically can be fixed, but I could imagine Team Blue having some way to change the config in dev builds, potentially in ways which could make it fail. 

Anyway they can then just send `Init` again, I guess.

acrrd (approved):

janpetschexain:
bors merge

PR #200, opened 2022-02-25
[Merged by Bors] - TY-2435 add page param to news query

finiteprods:
**Summary**

adds a `page` field to the `NewsQuery` struct and updates the `news` method to begin supporting active search.

### DIFF
--- a/discovery_engine_core/core/src/stack/ops/personalized.rs
+++ b/discovery_engine_core/core/src/stack/ops/personalized.rs
@@ -64,6 +64,7 @@ impl Ops for PersonalizedNews {
                     market,
                     filter: filter.clone(),
                     page_size: self.page_size,
+                    page: None,
                 };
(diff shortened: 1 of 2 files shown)


janpetschexain (approved):
lgtm

i'm lacking some overview of the providers, why is the task only about adding `page` to `NewsQuery` but not to `HeadlinesQuery` as well? it looks to me like the same reasoning applies for both, regarding the possibly too slowly changing content mentioned in the daily.

finiteprods:
> why is the task only about adding `page` to `NewsQuery` but not to `HeadlinesQuery` as well?

fair question @acrrd is it needed for headlines?

acrrd:
> fair question @acrrd is it needed for headlines?

Because we cannot do custom query on the headlines.

finiteprods:
bors merge

PR #203, opened 2022-02-25
[Merged by Bors] - TY-2315 filter previously seen articles

janpetschexain:
**References**

- [TY-2315]

**Summary**

- update `ArticleFilter` trait with `history: &[HistoricDocument]`
- impl `HistoryFilter` and apply it in `CommonFilter`

[TY-2315]: https://xainag.atlassian.net/browse/TY-2315?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine_core/core/src/document.rs
+++ b/discovery_engine_core/core/src/document.rs
@@ -283,6 +283,19 @@ mod tests {
         }
     }
 
+    impl TryFrom<Article> for HistoricDocument {
+        type Error = Error;
+
+        fn try_from(article: Article) -> Result<Self, Self::Error> {
+            Ok(Self {
+                id: Uuid::new_v4().into(),
+                url: Url::parse(&article.link)?,
+                snippet: article.excerpt,
+                title: article.title,
+            })
+        }
+    }
+
     #[test]
     fn test_news_resource_from_article() {
         let article = mock_article();
--- a/discovery_engine_core/core/src/stack/filters.rs
+++ b/discovery_engine_core/core/src/stack/filters.rs
@@ -30,35 +30,18 @@ pub(crate) trait ArticleFilter {
     ) -> Result<Vec<Article>, GenericError>;
 }
 
-struct HistoryFilter;
+struct DuplicateFilter;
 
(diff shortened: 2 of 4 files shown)


discovery_engine_core/core/src/document.rs
```diff
@@ -282,6 +282,19 @@ mod tests {
         }
     }
 
+    impl TryFrom<Article> for HistoricDocument {
```
acrrd, line 285: If we need  it make tests simpler we can define it in test module instead.

discovery_engine_core/core/src/stack/filters.rs
```diff
@@ -12,23 +12,51 @@
...
+    fn apply(
+        history: &[HistoricDocument],
+        stack: &[Document],
+        articles: Vec<Article>,
+    ) -> Result<Vec<Article>, GenericError>;
+}
+
+struct HistoryFilter;
+
+impl ArticleFilter for HistoryFilter {
```
acrrd, line 35: Can't we do this already in `Duplicatefilter`?

discovery_engine_core/core/src/document.rs
janpetschexain: not sure what you mean? this impl is already in a test module.

discovery_engine_core/core/src/stack/filters.rs
janpetschexain: sure we can. my impression of the previous design was that all filters should be separate such that they can be combined in the `CommonFilter` later on. but we can also combine the `HistoryFilter` and the `DuplicateFilter` into one, i don't mind much.

discovery_engine_core/core/src/document.rs
acrrd: 🙈 I'm blind sorry.

acrrd (approved):

janpetschexain:
bors merge

PR #207, opened 2022-02-28
[Merged by Bors] - TY-2505 Android example fixes

biernacki:
[Ref](https://xainag.atlassian.net/browse/TY-2505)

- Bumped kotlin and adroid sdk versions for the example app to make it compatible with latest flutter.
- Adjusted engine init procedure in the example

### DIFF
--- a/discovery_engine_flutter/example/android/app/build.gradle
+++ b/discovery_engine_flutter/example/android/app/build.gradle
@@ -26,7 +26,7 @@ apply plugin: 'kotlin-android'
 apply from: "$flutterRoot/packages/flutter_tools/gradle/flutter.gradle"
 
 android {
-    compileSdkVersion 30
+    compileSdkVersion 31
 
     compileOptions {
         sourceCompatibility JavaVersion.VERSION_1_8
--- a/discovery_engine_flutter/example/lib/main.dart
+++ b/discovery_engine_flutter/example/lib/main.dart
@@ -41,8 +41,14 @@ class _MyAppState extends State<MyApp> {
   double progress = .0;
   EngineState engineState = EngineState.notReady;
 
+  @override
+  void initState() {
+    super.initState();
+    _initEngine();
+  }
+
   // Platform messages are asynchronous, so we initialize in an async method.
-  Future<void> initEngine() async {
+  Future<void> _initEngine() async {
     if (engineState != EngineState.notReady) return;
     setState(() => engineState = EngineState.initializing);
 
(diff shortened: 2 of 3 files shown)


discovery_engine_flutter/example/android/app/build.gradle
```diff
@@ -26,7 +26,7 @@ apply plugin: 'kotlin-android'
 apply from: "$flutterRoot/packages/flutter_tools/gradle/flutter.gradle"
 
 android {
-    compileSdkVersion 30
+    compileSdkVersion 31
```
acrrd, line 29: Where you using flutter 2.5? I realized I was using flutter 2.10

biernacki: me too, I'll downgrade and check again

discovery_engine_flutter/example/lib/main.dart
```diff
@@ -114,17 +115,21 @@ class _MyAppState extends State<MyApp> {
...
-                  child: engineState == EngineState.notReady
-                      ? const Text('Init engine')
-                      : const Text('Initializing...'),
+                  onPressed: requestNews,
+                  child: const Text('Request News'),
+                ),
+              ],
+              if (engineState == EngineState.notReady)
+                ElevatedButton(
+                  onPressed: initEngine,
```
rustonaut, line 128: Can we automatically call `initEngine` when build is called. Also fetch one batch automatically.

It will fast exit if it was called before.

And there is no reason to wait for a user interaction.

I did that change ad-hoc on all branches where I tested things regarding the compiler bug.

rustonaut: We also probably still want to call `await engine.requestFeed();` in `initEngine`.

As a side note, we probably call it `restoreFeed`. I found it being named `requestFeed` prone to confusing with `requestNextFeedBatch`.

discovery_engine_flutter/example/android/app/build.gradle
biernacki: the bumped versions work for 2.5.3 and for 2.10, the old ones work only for 2.5.3, so let's maybe keep the bumped ones

discovery_engine_flutter/example/lib/main.dart
biernacki: I've dropped the need for user interaction, added `requestFeed`.

I agree we should rename it to `restoreFeed`, but maybe let's do it after the beta release, cause it's a breaking change, wdyt?

acrrd: We also talked with Frank about this renaming, we can do it after the release

rustonaut (approved):
lgtm

PR #210, opened 2022-02-28
[Merged by Bors] - TY-2474 add engine restore tests

Robert-Steiner:
Ticket:

- [TY-2474]

[TY-2474]: https://xainag.atlassian.net/browse/TY-2474?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #212, opened 2022-02-28
[Merged by Bors] - TY-2472 test user reacted event

Robert-Steiner:
Ticket:

- [TY-2472]

based on #210 

[TY-2472]: https://xainag.atlassian.net/browse/TY-2472?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine/test/integration/init_engine_test.dart
+++ b/discovery_engine/test/integration/init_engine_test.dart
@@ -19,8 +19,8 @@ import 'package:xayn_discovery_engine/discovery_engine.dart'
     show DiscoveryEngine;
 
 import '../logging.dart' show setupLogging;
-import 'utils/create_config.dart'
-    show TestEngineData, createConfig, setupTestEngineData;
+import 'utils/helpers.dart'
+    show TestEngineData, initEngine, setupTestEngineData;
 import 'utils/local_newsapi_server.dart' show LocalNewsApiServer;
 
 void main() {
@@ -41,21 +41,17 @@ void main() {
     });
 
     test('init engine with ai models', () async {
-      final engine = await DiscoveryEngine.init(
-        configuration: createConfig(data, server.port),
-      );
-
+      final engine = await initEngine(data, server.port);
       expect(engine, isA<DiscoveryEngine>());
+      await engine.dispose();
     });
 
     test('news api request error should not raise an engine init exception',
         () async {
       server.replyWithError = true;
-      final engine = await DiscoveryEngine.init(
-        configuration: createConfig(data, server.port),
-      );
-
+      final engine = await initEngine(data, server.port);
       expect(engine, isA<DiscoveryEngine>());
+      await engine.dispose();
     });
   });
 }
--- a/discovery_engine/test/integration/init_engine_test.dart
+++ b/discovery_engine/test/integration/init_engine_test.dart
@@ -32,6 +32,7 @@ void main() {
 
     setUp(() async {
       data = await setupTestEngineData();
+      server = await LocalNewsApiServer.start();
     });
 
     tearDown(() async {
@@ -40,7 +41,6 @@ void main() {
     });
 
     test('init engine with ai models', () async {
-      server = await LocalNewsApiServer.start();
       final engine = await DiscoveryEngine.init(
         configuration: createConfig(data, server.port),
       );
@@ -50,7 +50,6 @@ void main() {
 
     test('news api request error should not raise an engine init exception',
         () async {
-      server = await LocalNewsApiServer.start();
       server.replyWithError = true;
       final engine = await DiscoveryEngine.init(
         configuration: createConfig(data, server.port),
--- a/discovery_engine/test/integration/user_reacted_test.dart
+++ b/discovery_engine/test/integration/user_reacted_test.dart
@@ -1,0 +1,121 @@
+// Copyright 2022 Xayn AG
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, version 3.
+//
+// This program is distributed in the hope that it will be useful,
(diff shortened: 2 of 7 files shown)


discovery_engine/test/integration/user_reacted_test.dart
```diff
@@ -0,0 +1,121 @@
...
+    tearDown(() async {
+      await server.close();
+      await Directory(data.applicationDirectoryPath).delete(recursive: true);
+    });
+
+    test('change the user reaction of a document', () async {
+      server = await LocalNewsApiServer.start();
+      var engine = await DiscoveryEngine.init(
+        configuration: createConfig(data, server.port),
+      );
```
janpetschexain, line 57: this pattern repeats over several tests, should we encapsulate this in a utility function as well? (but only if this doesn't make too much work, as we don't have a native tuple type to return in dart)

janpetschexain (approved):
lgtm

discovery_engine/test/integration/user_reacted_test.dart
Robert-Steiner: Good idea :+1 

I also moved the server + engine (when possible) into the setup function and call `engine.dispose()` at the end.
Please have another look

discovery_engine/test/integration/utils/helpers.dart
```diff
@@ -15,7 +15,12 @@
 import 'dart:io';
 
 import 'package:xayn_discovery_engine/discovery_engine.dart'
-    show Manifest, createManifestReader, Configuration, FeedMarket;
+    show
+        Configuration,
+        DiscoveryEngine,
+        FeedMarket,
+        Manifest,
+        createManifestReader;
```
janpetschexain, line 23: it's weird that dart complains about unsorted `import`s, but not about unsorted `show`s.
```suggestion
        Configuration,
        createManifestReader,
        DiscoveryEngine,
        FeedMarket,
        Manifest;
```

discovery_engine/test/integration/init_engine_test.dart
```diff
@@ -40,23 +41,17 @@ void main() {
...
 
     test('init engine with ai models', () async {
-      server = await LocalNewsApiServer.start();
-      final engine = await DiscoveryEngine.init(
-        configuration: createConfig(data, server.port),
-      );
-
+      final engine = await initEngine(data, server.port);
       expect(engine, isA<DiscoveryEngine>());
+      await engine.dispose();
```
janpetschexain, line 46: this line could be moved to the `tearDown()` like in the other tests. similarly below.

discovery_engine/test/integration/init_engine_test.dart
```diff
@@ -40,23 +41,17 @@ void main() {
     });
 
     test('init engine with ai models', () async {
-      server = await LocalNewsApiServer.start();
-      final engine = await DiscoveryEngine.init(
-        configuration: createConfig(data, server.port),
-      );
-
+      final engine = await initEngine(data, server.port);
```
janpetschexain, line 44: this line could be moved to the `setUp()` like in the other tests. similarly below.

janpetschexain (commented):
lgtm. a test has timed out though, is it enough to increase the timeout or is this due to some external factors?

discovery_engine/test/integration/init_engine_test.dart
Robert-Steiner: unfortunately, that will break the second test case. it will pass but we don't check if the initialisation is   successful in case of an api error . but we can move the dispose into `tearDown()`

Robert-Steiner:
> lgtm. a test has timed out though, is it enough to increase the timeout or is this due to some external factors?

hm, that timeout is not so good. i also got that error locally after i rebased and added the `engine.dispose()` calls. I thought that happened because I was running the test while i was alter them. I will take a closer look

Robert-Steiner:
it would be possible that we spawn too many engines at the same time and thus everything is slowed down.

Robert-Steiner:
the changeUserReaction request takes on my machine around 9.5 seconds and out timeout is 10 seconds. maybe we should think about comiling the rust binary in release mode to speed things up

Robert-Steiner:
> the changeUserReaction request takes on my machine around 9.5 seconds and our timeout is 10 seconds. maybe we should think about compiling the rust binary in release mode to speed things up

we can also just change the opt-level of `tract-core` and `tract-data`
```
[profile.dev.package.tract-core]
opt-level = 3
[profile.dev.package.tract-data]
opt-level = 3
```
`tract-core` reduces the `time log_user_reaction` time from 15sec to 6sec
and `tract-data` reduces the time of the engine init from 9.5sec to 3sec

Robert-Steiner:
> Is the increase in compilation time noticeable?

yes, it takes 2 min more

what i saw is that they do 
```
[profile.dev.package."*"]
opt-level = 3
```

but this setting should not affect our build options or did i misunderstood you?

Robert-Steiner:
The new smbert model makes the initialisation of the engine much faster. Even without optimisations and running the tests in parallel it takes 4 sec instead of 15 sec. however `changeUserReaction` got slower from 9 sec to 14 sec (without opt, running in parallel). 

what we can do is: 
- enable `opt = 3` for `tract-core`
	- 7 sec running tests in parallel, 5 secs sequential
	- but that will take 2 min longer to compile (but it is not recompiled every time/ will be slower on the ci if we don't have cache hit) 

wdyt?

unfortunately, running the dart tests in sequentially will be to slow

Robert-Steiner:
like discussed in the daily we compile the tract-core crate with `opt=3` to make the dart test run faster

bors merge

Robert-Steiner:
bors merge

PR #213, opened 2022-02-28
Investigate what is needed to make discovery engine more robust to errors

rustonaut:
This PR commits a number of TODO mostly about how we could make the discovery engine more robust to errors, but also miscellaneous other things I noticed while going through the code.

I noticed that there are 3 kinds of errors:

- fatal ones (like initialization failed)
- expected non-fatal ones (like  network down)
- unexpected maybe-fatal ones (like some broken invariants)

Except for the first category in most other cases we still could "go one" and proceed running the discovery engine.

Sure in some cases we need to notify dart (e.g. so that it can display  that there are network problems to the user, or so that it can log a error/bug message which is somehow transmitted to us).

And sure in some cases it might degrade the usability of the app, but we still can go on.

For example:

- if de-serializing the state fails we can just not do so and pretend there was no state. (and report a bug)
- if serializing the state fails we also can go-on just every time the native context is recreated it forgets everything learned, but otherwise still works
- if some Url in an Article is unreadable we can skip that article, but keep the rest of the stack (or even just set the Url to `None` in some cases)
- for getting the next feed we already have documents before we update the stacks, so even if it fails we could return the current documents.
- if fetching fails we still can return a empty vec, until the stack runs out of documents that would work (and even after it would just be "useless" for the user but not "fail".
- in the sampler we can clamp alpha/beta instead of failing

Additional to that there are a bunch of "impossible" errors, where the used interfaces still return a result, but the error case is unreachable. Like ranking can't fail anymore as it will fallback to sorting by score. 

I believe it could make sense/be valuable to add a side-channel to report bugs, network errors and maybe later some stats and prune most error cases out of most use-cases. Always reporting them appropriately to the side-channel but then "going on" anyway.

I believe we can easily end up with a `update_stacks` not returning a `Result`  as well as less results in FFI.

With out question this is a bit of an unusual approach, especially for rust. But for our application might be the right choice. And by making the side-channel not just a logger we can still test everything properly.

### DIFF
--- a/discovery_engine/lib/src/domain/document_manager.dart
+++ b/discovery_engine/lib/src/domain/document_manager.dart
@@ -90,6 +90,9 @@ class DocumentManager {
         reaction: userReaction,
       ),
     );
+    //TODO do we want to try to save the engine state even if userReacted failes
+    //TODO If we can't save the engine state we might want to log report error,
+    //     but we still can continue with `notifyChanged()`
     await _engineStateRepo.save(await _engine.serialize());
     _changedDocsReporter.notifyChanged([doc]);
   }
@@ -100,6 +103,7 @@ class DocumentManager {
   Future<void> addActiveDocumentTime(
     DocumentId id,
     DocumentViewMode mode,
+    //TODO this should be `Duration time`
     int sec,
   ) async {
     if (sec < 0) {
@@ -130,6 +134,7 @@ class DocumentManager {
       ),
     );
 
+    //TODO do we want to try to save the engine state even if timeSpent failes
     await _engineStateRepo.save(await _engine.serialize());
   }
 }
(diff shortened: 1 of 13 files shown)


rustonaut:
Some other thinks I ran into:

- A bunch of small improvements around `Client` and `stack::Ops`
- I think it can make sense to decouple (spawn) `update_stacks`, this could make it more reactive (e.g. returning successful initialization before `update_stacks` complete, especially with long network delays this can be very valuable).
- Depending on how "stall" the feed is we might have the order of `update_stack` and sampling documents wrong.

rustonaut:
The only thing which probably should be done before beta is:

- when getting the next feed back we should return the documents we already sampled, even if update stack fails

Also maybe:

- filter out "bad" articles (url is not parsable), instead of discarding the whole list of articles returned

janpetschexain:
> The only thing which probably should be done before beta is:
>     * when getting the next feed back we should return the documents we already sampled, even if update stack fails

this case should be handled by #221 because `update_stacks()` doesn't throw anymore.

rustonaut:
Another thing:

Make allocation fallible in FFI?

acrrd:
Thanks for the comments!

PR #214, opened 2022-02-28
[Merged by Bors] - TY-2510 Assets checksum veryfication speed-up

biernacki:
[Ref](https://xainag.atlassian.net/browse/TY-2510)

Changes to the NativeSetupData provider made:

- bundled assets are copied to final location with a `.tmp` extension, waiting to be verified
- assets that are already in final location are guaranteed to be verified
- when there are no assets in the final location we check for `.tmp` assets and try to verify them
- if verification of temp assets fails, they are removed to speed up the next run
- if no assets are available locally, they are fetched from the server and then verified
- if the verification of fetched assets fails it will throw an exception


discovery_engine/lib/src/infrastructure/assets/native/data_provider.dart
```diff
@@ -65,36 +67,55 @@ class NativeDataProvider extends DataProvider {
...
-        asset,
-        onFetched: assetReporter.assetFetched,
-      );
-      await assetFile.writeAsBytes(bytes, flush: true);
-    } else if (asset.fragments.isEmpty) {
-      assetReporter.assetFetched(asset.urlSuffix);
-    } else {
+    // if the file exists it means it was verified and has a proper checksum
+    if (assetFileRef.existsSync()) {
+      if (asset.fragments.isEmpty) {
```
acrrd, line 75: If the file is in the proper place we should not care about the fragments. Is because how the the `assetReporter` works that we need to to send an event for each fragment anyway?

discovery_engine/lib/discovery_engine.dart
```diff
@@ -23,7 +23,7 @@ import 'package:xayn_discovery_engine/src/logger.dart' show initLogger, logger;
 export 'package:xayn_discovery_engine/src/api/api.dart';
 export 'package:xayn_discovery_engine/src/discovery_engine_base.dart';
 export 'package:xayn_discovery_engine/src/domain/assets/assets.dart'
-    show Manifest, kAssetsPath;
+    show Manifest, kAssetsPath, tmpFileExt;
```
acrrd, line 26: Why we need to expose `tmpfileExt`? (And why also `kAssetsPath` it seems it should be used only internally)

discovery_engine/lib/src/infrastructure/assets/native/data_provider.dart
```diff
@@ -65,36 +67,55 @@ class NativeDataProvider extends DataProvider {
...
       for (final fragment in asset.fragments) {
         assetReporter.assetFetched(fragment.urlSuffix);
       }
+      return assetFileRef.path;
+    }
+
+    // if the file doesn't exist we try to look for the temp file copied from
+    // bundled assets, waiting for checksum verification
+    final tmpFileRef = File('$filePath.$tmpFileExt');
+    final doesTmpFileExist = tmpFileRef.existsSync();
```
acrrd, line 87: There a race condition between this check and when we open the file with `_verifyChecksum`. If the file is not there an exception will be raised and we can use to to clean the temp file.

acrrd (commented):
only partial review

discovery_engine/lib/discovery_engine.dart
biernacki: it is used by the flutter plugin's `AssetCopier`

discovery_engine/lib/src/infrastructure/assets/native/data_provider.dart
biernacki: the AssetReporter sums all files or if the file is splitted all its fragments, and uses that info to calculate the progress, so if the asset has fragments, we need to report on each fragment.

acrrd (approved):

biernacki:
bors merge

PR #217, opened 2022-03-01
[Merged by Bors] - TY-2513 dedup by url or title

rustonaut:
**References:**

- [TY-2513](https://xainag.atlassian.net/browse/TY-2513)

### DIFF
--- a/discovery_engine_core/core/src/lib.rs
+++ b/discovery_engine_core/core/src/lib.rs
@@ -28,7 +28,7 @@
     clippy::missing_errors_doc,
     clippy::must_use_candidate,
     clippy::module_name_repetitions,
-    clippy::items_after_statements,
+    clippy::items_after_statements
 )]
 
 pub use xayn_discovery_engine_providers::Market;
--- a/discovery_engine_core/core/src/lib.rs
+++ b/discovery_engine_core/core/src/lib.rs
@@ -27,7 +27,8 @@
 #![allow(
     clippy::missing_errors_doc,
     clippy::must_use_candidate,
-    clippy::module_name_repetitions
+    clippy::module_name_repetitions,
+    clippy::items_after_statements,
 )]
 
 pub use xayn_discovery_engine_providers::Market;
--- a/discovery_engine_core/core/src/stack/filters.rs
+++ b/discovery_engine_core/core/src/stack/filters.rs
@@ -12,7 +12,10 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-use std::collections::HashSet;
+use std::{
+    cmp::Ordering,
+    collections::{BTreeSet, HashSet},
+};
 
 use url::Url;
 
@@ -39,12 +42,22 @@ impl PartialEq for UniqueArticle {
     }
 }
 
-impl std::hash::Hash for UniqueArticle {
-    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
-        self.0.link.hash(state);
-        self.0.title.hash(state);
+impl PartialOrd for UniqueArticle {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
     }
 }
+
+impl Ord for UniqueArticle {
+    fn cmp(&self, other: &Self) -> Ordering {
+        if self.eq(other) {
+            Ordering::Equal
+        } else {
+            (&self.0.title, &self.0.link).cmp(&(&other.0.title, &other.0.link))
+        }
+    }
+}
+
 struct DuplicateFilter;
 
 impl ArticleFilter for DuplicateFilter {
@@ -65,7 +78,7 @@ impl ArticleFilter for DuplicateFilter {
             .chain(stack.iter().map(|doc| &doc.resource.title))
             .collect::<HashSet<_>>();
 
-        let mut articles: HashSet<_> = articles.into_iter().map(UniqueArticle).collect();
+        let mut articles: BTreeSet<_> = articles.into_iter().map(UniqueArticle).collect();
 
         articles.retain(|article| {
             !(urls.contains(&article.0.link.as_str()) || titles.contains(&&article.0.title))
@@ -113,7 +126,7 @@ impl ArticleFilter for CommonFilter {
 
 #[cfg(test)]
 mod tests {
-    use std::convert::TryInto;
+    use std::{convert::TryInto, iter::FromIterator};
 
     use crate::{
         document::{document_from_article, Document},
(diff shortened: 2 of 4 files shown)


rustonaut:
**Open question:**

Currently we exclude articles in the history with the same url (and soon title). But we do not de-duplicate the same batch of articles itself.I just fetched a page with page size 100 and there where 100 unique urls, 93 unique titles, and 95 unique thumbnails.So should we dedup the articles themself by title?

**Probable answer:**

would say yes, because we filter it out when it's across page boundaries.
 
Side note:
 
Also in cases where the title was the same, the excerpt was somtimes but not always the same, I guess this is like when newspapers just copy the dpa announcements in some cases?

discovery_engine_core/core/src/stack/filters.rs
```diff
@@ -38,13 +38,26 @@ impl ArticleFilter for DuplicateFilter {
...
-        articles.retain(|article| !urls.contains(&article.link.as_str()));
+        let mut titles = history
+            .iter()
+            .map(|doc| Cow::from(&doc.title))
+            .chain(stack.iter().map(|doc| Cow::from(&doc.resource.title)))
+            .collect::<HashSet<_>>();
+
+        articles.retain(|article| {
+            let do_retain = !(urls.contains(&Cow::from(article.link.as_str()))
+                || titles.contains(&Cow::from(&article.title)));
```
acrrd, line 55: Not is missing for `titles.contains`. Can you please add a test for this case?

janpetschexain:
> Currently we exclude articles in the history with the same url (and soon title). But we do not de-duplicate the same batch of articles itself.I just fetched a page with page size 100 and there where 100 unique urls, 93 unique titles, and 95 unique thumbnails.So should we dedup the articles themself by title?

i agree, makes sense to dedup the articles themselves as well.

discovery_engine_core/core/src/stack/filters.rs
janpetschexain: the `!` is outside of the braces which enclose the `||`, so this should work as intended. we could rewrite the `!(url || titles)` to `!url && !titles` to make it easier to read.

discovery_engine_core/core/src/stack/filters.rs
```diff
@@ -38,13 +38,26 @@ impl ArticleFilter for DuplicateFilter {
...
+            urls.insert(Cow::from(article.link.to_string()));
+            titles.insert(Cow::from(article.title.clone()));
```
janpetschexain, line 57: with the example you gave this clones 2*100 strings just to filter out 7 articles. i get that we can't use the reference as it's not known outside the closure, so maybe we should address the issue in a different way: we could derive `Eq + Hash` for `Article` just for the `link` & `title` fields and pass around `HashSet<Article>` instead of `Vec<Article>` wherever we use articles. then the articles would already be unique and we can still use `HashSet::retain()` to filter them regarding the history & stack as before. wdyt?

discovery_engine_core/core/src/stack/filters.rs
acrrd: ah you are right. I thought the last `)` on the urls line was closing the not.

discovery_engine_core/core/src/stack/filters.rs
acrrd: Deriving will not work because we want `link` or `title`. Manually deriving a case like this seems dangerous since this a requirement of this filter and not a property of `Article`.

janpetschexain: fair point, makes sense to keep the logic in the filter. maybe another way to avoid the excessive cloning?

discovery_engine_core/core/src/stack/filters.rs
```diff
@@ -38,13 +38,26 @@ impl ArticleFilter for DuplicateFilter {
         stack: &[Document],
         mut articles: Vec<Article>,
     ) -> Result<Vec<Article>, GenericError> {
-        let urls = history
+        let mut urls = history
```
acrrd, line 41: We can move the type `HashSet<_>` here instead of at the end of the expression.

discovery_engine_core/core/src/stack/filters.rs
acrrd: I have a couple of idea, let me try.

acrrd: No failed. The only way is to trade lot's of small allocation with a couple of big one and use your idea on `Eq + Hash`

```rust
       struct UniqueArticle(Article);                                                                   
                                                                                                         
        impl Eq for UniqueArticle {}                                                                     
        impl PartialEq for UniqueArticle {                                                               
            fn eq(&self, other: &Self) -> bool {                                                         
                self.0.link == other.0.link || self.0.title == other.0.title                             
            }                                                                                            
        }                                                                                                
                                                                                                         
        impl std::hash::Hash for UniqueArticle {                                                         
            fn hash<H: std::hash::Hasher>(&self, state: &mut H) {                                        
                self.0.link.hash(state);                                                                 
                self.0.title.hash(state);                                                                
            }                                                                                            
        }                                                                                                
                                                                                                         
        let urls: HashSet<_> = history                                                                   
            .iter()                                                                                      
            .map(|doc| doc.url.as_str())                                                                 
            .chain(stack.iter().map(|doc| doc.resource.url.as_str()))                                    
            .collect();                                                                                  
                                                                                                         
        let titles: HashSet<_> = history                                                                 
            .iter()                                                                                      
            .map(|doc| &doc.title)                                                                       
            .chain(stack.iter().map(|doc| &doc.resource.title))                                          
            .collect();                                                                                  
                                                                                                         
        let mut articles: HashSet<_> = articles.into_iter().map(UniqueArticle).collect();                
                                                                                                         
        articles.retain(|article| {                                                                      
            !(urls.contains(&article.0.link.as_str())                                                    
                || titles.contains(&&article.0.title))                                                   
        });                                                                                              
                                                                                                         
        Ok(articles.into_iter().map(|u| u.0).collect())   
```

acrrd: Up to you guys decide. I'm fine with both.

rustonaut: I also had some other approaches but your approach seems to be the most reasonable one.

Through the way `Hash` is implemented lead it different hashes for equal types and through this inconsistencies. I switched it to a `BTreeSet` to avoid it.

discovery_engine_core/core/src/stack/filters.rs
```diff
@@ -30,22 +33,58 @@ pub(crate) trait ArticleFilter {
     ) -> Result<Vec<Article>, GenericError>;
 }
 
+struct UniqueArticle(Article);
```
acrrd, line 36: This only make sense for DuplicateArticle so I'll prefer to keep it somewhere related to that. This should be used by other filters.

rustonaut: I moved the definition into the function.

rustonaut:
:facepalm: I really should setup a check-fmt pre-commit hook message.

discovery_engine_core/core/src/lib.rs
```diff
@@ -27,7 +27,8 @@
 #![allow(
     clippy::missing_errors_doc,
     clippy::must_use_candidate,
-    clippy::module_name_repetitions
```
acrrd, line 30: This changes should be discussed with the team.
Not sure why allowing this `clippy::module_name_repetitions` sneaked in instead of discussing it with the others.

rustonaut: Because we had this discussion before as far as I remember, it's also why I requested a re-review.

rustonaut: Wait a moment something isn't right.

rustonaut: What was added was `clippy::items_after_statements` not `clippy::module_name_repetitions`.

rustonaut: In the `dart-api-dl` crate.

acrrd: `clippy::module_name_repetitions` was added recently after you asked me and I told you to discuss it with the team and this did not happen. You allow it on that struct but in general for project.
And for every change of the style is a team decision. For me is clear to have the struct declaration before the code. is also what often happen in the standard library code.

rustonaut: But this is about items in a function body, which if more then 1-3 lines makes the function supper unreadable as it splits up the function signature and the function body, or different parts of the function body.

Like IMHO having items before statements is an anti-patterns as long as the items aren't very very compact. like a simple drop guard.

rustonaut: Wrt. `clippy::module_name_repetitions` this seems to have accidentally slipt in I will start a discussion on slack.

rustonaut:
bors merge

rustonaut:
(Eventual changes to the linter will be made in a follow up PR as far as I understood.)

PR #221, opened 2022-03-01
TY-2502 update stacks errors

janpetschexain:
**References**

- [TY-2502]

**Summary**

- fail each `document_from_article()` conversion individually instead of all at once
- replace errors with traces in `update_stacks()`


discovery_engine_core/core/src/engine.rs
```diff
@@ -302,60 +298,66 @@ where
     /// Updates the stacks with data related to the top key phrases of the current data.
     ///
     /// Requires a threshold below which new items will be requested for a stack.
-    async fn update_stacks(
-        &mut self,
-        history: &[HistoricDocument],
-        request_new: usize,
-    ) -> Result<(), Error> {
+    #[instrument(skip(self))]
```
janpetschexain, line 301: this makes `just test` somewhat noisy again, should we also skip the other arguments of the method?

Robert-Steiner:
with this change, we can actually avoid the changes in #185. `get_feed_documents` can still fail if the beta sampling fails but that should not happen because we check that alpha/beta aren't `0`. I will adjust #185 after this pr.

discovery_engine_core/core/src/engine.rs
acrrd: We can avoid `instrument` completely. We will add this in another PR and see what we can do for the log from dart.

discovery_engine_core/core/src/engine.rs
```diff
@@ -302,60 +298,66 @@ where
...
+    async fn update_stacks(&mut self, history: &[HistoricDocument], request_new: usize) {
         let mut stacks = self.stacks.write().await;
-
-        let mut stacks: Vec<_> = stacks
+        let stacks = stacks
```
acrrd, line 304: it's more clear ihmo to have `:Vec<_>` here then having to go to look at the and of a multiline expression. It also reduce the changes. same below

discovery_engine/test/integration/request_feed_test.dart
```diff
@@ -63,20 +64,28 @@ void main() {
     });
 
     test(
-        'if requestNextFeedBatch fails due to a news api request error, requestFeed'
-        ' should return an empty list', () async {
+        'if requestNextFeedBatch traces a news api request error, then the'
+        ' depleted stacks have not been updated and subsequent calls should'
+        ' fail with FeedFailureReason.noNewsForMarket', () async {
```
acrrd, line 69: If the a trace is logged is not observable from the test so it should not be mentioned.
Returning an empty list is not an error.
If we want to return something we should signal an error iff we are completely unable to update the stacks.

acrrd (approved):

discovery_engine_core/core/src/engine.rs
janpetschexain: i'll revert it for the sake of fewer changed lines.

after all this is highly subjective, i for example find it much more readable if the `collect()` says what type is collected via the turbufish, otherwise i have to go back to the assignment to read the type after i have read the multiline operation.
also the type of the bound variable usually is displayed in the IDE via the rust-analyzer anyways. if we then write `: Vec<_>` we actually see less information, because the IDE doesn't give a type hint anymore, which would have also contained information about the `_` inferred type.

discovery_engine/test/integration/request_feed_test.dart
janpetschexain: i'll reformulate the trace part.

regarding the return value, i thought the whole purpose of the task was to make `update_stacks()` infallible, should this be reverted as well?

the test for the empty list was just a side effect of the thrown dart exception due to the rust error in `DiscoveryEngineFfi.getFeedDocuments()`, which hasn't been caught before. now the `FeedManager.nextFeedBatch()` can progress further due to the removed exception and properly checks for `feed.isEmpty` to return the corresponding event, this event is now checked in the test.

discovery_engine_core/core/src/engine.rs
acrrd: mmmm, I didn't know about this difference in the ide.

discovery_engine/test/integration/request_feed_test.dart
acrrd: the aim was to make it fail less, having it infallible is better. If `getFeedDocuments` returns an error is ok to return `NextBatchRequestFailed` but then it must contain the error we get from `getFeedDocument` otherwise we are loosing information.
If `getFeedDocument` returns an empty list that is not an error.
@Robert-Steiner @biernacki

janpetschexain: changed as discussed: `update_stacks()` is fallible again, but will only fail if all stacks fail, it succeeds if at least one stack succeeds.

@Robert-Steiner this also means that #185 is still necessary.

discovery_engine_core/core/src/engine.rs
```diff
@@ -324,35 +324,60 @@ where
...
+                        error
+                    })?;
+                    document_from_article(article, id, embedding).map_err(|error| {
+                        let error = Error::Document(error);
+                        error!("{}", error);
+                        error
+                    })
+                })
+                .partition_result::<Vec<_>, Vec<_>, _, _>();
+            if documents.is_empty() {
```
acrrd, line 366: This is not necessary an error, if `new_items` returns no articles or if the filter remove all of them we will have no documents and no errors.

discovery_engine_core/core/src/engine.rs
```diff
@@ -324,35 +324,60 @@ where
...
+                let error = Error::StackOpFailed(error);
+                error!("{}", error);
+                errors.push(error);
+            } else {
+                stack.data.retain_top(self.core_config.keep_top);
             }
         }
 
-        if errors.is_empty() {
+        if errors.len() < max_errors {
```
acrrd, line 380: Can't we inline `stack.len()` here. or assign `max_errors` before this if?

janpetschexain: the `stacks` is moved in l.328, that's why `max_errors` is already assigned in l.327.

instead of moving `stacks` we could also mutably iterate over it, what do you prefer?

acrrd: If is adding only `&mut` in front I would do it and keep the information near for the error. If it's more work this is fine.

discovery_engine/test/integration/request_feed_test.dart
janpetschexain: resolved merge conflicts, ready for rereview

acrrd (approved):

janpetschexain:
bors merge

PR #232, opened 2022-03-03
[Merged by Bors] - TY-2443...6 SearchManager

biernacki:
[new search](https://xainag.atlassian.net/browse/TY-2443)
[close search](https://xainag.atlassian.net/browse/TY-2444)
[restore search](https://xainag.atlassian.net/browse/TY-2445)
[search next batch](https://xainag.atlassian.net/browse/TY-2446)

PR adds a `SearchManager` class to handle all search releated events.

### DIFF
--- a/discovery_engine/lib/src/domain/event_handler.dart
+++ b/discovery_engine/lib/src/domain/event_handler.dart
@@ -68,7 +68,8 @@ import 'package:xayn_discovery_engine/src/infrastructure/box_name.dart'
         activeDocumentDataBox,
         changedDocumentIdBox,
         documentBox,
-        engineStateBox;
+        engineStateBox,
+        searchBox;
 import 'package:xayn_discovery_engine/src/infrastructure/repository/hive_active_document_repo.dart'
     show HiveActiveDocumentDataRepository;
 import 'package:xayn_discovery_engine/src/infrastructure/repository/hive_active_search_repo.dart';
@@ -294,7 +295,7 @@ class EventHandler {
       _openDbBox<ActiveDocumentData>(activeDocumentDataBox),
       _openDbBox<Uint8List>(changedDocumentIdBox),
       _openDbBox<Uint8List>(engineStateBox),
-      _openDbBox<ActiveSearch>(activeDocumentDataBox),
+      _openDbBox<ActiveSearch>(searchBox),
     ]);
   }
 
--- a/discovery_engine/lib/src/domain/event_handler.dart
+++ b/discovery_engine/lib/src/domain/event_handler.dart
@@ -44,11 +44,14 @@ import 'package:xayn_discovery_engine/src/domain/feed_manager.dart'
     show FeedManager;
 import 'package:xayn_discovery_engine/src/domain/models/active_data.dart'
     show ActiveDocumentData, ActiveDocumentDataAdapter;
-import 'package:xayn_discovery_engine/src/domain/models/active_search.dart';
+import 'package:xayn_discovery_engine/src/domain/models/active_search.dart'
+    show ActiveSearch, ActiveSearchAdapter;
 import 'package:xayn_discovery_engine/src/domain/models/configuration.dart'
     show Configuration;
 import 'package:xayn_discovery_engine/src/domain/models/document.dart'
     show Document, DocumentAdapter, UserReactionAdapter;
+import 'package:xayn_discovery_engine/src/domain/models/feed_market.dart'
+    show FeedMarketAdapter;
 import 'package:xayn_discovery_engine/src/domain/models/news_resource.dart'
     show NewsResourceAdapter;
 import 'package:xayn_discovery_engine/src/domain/models/view_mode.dart'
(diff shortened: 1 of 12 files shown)


discovery_engine/lib/src/domain/search_manager.dart
```diff
@@ -0,0 +1,217 @@
...
+        .where((doc) => doc.isSearched && doc.isActive)
+        .toList();
+
+    if (searchDocs.isEmpty) {
+      const reason = SearchFailureReason.noResultsAvailable;
+      return const EngineEvent.restoreSearchFailed(reason);
+    }
+
+    searchDocs.sort((doc1, doc2) {
+      final timeOrd = doc1.timestamp.compareTo(doc2.timestamp);
```
janpetschexain, line 165: this sorts older timestamps to the start and newer timestamps to the end, is this intended or do we want to have newer docs first? then we should revert the order.
```suggestion
      final timeOrd = doc2.timestamp.compareTo(doc1.timestamp);
```

discovery_engine/lib/src/domain/event_handler.dart
```diff
@@ -268,12 +286,16 @@ class EventHandler {
...
-    await _openDbBox<Document>(documentBox);
-    await _openDbBox<ActiveDocumentData>(activeDocumentDataBox);
-    await _openDbBox<Uint8List>(changedDocumentIdBox);
-    await _openDbBox<Uint8List>(engineStateBox);
+    await Future.wait([
+      _openDbBox<Document>(documentBox),
+      _openDbBox<ActiveDocumentData>(activeDocumentDataBox),
+      _openDbBox<Uint8List>(changedDocumentIdBox),
+      _openDbBox<Uint8List>(engineStateBox),
+      _openDbBox<ActiveSearch>(activeDocumentDataBox),
```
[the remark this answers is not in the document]
biernacki: great question! no it's not ok, it's my mistake :) I'll fix it, thanks!

discovery_engine/lib/src/domain/search_manager.dart
biernacki: good question, but I don't know the answer, maybe @acrrd can shed some light on it?

discovery_engine/lib/src/domain/search_manager.dart
```diff
@@ -0,0 +1,217 @@
...
+    if (search == null) {
+      const reason = SearchFailureReason.noActiveSearch;
+      return const EngineEvent.nextSearchBatchRequestFailed(reason);
+    }
+
+    // lets update active search params
+    search = search.copyWith(requestedPageNb: search.requestedPageNb + 1);
+    final docs = await _getSearchDocuments(search);
+
+    if (docs.isEmpty) {
```
acrrd, line 131: `noResultsAvailable` is not necessary true, we can have a network error or some other kind of error that we cannot necessary observe here. At the moment I would be more to return an empty list.

discovery_engine/lib/src/domain/search_manager.dart
```diff
@@ -0,0 +1,217 @@
...
+      const reason = SearchFailureReason.noResultsAvailable;
+      return const EngineEvent.nextSearchBatchRequestFailed(reason);
+    }
+
+    // TODO: should we persist updated search params before we get documents?
+    await _searchRepo.save(search);
```
acrrd, line 137: Yes, if we get an error back from rust we should try the same page again.

discovery_engine/lib/src/domain/search_manager.dart
acrrd: older first because we want to return them in the same order user saw them the first time. This should be the same as we are doing for the feed

discovery_engine/lib/src/domain/search_manager.dart
```diff
@@ -0,0 +1,204 @@
...
+
+    return searchDocs
+        .map((docWithData) => docWithData.document.toApiDocument())
+        .toList();
+  }
+
+  /// Obtain the first batch of search documents and persist to repositories.
+  Future<EngineEvent> searchRequested(
+    String queryTerm,
+    FeedMarket market,
```
acrrd, line 98: We will use the market that we already have so no need to take it here.

discovery_engine/lib/src/domain/search_manager.dart
biernacki: so then we should persist the search object after we get documents from rust, so we (the app) can send the same event again, and expect same results (especially for getting next page, the incremented page nb should be updated before we ask engine for next page but saved after we received the next page documents)

discovery_engine/lib/src/domain/event_handler.dart
```diff
@@ -268,12 +291,17 @@ class EventHandler {
     Hive.registerAdapter(DurationAdapter());
     Hive.registerAdapter(UriAdapter());
     Hive.registerAdapter(EmbeddingAdapter());
+    Hive.registerAdapter(FeedMarketAdapter());
```
biernacki, line 294: ❓we don't perists `FeedMarket` anymore, should I revert changes to this class so that Hive type adapters are not generated and we don't register it, or should we leave it, cause it might be usefull in the future and it doesn't break anything now?

discovery_engine/lib/src/domain/search_manager.dart
biernacki: I've changed that for now, so we persist search before we call the engine, but maybe I should revert it to the prev implementation, where we persisted after calling the engine?

janpetschexain: it looks like persisting it after the call to the engine would be better. currently if `_getSearchDocuments()` fails and we enter `nextSearchBatchRequested()` again, the counter `search.requestedPageNb` is already incremented but we didn't get the docs of the previous page. this could happen multiple times, which leaves us at later pages with worse results than necessary.

discovery_engine/lib/src/domain/event_handler.dart
janpetschexain: i guess we can leave it for now. if it's not needed in the future, it could be reverted in a follow up pr.

janpetschexain (approved):

biernacki:
bors merge

PR #233, opened 2022-03-03
[Merged by Bors] - TY-2527 parallelize api request

Robert-Steiner:
Ticket:

- [TY-2527]

Summary:

- performs the api requests for each market in parallel 

[TY-2527]: https://xainag.atlassian.net/browse/TY-2527?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine_core/core/Cargo.toml
+++ b/discovery_engine_core/core/Cargo.toml
@@ -12,7 +12,7 @@ derivative = "2.2.0"
 derive_more = { version = "0.99.17", default-features = false, features = ["display", "from"] }
 displaydoc = "0.2.3"
 figment = { version = "0.10.6", default-features = false, features = ["json"] }
-futures = "0.3.21"
+futures = { version = "0.3.21", default-features = false, features = ["alloc"] }
 rand = "0.8.5"
 rand_distr = "0.4.3"
 rayon = "1.5.1"
--- a/discovery_engine_core/core/Cargo.toml
+++ b/discovery_engine_core/core/Cargo.toml
@@ -12,6 +12,7 @@ derivative = "2.2.0"
 derive_more = { version = "0.99.17", default-features = false, features = ["display", "from"] }
 displaydoc = "0.2.3"
 figment = { version = "0.10.6", default-features = false, features = ["json"] }
+futures = "0.3.21"
 rand = "0.8.5"
 rand_distr = "0.4.3"
 rayon = "1.5.1"
--- a/discovery_engine_core/core/src/stack/ops/breaking.rs
+++ b/discovery_engine_core/core/src/stack/ops/breaking.rs
@@ -64,12 +64,13 @@ impl Ops for BreakingNews {
             let mut articles = Vec::new();
             let mut errors = Vec::new();
 
-            let mut requests = FuturesUnordered::new();
-            for market in markets.read().await.iter() {
-                let request =
-                    spawn_headlines_request(self.client.clone(), market.clone(), self.page_size);
-                requests.push(request);
-            }
+            let mut requests = markets
+                .read()
+                .await
+                .iter()
+                .cloned()
+                .map(|market| spawn_headlines_request(self.client.clone(), market, self.page_size))
+                .collect::<FuturesUnordered<_>>();
 
             while let Some(handle) = requests.next().await {
                 // should we also push handle errors?
--- a/discovery_engine_core/core/src/stack/ops/breaking.rs
+++ b/discovery_engine_core/core/src/stack/ops/breaking.rs
@@ -15,7 +15,8 @@
 use std::sync::Arc;
 
 use async_trait::async_trait;
-use tokio::sync::RwLock;
+use futures::{stream::FuturesUnordered, StreamExt};
+use tokio::{sync::RwLock, task::JoinHandle};
 use uuid::Uuid;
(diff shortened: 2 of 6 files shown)


discovery_engine_core/core/src/stack/ops/personalized.rs
```diff
@@ -66,18 +67,41 @@ impl Ops for PersonalizedNews {
...
-                    market,
-                    filter,
-                    page_size: self.page_size,
-                    page: None,
-                };
-                match self.client.news(&query).await {
-                    Ok(batch) => articles.extend(batch),
-                    Err(err) => errors.push(err),
+                // a lot of cloning
+                // if NewsQuery can own the market + filter we can create the
```
acrrd, line 74: I'm not sure having `NewsQuery` taking ownership will solve the problem. We still need to clone the market and the filter in that. 

The `Client` is stateless so we can put it in an `Arc` and avoid to clone it. But for the rest is not a problem we have max 3 markets.

acrrd: We can change `NewsQuery` to take `Deref<Target = Filter>` instead so we can avoid to clone it but seems more a micro optimization at the moment.

Robert-Steiner: > I'm not sure having NewsQuery taking ownership will solve the problem. We still need to clone the market and the filter in that.

Yes, we unfortunately need to clone markets and filters because tokio::spawn requires the future to be `'static` but if NewsQuery has the ownership of `market`, `filter` we can save some lines like 
```rust
let market_inner = market;
let filter_inner = filter;
```

What do you mean by `Deref<Target = Filter>`?

I will put the `Client` behind an `Arc` and move the creation of the request in a function. that should clean up some stuff.

discovery_engine_core/core/Cargo.toml
```diff
@@ -12,6 +12,7 @@ derivative = "2.2.0"
 derive_more = { version = "0.99.17", default-features = false, features = ["display", "from"] }
 displaydoc = "0.2.3"
 figment = { version = "0.10.6", default-features = false, features = ["json"] }
+futures = "0.3.21"
```
janpetschexain, line 15: we should pick the features, otherwise we will get an executor etc
```suggestion
futures = { version = "0.3.21", default-features = false, features = ["alloc"] }
```

discovery_engine_core/core/src/stack/ops/breaking.rs
```diff
@@ -60,16 +64,23 @@ impl Ops for BreakingNews {
...
-                    market,
-                    page_size: self.page_size,
-                };
-                match self.client.headlines(&query).await {
-                    Ok(batch) => articles.extend(batch),
-                    Err(err) => errors.push(err),
+                let request =
+                    spawn_headlines_request(self.client.clone(), market.clone(), self.page_size);
+                requests.push(request);
+            }
```
janpetschexain, line 72: the `requests` could be collected directly in a more functional style (i unfortunately can't mark it as a suggestion as the diff contains deleted lines). similarly for the personalized news query.
```rust
            let mut requests = markets
                .read()
                .await
                .iter()
                .cloned()
                .map(|market| spawn_headlines_request(self.client.clone(), market, self.page_size))
                .collect::<FuturesUnordered<_>>();
```

discovery_engine_core/core/src/stack/ops/breaking.rs
```diff
@@ -95,3 +106,18 @@ impl Ops for BreakingNews {
         Ok(res)
     }
 }
+
+fn spawn_headlines_request(
+    client: Arc<Client>,
+    market: Market,
```
janpetschexain, line 112: should we wrap each `market` in an `Arc`? otherwise this might become costly to clone since we have to do this for each query.

discovery_engine_core/core/src/stack/ops/personalized.rs
acrrd: > Yes, we unfortunately need to clone markets and filters because tokio::spawn requires the future to be `'static` but if NewsQuery has the ownership of `market`, `filter` we can save some lines like
> 
> ```rust
> let market_inner = market;
> let filter_inner = filter;
> ```

Maybe we can save some line but the clone will be there anyway no?

> What do you mean by `Deref<Target = Filter>`?

If we have something like
```rust
NewsQuery<'a, F> {
  market: &'a,
  filter: F
}

async fn headlines(q: NewsQuery<'_, impl Deref<Target = Filter>) { ... }
```

Then we can put the `Filter` in an `Arc` and construct the `NewsQuery` with that.

discovery_engine_core/core/src/stack/ops/breaking.rs
acrrd: We don't do query that often and market is 2 strings, with two char inside. the alloc will cost, but this is nowhere near a hot path.

discovery_engine_core/core/src/stack/ops/personalized.rs
Robert-Steiner: implemented it for the filter

janpetschexain (approved):
lgtm

PR #236, opened 2022-03-03
[Merged by Bors] - Don't log error that is not an error

acrrd:
Seeing this error in the console is a bit confusing. Nothing went really wrong, is normal that the file is not there, we rename it if the verification pass. If the file does not exists we go forward and download it.


janpetschexain (approved):

acrrd:
bors merge

PR #238, opened 2022-03-03
[Merged by Bors] - Remove invalid char from key phrase

acrrd:
`"` can interfere with the quoting that we do while constructing the filter, because ofthis we got some 406 from the endpoint.

### DIFF
--- a/discovery_engine_core/providers/src/filter.rs
+++ b/discovery_engine_core/providers/src/filter.rs
@@ -24,8 +24,9 @@ pub struct Filter {
 
 impl Filter {
     /// Add a keyword to filter with. All keyword are in "or" with each other.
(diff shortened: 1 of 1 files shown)


rustonaut (approved):
lgtm

maybe `"\\\""` or similar would also have worked, but then I don't know why 
we would care about `"` in a keyword tbh.

acrrd:
> lgtm
> 
> maybe `"\\\""` or similar would also have worked, but then I don't know why we would care about `"` in a keyword tbh.

I wasn't sure if the encoder work with escapes and it was already escaped is not enough to just add that but is a bit more complex to handle it correctly. Now I'll check if \ alone will create a problem with other chars.

acrrd:
bors merge

PR #239, opened 2022-03-03
[Merged by Bors] - Always update when positive reaction

acrrd:
We want to always update the stacks when the user react positively.


rustonaut (approved):
lgtm

acrrd:
bors merge

PR #223, opened 2022-03-01
[Merged by Bors] - TY-2434 add active search method to engine

finiteprods:
previously #200 

**Summary**

adds a method `active_search` to `Engine` that issues a news query and returns the `Document`s in the response.

PR #240, opened 2022-03-04
[Merged by Bors] - TY-2524 fix potential 32bit problems

rustonaut:
Does remove all usage of `usize`/`isize` in FFI.

Neither dart nor ffigen support that well when cross-compilation to 32bit targets is involved.

Once we have dart 1.16 we can replace this with a much better solutions using
`AbiSpecificInteger` and custom type-mapping options for `ffigen`.

**References:**

- [TY-2524](https://xainag.atlassian.net/browse/TY-2524)

PR #242, opened 2022-03-04
[Merged by Bors] - TY-2436 active search ffi

finiteprods:
see #223 for implementation of active search in engine.

**Summary**

exposes active search method via FFI.

### DIFF
--- a/discovery_engine/lib/src/domain/engine/mock_engine.dart
+++ b/discovery_engine/lib/src/domain/engine/mock_engine.dart
@@ -123,7 +123,7 @@ class MockEngine implements Engine {
     } else {
       return [
         DocumentWithActiveData(doc0, active0),
-        DocumentWithActiveData(doc1, active1)
+        DocumentWithActiveData(doc1, active1),
       ];
     }
   }
@@ -139,5 +139,19 @@ class MockEngine implements Engine {
     UserReacted userReacted,
   ) async {
     _incrementCount('userReacted');
+  }
+
+  @override
+  Future<List<DocumentWithActiveData>> activeSearch(
+    String query,
+    int page,
+    int pageSize,
+  ) async {
+    _incrementCount('activeSearch');
+
+    return [
+      DocumentWithActiveData(doc0, active0),
+      DocumentWithActiveData(doc1, active1),
+    ];
   }
 }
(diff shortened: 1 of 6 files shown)


discovery_engine_core/bindings/src/lib.rs
```diff
@@ -143,4 +143,22 @@ impl XaynDiscoveryEngineAsyncFfi {
                 .map_err(|error| error.to_string()),
         )
     }
+
+    /// Perform an active search with the given query parameters.
+    pub async fn active_search(
+        engine: &SharedEngine,
+        query: Box<String>,
+        page: usize,
+        page_size: usize,
```
janpetschexain, line 152: related to #240, we probably have to use fixed sized integers as well.

discovery_engine/lib/src/domain/engine/mock_engine.dart
```diff
@@ -140,4 +140,18 @@ class MockEngine implements Engine {
...
+  Future<List<DocumentWithActiveData>> activeSearch(
+    String query,
+    int page,
+    int pageSize,
+  ) async {
+    _incrementCount('activeSearch');
+
+    return [
+      DocumentWithActiveData(doc0, active0),
+      DocumentWithActiveData(doc1, active1),
```
janpetschexain, line 154: on the rust side we use the nil stack id for the documents of an active search, we probably have to do the same here? this would mean that we shouldn't reuse `doc0` and `doc1` since they have a non-nil stack id.

finiteprods: ah yes thanks, this totally escaped me..

finiteprods: to keep it simple it just returns empty for now.
I think returning some documents with nil id could be done separately to this pr.

rustonaut (approved):
lgtm,

I will add range checks for `page`/`pageSize` in the 32bit fixing PR,
as this PR has a helper function for that already.

finiteprods:
> I will add range checks for `page`/`pageSize` in the 32bit fixing PR, as this PR has a helper function for that already.

great, thanks for that.

finiteprods:
bors merge

PR #249, opened 2022-03-11
[Merged by Bors] - TY-2459-60-61 Handle excluded sources events

biernacki:
[handle add event](https://xainag.atlassian.net/browse/TY-2459)
[handle remove event](https://xainag.atlassian.net/browse/TY-2460)
[handle list event](https://xainag.atlassian.net/browse/TY-2461)

Added methods in the FeedManager to handle events related to excluded sources + added tests for these new methods.

### DIFF
--- a/discovery_engine/lib/src/api/events/client_events.dart
+++ b/discovery_engine/lib/src/api/events/client_events.dart
@@ -97,12 +97,14 @@ class ClientEvent with _$ClientEvent {
 
   /// Event created when a source is added to the list of excluded sources.
   @Implements<FeedClientEvent>()
-  const factory ClientEvent.excludedSourceAdded(Uri source) =
+  @Assert('source != ""')
+  const factory ClientEvent.excludedSourceAdded(String source) =
       ExcludedSourceAdded;
 
   /// Event created when a source is removed from the list of excluded sources.
   @Implements<FeedClientEvent>()
-  const factory ClientEvent.excludedSourceRemoved(Uri source) =
+  @Assert('source != ""')
+  const factory ClientEvent.excludedSourceRemoved(String source) =
       ExcludedSourceRemoved;
 
   /// Event created when a client requests for list of excluded sources.
--- a/discovery_engine/lib/src/domain/feed_manager.dart
+++ b/discovery_engine/lib/src/domain/feed_manager.dart
@@ -15,7 +15,7 @@
 import 'package:xayn_discovery_engine/src/api/events/client_events.dart'
     show FeedClientEvent;
 import 'package:xayn_discovery_engine/src/api/events/engine_events.dart'
-    show EngineEvent, EngineExceptionReason, FeedFailureReason;
+    show EngineEvent, FeedFailureReason;
 import 'package:xayn_discovery_engine/src/domain/engine/engine.dart'
     show Engine;
 import 'package:xayn_discovery_engine/src/domain/event_handler.dart'
--- a/discovery_engine/lib/src/domain/feed_manager.dart
+++ b/discovery_engine/lib/src/domain/feed_manager.dart
@@ -141,13 +141,11 @@ class FeedManager {
 
   /// Adds a source to excluded sources set.
   Future<EngineEvent> addExcludedSource(String source) async {
-    final uri = _sanitizeSource(source);
-    if (uri == null) {
-      const reason = EngineExceptionReason.genericError;
-      return const EngineEvent.engineExceptionRaised(reason);
+    if (source.isEmpty) {
+      throw ArgumentError('source "$source" can\'t be empty');
     }
     final sources = await _excludedSourcesRepository.getAll();
-    sources.add(uri);
+    sources.add(source);
     await _excludedSourcesRepository.save(sources);
     // TODO: send updated sources to the engine
     return const EngineEvent.clientEventSucceeded();
@@ -155,13 +153,11 @@ class FeedManager {
 
   /// Removes a source to excluded sources set.
   Future<EngineEvent> removeExcludedSource(String source) async {
-    final uri = _sanitizeSource(source);
-    if (uri == null) {
-      const reason = EngineExceptionReason.genericError;
-      return const EngineEvent.engineExceptionRaised(reason);
+    if (source.isEmpty) {
+      throw ArgumentError('source "$source" can\'t be empty');
     }
     final sources = await _excludedSourcesRepository.getAll();
-    sources.remove(uri);
+    sources.remove(source);
     await _excludedSourcesRepository.save(sources);
     // TODO: send updated sources to the engine
     return const EngineEvent.clientEventSucceeded();
@@ -171,11 +167,5 @@ class FeedManager {
   Future<EngineEvent> getExcludedSourcesList() async {
     final sources = await _excludedSourcesRepository.getAll();
     return EngineEvent.excludedSourcesListRequestSucceeded(sources);
-  }
-
-  Uri? _sanitizeSource(String source) {
-    final uri = Uri.tryParse(source);
-    if (uri == null) return uri;
-    return Uri(host: uri.host);
   }
 }
--- a/discovery_engine/lib/src/domain/feed_manager.dart
+++ b/discovery_engine/lib/src/domain/feed_manager.dart
@@ -32,6 +32,8 @@ import 'package:xayn_discovery_engine/src/domain/repository/document_repo.dart'
     show DocumentRepository;
 import 'package:xayn_discovery_engine/src/domain/repository/engine_state_repo.dart'
     show EngineStateRepository;
+import 'package:xayn_discovery_engine/src/domain/repository/excluded_sources_repo.dart'
+    show ExcludedSourcesRepository;
 
 /// Business logic concerning the management of the feed.
 class FeedManager {
@@ -41,6 +43,7 @@ class FeedManager {
   final ActiveDocumentDataRepository _activeRepo;
   final ChangedDocumentRepository _changedRepo;
   final EngineStateRepository _engineStateRepo;
+  final ExcludedSourcesRepository _excludedSourcesRepository;
 
   FeedManager(
     this._engine,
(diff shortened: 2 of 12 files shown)


discovery_engine/test/feed_manager_test.dart
```diff
@@ -196,4 +199,67 @@ Future<void> main() async {
...
+
+      expect(response, isA<ClientEventSucceeded>());
+      expect(excludedBox.values.first, equals({Uri(host: 'www.nytimes.com')}));
+    });
+
+    test('getExcludedSourcesList', () async {
+      final excludedSoures = {
+        Uri(host: 'theguardian.com'),
+        Uri(host: 'bbc.co.uk'),
+        Uri(host: 'wsj.com'),
```
Robert-Steiner, line 251: Will it be a problem that `wsj.com` and `www.wsj.com` lead to two entries?

Robert-Steiner (approved):
Looks good 👍
Just one question about the URLs, but this has probably already been discussed.

discovery_engine/test/feed_manager_test.dart
frankpepermans: Will it match subdomains?
Eg www.bbc.co.uk is excluded
The article is then sport.bbc.co.uk ?

discovery_engine/lib/src/api/events/client_events.dart
```diff
@@ -95,6 +95,21 @@ class ClientEvent with _$ClientEvent {
     Set<DocumentId> documentIds,
   ) = FeedDocumentsClosed;
 
+  /// Event created when a source is added to the list of excluded sources.
+  @Implements<FeedClientEvent>()
+  const factory ClientEvent.excludedSourceAdded(Uri source) =
```
acrrd, line 100: As specified in the story the only values that are allowed here are value present in the field `source_domain`. That is not a valid `Uri` so we cannot use this type here.

discovery_engine/test/feed_manager_test.dart
acrrd: It depends on NewCatcher. This will be an exact match on the source value. If they return articles with `www.example.com` and `example.com` then they will treated as different sources

docs/events.md
```diff
@@ -407,7 +407,7 @@ class ContentCategoriesAccepted extends ClientEvent {
 
 ### ClientEventSucceeded
 
-Event created to inform the client that a particular "fire and forget" event, like ie. `UserReactionChanged`, was successfuly processed by the engine.
+Event created to inform the client that a particular "fire and forget" event, like ie. `UserReactionChanged`, was successfully processed by the engine.
```
acrrd, line 410: I didn't notice this line before, but this kind of events are arguably "fire and forget". If `UserReactionChanged` fails we should not show it as green. Similarly for other events.

discovery_engine/lib/src/api/events/client_events.dart
biernacki: I've changed that to `String`

docs/events.md
biernacki: I've updated docs a bit and removed the "fire and forget" bit.

discovery_engine/test/feed_manager_test.dart
biernacki: the `www` subdomain is a "special" case, so we could consider to threat it as equal to apex domain, for example if we find that NewsCatcher sometimes returns articles from the same source with `www` and other times without it. For now I would leave it separate cases.

discovery_engine/lib/src/api/events/client_events.dart
rustonaut: Shouldn't it also be turned into a string in the other places, mainly the repository?

And if we treat it as an opaque value returned by news catcher then shouldn't
we also not sanitize it, because it's already in exact the format Newscatcher expects?

biernacki: this is only our assumption that the value sent by the app will be a `source_domain` from news catcher, so I think it's better to sanitize it and/or check if it's vaild. We could also change the value in the events to `Document` or `DocumentId` so then we know that we can get a proper `source_domain` and not a random `String`.

rustonaut: The thing is a domain is not a `Uri`.

The only reason this works is because dart `Uri`s are not necessary Uris :sob: 

And as far as I understood `source_domain`'s must be _exactly_ the value from the `source_domain` field as we got from Newscatcher. Not the domain of the link.

"sanitizing" something which must not be changed isn't a good idea IMHO.

Also we can't use `DocumentId` because the excluded `source_domain` values might not be associated with any document we currently do have stored anywhere (neither in dart nor rust). 

So the only alternative would be to "newtype" `source_domain`.  Maybe in a follow up PR. But I'm not sure it's necessary.

discovery_engine/test/feed_manager_test.dart
acrrd: Why special? It has the same dignity of `sport`. We cannot control this data is all up to the provider at this point.

discovery_engine/lib/src/api/events/client_events.dart
acrrd: One simple way to avoid wrong data would be to check when we receive an `AddExcludedSource` that the value is a value of the field `source_domain` in the database. We can do a quick test with 10k documents and see how long it takes. If it is too much we can restrict that to only active documents (that kinda make sense anyway)

discovery_engine/lib/src/domain/feed_manager.dart
```diff
@@ -131,4 +138,34 @@ class FeedManager {
 
     return const EngineEvent.clientEventSucceeded();
   }
+
+  /// Adds a source to excluded sources set.
+  Future<EngineEvent> addExcludedSource(String source) async {
+    if (source.isEmpty) {
+      throw ArgumentError('source "$source" can\'t be empty');
```
Robert-Steiner, line 145:
```suggestion
      throw ArgumentError('source can\'t be empty');
```
since `source` is empty we can remove `$source` here

discovery_engine/lib/src/api/events/client_events.dart
biernacki: @acrrd I've added that check, I think it should be fine, cause Hive loads all the documents into memory anyways. Let me know if the implementation is ok.

rustonaut: > value is a value of the field source_domain in the database

But this will still couple the source to documents, which means when we e.g. prune the history a excluded source can sneak back in but IMHO should not. It's also the kind of error we likely won't catch with testing and will have problem tracing down when getting bug reports.

rustonaut: > value of the field source_domain in the database

Through this still couples the source to a document, which IMHO we should avoid.

biernacki: > > value of the field source_domain in the database
> 
> Through this still couples the source to a document, which IMHO we should avoid.

ok, so what if we manage a map of sources and their exclusion state via `Map<String, bool>` (I can adapt existing repo to do that)? We can add new sources this map every time we get new documents using their `sourceDomain`s, we can include/exclude a source by toggling the flag, and we can check if a source is valid if it's present as a key in this map? This won't be coupled to the document, but it will still give us some assurance regarding validity of the sources send from the app.

```dart
final excludedSources = {
  'example.com': true, // excluded
  'bbc.co.uk': false, // included
};

if (excludedSources['theguardian.com'] == null) {
  // invalid source
}
```

discovery_engine/test/feed_manager_test.dart
biernacki: it's special in a sence that `www` and non-`www` domains are most of the time, for most websites, connected with each other through 301 redirects, which is not the case for other subdomains like `sport`. But like I said before, we can rely on the data from newscatcher, and this could be easily changed later on.

biernacki: @acrrd I don't see any actionable items here for the PR, is it ok to resolve this conversation?

frankpepermans: NC indeed always presents urls before any redirects, it actually makes sense indeed to just always drop www.

Eg, they crawl https://bbc.co.uk, and then find a relative link there which is /sport

Then, it might be that the crawled one actually becomes https://www.bbc.co.uk/sport after resolving it to an absolute link.

If you always drop www, then this is no issue

frankpepermans: ...and then https://www.bbc.co.uk/sport redirects maybe to sport.bbc.co.uk, but that then doesn't matter

biernacki: I think it might depend how this set of sources is used. If we use send it with requests to the API so they are doing filtering on their side, or if we use it to filter out documents we got back. In the first case we probably shouldn't change anything and rely on data received from newscatcher, in the second case we might want to always drop the `www`.

acrrd: The only valid values are the one in `source_domain` we cannot do anything with them.

discovery_engine/lib/src/domain/feed_manager.dart
```diff
@@ -131,4 +138,47 @@ class FeedManager {
...
+  /// Adds a source to excluded sources set.
+  Future<EngineEvent> addExcludedSource(String source) async {
+    if (source.isEmpty) {
+      throw ArgumentError('source can\'t be empty');
+    }
+
+    final allDocuments = await _docRepo.fetchAll();
+    final doesExist = allDocuments
+        .where((doc) => doc.isActive)
+        .map((doc) => doc.resource.sourceDomain)
```
acrrd, line 151: We can use `any` instead so we don't have to collect all the values.

discovery_engine/lib/src/domain/feed_manager.dart
```diff
@@ -131,4 +138,47 @@ class FeedManager {
...
+        .contains(source);
+
+    if (!doesExist) {
+      throw ArgumentError('source $source not found in database');
+    }
+
+    final sources = await _excludedSourcesRepository.getAll();
+    sources.add(source);
+    await _excludedSourcesRepository.save(sources);
+    // TODO: send updated sources to the engine
```
acrrd, line 162: Not sure about this.

acrrd (approved):

biernacki:
bors merge

PR #250, opened 2022-03-11
[Merged by Bors] - TY-2467 determine semantic clusters

Robert-Steiner:
Ticket: 

- [TY-2467]

Summary:

Implements the [`determine_semantic_clusters`](https://github.com/xaynetwork/xayn_ai_research/blob/eb7f1d0f25ca42fa512663eb41b62438f805cd0f/experiments/semantic_grouping/utils.py#L172-L191) method for the cosine metric

[TY-2467]: https://xainag.atlassian.net/browse/TY-2467?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


discovery_engine_core/core/src/stack/semantic_filters.rs
```diff
@@ -0,0 +1,158 @@
...
+    #[test]
+    fn test_cluster_empty_documents() {
+        let labels = determine_semantic_clusters(&[], Method::Average, 1.);
+        assert!(labels.is_empty());
+    }
+
+    #[test]
+    fn test_cluster_single_document() {
+        let labels = determine_semantic_clusters(&[Document::default()], Method::Average, 1.);
+        assert_eq!(labels, [0]);
```
Robert-Steiner, line 110: Here we deviate slightly from the `sklearn` implementation. `sklearn` always requires at least two samples but I think that shouldn't be a problem for us.

Robert-Steiner (commented):
@janpetschexain thanks for completing the task 👍

discovery_engine_core/core/src/stack/semantic_filters.rs
janpetschexain: i think so too. for sklearn one would need additional error handling for the case < 2, here we just baked that into the function with a trivial extension.

janpetschexain (approved):
approved on behalf of @Robert-Steiner

janpetschexain:
bors merge

PR #251, opened 2022-03-15
[Merged by Bors] - TY-2539 add to rank feature to newscatcher client

rustonaut:
**Based on:**

- #244 

**References:**

- [story TY-2528](https://xainag.atlassian.net/browse/TY-2528)
- [ticket TY-2539](https://xainag.atlassian.net/browse/TY-2539)

PR #252, opened 2022-03-15
[Merged by Bors] - TY-2540 default to hard coded market specific news quality rank limits

rustonaut:
**Based on:**

- #251
- #244 

**References:**

- [ticket TY-2540](https://xainag.atlassian.net/browse/TY-2540)

### DIFF
--- a/discovery_engine_core/bindings/src/types/market.rs
+++ b/discovery_engine_core/bindings/src/types/market.rs
@@ -40,17 +40,26 @@ pub unsafe extern "C" fn market_place_of_lang_code(place: *mut Market) -> *mut S
     unsafe { addr_of_mut!((*place).lang_code) }
 }
 
-/// Sets the `news_quality_rank_limit` field of a [`Market`] to `None`.
+/// Finish the initialization of a [`Market`] instance.
+///
+/// This sets defaults for fields not yet exposed to dart.
+///
 ///
 /// # Safety
 ///
-/// The pointer must point to a valid [`Market`] memory object, it
-/// might be uninitialized.
+/// - The pointer must point to a valid [`Market`] memory object, it
+///   might be uninitialized.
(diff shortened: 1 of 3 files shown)


discovery_engine_core/bindings/src/types/market.rs
```diff
@@ -40,17 +40,26 @@ pub unsafe extern "C" fn market_place_of_lang_code(place: *mut Market) -> *mut S
...
+/// - Must be called after all exposed fields have been initialized and
+///   not before that.
 #[no_mangle]
-pub unsafe extern "C" fn init_market_news_quality_rank_limit(place: *mut Market) {
+pub unsafe extern "C" fn finish_market_initialization(place: *mut Market) {
     unsafe {
-        addr_of_mut!((*place).news_quality_rank_limit).write(None);
-    }
+        //Note: I'm not sure if `&*place.country_code` is guaranteed to not construct a
+        //      intermediate `&place` (which would be unsound).
```
janpetschexain, line 58: from the [MIR of this playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d66e599a356baafe0c345c1240fddcce) it doesn't seem that an intermediate reference is created.

on the other hand it might be better to be on the safe side and use the address macro regardless.

janpetschexain (approved):

acrrd:
We can merge it so we have it but all the parts from binding to dart must be removed. If we want to keep `Market` like this that we need a `Market` to use as DTO between rust/dart and app

rustonaut:
Why?

We are not exposing any implementation details to dart.

The only change on the dart side is this line: `ffi.finish_market_initialization(place);`

Which is completely fine as far as I can tell.

rustonaut:
bors merge

PR #253, opened 2022-03-15
[Merged by Bors] - TY-2591 Fix for restoring search

biernacki:
[ref](https://xainag.atlassian.net/browse/TY-2591)

When getting documents from active search we need to make sure that we save them as "searched" documents, by setting the `isSearched` flag to true.

### DIFF
--- a/discovery_engine/lib/src/ffi/types/document/document_vec.dart
+++ b/discovery_engine/lib/src/ffi/types/document/document_vec.dart
@@ -70,13 +70,13 @@ extension DocumentSliceFfi on List<DocumentFfi> {
   }
 
   List<DocumentWithActiveData> toDocumentListWithActiveData({
-    bool searched = false,
+    bool isSearched = false,
   }) =>
       asMap()
           .entries
           .map(
             (e) => DocumentWithActiveData(
-              e.value.toDocument(batchIndex: e.key, isSearched: searched),
+              e.value.toDocument(batchIndex: e.key, isSearched: isSearched),
               e.value.toActiveDocumentData(),
             ),
           )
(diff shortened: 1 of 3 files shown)


discovery_engine/lib/src/domain/search_manager.dart
```diff
@@ -80,7 +80,9 @@ class SearchManager {
     );
     await _engineStateRepo.save(await _engine.serialize());
 
-    await _docRepo.updateMany(searchDocs.map((e) => e.document));
+    final documents = searchDocs.map((e) => e.document..isSearched = true);
```
x3ro, line 83: Interesting use of the cascade operator. I personally would split this into multiple lines, for readability sake, but it's not a must.

x3ro (approved):
LGTM 🚀

biernacki:
@x3ro I've changed the PR so that documents are marked as search already when thay are comming from the `engine.activeSearch` method, which I think is more correct, then trying to "fix" them in a manager. Please check again :)

discovery_engine/lib/src/ffi/types/document/document_vec.dart
```diff
@@ -69,13 +69,16 @@ extension DocumentSliceFfi on List<DocumentFfi> {
...
-      .entries
-      .map(
-        (e) => DocumentWithActiveData(
-          e.value.toDocument(batchIndex: e.key),
-          e.value.toActiveDocumentData(),
-        ),
-      )
-      .toList();
+  List<DocumentWithActiveData> toDocumentListWithActiveData({
+    bool searched = false,
```
x3ro, line 73: 🟢 `searched` could be `isSearched` here, to be consistent with the field name.

x3ro (approved):

biernacki:
bors merge

PR #254, opened 2022-03-15
TY-2466 determine date clusters

janpetschexain:
**References**

- [TY-2466]
- requires #250

**Summary**

- implements the [`determine_date_clusters`](https://github.com/xaynetwork/xayn_ai_research/blob/eb7f1d0f25ca42fa512663eb41b62438f805cd0f/experiments/semantic_grouping/utils.py#L278) method


discovery_engine_core/core/src/stack/semantic_filters.rs
```diff
@@ -52,6 +53,94 @@ fn condensed_cosine_distance(documents: &[Document]) -> Vec<f32> {
...
+        BTreeMap::<_, Vec<_>>::new(),
+        |mut clusters, (idx, &label)| {
+            clusters.entry(label).or_default().push(idx);
+            clusters
+        },
+    );
+
+    let (_, labels) = clusters.into_values().fold(
+        (0, vec![0; labels.len()]),
+        |(label, mut labels), cluster| {
```
Robert-Steiner, line 74:
```suggestion
        |(cluster_id, mut labels), cluster| {
```
🟢 not sure if `cluster_id` is a good alternative but the name `label` confused me a bit

Robert-Steiner (approved):
Nice work!👍

discovery_engine_core/core/src/stack/semantic_filters.rs
janpetschexain: i agree that it's quite hard to name. it's a counter and a label at the same time, and depending on if the next cluster has one or more samples it's more like an offset to a counter/label.

after all it's a label/sequence of labels assigned to the documents which are indexed by the cluster of the current iteration, so i guess label is a bit more fitting than id. what we sometimes call sample (the indices hold in a cluster) could be called cluster id instead.

janpetschexain:
bors merge

PR #259, opened 2022-03-17
[Merged by Bors] - TY-2454 news provider supports excluded sources

rustonaut:
**References:**

- [story TY-2528](https://xainag.atlassian.net/browse/TY-2528)
- [task TY-2454](https://xainag.atlassian.net/browse/TY-2454)

**Based on:**

- #258

PR #258, opened 2022-03-17
[Merged by Bors] - TY-2599	news provider do not support to_rank in headline queries.

rustonaut:
- newscatcher doesn't support `to_rank` in headline queries
- given the feedback from the previous PR it is unnecessary to store
  to limit anywhere (as it isn't meant to be configurable).

**References:**

- [TY-2599](https://xainag.atlassian.net/browse/TY-2599)

PR #260, opened 2022-03-17
[Merged by Bors] - TY-2457 provide excluded sources to provider

rustonaut:
**References:**

- [task TY-2457](https://xainag.atlassian.net/browse/TY-2457)

**Based on:**

- #258 
- #259

### DIFF
--- a/discovery_engine_core/core/src/engine.rs
+++ b/discovery_engine_core/core/src/engine.rs
@@ -140,7 +140,7 @@ impl From<InitConfig> for EndpointConfig {
             api_base_url: config.api_base_url,
             page_size: 100,
             markets: Arc::new(RwLock::new(config.markets)),
-            excluded_sources: Arc::new(RwLock::new(Vec::new())),
+            excluded_sources: Arc::default(),
         }
     }
 }
--- a/discovery_engine_core/core/src/engine.rs
+++ b/discovery_engine_core/core/src/engine.rs
@@ -27,31 +27,19 @@ use tracing::error;
 
 use xayn_ai::{
     ranker::{AveragePooler, Builder, CoiSystemConfig},
-    KpeConfig,
-    SMBertConfig,
+    KpeConfig, SMBertConfig,
 };
 use xayn_discovery_engine_providers::{Client, CommonQueryParts, Filter, Market, NewsQuery};
 
 use crate::{
     document::{
-        self,
-        document_from_article,
-        Document,
-        HistoricDocument,
-        TimeSpent,
-        UserReacted,
+        self, document_from_article, Document, HistoricDocument, TimeSpent, UserReacted,
         UserReaction,
     },
     mab::{self, BetaSampler, SelectionIter},
     ranker::Ranker,
     stack::{
-        self,
-        BoxedOps,
-        BreakingNews,
-        Data as StackData,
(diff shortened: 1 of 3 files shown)


rustonaut:
In this PR I treated it exactly like market, but we probably want to do some refactoring in a follow up PR to:

- reduce code duplication between the current Ops implementations
- clean up some code around the configuration/locks etc.

acrrd:
If it's based on other pr you can set them as the base branch

discovery_engine_core/core/src/engine.rs
```diff
@@ -138,6 +140,7 @@ impl From<InitConfig> for EndpointConfig {
             api_base_url: config.api_base_url,
             page_size: 100,
             markets: Arc::new(RwLock::new(config.markets)),
+            excluded_sources: Arc::new(RwLock::new(Vec::new())),
```
janpetschexain, line 143: 🟢  the wrapped types all implement `Default`, so we could simplify the call
```suggestion
            excluded_sources: Arc::default(),
```

discovery_engine_core/core/src/stack/ops/breaking.rs
```diff
@@ -60,16 +63,24 @@ impl Ops for BreakingNews {
     }
 
     async fn new_items(&self, _key_phrases: &[KeyPhrase]) -> Result<Vec<Article>, GenericError> {
-        if let Some(markets) = self.markets.as_ref() {
+        if let (Some(markets), Some(excluded_sources)) =
+            (self.markets.as_ref(), self.excluded_sources.as_ref())
```
janpetschexain, line 67: 🔴  this won't fetch any new items in the case of `(Some(markets), None)`, which is a viable case we need to support. same for personalized news.

discovery_engine_core/core/src/stack/ops/breaking.rs
```diff
@@ -112,6 +123,7 @@ fn spawn_headlines_request(
     client: Arc<Client>,
     market: Market,
     page_size: usize,
+    excluded_sources: Vec<String>,
```
janpetschexain, line 126: 🟡  we could wrap the vector in another `Arc` to avoid cloning the vector everytime we make a request, ie `BreakingNews` would have the field `excluded_sources: Option<Arc<RwLock<Arc<Vec<String>>>>>`. similarly for `PersonalizedNews`, wdyt?
```suggestion
    excluded_sources: Arc<Vec<String>>,
```

janpetschexain (commented):
i won't hit the request changes button to not block this pr next week, but the red comment needs to be adressed before merging.

discovery_engine_core/core/src/stack/ops/breaking.rs
rustonaut: Yes I wanted to fix the unnecessary cloning in a follow up refactoring (which also reduces the number of locks and arcs) through until then but this works, too.

rustonaut:
After (in the background) going through the borse -> rebase  -> borse cycle  for the PRs this is based on this is now finally fully re-based on `main`. All needed changes are applied. @Robert-Steiner

Robert-Steiner (approved):

rustonaut:
bors merge

PR #264, opened 2022-03-17
[Merged by Bors] - TY-2607 split pipeline into multiple workflows

x3ro:
TY-2607


.github/workflows/pull-request.yml
```diff
@@ -0,0 +1,77 @@
...
+              - 'discovery_engine/**'
+            flutter:
+              - *rust
+              - *dart
+              - 'discovery_engine_flutter/**'
+
+  rust-checks:
+    needs: selection
+    if: ${{ needs.selection.outputs.rust  == 'true' }}
+    uses: ./.github/workflows/_reusable.rust.yml
```
Robert-Steiner, line 55: ❔does this work now? I thought reusable workflow cannot be nested

.github/actions/setup-job-docker/action.yml
```diff
@@ -20,6 +20,14 @@ runs:
       with:
         working-directory: ${{ env.RUST_WORKSPACE }}
 
+    - name: Configure cargo
+      shell: bash
+      run: |
+        # We do this since multiple jobs are run on the same CI build server,
```
Robert-Steiner, line 26: ❔would it be possible to limit the cpu resources a single docker container can use instead of limit the cargo jobs?

Robert-Steiner (commented):
Looks good so far 👍 
I really like the new structure. much cleaner and easier to understand

.github/actions/setup-job-docker/action.yml
x3ro: I think that this would be possible in theory (with docker) but not in practice (with the github runners spawning the docker containers). It would certainly be the cleaner solution, though, so we could research this as a follow-up. 📌 maybe?

.github/workflows/pull-request.yml
x3ro: `pull-request` is a top-level workflow, so this would not qualify as nested, I think 🤔

.github/actions/setup-job-docker/action.yml
Robert-Steiner: > I think that this would be possible in theory (with docker) but not in practice (with the github runners spawning the docker containers).

Ah true that could be a problem. As far as I remember, the settings that you can set on the github runner are very limited

.github/workflows/pull-request.yml
Robert-Steiner: but `workflow_call` defines a resuable workflow. maybe `workflow_dispatch` was meant?

x3ro: Ah, I think that's just a misunderstanding. The `pull-request.yml` is what was previously the `ci.yml`, but I threw in the `workflow_call` so that it doesn't run automatically while I'm testing the full bors CI run :) So that `workflow_call` would not remain there.

.github/actions/setup-job-docker/action.yml
acrrd: github allows to set some options for the container including cpu. We might also try something with cgroups. but it could be an optimization of this.

How 6 was choosen?

x3ro: Fair question. Currently it's approx. `$number_of_cores / $number_of_runners` :) I definitely think this _could_ be something we might want to look into, but only if just setting a fixed number doesn't work well.

.github/workflows/_reusable.rust.yml
```diff
@@ -0,0 +1,73 @@
...
+      - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # v2.4.0
+
+      - name: Setup the rust CI.
+        uses: ./.github/actions/setup-job-docker
+        with:
+          rust: true
+
+      - name: cargo clippy
+        run: |
+          just rust-check
```
Robert-Steiner, line 38:
```suggestion
        run: just rust-check
```

.github/workflows/_reusable.rust.yml
```diff
@@ -0,0 +1,73 @@
...
+
+      - name: Setup the rust CI.
+        uses: ./.github/actions/setup-job-docker
+        with:
+          rust: true
+
+      - name: Run tests
+        run: |
+          env
+          just rust-test
```
Robert-Steiner, line 56:
```suggestion
      - name: Run tests
        run: just rust-test
```

.github/workflows/_reusable.rust.yml
```diff
@@ -0,0 +1,73 @@
...
+      - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # v2.4.0
+
+      - name: Setup the rust CI.
+        uses: ./.github/actions/setup-job-docker
+        with:
+          rust: true
+
+      - name: cargo doc
+        run: |
+          just rust-check-doc
```
Robert-Steiner, line 73:
```suggestion
        run: just rust-check-doc
```

.github/workflows/pull-request.yml
```diff
@@ -0,0 +1,74 @@
...
+    outputs:
+      dart: ${{ steps.filter.outputs.dart }}
+      rust: ${{ steps.filter.outputs.rust }}
+      flutter: ${{ steps.filter.outputs.flutter }}
+    timeout-minutes: 5
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # v2.4.0
+
+      # For pull requests it's not necessary to checkout the code
```
Robert-Steiner, line 30: I don't understand this comment because in the previous step we are checking out the code. maybe we can remove this line or check if this action works without checking out the repo first

.github/actions/setup-job-docker/action.yml
```diff
@@ -20,6 +20,14 @@ runs:
       with:
         working-directory: ${{ env.RUST_WORKSPACE }}
 
+    - name: Configure cargo
+      shell: bash
```
Robert-Steiner, line 24: should we also only run this if `rust=true`?

Robert-Steiner (approved):
added some small suggestions to improve formatting
from my side it is ready to be merged

.github/workflows/pull-request.yml
x3ro: Fair, I don't understand it either tbh 🤔 I'll take a look.

x3ro:
bors merge

x3ro:
bors merge-

x3ro:
bors merge

x3ro:
bors merge

x3ro:
bors merge

x3ro:
bors try

x3ro:
bors try-

x3ro:
bors try

x3ro:
bors try-

x3ro:
bors try

x3ro:
bors merge

x3ro:
bors merge-

x3ro:
bors merge

PR #266, opened 2022-03-18
[Merged by Bors] - TY-2543 refine dedup logic in article filtering

finiteprods:
**Summary**

refines the dedup logic of `filter_articles`:

1. articles that are duplicates (i.e. there is a common title or link) of each other are discarded, keeping only the best ranked among them.
2. articles that are duplicates of historic documents are discarded (as before).
3. articles that are duplicates of stack documents are discarded if not better ranked.

in case 3., stack documents that are worse ranked duplicates of articles are deduped later as part of `merge`.


discovery_engine_core/core/src/stack/filters.rs
```diff
@@ -39,52 +36,57 @@ impl ArticleFilter for DuplicateFilter {
...
-            .collect::<HashSet<_>>();
-
-        let titles = history
+            .map(|doc| (doc.resource.url.as_str(), doc.resource.rank))
+            .chain(history.iter().map(|doc| (doc.url.as_str(), 0)))
+            .fold(HashMap::new(), |mut urls, (url, rank)| {
+                let best_rank = urls.entry(url).or_insert(rank);
+                if rank < *best_rank {
+                    *best_rank = rank;
+                };
```
janpetschexain, line 49: 🟢  we could write this a bit more consise such that there is no additional check done in the or-case. similarly for the `titles`.
```suggestion
                urls.entry(url)
                    .and_modify(|best_rank| {
                        if rank < *best_rank {
                            *best_rank = rank;
                        }
                    })
                    .or_insert(rank);
```

discovery_engine_core/core/src/stack/filters.rs
```diff
@@ -39,52 +36,57 @@ impl ArticleFilter for DuplicateFilter {
...
+            .chain(history.iter().map(|doc| (&doc.title, 0)))
+            .fold(HashMap::new(), |mut titles, (title, rank)| {
+                let best_rank = titles.entry(title).or_insert(rank);
+                if rank < *best_rank {
+                    *best_rank = rank;
+                };
+                titles
+            });
+
+        // discard dups in the title keeping only the best ranked
```
janpetschexain, line 65: 🟢  we could reorder the code blocks such that first the title map & dedup is done, and afterwards the url map & dedup is done. this would place the map next to the dedup for better readability, also the compiler could drop the maps sooner. (the retaining by rank would come as a third block like it currently is.)

finiteprods: thanks. actually the `titles` and `urls` hash maps are used by the "retaining" block (rather than the dedup block) so what I'll do is move them down so they're closer together.

discovery_engine_core/core/src/stack/filters/article.rs
```diff
@@ -39,52 +36,64 @@ impl ArticleFilter for DuplicateFilter {
...
+        articles.dedup_by(|art1, art2| art1.link == art2.link);
 
-        impl Eq for UniqueArticle {}
-        impl PartialEq for UniqueArticle {
-            fn eq(&self, other: &Self) -> bool {
-                self.0.link == other.0.link || self.0.title == other.0.title
+        let urls = stack
+            .iter()
+            .map(|doc| (doc.resource.url.as_str(), doc.resource.rank))
+            .chain(history.iter().map(|doc| (doc.url.as_str(), 0)))
```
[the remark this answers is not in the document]
acrrd: Can't we move the code that remove articles that are already in the history as first pass? and only then dedup and keep the best rank? This should remove the need for the special `0` case.

discovery_engine_core/core/src/document.rs
```diff
@@ -222,6 +223,27 @@ pub(crate) fn document_from_article(
     })
 }
 
+/// Removes documents where the title or url is duplicated.
+pub(crate) fn dedup_documents(documents: &mut Vec<Document>) {
+    // discard dups in the title keeping only the best ranked
+    documents.sort_unstable_by(
+        |doc1, doc2| match doc1.resource.title.cmp(&doc2.resource.title) {
+            Ordering::Equal => doc1.resource.rank.cmp(&doc2.resource.rank),
```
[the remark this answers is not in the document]
finiteprods: there is indeed a `published_date` on articles but not sure whether that is a better way to choose between (say in this case) 2 articles with the same title. assuming we want to keep the "better" one (which may not necessarily be the earliest one), it seems to me sensible to let rank guide the choice?

discovery_engine_core/core/src/stack/filters/article.rs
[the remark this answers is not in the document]
finiteprods: @x3ro yes. you could also maybe think of `0` as `NIL_RANK` here, since historical documents don't have a rank. but yes, there was a particular reason `0` was chosen and not something else.
@acrrd that works too, and perhaps that will be clearer overall to just deal with it separately in a pass.

discovery_engine_core/core/src/stack/filters/article.rs
```diff
@@ -39,52 +36,64 @@ impl ArticleFilter for DuplicateFilter {
...
+                    })
+                    .or_insert(rank);
+                titles
+            });
+
+        // discard worse-ranked dups of documents
+        articles.retain(|article| {
+            match (urls.get(&article.link.as_str()), titles.get(&article.title)) {
+                (None, None) => true,
+                (Some(rank_u), None) => &article.rank < rank_u,
```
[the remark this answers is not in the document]
finiteprods: I agree they're not great names. if you can tolerate it, I'd propose to just call these `doc_rank`s and instead supplement with a few words in a comment which would explain it better
```rust
// a few words here to explain...
match (..., ...) {
    (Some(doc_rank), None) => art.rank < doc_rank,
    ...
```

discovery_engine_core/core/src/document.rs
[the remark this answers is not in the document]
x3ro: Yeah, I suppose this might also be a question we would want to direct at pink? 🤔 I'm certainly not equipped to answer it 😅 Maybe it also doesn't matter here, as we're talking about exact duplicates.

discovery_engine_core/core/src/stack/filters/article.rs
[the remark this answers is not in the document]
x3ro: Yeah, I think `NIL_RANK` would still not convey the semantics of the constant, i.e. what using the 0 means there. But I'd be happy with @acrrd's solution also.

discovery_engine_core/core/src/document.rs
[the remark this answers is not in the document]
acrrd: At the moment we want to try with the rank but this will be probably a subject for many tests to see what works better.

discovery_engine_core/core/src/document.rs
```diff
@@ -222,6 +223,27 @@ pub(crate) fn document_from_article(
...
+            ord => ord,
+        },
+    );
+    documents.dedup_by(|doc1, doc2| doc1.resource.title == doc2.resource.title);
+
+    // discard dups in the url keeping only the best ranked
+    documents.sort_unstable_by(
+        |doc1, doc2| match doc1.resource.url.cmp(&doc2.resource.url) {
+            Ordering::Equal => doc1.resource.rank.cmp(&doc2.resource.rank),
```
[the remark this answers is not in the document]
acrrd: Yes is a good point, if the url is the same there is not point in swapping because the rank will be the same.

discovery_engine_core/core/src/stack/filters/article.rs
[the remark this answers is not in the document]
acrrd: I think it will be clearer so we don't have to hack in rank values for history values.

So at the light of the latest comments we want:
remove an article if:
 * title or url are the same of one in the history.
 * url is the same of one in the stack documents or batch of articles

Then we want to keep the the article with higher rank if another article or document have same title.

To compare the title we should trim it and put it lowercase.

discovery_engine_core/core/src/document.rs
```diff
@@ -222,6 +223,23 @@ pub(crate) fn document_from_article(
...
 }
 
+/// Discards documents where the title is duplicated, keeping only the top ranked.
+pub(crate) fn dedup_documents(documents: &mut Vec<Document>) {
+    documents.sort_unstable_by(|doc1, doc2| {
+        match doc1
+            .resource
+            .title
+            .to_lowercase()
+            .cmp(&doc2.resource.title.to_lowercase())
```
acrrd, line 233: We should be able to use `then_with` or `then` instead of the explicit match

discovery_engine_core/core/src/stack/filters/article.rs
```diff
@@ -39,52 +39,70 @@ impl ArticleFilter for DuplicateFilter {
...
+        mut articles: Vec<Article>,
     ) -> Result<Vec<Article>, GenericError> {
-        let urls = history
-            .iter()
-            .map(|doc| doc.url.as_str())
-            .chain(stack.iter().map(|doc| doc.resource.url.as_str()))
-            .collect::<HashSet<_>>();
+        // discard dups in the title keeping only the best ranked
+        articles.sort_unstable_by(|art1, art2| {
+            match art1.title.to_lowercase().cmp(&art2.title.to_lowercase()) {
```
acrrd, line 46: Maybe we can a function to normalize the article for comparison? here are not trimming while below we are.

discovery_engine_core/core/src/stack/filters/article.rs
```diff
@@ -39,52 +39,70 @@ impl ArticleFilter for DuplicateFilter {
...
+                    .or_insert(rank);
+                titles
+            });
+
+        // discard worse-ranked dups of stack documents; more precisely, discard:
+        // * dups of stack documents in the url
+        // * dups of stack documents in the title when the rank is no better
+        articles.retain(|art| {
+            match (
+                stack_urls.contains(art.link.as_str()),
```
acrrd, line 96: Can we avoid to look into titles if we know that it is contained in the url?
Something like
`urls.contain(...) || titles.get(...).map(|doc| art.rank < doc.rank).unwrap_or(true)`

discovery_engine_core/core/src/document.rs
finiteprods: thanks for pointing that out

discovery_engine_core/core/src/stack/filters/article.rs
finiteprods: good catch. added a helper `normalize` to `utils`

discovery_engine_core/core/src/stack/filters/article.rs
finiteprods: nice suggestion. based on your idea,
```rust
!urls.contains(art.link) && titles.get(art.title).map_or(true, |doc_rank| art.rank < doc_rank)
```
should do the trick (clippy helpfully suggested `map_or` in place of `.map().unwrap_or()`)

discovery_engine_core/core/src/utils.rs
```diff
@@ -55,3 +55,8 @@ pub(crate) fn nan_safe_f32_cmp_base(a: &f32, b: &f32, nan_min: bool) -> Ordering
 pub(crate) fn nan_safe_f32_cmp(a: &f32, b: &f32) -> Ordering {
     nan_safe_f32_cmp_base(a, b, true)
 }
+
+/// Normalizes `text` to a trimmed lowercase string.
+pub(crate) fn normalize(text: &str) -> String {
```
acrrd, line 60: I'll prefer to keep it together with the other code, it does not seems that useful in general at the moment.
If we `trim` before `to_lowercase` we can avoid the last `to_string`.

acrrd (approved):

discovery_engine_core/core/src/utils.rs
finiteprods: thanks, moved back to `article`.

discovery_engine_core/core/src/stack/filters/article.rs
finiteprods: [now in `article`](https://github.com/xaynetwork/xayn_discovery_engine/pull/266#discussion_r835468158)

PR #269, opened 2022-03-21
[Merged by Bors] - TY 2642 Change release repo

acrrd:


acrrd:
Test run:
https://github.com/xaynetwork/xayn_discovery_engine/runs/5630445167?check_suite_focus=true
result:
https://gitlab.com/xayn/xayn_discovery_engine_release/-/tree/TY-2642_change_release_repo

.github/workflows/release.yml
```diff
@@ -19,22 +19,22 @@ jobs:
 
   release:
     runs-on: ubuntu-20.04
-    container:
-      image: xaynetci/yellow:9135de0
+    #container:
```
acrrd, line 22: docker image does not have rsync that is needed by `release.sh`

.github/workflows/release.yml
```diff
@@ -19,22 +19,22 @@ jobs:
...
 
       - name: Checkout repository
         uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # v2.4.0
 
       - name: Setup job
-        uses: ./.github/actions/setup-job-docker
+        uses: ./.github/actions/setup-job
         with:
           dart: true
           rust: true
```
Robert-Steiner, line 40: beacuse we use the old script. but maybe it doesn't matter because we probably won't get into that limit anymore
```suggestion
          rust: true
          github_token: ${{ secrets.GITHUB_TOKEN }}
```

Robert-Steiner (approved):

.github/workflows/release.yml
acrrd: I think since we had the problem with the size no one run the release CI and we didn't notice.

@rustonaut we have setup-job that install everything that is still used on the mac runner, and we have setup-job-docker that only loads env and setup cache for when we run on docker.

acrrd:
bors merge

PR #274, opened 2022-03-22
[Merged by Bors] - TY-2455 pass excluded sources with init config

rustonaut:
**References:**

- [story TY-2453](https://xainag.atlassian.net/browse/TY-2453)
- [task TY-2455](https://xainag.atlassian.net/browse/TY-2455)

PR #275, opened 2022-03-22
[Merged by Bors] - TY-2456 Allow setting excluded sources in engine after init.

rustonaut:
**References:**

- [task TY-2456](https://xainag.atlassian.net/browse/TY-2456)

**Based on:**

-  #274

### DIFF
--- a/discovery_engine/lib/src/domain/feed_manager.dart
+++ b/discovery_engine/lib/src/domain/feed_manager.dart
@@ -145,18 +145,20 @@ class FeedManager {
       throw ArgumentError('source can\'t be empty');
     }
 
-    final allDocuments = await _docRepo.fetchAll();
-    final doesExist = allDocuments
-        .any((doc) => doc.isActive && doc.resource.sourceDomain == source);
+    // FIXME BUG isActive is not correctly set as activity is tracked through another repo :/
+    // final allDocuments = await _docRepo.fetchAll();
+    // final doesExist = allDocuments
+    //     .any((doc) => doc.isActive && doc.resource.sourceDomain == source);
 
-    if (!doesExist) {
-      throw ArgumentError('source $source not found in database');
-    }
+    // if (!doesExist) {
+    //   throw ArgumentError('source $source not found in database');
+    // }
 
     final sources = await _excludedSourcesRepository.getAll();
     sources.add(source);
     await _excludedSourcesRepository.save(sources);
-    await _engine.setExcludedSources(sources);
+    final history = await _docRepo.fetchHistory();
+    await _engine.setExcludedSources(history, sources);
     return const EngineEvent.clientEventSucceeded();
   }
 
@@ -169,7 +171,8 @@ class FeedManager {
     final sources = await _excludedSourcesRepository.getAll();
     sources.remove(source);
     await _excludedSourcesRepository.save(sources);
-    await _engine.setExcludedSources(sources);
+    final history = await _docRepo.fetchHistory();
+    await _engine.setExcludedSources(history, sources);
     return const EngineEvent.clientEventSucceeded();
   }
 
--- a/discovery_engine/lib/src/domain/feed_manager.dart
+++ b/discovery_engine/lib/src/domain/feed_manager.dart
@@ -145,15 +145,6 @@ class FeedManager {
       throw ArgumentError('source can\'t be empty');
     }
 
-    // FIXME BUG isActive is not correctly set as activity is tracked through another repo :/
-    // final allDocuments = await _docRepo.fetchAll();
-    // final doesExist = allDocuments
-    //     .any((doc) => doc.isActive && doc.resource.sourceDomain == source);
-
-    // if (!doesExist) {
-    //   throw ArgumentError('source $source not found in database');
-    // }
-
     final sources = await _excludedSourcesRepository.getAll();
     sources.add(source);
     await _excludedSourcesRepository.save(sources);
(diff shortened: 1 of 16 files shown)


rustonaut:
The "source in history" check is temporary disabled as it has an bug. (Which needs some discussion and can be fixed in another PR).

rustonaut:
Ok problem seems to be simpler then I thought, I just seem to not use the test fixtures I thought it was using.

discovery_engine/lib/src/domain/feed_manager.dart
```diff
@@ -145,18 +145,11 @@ class FeedManager {
...
     }
 
-    final allDocuments = await _docRepo.fetchAll();
-    final doesExist = allDocuments
-        .any((doc) => doc.isActive && doc.resource.sourceDomain == source);
-
-    if (!doesExist) {
-      throw ArgumentError('source $source not found in database');
-    }
-
```
rustonaut, line 155: I removed this test as it's very prone to cause problems/complications for very little benefit.

It creates a implicit coupling with assumptions about the state a mostly unrelated part of the app will be in (the active documents).

This means any time we change any of the requirements around it it's prone to fail.

Furthermore it means in tests for add/rem excludedSources you have to recreate the right implicit state by happening to now how to combine the right fixtures and engine calls while avoiding filtering of any kind messing the state up. Which means there is a lot of implicit "state" in the test which makes it harder to maintain/understand.

(E.g. currently there is some unrelated problem somewhere, which makes creating that state hard/complicated.)

I.e. with this tests we can have false/spurious (but quite confusing) failures if:

- We allow adding sources in a different context (e.g. based on the /source endpoint of newscatcher).
- We change filtering of articles.
- We have a bug in filtering of sources, or the test server, or a unexpected change in the test fixture etc.
- We change how active documents are handled.

acrrd: The benefit is that we can get logic error from the user if they pass us invalid sources and there is not other way to know this otherwise.
The points you mentions are all valid but I don't agree with the solution. Instead of removing a check we should create better abstractions to allow to have simpler tests. We can introduce `SourcesRepository` to which we can ask if a source is ok or not. We can easily mock it in the tests and the real implementation will do what we are doing here and in the future we can change it to use the endpoint or some other data.

discovery_engine_core/core/src/engine.rs
```diff
@@ -447,6 +450,31 @@ where
...
+        &mut self,
+        history: &[HistoricDocument],
+        excluded_sources: Vec<String>,
+    ) -> Result<(), Error> {
+        let exclusion_set = excluded_sources.iter().cloned().collect::<HashSet<_>>();
+        *self.config.excluded_sources.write().await = excluded_sources;
+
+        let mut stacks = self.stacks.write().await;
+        for stack in stacks.values_mut() {
+            stack.data.prune_by_excluded_sources(&exclusion_set);
```
Robert-Steiner, line 465: ❔Maybe a filter function on the stack struct is a bit cleaner, or we just reset the stack like we do in setMarket

discovery_engine/lib/src/domain/feed_manager.dart
biernacki: I've mentioned a similar solution here https://github.com/xaynetwork/xayn_discovery_engine/pull/249#discussion_r828769317, which could be adopted to also take favorite sources into consideration. Maybe a smth like this?

```dart
enum SourceStatus {
  neutral, // default status
  excluded,
  favorite,
}

final sourcesRepo = {
  'example.com': SourceStatus.excluded,
  'bbc.co.uk': SourceStatus.favorite,
  'sport.de': SourceStatus.neutral,
};

final isValidSource = sourcesRepo['theguardian.com'] != null;
```

acrrd: I think we can refactor in this direction but using a repository instead of data structure. Maybe we don't need the neutral status depending on the api we want to expose but would be nice to have a central place for sources and where we can check that a source is only excluded or favorite and not both.

biernacki: Yes, ofc. I used data structure just as an example to describe how a Hive box could look like

rustonaut: >  Instead of removing a check we should create better abstractions to allow to have simpler tests

Yes but not in this PR.

>  We can introduce SourcesRepository to which we can ask if a source is ok or not. We can easily mock it in the tests and the real implementation will do what we are doing here and in the future we can change it to use the endpoint or some other data.

Its only preventing a systematic API misuse by Team Blue which stems from sources being stringly typed, something we can change relatively easy without needing a add a bunch of complexity through another repository + mocking. Especially giving that mocking in integration tests across isolate boundaries just doesn't work well.

discovery_engine/lib/src/infrastructure/type_adapters/hive_set_string_adapter.dart
```diff
@@ -0,0 +1,36 @@
...
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+import 'package:hive/hive.dart' show BinaryReader, BinaryWriter, TypeAdapter;
+import 'package:xayn_discovery_engine/src/domain/repository/type_id.dart'
+    show setStringTypeId;
+
+class SetStringAdapter extends TypeAdapter<Set<String>> {
```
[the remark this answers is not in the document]
rustonaut: sure

Robert-Steiner (approved):
🚀

biernacki (approved):

rustonaut:
bors merge

PR #278, opened 2022-03-22
[Merged by Bors] - Switch CI docker container from hard-coded tag to `v1` tag

x3ro:
[This commit](https://github.com/xaynetwork/xayn_docker/commit/f3bcb13d6ec3bdb0d14178106443caf3c6ca244b) from the `xayn_docker` repo adds rsync and allows the `release` tag to be created.

### DIFF
--- a/.github/workflows/ci_build_android.yml
+++ b/.github/workflows/ci_build_android.yml
@@ -17,7 +17,7 @@ jobs:
   build-android-libs:
     runs-on: ubuntu-20.04
     container:
-      image: xaynetci/yellow:9135de0
+      image: xaynetci/yellow:v1
     timeout-minutes: 45
     strategy:
       matrix:
--- a/.github/workflows/ci_flutter_example.yml
+++ b/.github/workflows/ci_flutter_example.yml
@@ -44,7 +44,7 @@ jobs:
     # that needs to exist during the linking process.
     runs-on: ubuntu-20.04
     container:
-      image: xaynetci/yellow:9135de0
+      image: xaynetci/yellow:v1
     timeout-minutes: 30
     steps:
       - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # v2.4.0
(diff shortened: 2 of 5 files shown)


acrrd:
Why call it `release` and not `latest`? `release` has a stronger meaning and are not attaching versioning to it.

x3ro:
I chose something else than latest because it's easy to accidentally tag things `latest`, as it's the docker default. I wanted to explicitly prevent this error scenario, where e.g. a local build or a modification in the pipeline accidentally overwrites the `latest` tag. In addition, `latest` would also be wrong semantically, as the `release` tag is manually generated (similar to the release job in the discovery engine), so it would not be `latest` most of the time.

acrrd:
Another question. How would be the workflow when a PR needs to update the image with a change that is not backward compatible? If we update the image other PRs will fail but them cannot rebased yet. Do we have a nice solution for this or we hope that this will not happen that often to create significant problem?

x3ro:
I think in that case I would create a new tag, e.g. `release-v2` or something like that, which would then allow us to do successive updates in the pipelines. If we discover that something like this happens regularly in the future, we could also think about a more elaborate tagging scheme (e.g. semver like), but I think that this would currently not be worth the effort.

acrrd:
Should we then start with `v1` (without the 'release' part, it don't think it adds much value if we have the version) and bump it when we have a breaking change?

x3ro:
Sure, let's go with that.

x3ro:
borst try-

x3ro:
bors try

x3ro:
bors try-

x3ro:
bors try

Robert-Steiner (approved):
🚀

x3ro:
bors merge

acrrd (approved):

x3ro:
bors merge-

x3ro:
bors merge

x3ro:
bors merge-

x3ro:
bors merge

PR #281, opened 2022-03-24
[Merged by Bors] - TY-2602 add favourite sources to headline query

finiteprods:
**Summary**

adds `sources` to `HeadlinesQuery`.

### DIFF
--- a/discovery_engine_core/providers/src/client.rs
+++ b/discovery_engine_core/providers/src/client.rs
@@ -62,7 +62,7 @@ pub struct CommonQueryParts<'a> {
     ///
     /// Paging starts with `1`.
     pub page: usize,
-    /// Exclude given sources
+    /// Exclude given sources.
     pub excluded_sources: &'a [String],
 }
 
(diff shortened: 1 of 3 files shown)


discovery_engine_core/providers/src/client.rs
```diff
@@ -119,15 +119,24 @@ where
...
 
 impl Query for HeadlinesQuery<'_> {
     fn setup_url(&self, url: &mut Url) -> Result<(), Error> {
-        self.common.setup_url(url, "_lh")
+        self.common.setup_url(url, "_lh")?;
+
+        if !self.sources.is_empty() {
+            let mut query = url.query_pairs_mut();
+            query.append_pair("sources", &self.sources.join(","));
+        };
```
rustonaut, line 137: :orange_circle:  As far as I can tell this should go into `CommonQueryParts` as it applies to both news and headlines.

Through the issue only says `HeadlineQuery` :thinking: 

@acrrd is meant to be applied to both kinds of queries?

acrrd: Now we plan to use it only for query to the headline endpoints but we can indeed add it to the common query so it is already there we decide to do something different later.

finiteprods:
bors merge

PR #283, opened 2022-03-25
[Merged by Bors] - TY-2603 add favourite sources to initconfig

finiteprods:
previously #281

**Summary**

- adds `sources` to `InitConfig`, `EndpointConfig`.
- adds corresponding FFI binding.


discovery_engine/lib/src/domain/repository/sources_repo.dart
```diff
@@ -0,0 +1,22 @@
...
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+/// Repository interface for source domains excluded from the feed.
+abstract class SourcesRepository {
+  /// Get a set of all favourite sources.
+  Future<Set<String>> getAll();
+
+  /// Persist a set of favourite sources.
+  Future<void> save(Set<String> sources);
+}
```
rustonaut, line 22: Will this be merged with the `ExcludedSourcesRepository` in a later PR.

If so maybe we should call the methods `getFavorites`/`saveFavorites` or similar.

If not this probably should still be called `FavoritSourceRepository`

discovery_engine/test/ffi/ffi_test.dart
```diff
@@ -59,6 +59,7 @@ void main() {
           engineState: null,
           history: [],
           aiConfig: null,
+          sources: {},
```
rustonaut, line 62: This probably shouldn't be called `sources` but `favoriteSources`. (also in other places)

We do not need to use the exact naming of the http parameter in newscatcher, I mean 
we also don't call `excludedSources` `notSources` :wink:

discovery_engine/lib/src/domain/event_handler.dart
```diff
@@ -308,6 +313,7 @@ class EventHandler {
       _openDbBox<ActiveDocumentData>(activeDocumentDataBox),
       _openDbBox<Uint8List>(engineStateBox),
       _openDbBox<ActiveSearch>(searchBox),
+      _openDbBox<Set<String>>(sourcesBox),
```
rustonaut, line 316: After merging #284 this will change to `Set<Sources>` same in other places.

discovery_engine/lib/src/domain/repository/sources_repo.dart
finiteprods: let's keep them separate for now. whoever ends up working on the dart side of this story can later refactor this if necessary - I only introduced this here to have the minimum amount of dart in place for the changes on the rust side to go through.

btw you didn't prefix your comments with PR-emojis :wink:

discovery_engine/test/ffi/ffi_test.dart
finiteprods: >We do not need to use the exact naming of the http parameter in newscatcher, I mean
we also don't call `excludedSources` `notSources`

true but I felt that `Sources` vs `ExcludedSources` was enough to disambiguate, and is consistent in the same way newscatcher has `sources` vs `not_sources` (even if we don't necessarily copy their exact naming as you mentioned).

but I can change it if you disagree (opinions from others also welcome!)

discovery_engine/lib/src/domain/event_handler.dart
finiteprods: thanks, will adjust once that's merged.

discovery_engine_core/core/src/stack/ops/breaking.rs
```diff
@@ -39,6 +39,7 @@ use super::{
 pub(crate) struct BreakingNews {
     client: Arc<Client>,
     markets: Arc<RwLock<Vec<Market>>>,
+    sources: Arc<RwLock<Vec<String>>>,
```
acrrd, line 42: :wrench: Favorite sources will only be used in their own stack no need to apply changes here.

discovery_engine/lib/src/domain/event_handler.dart
```diff
@@ -226,6 +230,7 @@ class EventHandler {
         engineState: engineState,
         history: history,
         aiConfig: aiConfig,
+        sources: sources,
```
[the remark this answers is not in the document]
finiteprods: you are right, was doing more than necessary!

discovery_engine/lib/src/domain/repository/sources_repo.dart
finiteprods: see [later comment](https://github.com/xaynetwork/xayn_discovery_engine/pull/283#discussion_r839719445)

finiteprods:
bors merge

PR #295, opened 2022-04-05
[Merged by Bors] - TY-2708 add cluster test

Robert-Steiner:
Ticket:
- [TY-2708]

Summary:
- adds additional semantic cluster test
- the test only checks the calculation of the normalized distances
- The calculation of the normalized distances was checked for consistency with the calculation of the Python implementation 

[TY-2708]: https://xainag.atlassian.net/browse/TY-2708?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


discovery_engine_core/core/src/stack/filters/semantic.rs
```diff
@@ -136,6 +136,15 @@ fn cut_tree(dendrogram: &Dendrogram<f32>, max_dissimilarity: f32) -> Vec<usize>
     )
 }
 
+/// Calculates the normalized distances.
+fn normalized_distance(documents: &[Document], config: &SemanticFilterConfig) -> Vec<f32> {
```
x3ro, line 140: Is there a link to pink's documentation that we can reference to? I feel like, for these sort of calculations, that makes sense. At least for me, it otherwise becomes difficult to determine whether we implemented the correct thing 🤔

discovery_engine_core/core/src/stack/filters/semantic.rs
```diff
@@ -317,4 +328,39 @@ mod tests {
...
+                resource: NewsResource {
+                    date_published: NaiveDateTime::from_timestamp(secs, 0),
+                    ..NewsResource::default()
+                },
+                ..Document::default()
+            }
+        }
+
+        let documents = vec![
+            new_doc(&[0.1, -0.5, 0.5], 100),
```
x3ro, line 346: Same as above, without knowing what's supposed to happen, it's difficult to verify whether these are correct / good test cases :D

discovery_engine_core/core/src/stack/filters/semantic.rs
Robert-Steiner: afaik there is no real documentation except [this](https://github.com/xaynetwork/xayn_ai_research/tree/main/experiments/semantic_grouping#semantic-similarity-with-publication-date---in-one-step). the `normalized_distance` function correspond to this [code block](https://github.com/xaynetwork/xayn_ai_research/blob/181042ca515ac4aba5de02437f92d298bcf2fddc/experiments/semantic_grouping/combine_similarity_scores.py#L53-L61)  with the exception that we don't create the clusters but only [normalize the scores](https://github.com/xaynetwork/xayn_ai_research/blob/022525203df04238567b3e621eaf183ae33e23c5/experiments/semantic_grouping/utils.py#L461). 

that would be the exact code
```python
combined_scores = get_combined_similarity(
        article_embeddings,
        news_df["published date"],
        max_time=args.THRESHOLD_DATE,
        threshold=args.COMBINED_THRESHOLD,
)
combined_scores_norm = 1 - normalize(combined_scores)
```

there are some fixes that have not been merged yet https://github.com/xaynetwork/xayn_ai_research/pull/184 but are already part of our code.

PR #297, opened 2022-04-05
[Merged by Bors] - [Merged by Bors] - TY 2690 move xayn ai

acrrd:
We move the relevant part of the xayn_ai project in this repository under `discovery_engine_core/ai/`.
The code under `xayn-ai/reranker` has been removed and with it all the parts that were not used anymore.
Most of removed code was under `xayn-ai`, there could still be code in the other crate that is unused but marked as `pub` so the compiler don't tell us. Looking for this code is out of the scope of this PR.

I left some comment on the parts that are important to review, a special look should be give to the second commit "Adapt path to assets" that modify the path to reach the assets in this repository, this is the only big change in the imported code.

The first commit import the crates, the second fix the path to reach the assets to run the tests for the parts that we are interested in. After this we have commits that removes components that we don't want to use anymore followed by commits that remove code that the compiler tell us that is unused.

### DIFF
--- a/discovery_engine_core/ai/rubert/src/config.rs
+++ b/discovery_engine_core/ai/rubert/src/config.rs
@@ -1,3 +1,17 @@
+// Copyright 2022 Xayn AG
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, version 3.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
 use std::{
     fs::File,
     io::{BufRead, BufReader, Read},
--- a/discovery_engine_core/ai/xayn-ai/src/lib.rs
+++ b/discovery_engine_core/ai/xayn-ai/src/lib.rs
@@ -1,3 +1,17 @@
+// Copyright 2021 Xayn AG
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, version 3.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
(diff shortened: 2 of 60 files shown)


discovery_engine_core/ai/xayn-ai/src/lib.rs
```diff
@@ -0,0 +1,41 @@
...
+        DocumentHistory,
+        DocumentId,
+        QueryId,
+        Relevance,
+        SessionId,
+        UserAction,
+        UserFeedback,
+    },
+    error::Error,
+    embedding::utils::COSINE_SIMILARITY_RANGE,
```
acrrd, line 24: This is needed because otherwise it complains that we have reference to private types in the public documentations here: ai/xayn-ai/src/coi/config.rs

discovery_engine_core/ai/rubert/src/config.rs
```diff
@@ -0,0 +1,103 @@
...
+    ///
+    /// Defaults to `true`.
+    pub fn with_lowercase(mut self, lowercase: bool) -> Self {
+        self.lowercase = lowercase;
+        self
+    }
+
+    /// Sets the token size for the tokenizer and the model.
+    ///
+    /// Defaults to [`BertModel::TOKEN_RANGE`].
```
acrrd, line 76: Here it was ``[`K::TOKEN_RANGE`]``, I put `BertModel` since it is the bound on `K`. Maybe we can find a better way to fix this.

NOTICE
```diff
@@ -6,3 +6,7 @@ Xayn AG (https://www.xayn.com/).
 
 Except as otherwise noted (below and/or in individual files), this project is
 licensed under the GNU Affero General Publc License v3.0 only <LICENSE>.
+
+- discovery_engine_core/ai/rubert-tokenizer
+  This crate is a fork of https://github.com/huggingface/tokenizers and is licensed
+  under the Apache License, Version 2.0 <discovery_engine_core/ai/rubert-tokenizer/LICENSE>.
```
acrrd, line 12: Added this because of the different license of rubert-tokenizer

janpetschexain (commented):
🟢  should we also remove the unused models from our S3 buckets after this is merged and the old `xayn_ai` repository is archived?

🟢 should we remove the `obake` versioning to simplify code (maybe in another pr) since we have a fresh start here anyways?

NOTICE
janpetschexain: do we already know how to handle the ci license check? are we going to exclude this directory from the check or are we going to check for various license headers?

acrrd:
> green_circle should we also remove the unused models from our S3 buckets after this is merged and the old `xayn_ai` repository is archived?

I'm not sure that what we have there will stay unused so I'll prefer to keep it for a while.

> green_circle should we remove the `obake` versioning to simplify code (maybe in another pr) since we have a fresh start here anyways?

it can be removed when we switch to the internal database, I don't think is worth to spend time on this now.

janpetschexain (approved):
lgtm 👍

acrrd:
bors merge

PR #299, opened 2022-04-06
[Merged by Bors] - TY-2683 simplify relevance/key phrases map

janpetschexain:
**References**

- [TY-2683]
- requires #297, #303
- followed by #302

**Summary**

- replace `RelevanceMap` with `KeyPhrases` which is a simple mapping from coi ids to key phrases
  - this removes all internal bookkeeping of the various relevance values, now they are computed on the fly when needed
  - remove related dead code and update tests
- move arcing out of the `KeyPhrase`: previously the `Arc` only spanned the embedding point, now it spans the whole key phrase including the words, in the next prs this will also span the market info
- fix bugs in the similarity matrix normalization:
  - add a boolean flag for the `finalize`: the divisor of the mean and stddev was off by one if the respective matrix lane was not part of the main square part of the rectangular matrix
  - add a `prepare`: the `reduce` on itself can't account for non-linear transformations of the first element of the reduced matrix lane, for example for squaring in the stddev computation

[TY-2683]: https://xainag.atlassian.net/browse/TY-2683?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


discovery_engine_core/ai/xayn-ai/src/coi/key_phrase.rs
```diff
@@ -835,88 +682,84 @@ mod tests {
...
         }
+    }
 
-        #[test]
-        fn adjust_spaces() {
-            assert_eq!(clean_key_phrase("  a  b  c  "), "a b c");
-        }
+    #[test]
+    fn adjust_spaces() {
+        assert_eq!(clean_key_phrase("  a  b  c  "), "a b c");
```
acrrd, line 763: :wrench: What's the reason to move this out of the their own mod? it is easier to identify what part this group of functions are testing and I don't see any reason to move them out in this PR.

discovery_engine_core/ai/xayn-ai/src/ranker/system.rs
```diff
@@ -47,7 +52,7 @@ pub(super) struct State {
     pub(super) user_interests: UserInterests,
 
     /// Key phrases.
-    pub(super) relevances: RelevanceMap,
+    pub(super) key_phrases: KeyPhrases,
```
acrrd, line 55: :wrench: we need to change `serialize`/`deserialize`.

acrrd (commented):
A couple of additional points:
- I was wandering if we should move the `Arc` inside `KeyPhrase`, something like `KeyPhrase(Arc<KeyPhraseInner>)` to avoid to have to change the signature of every method that use it.
- From my understanding the mentioned fixes are independent from what this PR is about and could they could be in their own.

discovery_engine_core/ai/xayn-ai/src/coi/key_phrase.rs
```diff
@@ -154,50 +172,64 @@ impl RelevanceMap {
...
         .into_iter()
         .enumerate()
         .map(|(i, lane)| {
             lane.iter()
                 .enumerate()
-                .filter_map(|(j, element)| (i != j).then(|| *element))
-                .reduce(reduce)
-                .map(finalize)
+                .filter_map(|(j, element)| (i != j).then(|| prepare(*element)))
+                .reduce(|x, y| reduce(x, y))
```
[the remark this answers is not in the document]
janpetschexain: this doesn't work because of
```
cannot move out of `reduce`, a captured variable in an `FnMut` closure
move occurs because `reduce` has type `R`, which does not implement the `Copy` trait
```
i removed the `Copy` bounds on the arguments to allow for more general closures than just function pointers. the small downside is that we have to type this slightly more redundant.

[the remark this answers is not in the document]
acrrd: ok, thanks for checking!

discovery_engine_core/ai/xayn-ai/src/ranker/system.rs
janpetschexain: i don't understand what you are trying to say. the new `KeyPhrases` are implementing ser/de. is the way how `KeyPhrases` implements ser/de not the correct one?

janpetschexain:
>     * I was wandering if we should move the `Arc` inside `KeyPhrase`, something like `KeyPhrase(Arc<KeyPhraseInner>)` to avoid to have to change the signature of every method that use it.

i'll change it as suggested.

>     * From my understanding the mentioned fixes are independent from what this PR is about and could they could be in their own.

they are somewhat independent, the updated tests revealed the bugs (which haven't been caught by the old tests before). we could fix them in a separate pr, but that would require us to impl wrong tests (or to comment them out). it seems preferable to fix them right away even though this means to add a little extra scope to the pr. i have no strong preference on this though.

discovery_engine_core/ai/xayn-ai/src/ranker/system.rs
acrrd: We are changing the state so we need to bump the version number for the de/serialization and migrate the content during deserialization

janpetschexain: i see. afaik we only have beta user currently, so do we actually need a full scale migration scheme?

i'll bump the version and add a outdated version handler for now. otherwise we would need to follow deep down the rabit hole and introduce versioning like `obake` for the `State` including an actual migration scheme from `RelevanceMap` to `KeyPhrases`.

acrrd:
> they are somewhat independent, the updated tests revealed the bugs (which haven't been caught by the old tests before). we could fix them in a separate pr, but that would require us to impl wrong tests (or to comment them out). it seems preferable to fix them right away even though this means to add a little extra scope to the pr as we have to touch the tests anyways. i have no strong preference on this though.
 
 Ok I see, we only knew after we modified the tests. I honestly don't understand how much the tests should be changed to get the error so it depends a bit but I think a good approach could have been to open a PR with the fix with new test to highlight the problem and base this one on that. But we don't have to do it right now.

janpetschexain:
> Ok I see, we only knew after we modified the tests. I honestly don't understand how much the tests should be changed to get the error so it depends a bit but I think a good approach could have been to open a PR with the fix with new test to highlight the problem and base this one on that. But we don't have to do it right now.

i guess we should write tests for all the subfunctions, otherwise it will be hard to notice. for example the result might be ordered correctly in the current test, but we would never notice that the values used to determine the order are slightly off like here. unfortunately, we didn't have time yet to extend the test coverage, but we should definitely do so in the near future.

discovery_engine_core/ai/xayn-ai/src/ranker/system.rs
janpetschexain: for outdated state versions i changed the error to the default state instead.

acrrd (approved):

janpetschexain:
bors merge

PR #300, opened 2022-04-06
[Merged by Bors] - Add value getter to Source

acrrd:
In dart `toString` should be used only to inspect the object for debug.
This PR add `value` to get the inner string value of the source.

### DIFF
--- a/discovery_engine/lib/src/domain/models/source.dart
+++ b/discovery_engine/lib/src/domain/models/source.dart
@@ -36,6 +36,8 @@ class Source extends Equatable {
   @override
   String toString() => _repr;
 
+  String get value => _repr;
+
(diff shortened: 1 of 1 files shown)


biernacki (approved):

discovery_engine/lib/src/domain/models/source.dart
```diff
@@ -36,6 +36,8 @@ class Source extends Equatable {
   @override
   String toString() => _repr;
 
+  String get value => _repr;
```
biernacki, line 39: We could also just make the `_repr` member public (maybe rename it also to `value`), but this is also fine

acrrd:
bors merge

PR #303, opened 2022-04-06
TY 2690 add license

acrrd:
Add license header to the files imported from xayn-ai. 
Exclude `rubert-tokenizer` from the copyright checks since the code there is forked from another project. In that copyright headers are not used and it is not clear how and to who we should attribute the copyright to.

I use the following snippet to extract the year from git.
```bash
for FILE in $(find ./ -iname "*.rs") 
do
  echo "$FILE $(git log --format=%ad --date=format:'%Y' $FILE | tail -1)"
done
```


janpetschexain (approved):
lgtm

PR #301, opened 2022-04-06
[Merged by Bors] - TY-2713 make market optional in endpoint query

finiteprods:
related #281 #283 #292

**Summary**

allow `market` to be optional in queries to the newscatcher endpoint.

PR #292, opened 2022-04-04
[Merged by Bors] - TY-2604 set favourite sources

finiteprods:
previously #281 #283

**Summary**

allows the favourite sources to be updated post-initialization.

PR #304, opened 2022-04-06
[Merged by Bors] - TY-2606 favourite news sources stack

finiteprods:
previously related #281 #283 #292 #301

**Summary**

adds a `FavouriteNews` struct, added to the discovery engine's stacks.
also covers TY-2605 (pass sources to favourite stack).

### DIFF
--- a/discovery_engine_core/core/src/stack/ops/favourite.rs
+++ b/discovery_engine_core/core/src/stack/ops/favourite.rs
@@ -12,26 +12,35 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-use std::sync::Arc;
+use std::{iter, sync::Arc};
 
 use async_trait::async_trait;
+use futures::stream::FuturesUnordered;
+use itertools::chain;
 use tokio::{sync::RwLock, task::JoinHandle};
+use uuid::Uuid;
 use xayn_ai::ranker::KeyPhrase;
 use xayn_discovery_engine_providers::{Article, Client, CommonQueryParts, HeadlinesQuery};
 
 use crate::{
-    document::{Document, HistoricDocument},
+    document::{dedup_documents, Document, HistoricDocument},
     engine::{EndpointConfig, GenericError},
-    stack::Id,
+    stack::{
+        filters::{filter_semantically, ArticleFilter, CommonFilter, SemanticFilterConfig},
+        Id,
+    },
 };
 
-use super::Ops;
+use super::{common::request_min_new_items, Ops};
 
 /// Stack operations customized for favourite news.
 pub(crate) struct FavouriteNews {
     client: Arc<Client>,
     sources: Arc<RwLock<Vec<String>>>,
     page_size: usize,
+    semantic_filter_config: SemanticFilterConfig,
+    max_requests: u32,
+    min_articles: usize,
 }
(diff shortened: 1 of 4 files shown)


discovery_engine_core/core/src/stack/ops/favourite.rs
```diff
@@ -0,0 +1,129 @@
...
+        Id,
+    },
+};
+
+use super::{common::request_min_new_items, Ops};
+
+/// Stack operations customized for favourite news.
+pub(crate) struct FavouriteNews {
+    client: Arc<Client>,
+    sources: Arc<RwLock<Vec<String>>>,
```
janpetschexain, line 39: 🎨 we could change this type to
```suggestion
    sources: Arc<RwLock<Arc<Vec<String>>>>,
```
(the endpoint config could already be changed to hold an `Arc<Vec<String>>`)

then in l.84 everytime we read the sources to get new items, the cloning will be cheap. i guess we should do the same for the excluded sources fields of the other stacks as well (maybe in another pr).

discovery_engine_core/core/src/stack/ops/favourite.rs
```diff
@@ -0,0 +1,129 @@
...
+            self.min_articles,
+            |request_num| {
+                let page = request_num as usize + 1;
+                let future = spawn_favourites_request(
+                    self.client.clone(),
+                    self.page_size,
+                    page,
+                    sources.clone(),
+                );
+                iter::once(future).collect::<FuturesUnordered<_>>()
```
janpetschexain, line 96: :art: currently the `requests_fn` closure of `request_min_new_items()` is required to return a `FuturesUnordered` (hence you had do collect one here). if you like to, we could try to change the return type of `requests_fn` to `impl futures::stream::Stream`, then we could use `futures::stream::once()` to return a `futures::stream::Once` which should have less overhead than a `FuturesUnordered` and also doesn't require collecting (i'm not sure if there are further caveats though).
```suggestion
                stream::once(future)
```

discovery_engine_core/core/src/stack/ops/favourite.rs
```diff
@@ -0,0 +1,129 @@
...
+            |articles| Self::filter_articles(history, stack, articles),
+        )
+        .await
+    }
+
+    fn merge(&self, stack: &[Document], new: &[Document]) -> Result<Vec<Document>, GenericError> {
+        let mut merged = chain!(stack, new).cloned().collect();
+        dedup_documents(&mut merged);
+        let filtered = filter_semantically(merged, &self.semantic_filter_config);
+        Ok(filtered)
```
janpetschexain, line 107: :art: newline
```suggestion

        Ok(filtered)
```

janpetschexain (approved):
lgtm, some optional suggestions

discovery_engine_core/core/src/stack/ops/favourite.rs
finiteprods: I agree this would be better in a separate pr

discovery_engine_core/core/src/stack/ops/favourite.rs
finiteprods: will open a pr for this and can be discussed further there

finiteprods:
bors merge

finiteprods:
bors merge

PR #305, opened 2022-04-07
[Merged by Bors] - TY-2589 Updated to tract 0.16.3 and use a patched tract-linalg. [1]

rustonaut:
Reviewers should probably also take a look at the patch branch: 

https://github.com/xaynetwork/tract/compare/v0.16.3...xaynetwork:android-relocation-patch

Be aware that the solution implemented by sonos/tract will likely be somewhat different
(i.e. only changing the asm for android, maybe also using different asm).

This was tested with `cross` for the non-neon implementation and with `cargo dinghy`
on the armv7a phone we have for the neon implementation.

**References:**

- [TY-2589](https://xainag.atlassian.net/browse/TY-2589)

### DIFF
--- a/justfile
+++ b/justfile
@@ -228,8 +228,7 @@ _compile-android target:
     cd "$RUST_WORKSPACE"; \
         cargo ndk --bindgen -t $(echo "{{target}}" | sed 's/[^ ]* */&/g') -p $ANDROID_PLATFORM_VERSION \
         -o "{{justfile_directory()}}/$FLUTTER_WORKSPACE/android/src/main/jniLibs" build \
-        --release \
-        -p xayn-discovery-engine-bindings --locked
+        --release -p xayn-discovery-engine-bindings --locked
 
 compile-android-local: _codegen-order-workaround
     #!/usr/bin/env sh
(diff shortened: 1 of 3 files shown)


justfile
```diff
@@ -228,7 +228,8 @@ _compile-android target:
     cd "$RUST_WORKSPACE"; \
         cargo ndk --bindgen -t $(echo "{{target}}" | sed 's/[^ ]* */&/g') -p $ANDROID_PLATFORM_VERSION \
         -o "{{justfile_directory()}}/$FLUTTER_WORKSPACE/android/src/main/jniLibs" build \
-        --release -p xayn-discovery-engine-bindings --locked
+        --release \
+        -p xayn-discovery-engine-bindings --locked
```
janpetschexain, line 232: :art: if we split the line by arguments then we should do so completely
```suggestion
        --release \
        -p xayn-discovery-engine-bindings \
        --locked
```
this would also apply for l.229

janpetschexain (commented):
> Reviewers should probably also take a look at the patch branch: 
> https://github.com/xaynetwork/tract/compare/v0.16.3...xaynetwork:android-relocation-patch

looks about right, but my assembler knowledge is way to basic, so another reviewer is definitely required

justfile
janpetschexain: done, also rebased to resolve merge conflicts.

acrrd (approved):

janpetschexain:
bors merge

PR #306, opened 2022-04-07
[Merged by Bors] - TY-2589 check android so in ci [3]

rustonaut:
**Requires:**

- https://github.com/xaynetwork/xayn_docker/pull/8
- #305 

**References:**

- [TY-2589](https://xainag.atlassian.net/browse/TY-2589)

### DIFF
--- a/.github/scripts/check_android_so.sh
+++ b/.github/scripts/check_android_so.sh
@@ -1,0 +1,57 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+ERROR_COUNT=0
+
+function error() {
+    ((ERROR_COUNT++))
+    echo "$@" 1>&2
+}
+
+function warn() {
+    echo "$@" 1>&2
+}
+
+function check_so() {
+    SO_FILE="$1"
+    local OLD_ERROR_COUNT="$ERROR_COUNT"
+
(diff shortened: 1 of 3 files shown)


.github/scripts/check_android_so.sh
```diff
@@ -0,0 +1,57 @@
...
+        error "Shared object must be a file."
+        return
+    fi
+
+    SCAN_RESULT="$(scanelf -qT "$SO_FILE")"
+
+    if [ "$SCAN_RESULT" != "" ]; then
+        error "RELOCATIONS in: '$SO_FILE'"
+        warn "$SCAN_RESULT"
+    fi
```
rustonaut, line 30: This is the relevant check.

The other check blow is still relevant but less pressing.

x3ro (approved):
LGTM 🚀 

🎨  The only thing that stands out to me is that I don't like abbreviating `so`, as I think it'd be hard for me to understand what it stands for without context. For example, in `check-android-so`. However, I'm also failing to come up with a much better name that's not super annoying to type. So, if possible, maybe add a comment to the Justfile that shortly outlines what's happening in that task?

janpetschexain:
bors merge

PR #314, opened 2022-04-11
[Merged by Bors] - TY-2658 replace real domains

janpetschexain:
**References**

- [TY-2658]

**Summary**

- replace real domains with example domains

[TY-2658]: https://xainag.atlassian.net/browse/TY-2658?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


discovery_engine/test/feed_manager_test.dart
```diff
@@ -247,25 +247,26 @@ Future<void> main() async {
...
 
       final excludedSoures = {
-        Source('theguardian.com'),
-        Source('bbc.co.uk'),
-        Source('wsj.com'),
-        Source('www.nytimes.com'),
+        Source('example.com'),
+        Source('example.org'),
+        Source('example.net'),
+        Source('www.example.info'),
```
acrrd, line 269: :wrench: This is a valid domain. a part from `example.com`, `example.org`, `example.net`, other domain must in the form `*.example`, unless a standard is saying that they should be used only for documentation/testing pupose.

discovery_engine_core/core/test-fixtures/articles-some-malformed-media-urls.json
```diff
@@ -18,7 +18,7 @@
     ],
     "media": "",
     "is_opinion": false,
-    "twitter_account": "@XaynHQ",
+    "twitter_account": "@Example",
```
acrrd, line 21: :wrench: This is a valid twitter account. I guess we can use `xaynhq` here. I'm not sure if there is some handle reserved for this use case.

discovery_engine/test/feed_manager_test.dart
janpetschexain: good point. i'll also change other occurences like example.de

discovery_engine_core/core/test-fixtures/articles-some-malformed-media-urls.json
janpetschexain: well that is unexpected :O looks like the twitter api doesn't provide a dummy handle, i'll revert this change.

acrrd (approved):

janpetschexain:
bors merge

PR #315, opened 2022-04-11
TY-2718 breaking news error if empty key phrases

janpetschexain:
**References**

- [TY-2718]

**Summary**

- constify stack ids
- fail `update_stacks()` if the `BreakingNews` stack failed and the key phrases for the `PersonalizedNews` stack are empty
- update integration test


discovery_engine_core/core/src/engine.rs
```diff
@@ -548,13 +549,31 @@ async fn update_stacks<'a>(
 
     let mut errors = Vec::new();
     for stack in &mut stacks {
+        // bail if breaking news failed and key phrases for personalized news are empty
+        macro_rules! bail_if_breaking_news_failed_without_key_phrases {
```
janpetschexain, line 553: how are we going to handle this once #304 introduced `TrustedNews` which also doesn't depend on key phrases? i guess we should only fail if all stacks which don't need key phrases failed and the key phrases are empty?

acrrd (commented):
:wrench: The engine should not know what the stacks are. 
This can be fixed by returning an error if key_phrases is empty in `personalized.rs` since we cannot get new items without them. Optionally we can retain only the stacks for which `needs_key_phrases` is false if `key_phrases` is empty in `update_stacks`.

janpetschexain:
> 🔧 The engine should not know what the stacks are. This can be fixed by returning an error if key_phrases is empty in `personalized.rs` since we cannot get new items without them. Optionally we can retain only the stacks for which `needs_key_phrases` is false if `key_phrases` is empty in `update_stacks`.

i'll go for the second suggestion, this will probably scale better once we add new stacks like `TrustedNews`

discovery_engine_core/core/src/engine.rs
```diff
@@ -545,6 +545,14 @@ async fn update_stacks<'a>(
         .any(|stack| stack.ops.needs_key_phrases())
         .then(|| ranker.select_top_key_phrases(select_top))
         .unwrap_or_default();
+    if key_phrases.is_empty() {
+        // only stacks which don't need key phrases remain. eventually, if they all fail, then an
+        // error is returned. if there are no stacks left to be updated, then a success is returned.
+        stacks = stacks
```
acrrd, line 551: :art: We can use `retain`, we avoid an allocation. Even if a bit of memory will be wasted we have only references here so is not much.

discovery_engine_core/core/src/stack/ops/breaking.rs
```diff
@@ -70,10 +69,15 @@ impl BreakingNews {
     }
 }
 
+/// 1ce442c8-8a96-433e-91db-c0bee37e5a83 (valid uuid)
+pub(crate) const BREAKING_NEWS_ID: Id = Id::from_bytes([
```
acrrd, line 73: We don't need this, hopefully we will never have to inspect the id of the stack after creation and when uuid 1.0 will be released we can use `uuid!` that will parse the string and create the uuid at compile time. At the moment there is an alpha version.

janpetschexain: i'll revert this, it's a leftover of the first impl iteration.

acrrd (approved):

PR #316, opened 2022-04-13
[Merged by Bors] - TY-2710 macos runner

Robert-Steiner:
Ticket:
- [TY-2710]

**Job dependencies**

- dotenv (npm action)
- flutter (action installs it in the runner tool cache dir)
- just (action installs in it the runner tool cache dir)
- async_bindgen (is installed in the workspace dir)
- rustup toolchain + targets (is installed in the runner tool cache dir)
- cocoapods (is installed in the runner tool cache dir)

**Installed on the runner and shared between them**

- `xcode llvm`
	- dart ffi generation (in ios-flutter-example) requires `llvm`

**Data removed after each run** 

- workspace (`${{ github.workspace }}`)
- `$RUNNER_TOOL_CACHE`
  - rustup and cargo
    - (`$RUNNER_TOOL_CACHE/.rustup`) and (`$RUNNER_TOOL_CACHE/.cargo`)
  - `just` (`$RUNNER_TOOL_CACHE/just`)
  - `flutter` (`$RUNNER_TOOL_CACHE/flutter`)
  - `cocoapods` (`$RUNNER_TOOL_CACHE/.gem`)
- `$RUNNER_TEMP` (folder where the build artifacts will be downloaded) 

_deps that we can cache between runs and only remove them after some time (maybe by a cron job on the runner)_

- `just` (`$RUNNER_TOOL_CACHE/just`)
- `flutter` (`$RUNNER_TOOL_CACHE/flutter`)
- `cocoapods` (`$RUNNER_TOOL_CACHE/.gem`)

**Some small optimization i tried**

- `--default-toolchain none` when installing `rustup`. the problem is that the installation of `xayn_async_bindgen` is called in the root of the repo and here we don't have a `rust-toolchain` file which means `rustup`/`cargo` fails because it can't find a default toolchain.

Release CI
- https://github.com/xaynetwork/xayn_discovery_engine/actions/runs/2161195883

Bors
- https://github.com/xaynetwork/xayn_discovery_engine/actions/runs/2161090771

 checked the release artefact (x86_64) on the iphone simulator

[TY-2710]: https://xainag.atlassian.net/browse/TY-2710?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Robert-Steiner:
bors try

.github/actions/setup-job-macos/action.yml
```diff
@@ -0,0 +1,94 @@
...
+    - name: Install rust target
+      if: inputs.rust-target != ''
+      working-directory: ${{ env.RUST_WORKSPACE }}
+      shell: bash
+      run: rustup target add ${{ inputs.rust-target }}
+
+    - uses: Swatinem/rust-cache@d12701459954fec471b2d34cdf7ea3374b026383 # v1
+      with:
+        key: ${{ inputs.cache-key }}
+        working-directory: ${{ env.RUST_WORKSPACE }}
```
biernacki, line 94: ❔ sorry if it's a stupid question, but it will help me to understand :) what this cache is for? for builds only or for dependecies as well? isn't it better to put this step before installing dependencies?

Also, don't we want to add a cache for flutter too?

.github/workflows/_reusable.build-flutter-example.yml
```diff
@@ -13,15 +13,19 @@ permissions:
...
 jobs:
   flutter-build-example-ios:
-    runs-on: macos-11
+    runs-on: hetzner-mac-pm
     timeout-minutes: 30
     steps:
+      - name: Clean up workspace
+        shell: bash
+        working-directory: ${{ github.workspace }}
+        run: rm -rf *
```
biernacki, line 22: ❔ also just to help me understand.. why we need this? it this because of that "leftover" issue between runs that you've mentioned during planning?

biernacki (commented):
Looks good, left just a couple of questions

.github/actions/setup-job-macos/action.yml
```diff
@@ -0,0 +1,94 @@
...
+  github_token:
+    description: 'The github token used for the `just` action to avoid exceeding API rate limits'
+    required: true
+runs:
+  using: "composite"
+  steps:
+    - name: Clean up runner tool cache and temp dir
+      shell: bash
+      run: |
+        cd $RUNNER_TOOL_CACHE
```
x3ro, line 25: 🔧  We should verify here that `RUNNER_TOOL_CACHE` and `RUNNER_TEMP` aren't empty. If they were empty for whatever reason, we'd be deleting the entire home directory of the M1 Mac

🤔 Optimally we would also check whether they are subdirectories of the runner main directory.

.github/actions/setup-job-macos/action.yml
```diff
@@ -0,0 +1,94 @@
...
+
+    - name: Setup flutter dependencies
+      if: inputs.flutter == 'true'
+      shell: bash
+      run: just flutter-deps
+
+    - if: inputs.flutter == 'true'
+      shell: bash
+      run: |
+        echo "GEM_HOME=$RUNNER_TOOL_CACHE/.gem" >> $GITHUB_ENV
```
x3ro, line 54: 🎨 I would merge this with the other path-setting code below. We don't gain anything, runtime wise, by keeping them apart / conditional, and it'll provide a better overview of the environment (imo).

.github/workflows/_reusable.build-flutter-example.yml
x3ro, line 22: 🤔  Similar to what I wrote above. I don't know if "working directory" accepts an empty value or what it does with it, but I suppose it's worth making sure we're not deleting things we don't want to delete, like the home dir.

.github/workflows/_reusable.build-ios.yml
```diff
@@ -15,21 +15,26 @@ on:
...
     timeout-minutes: 60
     strategy:
       matrix:
         target: ["aarch64-apple-ios", "x86_64-apple-ios"]
     outputs:
       artifact-dir-base: ${{ steps.build.outputs.artifact-dir-base }}
     steps:
+      - name: Clean up workspace
+        shell: bash
+        working-directory: ${{ github.workspace }}
```
x3ro, line 28: 🤔 See above.

discovery_engine/ffigen.yaml
```diff
@@ -9,3 +9,10 @@ headers:
 compiler-opts-automatic:
   macos:
     include-c-standard-library: false
+# Can be removed once the github runner is available for the aarch64 on the M1 +
+# https://github.com/dart-lang/ffigen/pull/278 has been released.
+# Because the github runner runs on rosetta (x86_64) we can't use the homebrew
+# llvm version (aarch64) that is installed on the M1. However xcode is available
+# for both architectures so we can use its llvm version.
+llvm-path:
+  - '/Library/Developer/CommandLineTools/usr/lib/libclang.dylib'
```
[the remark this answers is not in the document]
x3ro: 🌱  It's possible to install a x64 version of homebrew on M1 macs, but that's just FYI. I don't think it's worth looking into here.

x3ro (changes requested):

.github/actions/setup-job-macos/action.yml
Robert-Steiner: > sorry if it's a stupid question

there are no stupid questions :) 

> what this cache is for? for builds only or for dependecies as well?

the rust cache caches the build artefacts and the dependencies (but not the installed rust-toolchain)

> isn't it better to put this step before installing dependencies?

good point otherwise it could happen that the deps of the `rust-deps` will be overwritten

.github/workflows/_reusable.build-flutter-example.yml
Robert-Steiner: yes 👍
If you use the github-hosted runner you always get a runner with a clean state. What I say now might not 100% accurate but Github runs its runner in a VM and restores the VM to a clean state after a job completes. Our self-hosted runner, on the other hand, run on the host without isolation. This means that we have to "manually" clean some parts of the runner between jobs. At the moment there is no easy way to clean up the jobs automatically via the runner. I think that's because Github runs the runner in the VM and therefore doesn't really need this feature. But Github now provides a way to run scripts before and after a job, so later in the future we can at least move that specific runner detail into such a script. The reason we only need to clean the workspace on the mac runner is because on linux we run the jobs in docker containers. So we just start with a clean container, run the steps and remove it after the job is done. There is one exception, however, that we have to empty the RUNNER_TOOL_CACH and RUNNER_TEMP folders, as these are mounted on the host system and are not automatically cleaned up after a job.

.github/actions/setup-job-macos/action.yml
Robert-Steiner: I prefer to keep them separate because I think it's easier to understand which tool belongs to which toolchain

Robert-Steiner:
bors try

x3ro (approved):

x3ro:
LGTM! thanks for finishing this, and have a great weekend 🎉

.github/actions/setup-job-macos/action.yml
Robert-Steiner: I haven't found in the documentation whether the string can be empty or whether it is never empty :/
> If they were empty for whatever reason, we'd be deleting the entire home directory of the M1 Mac

In this case I think we should tweak the runner's permissions on the host so if this happens it will fail due to insufficient permissions. wdyt?

For know I have added a check that the last dir name is `_tool` or `_temp`. Do you think that is enough?

.github/workflows/_reusable.build-flutter-example.yml
Robert-Steiner: We can actually remove `working-directory` because the default value is the path of the workspace.  😅

Robert-Steiner:
bors try

.github/actions/setup-job-macos/action.yml
x3ro: It could also be empty due to a bug or an error on our side, so even if the documentation says no, it would make sense to check it I think. 

You're right in that more granular permissions would be nice, and we could look into doing that. However, I fear it might not be as simple as it'd be on linux. At least I have bad experiences with multi-user macOS, so maybe that's just why I'm skeptical 😅

It just occurred to me that, for good measure, we could throw in `set -euo pipefail` ([bash 'safe mode'](http://redsymbol.net/articles/unofficial-bash-strict-mode/)). That should catch problems. Sorry this occurred to me a little late :D

Robert-Steiner: Good idea 👍 i have added the `-u` option. `-eo pipefail` are already set by the runner https://github.com/xaynetwork/xayn_discovery_engine/runs/6025070446?check_suite_focus=true#step:4:30

Robert-Steiner:
bors try

biernacki (approved):
🚀

Robert-Steiner:
bors merge

Robert-Steiner:
bors merge

Robert-Steiner:
bors merge

PR #319, opened 2022-04-13
TY-2670 Implement gnews

x3ro:
[TY-2670]

### What did I do?

* Introduce a common representation for an article, which closely maps to the the `NewsResource` we're already using in `core/src/document.rs`.
    * moved `excerpt_or_title` to this common representation, as `snippet_or_title`
* Implement the Gnews `Client`, `Response` and `Article` and `NewsQuery`.
    * `Client` takes `NewsQuery` in order to query the endpoint, returning a `Response` containing `Article`s.
* I wanted to avoid changing all of the fixtures based on the newscatcher representation, but also did not want to continue to leak the newscatcher-specific `Article` so I introduced `Article::load_from_newscatcher_json_representation` which reads a JSON in newscatcher format and returns a generic `Article`. 
* Introduced the `direct_news_client_access` feature, only intended to be used for `tooling/`

Apologies, this PR unfortunately ended up being a bit larger than I would've hoped, especially because git didn't pick up on the fact that I renamed `client.rs` to `newscatcher_client.rs`.

[TY-2670]: https://xainag.atlassian.net/browse/TY-2670?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


discovery_engine_core/providers/src/newscatcher_client.rs
```diff
@@ -0,0 +1,409 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+//! Client to get new documents.
+
+use std::{ops::Deref, time::Duration};
```
x3ro, line 17: Git picks this up as a new file, but this is just the previous `client.rs` renamed to `newscatcher_client.rs`

discovery_engine_core/providers/src/client.rs
```diff
@@ -46,383 +49,141 @@ pub enum Error {
...
-
-        let mut query = url.query_pairs_mut();
-        if !self.trusted_sources.is_empty() {
-            query.append_pair("sources", &self.trusted_sources.join(","));
-        };
-        if let Some(topic) = self.topic {
-            query.append_pair("topic", topic);
+impl From<GnewsArticle> for Article {
+    fn from(source: GnewsArticle) -> Self {
+        let source_domain = Url::parse(&source.url)
```
x3ro, line 126: Not sure if we need better handling for this. If the article URL is invalid, returning an empty source domain isn't optimal. However, our other filtering steps should definitely catch articles with invalid URLs, so this article should be filtered out down the line anyway 🤔

discovery_engine_core/tooling/bin/news_client.rs
```diff
@@ -23,11 +35,63 @@ async fn main() -> Result<()> {
                   The token can be found in 1Password",
     )?;
 
+    let args: Vec<String> = env::args().collect();
+    let mode = args
+        .get(1)
+        .map(|s| s.to_owned())
+        .unwrap_or_else(|| "".to_string());
```
janpetschexain, line 42: 🔧 could be simplified
```suggestion
        .unwrap_or_default();
```

discovery_engine_core/tooling/bin/news_client.rs
```diff
@@ -23,11 +35,63 @@ async fn main() -> Result<()> {
...
+
+    if mode == "newscatcher" {
```
janpetschexain, line 44: :art: limited number of if-else branches is usually ok, though this might look a bit cleaner with a `match mode {}`

janpetschexain (commented):
looks good, i left some suggestions

discovery_engine_core/providers/src/gnews_client.rs
```diff
@@ -0,0 +1,184 @@
...
+    pub page: usize,
+    /// Exclude given sources.
+    pub excluded_sources: &'a [String],
+    /// Search filter for selecting the results
+    pub filter: &'a Filter,
+}
+
+/// Client that can provide documents.
+pub struct Client {
+    token: String,
```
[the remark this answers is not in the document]
x3ro: As discussed w/ @janpetschexain, this is not currently worth the effort, given that the token is embedded in the binary anyway, and thus will inevitably end up in memory.

Robert-Steiner:
There is a small issue with the feature flag. When building/checking a single package (e.g. `cargo check -p xayn-discovery-engine-binding`) the compiler prints warnings regarding `unreachable pub item` and `unused code`. Because of that the release CI will fail. We can discuss on the next daily what to do here

acrrd (commented):
Some of the comments are stuff that are due to changes I did, but I didn't do them in all the places, sorry!

To fix the dart tests we need to return data from the proper directory depending on the provider.

discovery_engine_core/providers/src/gnews.rs
```diff
@@ -0,0 +1,63 @@
...
+    #[serde(deserialize_with = "deserialize_null_default")]
+    pub url: String,
+
+    #[serde(deserialize_with = "deserialize_null_default")]
+    pub image: String,
+
+    #[serde(deserialize_with = "deserialize_null_default")]
+    pub source: Source,
+
+    #[serde(rename(deserialize = "publishedAt"), alias="date_published")]
```
acrrd, line 42: We can now remove the alias since it as been fixed on the backend.

discovery_engine_core/providers/src/gnews.rs
```diff
@@ -0,0 +1,63 @@
...
+    pub published_at: DateTime<Utc>,
```
acrrd, line 43: We should handle this field as possibly missing or null like for the other provider.

discovery_engine_core/providers/src/gnews_client.rs
```diff
@@ -0,0 +1,235 @@
...
+    /// Run a query for fetching `Article`s
+    pub async fn query_articles(&self, query: &NewsQuery<'_>) -> Result<Vec<Article>, Error> {
+        let url = self.build_news_url(query)?;
+        self.query(url).await.map(|news| news.articles)
+    }
+
+    /// Run a query for fetching `Article`s
+    pub async fn query_headlines(&self, query: &HeadlinesQuery<'_>) -> Result<Vec<Article>, Error> {
+        let url = self.build_headlines_url(query)?;
+        self.query(url).await.map(|news| news.articles)
```
acrrd, line 99: We can set `country` and `lang` of each article from the market in the query.

x3ro:
This is not intended to be merged at the moment, since there has been no decision on how to move forward with gnews.

Robert-Steiner:
Rebase on the lastest main commit. It might have outdated some comments. We should review them them once we get back to this pr.

acrrd:
bors cancel

acrrd:
We don't plan to proceed on this.

PR #317, opened 2022-04-13
[Merged by Bors] - TY-2723 rename to trusted news sources

finiteprods:
previously related #281 #283 #292 #301 #304

**Summary**

renames "favourite" to "trusted" sources / news

PR #320, opened 2022-04-14
[Merged by Bors] - TY-2605 initialize trusted stack

finiteprods:
previously related #281 #283 #292 #301 #304 #317

**Summary**

Initializes the trusted stack for engine initialization.

note: a dart integration test fails if the trusted stack is added as the last of the stacks.
the test should be fixed, perhaps separately.


finiteprods:
bors merge

PR #322, opened 2022-04-14
[Merged by Bors] - Only return errors when there are some

acrrd:
It could happen that we have no stacks and no errors, in that case an error with an empty list of error was returned.
This PR change the condition to make it a bit more explicit on the cases in which we want to return an error.

### DIFF
--- a/discovery_engine_core/core/src/engine.rs
+++ b/discovery_engine_core/core/src/engine.rs
@@ -602,10 +602,10 @@ async fn update_stacks<'a>(
     }
 
     // only return an error if all stacks failed
-    if errors.len() < stacks.len() {
-        Ok(())
-    } else {
+    if !errors.is_empty() && errors.len() >= stacks.len() {
         Err(Error::Errors(errors))
+    } else {
+        Ok(())
     }
 }
 


discovery_engine_core/core/src/engine.rs
```diff
@@ -602,10 +602,10 @@ async fn update_stacks<'a>(
     }
 
     // only return an error if all stacks failed
-    if errors.len() < stacks.len() {
-        Ok(())
-    } else {
+    if !errors.is_empty() && errors.len() >= stacks.len() {
```
x3ro, line 605: The implication here is that we return an error if all stacks returned an error, right? Without having looked at the rest of the code, it's not possible for a stack to return more than one error, right?

acrrd: is not possible, as soon as we have an error we log it, add it and we break out of the loop.

acrrd: The `>` part in the check is to avoid to leave unexpected cases out more then something that is really needed with the current code.

acrrd:
bors merge

PR #325, opened 2022-04-19
TY-2768 exclude trusted stack

finiteprods:
**Summary**

temporarily exclude the stack added in #320 
due to errors occurring when it uses the headlines endpoint


acrrd:
#327 seems to fix so we can keep it

PR #327, opened 2022-04-19
[Merged by Bors] - Don't do query if there are no trusted sources

acrrd:


x3ro (approved):
LGTM! 🚀

finiteprods (approved):

acrrd:
bors merge

acrrd:
bors cancel

acrrd:
For how our code works at the moment we cannot return an empty list if we don't have trusted sources because otherwise errors from stacks that can run will never returned because we have at least one stack that can run correctly.
We need to rework what we are doing here. In the mean while we return an error.

x3ro:
Makes sense. Do we report the errors from individual stacks somewhere, even if others succeed? 🤔 Just curious if this would flood error logs somewhere :D

acrrd:
bors merge

acrrd:
> Makes sense. Do we report the errors from individual stacks somewhere, even if others succeed? thinking Just curious if this would flood error logs somewhere :D

Yes they are sent to logcat on android. We don't run the update that often to flood them but we will send it every time till the user add a trusted source.

PR #331, opened 2022-04-20
TY-2755 SQLite in rust demo

biernacki:


acrrd:
@x3ro please ignore the request, was only a test for the automatic assignment

PR #335, opened 2022-04-21
[Merged by Bors] - Handle stacks that are not ready to ask new data

acrrd:
If a stack is not ready to return data we don't count it as successful or failed stacks.

At the moment if `TrustedSources` has no  sources and is the only stack that required to be updated we are returning an error but we should not.

This PR allows a stack to return `NotReady` as a variant of the error, if a stack is not ready we exclude it from the count of stacks w.r.t. the return of errors.

Having `new_items` returning something like
```rust
enum NewItems {
  NotReady,
  Items(Vec<Article>),
  Err(GenericError),
}
```
would be nicer but for how we are handling errors in the stack and engine it will require more changes.

### DIFF
--- a/discovery_engine_core/core/src/stack/ops/trusted.rs
+++ b/discovery_engine_core/core/src/stack/ops/trusted.rs
@@ -15,10 +15,8 @@
 use std::{iter, sync::Arc};
 
 use async_trait::async_trait;
-use displaydoc::Display;
 use futures::stream::FuturesUnordered;
 use itertools::chain;
-use thiserror::Error;
 use tokio::{sync::RwLock, task::JoinHandle};
 use uuid::Uuid;
 use xayn_ai::ranker::KeyPhrase;
@@ -33,7 +31,7 @@ use crate::{
     },
 };
 
-use super::{common::request_min_new_items, Ops};
+use super::{common::request_min_new_items, NewItemsError, Ops};
 
 /// Stack operations customized for trusted news.
 pub(crate) struct TrustedNews {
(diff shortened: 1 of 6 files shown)


discovery_engine_core/core/src/stack/ops/trusted.rs
```diff
@@ -89,10 +81,10 @@ impl Ops for TrustedNews {
         _key_phrases: &[KeyPhrase],
         history: &[HistoricDocument],
         stack: &[Document],
-    ) -> Result<Vec<Article>, GenericError> {
+    ) -> Result<Vec<Article>, NewItemsError> {
         let sources = Arc::new(self.sources.read().await.clone());
         if sources.is_empty() {
-            return Err(Box::new(Error::NoTrustedSources));
+            return Err(NewItemsError::NotReady);
```
rustonaut, line 87: :grey_question: Why couldn't we just return `return Ok(vec![]);`

janpetschexain: this should indeed result in the same logical behavior in `update_stacks()` as this is basically what we did in the personalized news stack before. but this wouldn't allow us to differentiate why the articles are empty, either if they are actually empty or if `new_items()` failed for specific reasons, like missing key phrases. imo it's clearer to encode this in various errors and then decide if/how to succeed/fail based on the specific error, like continue on `NotReady`.

rustonaut (approved):
lgtm

discovery_engine_core/core/src/stack/ops/trusted.rs
acrrd: We are not returning an empty list because that will count as a stack that correctly did a call but there was no data to return. If we return an empty array when we are not able to request new data we will count that as a stack that ended successfully and if, for example, there is no network we will not return an error at all from other stacks that failed.

acrrd:
bors merge

PR #339, opened 2022-04-25
[Merged by Bors] - TY-2692 persist active search by topic

finiteprods:
previously #330 added a `searchBy` field to `ActiveSearch`

**Summary**

support persistence of the current topic-based (or query-based) search to the `ActiveSearchRepository`.
in particular, migrate an "old" active search to the new-style (possibly topic-based) active search.

hive reference for [updating a class](https://docs.hivedb.dev/#/custom-objects/generate_adapter?id=updating-a-class).

PR #330, opened 2022-04-20
[Merged by Bors] - TY-2691 topic search events

finiteprods:
**Summary**

extends `ClientEvent.searchRequested` to be _either_ a search query term or a topic.
extends the `ActiveSearch` class accordingly.

PR #343, opened 2022-04-27
[Merged by Bors] - TY-2722 method on engine to search by topic

finiteprods:
previously #330 #339

**Summary**

adds a method to the engine for searching by topic. also:

- adds ffi binding for above
- corresponding changes on the dart side to accommodate new engine method
- search manager calls the new method when search requested is by topic

### DIFF
--- a/discovery_engine/lib/src/domain/models/active_search.dart
+++ b/discovery_engine/lib/src/domain/models/active_search.dart
@@ -26,7 +26,7 @@ part 'active_search.g.dart';
 class ActiveSearch with _$ActiveSearch {
   @HiveType(typeId: searchTypeId)
   const factory ActiveSearch({
-    @HiveField(0) required String queryTerm,
+    @HiveField(0) required String searchTerm,
     @HiveField(1) required int requestedPageNb,
     @HiveField(2) required int pageSize,
     @HiveField(3, defaultValue: SearchBy.query) required SearchBy searchBy,
--- a/discovery_engine/lib/src/domain/search_manager.dart
+++ b/discovery_engine/lib/src/domain/search_manager.dart
@@ -21,6 +21,8 @@ import 'package:xayn_discovery_engine/src/domain/engine/engine.dart'
     show Engine;
 import 'package:xayn_discovery_engine/src/domain/event_handler.dart'
     show EventConfig;
+import 'package:xayn_discovery_engine/src/domain/models/active_data.dart'
+    show DocumentWithActiveData;
 import 'package:xayn_discovery_engine/src/domain/models/active_search.dart'
     show ActiveSearch, SearchBy;
 import 'package:xayn_discovery_engine/src/domain/models/document.dart'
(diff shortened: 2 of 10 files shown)


discovery_engine/lib/src/domain/models/active_search.dart
```diff
@@ -26,7 +26,7 @@ part 'active_search.g.dart';
 class ActiveSearch with _$ActiveSearch {
   @HiveType(typeId: searchTypeId)
   const factory ActiveSearch({
-    @HiveField(0) required String queryTerm,
+    @HiveField(0) required String searchTerm,
```
Robert-Steiner, line 29: ❔I know too little about the hive db but this change doesn't require any mirgration, does it?

discovery_engine/lib/src/domain/search_manager.dart
```diff
@@ -100,7 +113,7 @@ class SearchManager {
     await searchClosed();
 
     final search = ActiveSearch(
-      queryTerm: queryTerm,
+      searchTerm: queryTerm,
```
Robert-Steiner, line 116: ❔Should we also rename `queryTerm` to `searchTerm`?

discovery_engine_core/core/src/engine.rs
```diff
@@ -457,6 +464,66 @@ where
...
+        }
+
+        let mut errors = Vec::new();
+        let mut articles = Vec::new();
+
+        let markets = self.config.markets.read().await;
+        let scaled_page_size = page_size as usize / markets.len() + 1;
+        let excluded_sources = self.config.excluded_sources.read().await.clone();
+        for market in markets.iter() {
+            let news_query = HeadlinesQuery {
```
Robert-Steiner, line 485:
```suggestion
            let headlines_query = HeadlinesQuery {
```

discovery_engine_core/core/src/engine.rs
```diff
@@ -457,6 +464,66 @@ where
...
     }
 
+    /// Perform an active search by topic.
+    pub async fn search_by_topic(
+        &mut self,
+        topic: &str,
+        page: u32,
+        page_size: u32,
+    ) -> Result<Vec<Document>, Error> {
+        if topic.trim().is_empty() {
```
Robert-Steiner, line 474: 🌱 Except for the query, both methods ( `active_search` and `search_by_topic`) share the same code. Maybe it makes sense to refactor the code a bit in a follow-up future.

Robert-Steiner (approved):
🚀

discovery_engine/lib/src/domain/models/active_search.dart
finiteprods: good question. according to the hive docs, renaming (while keeping the `HiveField` number the same) should be fine.

biernacki: yes, it should be fine, field name is not important as long as it uses the same `HiveField` identifier.

finiteprods:
bors merge

PR #347, opened 2022-04-28
[Merged by Bors] - Expose SearchBy enum

acrrd:
Expose SearchBy since it appears in the public interface of ActiveSearch but is not exported.


acrrd:
bors merge

PR #349, opened 2022-05-02
[Merged by Bors] - TY-2743 clean sources csv

janpetschexain:
**References**

- [TY-2743]

**Summary**

- add script to clean the sources csv file from empty names and irrelevant fields
- the cleaned sources csv file is attached to the task

[TY-2743]: https://xainag.atlassian.net/browse/TY-2743?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine_core/tooling/bin/clean_sources.rs
+++ b/discovery_engine_core/tooling/bin/clean_sources.rs
@@ -1,0 +1,119 @@
+// Copyright 2022 Xayn AG
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, version 3.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+use std::{
+    fs::{remove_file, File},
+    io::{BufRead, BufReader, BufWriter, Write},
+    path::Path,
+};
+
+use anyhow::{bail, Result};
+use csv::{QuoteStyle, ReaderBuilder, WriterBuilder};
+
+fn main() -> Result<()> {
+    let in_file = "sources_filled.csv";
+    let tmp_file = "sources_fixed.csv";
+    let out_file = "sources_cleaned.csv";
+
+    fix_sources(in_file, tmp_file)?;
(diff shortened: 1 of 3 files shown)


discovery_engine_core/tooling/bin/clean_sources.rs
```diff
@@ -0,0 +1,119 @@
...
+    let mut writer = BufWriter::new(File::create(out_file)?);
+
+    let mut broken_line = String::new();
+    for line in reader.lines() {
+        let line = line?;
+        if line.split(';').count() == 4 {
+            writer.write_all(line.as_bytes())?;
+            writer.write_all(&[b'\n'])?;
+        } else {
+            broken_line.push_str(&line);
```
Robert-Steiner, line 49: 🔧 It looks like that the illegal linebreaks mostly occur in the name of the record.
e.g.

```
Aktuelles Programm
- SWR Fernsehen;swrfernsehen.de;False;47911
```

The problem is that the next line contains three `;` so we only keep `- SWR Fernsehen;swrfernsehen.de` and `broken_line` will just append the lines that contain the linebreak.

Robert-Steiner (commented):
looks good! there is only one small issue with the linebreaks

discovery_engine_core/tooling/bin/clean_sources.rs
janpetschexain: good catch 👍 i'll change it such that we merge consecutive lines if there have been broken lines before, this should fit the current issue.

discovery_engine_core/tooling/bin/clean_sources.rs
```diff
@@ -0,0 +1,125 @@
...
+                broken_line = line;
+            } else {
+                broken_line.push(' ');
+                broken_line.push_str(&line);
+            }
+            if broken_line.split(';').count() == 4 {
+                writer.write_all(broken_line.as_bytes())?;
+                writer.write_all(&[b'\n'])?;
+                broken_line.clear();
+            }
```
Robert-Steiner, line 57: ❔I'm not sure if it makes sense to return an error/panic, if we for some reason get `broken_line.split(';').count() > 4` wdyt?

Robert-Steiner (approved):

discovery_engine_core/tooling/bin/clean_sources.rs
janpetschexain: in general we should get wellformed csv files, even the broken lines shouldn't be there, but unfortunately it's not up to us to ensure this. for now i would just keep the script as is, if we need to we can always update it later on.

janpetschexain:
bors merge

PR #351, opened 2022-05-03
[Merged by Bors] - TY-2763 remove key phrases when dropping markets

rustonaut:
**References:**

- [TY-2763](https://xainag.atlassian.net/browse/TY-2763)

**Based On:**

- #350

### DIFF
--- a/discovery_engine_core/core/src/engine.rs
+++ b/discovery_engine_core/core/src/engine.rs
@@ -337,9 +337,19 @@ where
     pub async fn set_markets(
         &mut self,
         history: &[HistoricDocument],
-        markets: Vec<Market>,
+        new_markets: Vec<Market>,
     ) -> Result<(), Error> {
-        *self.config.markets.write().await = markets;
+        let new_markets_set = new_markets.iter().collect::<HashSet<_>>();
+        let mut markets_guard = self.config.markets.write().await;
+        let removed = markets_guard
+            .iter()
+            .filter_map(|market| (!new_markets_set.contains(market)).then(|| market.clone()))
+            .collect::<Vec<_>>();
+
+        *markets_guard = new_markets;
+
+        self.ranker.remove_key_phrases(&removed);
+        drop(markets_guard);
 
         let mut stacks_guard = self.stacks.write().await;
         for stack in stacks_guard.values_mut() {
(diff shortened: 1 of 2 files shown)


discovery_engine_core/core/src/ranker.rs
```diff
@@ -97,6 +100,10 @@ impl Ranker for xayn_ai::ranker::Ranker {
     fn negative_cois(&self) -> &[NegativeCoi] {
         self.negative_cois()
     }
+
+    fn remove_key_phrases(&mut self, markets: &[Market]) {
+        self.remove_key_phrases(markets)
```
janpetschexain, line 105: 🔧 `just check` locally complains about this, not sure why the ci doesn't
```suggestion
        self.remove_key_phrases(markets);
```

discovery_engine_core/core/src/engine.rs
```diff
@@ -337,11 +338,15 @@ where
...
         &mut self,
         history: &[HistoricDocument],
-        markets: Vec<Market>,
+        new_markets: Vec<Market>,
     ) -> Result<(), Error> {
-        *self.config.markets.write().await = markets;
-
         {
+            let markets = &mut *self.config.markets.write().await;
+            let mut old_markets = replace(markets, new_markets);
```
janpetschexain, line 345: :art: imo it looks a bit cleaner to use the trait instead of de- and re-referencing
```suggestion
            let mut markets = self.config.markets.write().await;
            let mut old_markets = replace(markets.deref_mut(), new_markets);
```

discovery_engine_core/core/src/engine.rs
```diff
@@ -337,11 +338,15 @@ where
...
+            let new_markets_set = markets.iter().collect::<HashSet<_>>();
+            old_markets.retain(|market| !new_markets_set.contains(market));
```
janpetschexain, line 347: ❓ we could use the vector directly instead of collecting into a set before. this would mean linear instead of constant lookup time, but the number of markets is very limited and we save the set allocation.
```suggestion
            old_markets.retain(|market| !markets.contains(market));
```

janpetschexain (approved):
lgtm

rustonaut:
bors merge

PR #354, opened 2022-05-03
[Merged by Bors] - TY-2747 define available sources events

janpetschexain:
**References**

- [TY-2747]
- followed by #356

**Summary**

- add `AvailableSourcesListRequested` to the `ClientEvent`s
- add `AvailableSourcesListRequestSucceeded` and `AvailableSourcesListRequestFailed` to the `EngineEvent`s
- impl stubs for the engine manager and the discovery engine to get the available sources

[TY-2747]: https://xainag.atlassian.net/browse/TY-2747?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #356, opened 2022-05-03
[Merged by Bors] - TY-2749 handle available sources event

janpetschexain:
**References**

- [TY-2749]
- [TY-2748]
- requires #354

**Summary**

- impl fuzzy search for available sources, we don't want to expose the configuration but it might likely be changed in agreement with blue once they do some experiments
- move the available sources from the engine to the feed manager (review comment from previous pr)

[TY-2749]: https://xainag.atlassian.net/browse/TY-2749?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[TY-2748]: https://xainag.atlassian.net/browse/TY-2748?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine/lib/src/domain/feed_manager.dart
+++ b/discovery_engine/lib/src/domain/feed_manager.dart
@@ -208,6 +208,11 @@ class FeedManager {
 
   Future<EngineEvent> getAvailableSourcesList(String fuzzySearchTerm) async {
     final sources = await _engine.getAvailableSources(fuzzySearchTerm);
+
+    if (sources.isEmpty) {
+      return const EngineEvent.availableSourcesListRequestFailed();
+    }
+
     return EngineEvent.availableSourcesListRequestSucceeded(sources);
   }
 }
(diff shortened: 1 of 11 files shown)


discovery_engine/lib/src/domain/feed_manager.dart
```diff
@@ -204,4 +205,9 @@ class FeedManager {
     final sources = await _sourcePreferenceRepository.getTrusted();
     return EngineEvent.trustedSourcesListRequestSucceeded(sources);
   }
+
+  Future<EngineEvent> getAvailableSourcesList(String fuzzySearchTerm) async {
+    final sources = await _engine.getAvailableSources(fuzzySearchTerm);
+    return EngineEvent.availableSourcesListRequestSucceeded(sources);
```
biernacki, line 211: If we use only this one event, maybe we don't need `AvailableSourcesListRequestFailed`? otherwise maybe we could check if the list is empty, and return it in that case, wdyt?

biernacki (approved):
very nice!

janpetschexain:
bors merge

PR #357, opened 2022-05-03
[Merged by Bors] - TY-2789 Handle TrendingTopics events

biernacki:
[task](https://xainag.atlassian.net/browse/TY-2789)

This PR adds a handler for `TrandingTopicRequested` event and a unit test for it.

### DIFF
--- a/discovery_engine/lib/src/domain/search_manager.dart
+++ b/discovery_engine/lib/src/domain/search_manager.dart
@@ -66,6 +66,7 @@ class SearchManager {
         restoreSearchRequested: restoreSearchRequested,
         searchClosed: searchClosed,
         searchTermRequested: searchTermRequested,
+        trendingTopicsRequested: trendingTopicsRequested,
         orElse: () =>
             throw UnimplementedError('handler not implemented for $event'),
       );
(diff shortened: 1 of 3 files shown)


discovery_engine/lib/src/domain/search_manager.dart
```diff
@@ -185,6 +186,21 @@ class SearchManager {
     return EngineEvent.searchTermRequestSucceeded(search.searchTerm);
   }
 
+  /// Return the current trending topics.
+  Future<EngineEvent> trendingTopicsRequested() async {
+    final topics = await _engine.getTrendingTopics();
+
+    // TODO: do we need to persist the engine state??
```
janpetschexain, line 193: this is difficult to answer right now, because this is not yet implemented in the rust DE, so it might change later on.

discovery_engine/lib/src/domain/search_manager.dart
```diff
@@ -185,6 +186,21 @@ class SearchManager {
     return EngineEvent.searchTermRequestSucceeded(search.searchTerm);
   }
 
+  /// Return the current trending topics.
+  Future<EngineEvent> trendingTopicsRequested() async {
+    final topics = await _engine.getTrendingTopics();
```
janpetschexain, line 191: :art: it might be good to wrap this in a try-catch and to throw either an `EngineEvent.trendingTopicsRequestFailed()` or an `EngineEvent.engineExceptionRaised()`. but this depends somewhat on the internals of the DE call, which we don't know yet.

janpetschexain (approved):
lgtm

discovery_engine/lib/src/domain/search_manager.dart
biernacki: Fortunately we should be prepared for that already :) If this throws it will be caught above, cause all handlers inside the worker are wrapped and converted to proper `EngineExceptionRaised`.

discovery_engine/lib/src/domain/search_manager.dart
biernacki: https://xainag.atlassian.net/browse/TY-2807

biernacki:
bors merge

PR #358, opened 2022-05-04
[Merged by Bors] - TY-2727 excluded sources filter

finiteprods:
**Summary**

adds `SourcesFilter` with a method to discard articles with an excluded source domain.

### DIFF
--- a/discovery_engine_core/core/src/stack/filters/article.rs
+++ b/discovery_engine_core/core/src/stack/filters/article.rs
@@ -141,13 +141,11 @@ impl ArticleFilter for CommonFilter {
 pub(crate) struct SourcesFilter;
 
 impl SourcesFilter {
-    #[allow(unused)]
-    /// Filter out any articles with an excluded source domain.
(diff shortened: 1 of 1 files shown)


discovery_engine_core/core/src/stack/filters/article.rs
```diff
@@ -282,6 +293,21 @@ mod tests {
...
 
+    #[test]
+    fn test_filter_sources() {
+        let articles: Vec<Article> =
+            serde_json::from_str(include_str!("../../../test-fixtures/articles-valid.json"))
+                .unwrap();
+        assert_eq!(articles.len(), 4);
+
+        // all 4 articles have source domain example.com
+        let filtered1 = SourcesFilter::apply(articles, &["bbc.com".to_string()]);
```
janpetschexain, line 304: :art: we should refrain from using real domains, to make the test still work we could just use a different top level domain
```suggestion
        let filtered1 = SourcesFilter::apply(articles, &["example.org".to_string()]);
```

janpetschexain (approved):
lgtm

finiteprods:
bors merge

PR #361, opened 2022-05-04
[Merged by Bors] - TY-2802 improve error handling for invalid engine state

Robert-Steiner:
Ticket: 
- [TY-2802]

Summary:

Currently, if the engine cannot be restored from the saved state (if the version of the saved state > the expected version), the error is simply propagated to the caller. However, the caller has no way of clearing the invalid engine state or starting the engine from scratch. Therefore, we need to improve the error handling for this specific error. The idea is to just clear invalid engine status when this error occurs and return an error to the caller saying that they should try again. 

The improved error handling is implemented in Dart so that team blue can catch those errors and emit events for them. 

[TY-2802]: https://xainag.atlassian.net/browse/TY-2802?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine/lib/src/domain/event_handler.dart
+++ b/discovery_engine/lib/src/domain/event_handler.dart
@@ -163,9 +163,14 @@ class EventHandler {
         );
       } catch (e, st) {
         logger.e('failed to initialize the engine', e, st);
-        final reason = e is AssetFetcherException
-            ? EngineExceptionReason.failedToGetAssets
-            : EngineExceptionReason.genericError;
+        final EngineExceptionReason reason;
+        if (e is AssetFetcherException) {
+          reason = EngineExceptionReason.failedToGetAssets;
+        } else if (e is InvalidEngineStateException) {
+          reason = EngineExceptionReason.invalidEngineState;
+        } else {
+          reason = EngineExceptionReason.genericError;
+        }
         return EngineEvent.engineExceptionRaised(
           reason,
           message: '$e',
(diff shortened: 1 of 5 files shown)


rustonaut (approved):
lgtm

discovery_engine/lib/src/domain/event_handler.dart
```diff
@@ -228,17 +233,26 @@ class EventHandler {
...
-        config: config,
-        setupData: setupData,
-        engineState: engineState,
-        history: history,
-        aiConfig: aiConfig,
-        trustedSources: trustedSources,
-        excludedSources: excludedSources,
-      ),
-    );
+    final Engine engine;
```
biernacki, line 236: I'm a bit surprised that that the analyzer doesn't complaing about not using `late` here 😲

discovery_engine/lib/src/domain/event_handler.dart
```diff
@@ -342,3 +356,14 @@ class EventHandler {
...
+
+/// Thrown when the engine cannot be recovered from the given state.
+class InvalidEngineStateException implements Exception {
+  final String message;
+
+  InvalidEngineStateException(this.message);
+
+  @override
+  String toString() =>
+      'InvalidEngineStateException: $message. Try to restart the engine.';
```
biernacki, line 368: > Try to restart the engine.

Do you think it makes sense for us to try again after clearing state? or is should be tried in the Manager, or the app side?

biernacki (approved):
nice improvement 👍

discovery_engine/lib/src/domain/event_handler.dart
Robert-Steiner: Yes we could do that. the reason why we don't do it is that we want to let team blue log the error/event.

discovery_engine/lib/src/domain/event_handler.dart
Robert-Steiner: Would it be better to use `late`?

biernacki: if the analyzer doesn't complain, then it should be fine

Robert-Steiner:
bors merge

PR #362, opened 2022-05-04
[Merged by Bors] - TY-2731 Use semantic filtering across stacks 

x3ro:
[TY-2731]

[TY-2731]: https://xainag.atlassian.net/browse/TY-2731?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine_core/core/src/engine.rs
+++ b/discovery_engine_core/core/src/engine.rs
@@ -57,7 +57,7 @@ use crate::{
     ranker::Ranker,
     stack::{
         self,
-        filters::DuplicateFilter,
+        filters::{filter_semantically, DuplicateFilter, SemanticFilterConfig},
         BoxedOps,
         BreakingNews,
         Data as StackData,
@@ -641,6 +641,7 @@ async fn update_stacks<'a>(
     // Separate stack documents (via the `stack` parameter) are not needed here, since they are
     // already contained in `all_documents`.
     all_documents = DuplicateFilter::apply(history, &[], all_documents);
+    all_documents = filter_semantically(all_documents, &SemanticFilterConfig::default());
 
     // Finally, we can update the stacks with their respective documents. To do this, we
     // have to group the fetched documents by `stack_id`, then `update` the stacks.
--- a/discovery_engine_core/core/src/stack/ops/breaking.rs
+++ b/discovery_engine_core/core/src/stack/ops/breaking.rs
@@ -32,7 +32,7 @@ use crate::{
     document::{Document, HistoricDocument},
     engine::{EndpointConfig, GenericError},
     stack::{
-        filters::{filter_semantically, ArticleFilter, CommonFilter, SemanticFilterConfig},
+        filters::{ArticleFilter, CommonFilter},
         Id,
     },
 };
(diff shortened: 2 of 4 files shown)


discovery_engine_core/core/src/stack/ops/breaking.rs
```diff
@@ -120,8 +118,7 @@ impl Ops for BreakingNews {
 
     fn merge(&self, stack: &[Document], new: &[Document]) -> Result<Vec<Document>, GenericError> {
         let merged: Vec<Document> = chain!(stack, new).cloned().collect();
-        let filtered = filter_semantically(merged, &self.semantic_filter_config);
-        Ok(filtered)
+        Ok(merged)
```
janpetschexain, line 121: :art: since this is essentially a one liner now, we could write it as
```suggestion
        Ok(chain!(stack, new).cloned().collect())
```

discovery_engine_core/core/src/stack/ops/personalized.rs
```diff
@@ -129,8 +127,7 @@ impl Ops for PersonalizedNews {
 
     fn merge(&self, stack: &[Document], new: &[Document]) -> Result<Vec<Document>, GenericError> {
         let merged = chain!(stack, new).cloned().collect();
-        let filtered = filter_semantically(merged, &self.semantic_filter_config);
-        Ok(filtered)
+        Ok(merged)
```
janpetschexain, line 130: :art: similarly this could be written as a one liner now
```suggestion
        Ok(chain!(stack, new).cloned().collect())
```

discovery_engine_core/core/src/stack/ops/trusted.rs
```diff
@@ -114,8 +112,7 @@ impl Ops for TrustedNews {
 
     fn merge(&self, stack: &[Document], new: &[Document]) -> Result<Vec<Document>, GenericError> {
         let merged = chain!(stack, new).cloned().collect();
-        let filtered = filter_semantically(merged, &self.semantic_filter_config);
-        Ok(filtered)
+        Ok(merged)
```
janpetschexain, line 115: :art: similarly this could be written as a one liner now
```suggestion
        Ok(chain!(stack, new).cloned().collect())
```

discovery_engine_core/core/src/engine.rs
```diff
@@ -641,6 +641,7 @@ async fn update_stacks<'a>(
     // Separate stack documents (via the `stack` parameter) are not needed here, since they are
     // already contained in `all_documents`.
     all_documents = DuplicateFilter::apply(history, &[], all_documents);
+    all_documents = filter_semantically(all_documents, &SemanticFilterConfig::default());
```
janpetschexain, line 644: ❓ should we group those filters into a common filter struct? similarly as we do with the `CommonFilter` for the per-stack filtering? not sure what the best name for this would be, maybe `GlobalCommonFilter` or sth similar?

x3ro: If you ask me, it's fine like this for the time being. The cross-stack filtering is relatively new, and we don't yet know where it's headed. It's already not a unified interface, because the different filters require different parameters. I would  see what changes are in store in the next month or two, and if we see that we added / are about to add a bunch of functionality here, then we can still refactor it to a struct / function. wdyt?

janpetschexain (approved):

x3ro:
bors merge

PR #363, opened 2022-05-04
TY-2728 use sources filter in personalized and breaking stacks

finiteprods:
previously #358 added `SourcesFilter`

**Summary**

adjusts `PersonalizedNews` and `BreakingNews` to apply the `SourcesFilter` on fetched articles.


discovery_engine_core/core/src/stack/ops/breaking.rs
```diff
@@ -139,6 +145,10 @@ fn spawn_headlines_request(
             trusted_sources: &[],
             topic: None,
         };
-        client.query_articles(&query).await.map_err(Into::into)
+        client
+            .query_articles(&query)
+            .await
+            .map(|arts| SourcesFilter::apply(arts, &excluded_sources))
```
janpetschexain, line 151: 🔧 this seems odd within the `spawn_headlines_request()`. shouldn't this be called along the other filters in `filter_articles()` following l.76? we would need to pass `excluded_sources` there as well though. similar question for the personalized news.

finiteprods: yeah, was undecided where to call this - in `filter_articles` seems the natural place for it, but I also thought about having it "close" to where the source exclusion is already happening. as you mentioned it also avoids changing the signature of `filter_articles` which might be desirable as source exclusion won't be relevant for `TrustedNews`, for example. more thoughts welcome.

x3ro: Mmh, I think I'd also prefer the `SourcesFilter` to be executed in `filter_articles`. Could we make `filter_articles` a method? That way it'd gain access to the excluded sources without having to bloat the function signature 🤔

finiteprods: ah apologies I had in my mind that `filter_articles` was still part of the `Ops` trait (so was a little reluctant to change the signature)! calling from `filter_articles` does make the most sense, have changed this now accordingly.

janpetschexain (approved):
lgtm

finiteprods:
bors merge

PR #365, opened 2022-05-05
[Merged by Bors] - TY-2746 available sources asset

janpetschexain:
**References**

- [TY-2746]
- requires #356
- followed by #372

**Summary**

- add the available sources asset to the setup data, the asset itself is already uploaded to the yellow and staging buckets
- load it during engine initialization and parse the csv content

[TY-2746]: https://xainag.atlassian.net/browse/TY-2746?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine/lib/src/domain/assets/data_provider.dart
+++ b/discovery_engine/lib/src/domain/assets/data_provider.dart
@@ -15,6 +15,8 @@
 import 'package:equatable/equatable.dart' show EquatableMixin;
 import 'package:xayn_discovery_engine/src/domain/assets/assets.dart'
     show AssetFetcher, AssetReporter, Manifest;
+import 'package:xayn_discovery_engine/src/domain/models/source.dart'
+    show AvailableSources;
 
 const _kEnginePath = 'engine_data';
 const kAssetsPath = '$_kEnginePath/assets';
@@ -41,6 +43,10 @@ abstract class SetupData with EquatableMixin {
         kpeClassifier,
         availableSources,
       ];
+
+  Future<AvailableSources> getAvailableSources() {
+    throw UnsupportedError('Unsupported plattform.');
+  }
 }
 
 /// Reads the assets manifest and provides the [SetupData] to further use.
--- a/discovery_engine/lib/src/domain/assets/data_provider.dart
+++ b/discovery_engine/lib/src/domain/assets/data_provider.dart
@@ -29,10 +29,18 @@ abstract class SetupData with EquatableMixin {
   Object get kpeModel;
   Object get kpeCnn;
   Object get kpeClassifier;
+  Object get availableSources;
 
   @override
-  List<Object?> get props =>
-      [smbertVocab, smbertModel, kpeVocab, kpeModel, kpeCnn, kpeClassifier];
+  List<Object?> get props => [
+        smbertVocab,
+        smbertModel,
+        kpeVocab,
+        kpeModel,
+        kpeCnn,
+        kpeClassifier,
+        availableSources,
+      ];
 }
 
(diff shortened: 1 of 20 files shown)


discovery_engine/lib/src/domain/assets/data_provider.dart
```diff
@@ -29,10 +31,22 @@ abstract class SetupData with EquatableMixin {
...
+        smbertVocab,
+        smbertModel,
+        kpeVocab,
+        kpeModel,
+        kpeCnn,
+        kpeClassifier,
+        availableSources,
+      ];
+
+  Future<AvailableSources> getAvailableSources() {
```
rustonaut, line 47: :art: I think this should be an abstract method. `SetupData` is already `abstract`.

discovery_engine/lib/src/domain/models/source.dart
```diff
@@ -12,6 +12,9 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
+import 'dart:convert' show utf8;
+
+import 'package:csv/csv.dart' show CsvToListConverter;
 import 'package:equatable/equatable.dart' show Equatable;
 import 'package:freezed_annotation/freezed_annotation.dart';
 import 'package:fuzzy/fuzzy.dart' show Fuzzy, FuzzyOptions;
```
rustonaut, line 20: :wrench: ? `just check` complains locally that the _next_ line (which I can't comment on) is unneeded because the protected in `meta.dart` and in `freezed_annotation.dart` is the same. But this is already the way on master :thinking: ?

Anyway we could just fix it here.

discovery_engine/lib/src/domain/assets/data_provider.dart
janpetschexain: i'm not sure what you mean? the method is not abstract because it has a body, it throws an `UnsupportedError` like in our other abstract classes. also adding an `abstract` keyword to a method is illegal syntax in dart:
```
Members of classes can't be declared to be 'abstract'.
Try removing the 'abstract' keyword. You can add the 'abstract' keyword before the class declaration.
dart(abstract_class_member)
```

rustonaut:
```suggestion
  Future<AvailableSources> getAvailableSources();
```
(and removing the next lines of the body)

There is little reasons to implement a method which is meant to always be override in all concrete instances. If we do so in other cases we should change it there, too.  Throwing `UnsupportedError` is only for cases where a method is implemented in some of the (non concrete) sub-classes but not in others. E.g. in out case with not implementing the
method we lint/type-check that it will be overridden. By giving it a body like that we do lose the type checks while gaining nothing (as far as I can tell) and having more code.

IMHO `UnsupportedError` is most times an anti-pattern through it sometimes can be use-full to avoid increases of complexity.

discovery_engine/lib/src/domain/models/source.dart
rustonaut: I just checked it. I'm running the same version.

discovery_engine/lib/src/domain/assets/data_provider.dart
janpetschexain: we can change that, but we should do it in a separate pr for all occurences throughout our code base. for the current implementation i would prefer consistency of the local improvement.

rustonaut (approved):
lgtm

janpetschexain:
bors merge

PR #366, opened 2022-05-09
TY-2804 On disk persistance for database usage in tests

biernacki:
**Task:**
 - [TY-2804](https://xainag.atlassian.net/browse/TY-2804)

**Summary:**
This PR changes database usage in tests from in-memory to on-disk persistance in order to determine early if all needed `TypeAdapters` are registered.

**Context:**
Hive in-memory boxes don't use binary readers and writers, hence they don't use `TypeAdapters`, so if we forget to register one, then we will find out about it in the runtime.


finiteprods (approved):
looks good :+1:

biernacki:
bors merge

PR #367, opened 2022-05-09
[Merged by Bors] - Disable soname

acrrd:
For unknown reasons the lib is corrupted during the `flutter build` process on the CI and some of the local machine.
Setting the soname using the compiler flag seems to avoid this.

Edit:
Removed soname but kept the warning.

### DIFF
--- a/.env
+++ b/.env
@@ -16,7 +16,7 @@ ANDROID_PLATFORM_VERSION="21"
 IOS_TARGETS="aarch64-apple-ios x86_64-apple-ios"
 IOS_LIB_BASE="libxayn_discovery_engine_bindings"
 
-PRODUCTION_RUSTFLAGS="-Ccodegen-units=1 -Clto=on -Cembed-bitcode=yes -Clink-arg=-Wl,-soname,libxayn_discovery_engine_bindings.so"
+PRODUCTION_RUSTFLAGS="-Ccodegen-units=1 -Clto=on -Cembed-bitcode=yes"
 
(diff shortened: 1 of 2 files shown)


rustonaut (approved):
lgtm

the reason I didn't use linker args before is because that can also blow up as this args are used for all linked artifacts instead of just the final shared object. But I guess for us in the way we use it it should be fine.

acrrd:
bors merge

acrrd:
> the reason I didn't use linker args before is because that can also blow up as this args are used for all linked artifacts instead of just the final shared object. But I guess for us in the way we use it it should be fine.

I read about that but for us should be ok. In alternative we can avoid to set it since newer version of android are loading it even if the min api level is set to 30 and they claim that for level 23>= the soname must be present.

acrrd:
bors merge

acrrd:
I downgraded the soname check back to a warning and removed the setting of the soname from production flags.
When we normally build the flags are not used and the lib has no soname. Only are only setting then when we build for production.
I tried with `[target. ...] rustflags` in `Cargo.toml` and in `.cargo/config.toml` but nothing worked.

Android seems to not really care about the soname even in newer version where it should be enforced. To avoid to have a check that only warn us even if in production we would like to have it set I prefer to remove it completely and look for a proper to set it.

@rustonaut

rustonaut (approved):
should be fine for now

let's hope this doesn't brake again

acrrd:
bors merge

PR #371, opened 2022-05-09
[Merged by Bors] - TY-2805 Do not expose paging info of ActiveSearch to API.

rustonaut:
This is archived to have an `ActiveSearch` in the API and another internal one.

Also change the internal `ActiveSearch` to no longer use `freezed` due to the shenanigans needed to be able to add additional methods on a `freezed` object (and it's  not needed anymore).

**References:**

- [TY-2805](https://xainag.atlassian.net/browse/TY-2805)

### DIFF
--- a/discovery_engine/lib/src/domain/models/active_search.dart
+++ b/discovery_engine/lib/src/domain/models/active_search.dart
@@ -15,7 +15,6 @@
 import 'package:equatable/equatable.dart';
 import 'package:hive/hive.dart'
     show HiveType, HiveField, TypeAdapter, BinaryReader, BinaryWriter;
-import 'package:xayn_discovery_engine/src/api/models/active_search.dart' as api;
 import 'package:xayn_discovery_engine/src/domain/repository/type_id.dart'
     show searchTypeId, searchByTypeId;
 
@@ -27,28 +26,25 @@ class ActiveSearch with EquatableMixin {
   @HiveField(0)
   final String searchTerm;
   @HiveField(1)
-  final int requestedPageNb;
+  int requestedPageNb;
   @HiveField(2)
   final int pageSize;
   @HiveField(3, defaultValue: SearchBy.query)
   final SearchBy searchBy;
 
-  const ActiveSearch({
+  ActiveSearch({
     required this.searchTerm,
     required this.requestedPageNb,
     required this.pageSize,
     required this.searchBy,
   });
 
-  ActiveSearch nextPageSearch() => ActiveSearch(
+  ActiveSearch nextPage() => ActiveSearch(
         searchTerm: searchTerm,
         requestedPageNb: requestedPageNb + 1,
         pageSize: pageSize,
         searchBy: searchBy,
       );
-
-  api.ActiveSearch toApiRepr() =>
-      api.ActiveSearch(searchBy: searchBy, searchTerm: searchTerm);
 
   @override
   List<Object?> get props => [searchTerm, requestedPageNb, pageSize, searchBy];
--- a/discovery_engine/lib/src/domain/models/active_search.dart
+++ b/discovery_engine/lib/src/domain/models/active_search.dart
@@ -12,28 +12,46 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-import 'package:freezed_annotation/freezed_annotation.dart';
+import 'package:equatable/equatable.dart';
 import 'package:hive/hive.dart'
     show HiveType, HiveField, TypeAdapter, BinaryReader, BinaryWriter;
+import 'package:xayn_discovery_engine/src/api/models/active_search.dart' as api;
 import 'package:xayn_discovery_engine/src/domain/repository/type_id.dart'
     show searchTypeId, searchByTypeId;
 
-part 'active_search.freezed.dart';
 part 'active_search.g.dart';
 
 /// [ActiveSearch] represents attributes of a performed search.
-@freezed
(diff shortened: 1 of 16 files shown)


discovery_engine/lib/src/domain/models/active_search.dart
```diff
@@ -12,28 +12,46 @@
...
+  final SearchBy searchBy;
+
+  const ActiveSearch({
+    required this.searchTerm,
+    required this.requestedPageNb,
+    required this.pageSize,
+    required this.searchBy,
+  });
+
+  ActiveSearch nextPageSearch() => ActiveSearch(
```
janpetschexain, line 43: ❓ is it required (by hive or for other reasons) to create a new object? otherwise we could just mutate the current object by incrementing the `requestedPageNb`. this would be more in line with the updating character of the function as it is used in the search manager.

janpetschexain, line 43: :art: the naming seems a bit redundant
```suggestion
  ActiveSearch nextPage() => ActiveSearch(
```

discovery_engine/lib/src/domain/models/active_search.dart
```diff
@@ -12,28 +12,46 @@
...
+        searchTerm: searchTerm,
+        requestedPageNb: requestedPageNb + 1,
+        pageSize: pageSize,
+        searchBy: searchBy,
+      );
+
+  api.ActiveSearch toApiRepr() =>
```
acrrd, line 50: This object should be completely agnostic of the api models we are currently using

discovery_engine/lib/src/domain/models/active_search.dart
acrrd: Also this data is a basically a POD, it should not contain any logic about what the next page is.

rustonaut: > question is it required (by hive or for other reasons) to create a new object? 

No, just less likely to cause problems due to accidental shared references in a GCed language.
And at least new/before the fields had been final.

> Also this data is a basically a POD, it should not contain any logic about what the next page is.

This is roughly the same design as `Document` which also has small helpers (in form of computed fields/getters).

Before we used `foo.copyWith(requestedPageNb: foo.requestedPageNb + 1)` ad-hoc, but that's a bit prone to mistakes.

Furthermore given that this type already contains the concept of paging it's not like it's some logic specific to some implementation (or at least not more then the fields already are).

If you have a better place to put the method I will move it there. I mean I did consider an extension trait but it's just not worth the additional overhead IMHO.

discovery_engine/lib/src/domain/models/active_search.dart
rustonaut: It's the same approach as currently used by `Document`, it that wouldn't already have been this way in `Document` I would have placed it into `api/models/active_search.dart`. 

When doing so should I also move to `toApiDocument` method on `Document` from `domain/models` to `api/models`?

rustonaut: Through it also add some complexity  as you have to either write `ActiveSearch.fromInternalRepr(search)` instead of `search.toApiRepr()` or have to write a extension trait you then also have to import.

rustonaut: It's now a extension method.

discovery_engine/lib/src/domain/search_manager.dart
```diff
@@ -121,23 +123,26 @@ class SearchManager {
...
+    final search = await _searchRepo.getCurrent();
 
     if (search == null) {
       const reason = SearchFailureReason.noActiveSearch;
       return const EngineEvent.nextSearchBatchRequestFailed(reason);
     }
 
     // lets update active search params
-    search = search.copyWith(requestedPageNb: search.requestedPageNb + 1);
+    search.requestedPageNb += 1;
```
janpetschexain, line 139: 🔧 if we change the field directly, then we could remove the unused `ActiveSearch.nextPage()` method completely. is there a specific reason why the method is not used anymore (independent of whether the method mutates the object or creates a new object)?

❓ we probably can't mark the fields of `ActiveSearch` as private because of `hive`, right? otherwise we could better contain the changes to the object.

discovery_engine/lib/src/domain/models/document.dart
```diff
@@ -69,6 +69,8 @@ class Document {
     this.isSearched = false,
   }) : timestamp = DateTime.now().toUtc();
 
+  //FIXME move this conversion to `api/models/document.dart` e.g. as an
```
janpetschexain, line 72: ❓ i wouldn't mind if this was done already in this pr, but we can also leave it for another pr. in the latter case, please add a ticket to the backlog.

discovery_engine/lib/src/domain/search_manager.dart
rustonaut: > we probably can't mark the fields [...] private [...]

Also we use the fields in various places outside of the module so we would need to add getters and and the `nextPage` method again which conflicts with the feedback from @acrrd :

> Also this data is a basically a POD,  it should not contain any logic about what the next page is.

It's also not in line with the other types in `(domain|api)/models/` which are all PODs.

rustonaut: >  [..] could remove [..]

done

discovery_engine/lib/src/domain/models/document.dart
rustonaut: Ok did it.

janpetschexain (approved):
lgtm

rustonaut:
bors merge

PR #372, opened 2022-05-09
TY-2750 available sources tests

janpetschexain:
**References**

- [TY-2750]
- requires #365

**Summary**

- fix the searchable keys definition
- add tests


discovery_engine/lib/src/domain/models/source.dart
```diff
@@ -82,6 +82,18 @@ class AvailableSources extends Fuzzy<AvailableSource> {
...
+                name: 'name',
+                getter: (availableSource) => availableSource.name,
+                weight: 1.0,
+              ),
+              WeightedKey(
+                name: 'domain',
+                getter: (availableSource) => availableSource.domain,
+                weight: 1.0,
+              ),
+            ],
```
biernacki, line 96: ❔ just out of curiosity, how does adding this option affects the outcome?

discovery_engine/test/integration/available_sources_test.dart
```diff
@@ -0,0 +1,90 @@
...
+        engine.engineEvents,
+        emitsInOrder(<Matcher>[isA<AvailableSourcesListRequestSucceeded>()]),
+      );
+      final response = await engine.getAvailableSourcesList('');
+      expect(response, isA<AvailableSourcesListRequestSucceeded>());
+      expect(
+        (response as AvailableSourcesListRequestSucceeded)
+            .availableSources
+            .length,
+        equals(27221),
```
biernacki, line 61: 🎨 maybe we can check againt the asset that we are using, so the test won't be outdated when we change the list of sources, wdyt?

discovery_engine/test/integration/available_sources_test.dart
```diff
@@ -0,0 +1,90 @@
...
+      expect(
+        engine.engineEvents,
+        emitsInOrder(<Matcher>[isA<AvailableSourcesListRequestSucceeded>()]),
+      );
+      const fuzzySearchTerm = 'example';
+      final response = await engine.getAvailableSourcesList(fuzzySearchTerm);
+      expect(response, isA<AvailableSourcesListRequestSucceeded>());
+      final availableSources =
+          (response as AvailableSourcesListRequestSucceeded).availableSources;
+      expect(availableSources.length, equals(4));
```
biernacki, line 84: ❓ do we need to check this? feels like it might get out of date when we use different assets, or is this using a mocked fixture?

biernacki (commented):
looks good, just left a few questions

discovery_engine/lib/src/domain/models/source.dart
janpetschexain: without the key we don't get any results (or all result if the search term is the empty string), this was an oversight on my end when i implemented the search. basically, since we use a `List<AvailableSource>` instead of a `List<String>` we have to tell the searcher how to get to the string fields of the custom class.

the weights are required to give a weighting to each field/key of the searched object in the overall score computation, i've chosen them equally (they are normalized by the lib) due to lack of better values for now. eg we could give the name of the source a higher weighting than the domain of the source, depending on blue's needs.

discovery_engine/test/integration/available_sources_test.dart
janpetschexain: it's using the real asset, i agree that this is bound to break, i'll replace it with a not-empty check.

biernacki (approved):
lgtm 🚀

janpetschexain:
bors merge

PR #374, opened 2022-05-10
[Merged by Bors] - TY-2787 rankable trending topics

finiteprods:
**Summary**

adds a `TrendingTopic` struct, implementing `xayn_ai::ranker::Document` to allow it to be ranked.

### DIFF
--- a/discovery_engine_core/core/src/document.rs
+++ b/discovery_engine_core/core/src/document.rs
@@ -238,6 +238,20 @@ pub struct HistoricDocument {
     pub title: String,
 }
 
+/// A trending topic.
+pub struct TrendTopic {
+    /// Id of the topic.
+    pub id: Id,
+    /// Precomputed S-mBert of the topic.
+    pub smbert_embedding: Embedding,
+    /// Title of the topic.
+    pub name: String,
+    /// Query term that returns this topic.
+    pub query: String,
+    /// Link to a related image.
+    pub image: Option<Url>,
+}
+
 #[cfg(test)]
 mod tests {
     use chrono::NaiveDate;
--- a/discovery_engine_core/core/src/document.rs
+++ b/discovery_engine_core/core/src/document.rs
@@ -239,7 +239,7 @@ pub struct HistoricDocument {
 }
 
 /// A trending topic.
-pub struct TrendTopic {
+pub struct TrendingTopic {
     /// Id of the topic.
     pub id: Id,
     /// Precomputed S-mBert of the topic.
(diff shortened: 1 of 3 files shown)


discovery_engine_core/core/src/ranker.rs
```diff
@@ -36,6 +36,9 @@ pub trait Ranker {
     /// Performs the ranking of [`Document`] items.
     fn rank(&mut self, items: &mut [Document]) -> Result<(), GenericError>;
 
+    /// Performs the ranking of [`TrendTopic`] items.
+    fn rank_topics(&mut self, items: &mut [TrendTopic]) -> Result<(), GenericError>;
```
x3ro, line 40: 🎨 We could also modify `rank` above to accept all `ranker::Document`s:

```
fn rank<T>(&mut self, items: &mut [T]) -> Result<(), GenericError>
    where
      T: xayn_ai::ranker::Document,
```

discovery_engine_core/core/src/document.rs
```diff
@@ -238,6 +238,20 @@ pub struct HistoricDocument {
     pub title: String,
 }
 
+/// A trending topic.
+pub struct TrendTopic {
```
x3ro, line 242: 🎨 I'd personally prefer this to be called `TrendingTopic`. I'm assuming `TrendTopic` was chosen because it's shorter, or is there another reason?

x3ro (approved):
Some minor quibbles in the comments, but LGTM 🚀

discovery_engine_core/core/src/ranker.rs
Robert-Steiner: The disadvantage would be that this trait would no longer be object safe. But there is also the question of whether it has to be object safe.

x3ro: Fair point. I don't think it would currently be relevant, and I wouldn't be able to gauge whether it could be in the future.

finiteprods: it doesn't seem to much disadvantage for now (there was one place that required a type annotation, but that was it) so I will take this suggestion in, thanks.

discovery_engine_core/core/src/document.rs
finiteprods: yes just because it was shorter, but I'd be fine with `TrendingTopic`. perhaps it may change again, when I come to name the more bing-specific type!

finiteprods:
bors merge

PR #376, opened 2022-05-10
[Merged by Bors] - Be very liberal in what kinds of articles we accept, including those w/ missing fields

x3ro:
I'm getting very confusing compile errors related to deserialization of our articles, and want to investigate whether this also occurs in the CI.


rustonaut (approved):
lgtm

PR #377, opened 2022-05-10
[Merged by Bors] - TY-2809 remove unnecessary UnsupportedError usages

rustonaut:
We have some places where we `throw UnsupportedError` where it's not necessary:

1. methods of abstract classes which always need to be overridden by all implementations
    should just be abstract instead of having a body which always throw an error.
2. exhaustive switch statements over enums (with the lints we enabled dart has roughly
    exhaustive enum switch statements. Just roughly, but what matters is, it fails linting/CI
    when the switch is not exhaustive).
    
 **References:**
 
 - [TY-2809](https://xainag.atlassian.net/browse/TY-2809)

### DIFF
--- a/discovery_engine/lib/src/domain/assets/data_provider.dart
+++ b/discovery_engine/lib/src/domain/assets/data_provider.dart
@@ -44,9 +44,7 @@ abstract class SetupData with EquatableMixin {
         availableSources,
       ];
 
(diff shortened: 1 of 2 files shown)


janpetschexain (approved):
lgtm

rustonaut:
bors merge

PR #384, opened 2022-05-12
[Merged by Bors] - TY-2778 port xayn_ai crate to DE conventions

janpetschexain:
**References**

- [TY-2778]

**Summary**

- the first two commits handle renaming
- the following commits fix a lint each

[TY-2778]: https://xainag.atlassian.net/browse/TY-2778?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


discovery_engine_core/ai/ai/src/coi/key_phrase.rs
```diff
@@ -243,6 +243,7 @@ fn unify(
 /// The elements can be prepared before they are reduced. The reduced result can be finalized
 /// whereas the finalization conditially depends on whether the reduced lane overlaps with the main
 /// square block of the matrix.
+#[allow(clippy::needless_pass_by_value)] // arrayview instead of reference
```
acrrd, line 246: I don't understand the comment, we are getting an `ArrayBase` that could be an ArrayView but it could also be something else.

discovery_engine_core/ai/kpe/extract_params.py
```diff
@@ -29,14 +29,17 @@
 
 BYTE_ORDER: str = "little"
 
+
```
acrrd, line 32: Now sure why we are adding white space. We didn't discuss any style convention or formatting tool so we avoid to change this file more then necessary.

discovery_engine_core/ai/ai/src/coi/key_phrase.rs
janpetschexain: this is because of a conflict between the clippy lint and the ndarray design. the function signature uses `ArrayBase` to accomodate for owned as well as viewed arrays. now clippy wants us to write it as `&ArrayBase`, but the whole concept of "referenced arrays" is tackled by views in ndarray. essentially, we could have a `&ArrayBase` which would eg be a `&Array` or `&ArrayView`, but the latter is just a redundant double reference in a sense.

i'm more in favor of following the ndarray design instead of the clippy lint in this case.

discovery_engine_core/ai/kpe/extract_params.py
janpetschexain: good question, happened by mistake, not sure why it happened at all, it's a python file after all 🤔 i'll revert it.

discovery_engine_core/ai/ai/src/coi/key_phrase.rs
acrrd: I agree with you but I find the comment confusing since the signature does not request a view and an owned value can be passed in and in this case clippy is right.
I think a simple solution is to rephrase the comment, another is to actually request an ArrayView and modify the case in which we don't use it.

janpetschexain: i rephrased the comment

acrrd (approved):

PR #390, opened 2022-05-13
[Merged by Bors] - TY-2848 Allow missing doc comments in publicly exported members

x3ro:


janpetschexain (commented):
looks good, one question though: removing the lint does not only removes the need of doc comments of structs/enums, but also the need of module doc comments etc. unfortunately our choices of lints is pretty limited in this regard and we should still remove `missing_docs`. the only other knob is [`missing_crate_level_docs`](https://doc.rust-lang.org/rustdoc/lints.html#missing_crate_level_docs) which we could enable to at least require some docs in the crate module. wdyt?

x3ro:
Added `missing_crate_level_docs` :)

x3ro:
bors merge

PR #391, opened 2022-05-16
[Merged by Bors] - TY-2791 query bing trending topics

finiteprods:
**Summary**

extends `Client` with methods to query bing for trending topics.

### DIFF
--- a/discovery_engine_core/providers/src/bing.rs
+++ b/discovery_engine_core/providers/src/bing.rs
@@ -35,8 +35,8 @@ impl Query for TrendingQuery<'_> {
 
         let query = &mut url.query_pairs_mut();
 
-        let country = &self.market.country_code;
         let lang = &self.market.lang_code;
+        let country = &self.market.country_code;
         query.append_pair("mkt", &format!("{}-{}", lang, country));
 
         Ok(())
@@ -127,6 +127,13 @@ where
 
 #[cfg(test)]
 mod tests {
+    use wiremock::{
+        matchers::{header, method, path, query_param},
+        Mock,
+        MockServer,
+        ResponseTemplate,
+    };
+
     use super::*;
 
     #[test]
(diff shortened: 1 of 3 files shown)


discovery_engine_core/providers/src/bing.rs
```diff
@@ -0,0 +1,139 @@
...
+impl Query for TrendingQuery<'_> {
+    fn setup_url(&self, url: &mut Url) -> Result<(), Error> {
+        url.path_segments_mut()
+            .map_err(|_| Error::InvalidUrlBase(None))?
+            .push(URL_SUFFIX);
+
+        let query = &mut url.query_pairs_mut();
+
+        let country = &self.market.country_code;
+        let lang = &self.market.lang_code;
```
janpetschexain, line 39: :art: since `lang` is used first before `country` in the format string, the order of those two bindings could be switched to logically match it.

discovery_engine_core/providers/src/bing.rs
```diff
@@ -0,0 +1,139 @@
...
+
+impl Client {
+    /// Run query for fetching trending topics from Bing.
+    pub async fn query_trending(&self, query: &impl Query) -> Result<Vec<TrendingTopic>, Error> {
+        self.query_bing(query).await.map(|trending| trending.value)
+    }
+
+    /// Run a query against Bing.
+    pub async fn query_bing(&self, query: &impl Query) -> Result<Response, Error> {
+        let mut url = Url::parse(&self.url).map_err(|e| Error::InvalidUrlBase(Some(e)))?;
```
janpetschexain, line 56: :art: the client could parse the url once on construction instead of parsing it everytime on query. (i get that this is the same for the newscatcher, not sure why it wasn't changed there because i remember that i made a similar comment a while ago.)

discovery_engine_core/providers/src/bing.rs
```diff
@@ -0,0 +1,139 @@
...
+    #[serde(default, deserialize_with = "deserialize_null_default")]
+    pub(crate) name: String,
+
+    /// Search query that returns this topic.
+    #[serde(default)]
+    pub(crate) query: SearchQuery,
+
+    /// Link to a related image.
+    #[serde(default)]
+    pub(crate) image: Image,
```
janpetschexain, line 98: ❓ are the `SearchQuery` and `Image` newtypes actually required for the serde? it looks like we could just make them `String`s  and put the respective serde attributes on them which are currently within the newtypes.

discovery_engine_core/providers/src/client.rs
```diff
@@ -454,4 +458,44 @@ mod tests {
 
         assert_eq!(format!("{:?}", doc), format!("{:?}", expected));
     }
+
+    #[tokio::test]
+    async fn test_trending() {
```
janpetschexain, line 463: ❓ this is more a general question which might warrant a follow up pr: the `crate::client` module currently contains a lot of `newscatcher` provider related code. since we now also have the `bing` provider (and in the future maybe also `gnews`), should we clean up the `crate::client` module to only contain common client code and move the `newscatcher` related code to `crate::newscatcher`? this would better match with the new `crate::bing` module.

for now we could already move this test to the `crate::bing` module.

discovery_engine_core/providers/src/bing.rs
finiteprods: oh that's true. if you don't mind let me do this little refactoring in a separate pr

discovery_engine_core/providers/src/bing.rs
finiteprods: they seem to be necessary (did try the suggestion, adjusting the tests, but they break). I suppose it is because, for example under `image` in the json there are fields `url` and `provider` and the deserializer needs to be told which one we want. perhaps someone who knows more serde knows a more economical way to do it...

janpetschexain (approved):
lgtm

finiteprods:
bors merge

PR #392, opened 2022-05-16
[Merged by Bors] - TY-2635 add reset ai event

rustonaut:
**References**

- [TY-2635](https://xainag.atlassian.net/browse/TY-2635)

PR #393, opened 2022-05-16
[Merged by Bors] - TY-2636 handle reset ai in dart

rustonaut:
**References:**

- [TY-2636](https://xainag.atlassian.net/browse/TY-2636)
- based on #392


discovery_engine/lib/src/domain/ai_state_holder.dart
```diff
@@ -0,0 +1,18 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+/// Interface implemented by anything holding AI state.
+abstract class AIStateHolder {
+  Future<void> clearAIState();
+}
```
rustonaut, line 18: For now this was the simplest way to abstract over multiple places which need to be cleared.

But I noticed that `FeedManager`, `EventHandler` and the other `Managers`  are not really distinct objects in the classical sense as all of their state is the same shared overlapping state (the repos, engine, etc.). They are more different categories of implementation based on the same state object.

This makes me wonder if it makes sense to not make them separate classes but extensions or better mixins into the same `System` class (or more specific the class we currently call EventHandler and then could rename to mixin).

But  this is out of scope of this PR.

discovery_engine/lib/src/domain/event_handler.dart
```diff
@@ -225,6 +225,12 @@ class EventHandler {
     final activeDataRepository = HiveActiveDocumentDataRepository();
     final activeSearchRepository = HiveActiveSearchRepository();
     final engineStateRepository = HiveEngineStateRepository();
+    final aiStateHolders = [
+      documentRepository,
+      activeDataRepository,
+      activeSearchRepository,
+      engineStateRepository
+    ];
```
rustonaut, line 233: I believe that by grouping them in this way here it's less likely to create a situation where we forget the clear a bit of AI state.

discovery_engine/lib/src/domain/event_handler.dart
```diff
@@ -286,7 +292,12 @@ class EventHandler {
       activeDataRepository,
       engineStateRepository,
     );
-    _systemManager = SystemManager(engine, eventConfig, documentRepository);
+    _systemManager = SystemManager(
+      engine,
+      eventConfig,
+      documentRepository,
+      aiStateHolders,
```
acrrd, line 299: :grey_question: why passing this list instead of a function to call when we want to delete the state?
Here we know about hive so we can create it directly on the box that are involved and we don't expose `clearAIState` on each repository. We will also avoid that on `clearAIState` is called on some repository and not others when it only make sense when is called on all of them.

rustonaut: callbacks don't scale well, but I guess in this case we could do it

acrrd (approved):

discovery_engine/lib/src/domain/event_handler.dart
acrrd: In this case the implementation of each `clearAIState` is one line so is not much of a problem. Also we are trying to move away from dart so I would like to minimize changes here when possibile.

rustonaut: yes, exactly I just wanted to explain how I ended up where I ended up :wink:

PR #398, opened 2022-05-16
TY-2835 deep search ffi

janpetschexain:
**References**

- [TY-2835]
- requires #397

**Summary**

- expose deep search in rust ffi & impl in dart ffi
- dedup mocked documents: we had a lot of duplicate mocked documents in the `MockEngine`, also those mocks are purely test related and hence moved from the lib to the tests


discovery_engine/lib/src/domain/engine/engine.dart
```diff
@@ -88,6 +88,12 @@ abstract class Engine {
     int pageSize,
   );
 
+  /// Performs a deep search regarding a document.
```
finiteprods, line 91: 🎨 it's not clear to me how the arguments relate to the mentioned "document"? is it something along the lines:
Performs a deep search for documents relating to... an existing one described by `term`?
similarly in other places with this docstring

finiteprods (approved):
one minor comment re clarity of the method docstring, otherwise looking all good.

discovery_engine/lib/src/domain/engine/engine.dart
janpetschexain: good point, the method is meant to be called on document snippets, but it can actually be called on any search term and market, i'll adjust the docs.

janpetschexain:
bors merge

PR #399, opened 2022-05-16
[Merged by Bors] - TY-2850 Clean query from user

acrrd:
A query from the user can contain symbols.  We remove them.
At the moment if the user put a symbol in the query the request will fail.


Robert-Steiner (approved):

acrrd:
bors merge

PR #400, opened 2022-05-16
[Merged by Bors] - TY-2833 deep search events

janpetschexain:
**References**

- [TY-2833]
- requires #398
- followed by #401

**Summary**

- rename all active search related things to avoid ambiguity with deep search
- define and expose deep search events

[TY-2833]: https://xainag.atlassian.net/browse/TY-2833?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


discovery_engine/lib/src/discovery_engine_base.dart
```diff
@@ -364,43 +364,44 @@ class DiscoveryEngine {
...
   /// In response it can return:
-  /// - [SearchRequestSucceeded] for successful response, containing a list of
+  /// - [ActiveSearchRequestSucceeded] for successful response, containing a list of
   /// [Document]s
-  /// - [SearchRequestFailed] for failed response, with a reason for failure
+  /// - [ActiveSearchRequestFailed] for failed response, with a reason for failure
   /// - [EngineExceptionReason] for unexpected exception raised, with a reason
   /// for such failure.
-  Future<EngineEvent> requestSearch(String queryTerm) {
+  Future<EngineEvent> requestQuerySearch(String queryTerm) {
```
janpetschexain, line 375: i guess we have to inform blue about all the name changes in this public API?

rustonaut: :wrench:  Given that we call it active search everywhere we probably should call it active search here too:

```suggestion
  Future<EngineEvent> requestActiveSearch(String queryTerm) {
```

janpetschexain: both `QuerySearch` and `TopicSearch` are subtypes of `ActiveSearch`, this is why i deviated from the naming here. we could also name these functions `requestActiveQuerySearch` and `requestActiveTopicSearch`, but i'm not sure if this is too verbose?

rustonaut: I guess we could keep it that way.

janpetschexain:
bors merge

PR #401, opened 2022-05-17
TY-2834 deep search events handler

janpetschexain:
**References**

- [TY-2834]
- requires #400

**Summary**

- impl event handler for `DeepSearchRequested`
- add tests for deep search
- extend `LocalNewsApiServer` with additional test utility
- fix debug assertion of `Encoding` in KPE


discovery_engine_core/ai/kpe/src/tokenizer/encoding.rs
```diff
@@ -130,7 +120,6 @@ impl<const KEY_PHRASE_SIZE: usize> Tokenizer<KEY_PHRASE_SIZE> {
             valid_mask,
             active_mask,
         };
-        debug_assert!(encoding.is_valid(self.tokenizer.vocab_size(), KEY_PHRASE_SIZE));
```
janpetschexain, line 133: this assertion actually includes a handled error for the `valid_mask` sub-check, which is handled in the CNN. but since we run the tests with a debug build, we get a panic and hence a canceled future when running the dart tests. each of the validity checks are done in the respective Bert/CNN/Classifier modules anyways, so we don't loose anything by removing this assertion.

discovery_engine/lib/src/domain/search_manager.dart
```diff
@@ -191,6 +194,40 @@ class SearchManager {
...
+  ) async {
+    final List<DocumentWithActiveData> docs;
+    try {
+      docs = await _engine.deepSearch(term, market);
+    } catch (e) {
+      const fewWords =
+          'The sequence must contain at least `KEY_PHRASE_SIZE` valid words';
+      const notFound = 'HTTP status client error (404 Not Found) for url';
+      final message = e.toString();
+      if (message.contains(fewWords) || message.contains(notFound)) {
```
finiteprods, line 212: ❓ just wondering about "robustness" of checking for specific exception text... but I know we are probably limited in what we can do here...

discovery_engine/test/integration/init_engine_test.dart
```diff
@@ -48,7 +48,7 @@ void main() {
 
     test('news api request error should not raise an engine init exception',
         () async {
-      server.replyWithError = true;
+      server.replyWith = ReplyWith.error;
```
finiteprods, line 51: ❓ was there a reason setting it this way is preferred?

finiteprods (approved):
some minor ponderings as I read through but looks good.

discovery_engine/test/integration/init_engine_test.dart
janpetschexain: due to the newly introduced `server.replyWithEmpty` we would need to manage to exclusive booleans instead (only at  most one of `replyWithError` and `replyWithEmpty` must be true at the same time). hence i replaced it with an enum.

discovery_engine/lib/src/domain/search_manager.dart
janpetschexain: as long as we don't change the error messages on the rust side we should be fine, but i agree that this is not very robust. unfortunately all rust error types are just stringified on the dart side and we currently have no better way to handle this. it's probably also no worth to improve on the ffi error types now, since we move to protobufs.

discovery_engine/test/integration/init_engine_test.dart
finiteprods: I see, thanks

janpetschexain:
bors merge

PR #415, opened 2022-05-23
TY-2866 rebase work from TY-2670 on main and fix rebase problems

rustonaut:
This is only re-basing TY-2670 and then fixing follow up problems (mainly switching the code to use only newscatcher to not brake tests added since TY-2670 was implemented).

**References:**

- [TY-2866](https://xainag.atlassian.net/browse/TY-2866)
- [TY-2810](https://xainag.atlassian.net/browse/TY-2810)

Note that this task originally had no own issue, hence why the branch name starts with TY-2810 (the story).

Also be aware that I had to squash TY-2670 to reduce the amount of rebase problems.


discovery_engine_core/core/src/stack/ops/breaking.rs
```diff
@@ -136,17 +136,20 @@ fn spawn_headlines_request(
...
+
         let query = HeadlinesQuery {
             common: CommonQueryParts {
                 market: Some(&market),
                 page_size,
                 page,
-                excluded_sources: &excluded_sources,
+                excluded_sources: excluded_sources.as_slice(),
             },
+            //TODO pass in trusted sources if available
```
rustonaut, line 147: ignore this I already removed it in the follow up PR.

janpetschexain:
> This should not target main but the branch with gnews

@rustonaut please let me know once this is rebased onto `TY-2670-implement-gnews` as this will probably change a few things again before i start reviewing.

rustonaut:
> This should not target main but the branch with gnews

Well, that explains a lot.

rustonaut:
This PR rebases TY-2670-implement-gnews onto main.

rustonaut:
> if this is merely a rebase of the original branch, what's the reason for this pr then? i would assume that all following work should be merged into the rebased feature branch.

I didn't know we where tracing the feature branch or more correct I somehow thought we would merge the feature branch into master after my PR. But as my PR touches many things it would be easier to first merge this then my PR.

Anyway it was a misunderstanding on my side.

PR #416, opened 2022-05-23
TY-2864 no negative cois required

janpetschexain:
**References**

- [TY-2864]

**Summary**

- remove the `min_negative_cois` check & config, but return as long as there are enough positive cois


janpetschexain:
> We don't want to remove it from the configuration but only set it to zero.

i thought about both options. just setting it to zero makes sense if we want to change it back soon, is this the case? otherwise it doesn't make sense because checking a `usize >= 0` is in vain, in this case it's better to remove it. we could always bring it back by reverting this commit if we ever choose to have a minimum coi requirement again.

acrrd:
I don't know so I prefer to keep it in in case we need it again. checking `usize >= 0` is a noop and even if it wasn't the cost of doing it is non existent.

janpetschexain:
> I don't know so I prefer to keep it in in case we need it again. checking `usize >= 0` is a noop and even if it wasn't the cost of doing it is non existent.

i changed it accordingly. the previous removal was in the sense of reduced code complexity, not in the sense of cheap operations.

discovery_engine_core/ai/ai/src/ranker/context.rs
```diff
@@ -129,8 +198,6 @@ pub(super) fn compute_score_for_docs(
     documents
         .iter()
         .map(|document| {
-            // `compute_score_for_embedding` cannot panic because we calculate
-            // the relevances for all positive cois before
```
finiteprods, line 133: ❓ did you remove this comment because it's no longer relevant? (I don't really understand it / why it's there in the first place)

finiteprods (approved):
looks good, just 1 v minor clarification question

discovery_engine_core/ai/ai/src/ranker/context.rs
janpetschexain: correct, this comment is outdated, i guess i forgot to remove it when i refactored the relevance maps. it doesn't apply anymore because the relevances are computed on the fly whithin the mentioned function call now.

acrrd:
bors merge

PR #417, opened 2022-05-23
[Merged by Bors] - TY-2867 fix market handling when doing requests

janpetschexain:
**References**

- [TY-2867]

**Summary**

- pass the current `Market` when fetching `new_items()` for a `Stack` and use it for the `BreakingNews` & `PersonalizedNews` stacks

[TY-2867]: https://xainag.atlassian.net/browse/TY-2867?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine_core/core/src/stack/ops/common.rs
+++ b/discovery_engine_core/core/src/stack/ops/common.rs
@@ -12,58 +12,32 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-use futures::{stream::FuturesUnordered, StreamExt};
 use tokio::task::JoinHandle;
-use xayn_discovery_engine_providers::Market;
 
 use crate::engine::GenericError;
 
 type ItemsResult<I> = Result<Vec<I>, GenericError>;
-type Requests<I> = FuturesUnordered<JoinHandle<ItemsResult<I>>>;
-
-async fn request_new_items<I: Send>(
-    requests_fn: impl FnOnce() -> Requests<I> + Send,
-) -> ItemsResult<I> {
-    let mut requests = requests_fn();
-    let mut items = Vec::new();
-    let mut error = None;
-
-    while let Some(handle) = requests.next().await {
-        // should we also push handle errors?
-        if let Ok(result) = handle {
-            match result {
-                Ok(batch) => items.extend(batch),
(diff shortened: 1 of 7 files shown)


discovery_engine_core/core/src/stack/ops/common.rs
```diff
@@ -12,58 +12,32 @@
...
-        }
-    }
-
-    if items.is_empty() && error.is_some() {
-        Err(error.unwrap(/* nonempty error */))
-    } else {
-        Ok(items)
-    }
-}
+type Request<I> = JoinHandle<ItemsResult<I>>;
```
Robert-Steiner, line 20: 🌱 Since we only run a single request we could just return a `Future`. I just tried to replace `JoinHandle` with `Future` but it seems that it requires a bit more thinking and it is probably not worth it for the moment.

discovery_engine_core/core/src/stack/ops/common.rs
```diff
@@ -12,58 +12,32 @@
...
     let mut error = None;
 
     for request_num in 0..max_requests {
-        match request_new_items(|| requests_fn(request_num)).await {
+        match request_fn(request_num).await {
             // if the API doesn't return any new items, we stop requesting more pages
-            Ok(batch) if batch.is_empty() => break,
-            Ok(batch) => items.extend(batch),
-            Err(err) => {
+            Ok(Ok(batch)) if batch.is_empty() => break,
```
Robert-Steiner, line 34: 🔧 we can pass the page size to this function and check if `batch < page_size` and break earlier to avoid a request to the next empty page. that was a problem before that we for example got 10 items on page 1 and then we did another request because `batch` was not empty.

Robert-Steiner (approved):
one small thing but the rest looks good 👍

discovery_engine_core/core/src/stack/ops/common.rs
janpetschexain: i also gave that a try with `futures::future::Lazy` which can hold closures to be awaited. this is pretty much what we need, but since we can't name the closure type this would require quite a lot of name fiddling for the type alias. at this point i stopped and went the easy way :/

janpetschexain:
bors merge

PR #419, opened 2022-05-24
[Merged by Bors] - TY-2856 fix dart example readme

janpetschexain:
**Summary**

update readme of the dart example:
- the path resolution of the loaded dynamic library fails (on macos, maybe also other) if the example is executed from its directory (used to work before, maybe due to dart update)
- same holds for the flutter example but the fix is not applicable because additional required flutter project files can't be found anymore from a different path than the example dir

### DIFF
--- a/discovery_engine/.gitignore
+++ b/discovery_engine/.gitignore
@@ -6,6 +6,9 @@
 build/
 **/doc/api/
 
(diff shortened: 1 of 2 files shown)


biernacki (approved):

janpetschexain:
bors merge

PR #420, opened 2022-05-24
[Merged by Bors] - TY 2873 Fetch from all market before update the stack

acrrd:
This PR fetch articles from all the markets before updating the stack.
The problem with the current code is that if updating from a market can bring the stack to a state where it does not need to be updated. This means that when we try to update the other markets this stack will be skipped.

### DIFF
--- a/discovery_engine_core/core/src/engine.rs
+++ b/discovery_engine_core/core/src/engine.rs
@@ -347,29 +347,17 @@ where
         let markets = self.config.markets.read().await;
         let mut stacks = self.stacks.write().await;
 
-        let mut errors = vec![];
-        for market in markets.iter() {
-            if let Err(error) = update_stacks(
-                &mut stacks,
-                &mut self.exploration_stack,
-                &mut self.ranker,
-                history,
-                self.core_config.take_top,
-                self.core_config.keep_top,
-                request_new,
-                market,
-            )
-            .await
-            {
-                errors.push(error);
-            }
-        }
-
-        if errors.is_empty() {
-            Ok(())
-        } else {
-            Err(Error::Errors(errors))
-        }
+        update_stacks(
+            &mut stacks,
+            &mut self.exploration_stack,
+            &mut self.ranker,
+            history,
+            self.core_config.take_top,
+            self.core_config.keep_top,
+            request_new,
+            &*markets,
+        )
+        .await
     }
 
     /// Serializes the state of the `Engine` and `Ranker` state.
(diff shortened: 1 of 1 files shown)


discovery_engine_core/core/src/engine.rs
```diff
@@ -810,18 +798,29 @@ async fn update_stacks<'a>(
...
     // Here we gather new documents for all relevant stacks, and put them into a vector.
     // We don't update the stacks immediately, because we want to de-duplicate the documents
     // across stacks first.
+    let no_key_phrases = vec![];
     let new_document_futures = needy_stacks
         .iter()
-        .map(|stack| fetch_new_documents_for_stack(stack, ranker, &key_phrases, history, market))
+        .flat_map(|stack| {
+            markets.iter().map(|market| {
+                let key_phrases = key_phrases_by_market.get(market).unwrap_or(&no_key_phrases);
```
janpetschexain, line 820: :art: it would be nice to use `Option::unwrap_or_default()` but there seems no way to get from an `Option<&Vec<_>>` to an `Option<&[_]>` even with `Option::as_deref()`. still there is another possibility which at least allows us to avoid binding the empty `no_key_phrases`:
```suggestion
                let key_phrases = key_phrases_by_market.get(market).map_or_else(<&[_]>::default, Vec::as_slice);
```
or maybe better without the closure
```suggestion
                let key_phrases = key_phrases_by_market.get(market).map_or(&[] as &[_], Vec::as_slice);;
```

Robert-Steiner (approved):

discovery_engine_core/core/src/engine.rs
```diff
@@ -347,29 +347,17 @@ where
...
-        }
+        update_stacks(
+            &mut stacks,
+            &mut self.exploration_stack,
+            &mut self.ranker,
+            history,
+            self.core_config.take_top,
+            self.core_config.keep_top,
+            request_new,
+            &*markets,
```
janpetschexain, line 358: :art: the additional deref is not needed for inference to work
```suggestion
            &markets,
```

discovery_engine_core/core/src/engine.rs
```diff
@@ -810,18 +798,29 @@ async fn update_stacks<'a>(
         return Ok(());
     }
 
-    let key_phrases = needy_stacks
+    let key_phrases_by_market = needy_stacks
```
janpetschexain, line 801: ❓ it seems a bit unfortunate that we have to collect a hashmap in this first loop just to pick the entries one by one in the second loop below. i get that we can't just move the `ranker.take_key_phrases()` into the second loop because the captured `&mut ranker` can't be passed along with the constructed future. instead of the current `join_all(Vec<Future>)`, would this maybe work with sth like a `FuturesUnordered<_>` collection? or if we set up the collecting and joining of the futures differently eg. with a join macro?

janpetschexain (approved):

PR #422, opened 2022-05-25
[Merged by Bors] - TY-2822/TY-2866 refactor/fix tests to make follow up PR easier to review

rustonaut:
The first commit does some refactors, the second fixes failing tests of the base branch (not failing because of the refactor they already did fail before).

**References:**

- [TY-2866](https://xainag.atlassian.net/browse/TY-2866) (repurposed ticket)
- [TY-2822](https://xainag.atlassian.net/browse/TY-2822) (this pr is in preparation for this ticket)
- [TY-2810](https://xainag.atlassian.net/browse/TY-2810) (the story)

Things refactored:
 
1. newscatcher + newscatcher_cline => newscatcher + newscatcher::models
2. gnews + gnews_client => gnews + gnews::models
3. move some public top level parts to from client to lib (they wont change,
    client will be gone in follow up PR)
4. rename some functions using gnews to have gnews in the name
5. move code specific to newscatcher/gnews from client to the
     newscatcher/gnews modules


discovery_engine_core/providers/src/lib.rs
```diff
@@ -50,3 +53,69 @@ pub use filter::{Filter, Market};
...
+
+    /// The language of the article.
+    pub language: String,
+
+    /// Main topic of the publisher.
+    pub topic: String,
+}
+
+impl Article {
+    /// Gets the excerpt or falls back to the title if the excerpt is empty.
```
janpetschexain, line 115: :wrench: the doc naming should be updated as well
```suggestion
    /// Gets the snippet or falls back to the title if the snippet is empty.
```

discovery_engine_core/providers/src/gnews/models.rs
```diff
@@ -0,0 +1,89 @@
...
+            title: source.title,
+            snippet: source.description,
+            url: source.url,
+            source_domain,
+            date_published: source.published_at.naive_local(),
+            image: source.image,
+            rank: 0,
+            score: None,
+            country: String::new(),
+            language: String::new(),
```
janpetschexain, line 85: ❓ this might be an issue, in a different pr i asked if we can assume that this information can be used to create valid `Market`s (since we don't have any check on the `Market` constructor), i'm not sure though if this is actually exploited, so we need to double check. or better, to finally have a validity check on the `Market` constructor (in a different pr). @acrrd

discovery_engine_core/providers/src/newscatcher/models.rs
```diff
@@ -0,0 +1,203 @@
...
+        }
+    }
+}
+
+impl crate::Article {
+    /// Loads newscatcher articles from JSON and returns them in `Article` representation
+    pub fn load_from_newscatcher_json_representation(json: &str) -> Result<Vec<Self>, Error> {
+        let articles: Vec<Article> = serde_json::from_str(json).map_err(Error::Decoding)?;
+
+        Ok(articles.into_iter().map(Into::into).collect())
```
janpetschexain, line 173: :art: the result could be mapped directly (`map_into()` requires `Itertools`)
```suggestion
        serde_json::from_str::<Vec<Article>>(json)
            .map(|articles| articles.into_iter().map_into().collect())
            .map_err(Error::Decoding)
```

discovery_engine_core/tooling/bin/news_client.rs
```diff
@@ -12,133 +12,143 @@
...
+//     newscatcher,
+//     CommonQueryParts,
+//     Filter,
+//     GnewsHeadlinesQuery,
+//     GnewsNewsQuery,
+//     HeadlinesQuery,
+//     Market,
+// };
+
+//FIXME[TY-2810] Depending on what this tool is meant for fix it appropriately.
```
janpetschexain, line 29: afaik this tool was used for two purposes, 1) to experiment with the respective provider apis, and 2) to fetch some real data for the test fixtures.

if we don't need it anymore, we might just remove it, but it should be double checked with @x3ro

discovery_engine_core/providers/src/newscatcher/models.rs
```diff
@@ -0,0 +1,203 @@
...
+    }
+}
+
+/// Query response from the Newscatcher API
+#[derive(Deserialize, Debug)]
+pub(crate) struct Response {
```
janpetschexain, line 179: 🔧 this seems split from the query structs, i'd suggest to either move the query related structs here or this struct there (both would fit in one way or another)

discovery_engine_core/providers/src/lib.rs
```diff
@@ -25,18 +25,21 @@
...
+pub mod gnews;
+pub mod newscatcher;
 mod utils;
 
-pub use client::{Article, Client, Error};
-pub use gnews_client::{HeadlinesQuery as GnewsHeadlinesQuery, NewsQuery as GnewsNewsQuery};
-pub use newscatcher_client::{
+pub use client::Client;
+pub use gnews::{HeadlinesQuery as GnewsHeadlinesQuery, NewsQuery as GnewsNewsQuery};
+pub use newscatcher::{
```
janpetschexain, line 42: 🔧 this exports the items from those two modules, but at the same time these modules are public anyways, one of those options should be enough.

discovery_engine_core/providers/src/gnews/models.rs
```diff
@@ -0,0 +1,89 @@
...
+    #[serde(deserialize_with = "deserialize_null_default")]
+    pub name: String,
+
+    #[serde(deserialize_with = "deserialize_null_default")]
+    pub url: String,
+}
+
+/// Query response from the Gnews API
+#[derive(Deserialize, Debug)]
+pub(crate) struct Response {
```
janpetschexain, line 58: 🔧 this seems split from the query structs, i'd suggest to either move the query related structs here or this struct there (both would fit in one way or another)

discovery_engine_core/providers/src/gnews.rs
```diff
@@ -12,52 +12,231 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-use chrono::{DateTime, Utc};
+//! Client to get new documents.
```
janpetschexain, line 15: ❓ is there a specific reason to move the client related code to the parent module? instead of a module tree like
```
- gnews
  - models
```
i would have expected sth like
```
- gnews
  - client
  - models
```
or if we want to be more modular even sth like
```
- gnews
  - client
  - models
  - query
```
similarly for the newscatcher modules.

janpetschexain (commented):
some of the comments might be related to old code, but since we move it around anyways we could just change it right away.

discovery_engine_core/tooling/bin/news_client.rs
x3ro: Mmh, I think that having a tool that allows us to use our news client(s) from the command line would continue to be useful in the future. For example, I used this rather recently when downloading test data for elastic search. It's also a nice way of doing manual testing on large amounts of news data, to catch e.g. deserialisation problems that only happen occasionally. 

As a result, I'd "soft vote" to keep it, but also haven't taken a look at the changes, and thus at how difficult it would be to adapt it. Do you have a stronger opinion, @acrrd ?

discovery_engine_core/providers/src/gnews/models.rs
rustonaut: Agree, but outside of the scope of this PR.

I added: https://xainag.atlassian.net/browse/TY-2899

discovery_engine_core/tooling/bin/news_client.rs
rustonaut: So it wants to fetch the input to the providers crate, not the output.

This means it

- can use the "encode query parameters" part
- but NOT the rest of the fetching (as it does things like filling in default values)

Furthermore non of this code really takes advantage by any abstractions this lead to a situations where the code is in practice more complicated then a small bash snippet doing the fetching.

I would at most replace it with a bash script doing  `for x in {1..10}; do curl .... > ...; sleep 0.1;  done` or remove it completely.

discovery_engine_core/providers/src/newscatcher/models.rs
rustonaut: It's split because that is an internal implementation detail of the newscatcher provider while the query structs are a public interface used across all provider. Or at least after the next PR that is the case. This is also not a split I introduced.

So I would say wont fix in this PR as it's gone with the follow up PR anyway.

discovery_engine_core/providers/src/lib.rs
rustonaut: Not code I wrote, and will be changed anyway in follow up PR so wont fix in this PR.

discovery_engine_core/providers/src/gnews/models.rs
rustonaut: see same comment on newscatcher

discovery_engine_core/providers/src/gnews.rs
rustonaut: I guess it's a stylistic question, but I don't see the use for it as it only creates an "empty" gnews module which does nothing else but reexport the submodules.

Now putting this in context of the follow up PR:

- queries will be the same for all providers so there is no point for a `<provider>/query` module.
- models will be a purely internal implementation detail of the `gnews` provider
- everything else is the entry points of the module, we could move all of that into client, but I don't see any benefits in doing so, I mean it's like moving the whole main code of a module into a sub-module just because there are other sub-modules IMHO.

rustonaut:
> some of the comments might be related to old code,

yes, everything but one comment :rofl:

janpetschexain (approved):

rustonaut:
bors merge

PR #431, opened 2022-06-02
[Merged by Bors] - TY-2869 Move query time window to EndpointConfig

x3ro:
[TY-2869]

[TY-2869]: https://xainag.atlassian.net/browse/TY-2869?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine_core/providers/src/client.rs
+++ b/discovery_engine_core/providers/src/client.rs
@@ -110,6 +110,20 @@ pub struct NewsQuery<'a, F> {
     pub from: Option<String>,
 }
 
+impl<F> NewsQuery<'_, F> {
+    /// See [`NewsQuery::from`]
+    #[must_use]
+    pub fn max_article_age(mut self, days: usize) -> Self {
+        // (lj): This _could_ overflow if we specified trillions of days, but I don't
+        // think that's worth guarding against.
+        let days = days as i64;
+
+        let from = Utc::today() - chrono::Duration::days(days);
+        self.from = Some(from.format("%Y/%m/%d").to_string());
+        self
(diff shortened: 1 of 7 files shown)


discovery_engine_core/providers/src/client.rs
```diff
@@ -110,6 +110,20 @@ pub struct NewsQuery<'a, F> {
...
+    pub fn max_article_age(mut self, days: usize) -> Self {
+        // (lj): This _could_ overflow if we specified trillions of days, but I don't
+        // think that's worth guarding against.
+        let days = days as i64;
+
+        let from = Utc::today() - chrono::Duration::days(days);
+        self.from = Some(from.format("%Y/%m/%d").to_string());
+        self
+    }
+}
```
rustonaut, line 125: :art:, maybe :wrench: 

This is a reasonable approach but there are two reasons why I would do it slightly different:

- The query types are basically POD and we first explicitly set when/from to `None` and then to another value, it's kinda okay here but stylistically not grate and theoretically slightly bug prone, not that it matters here.

- `from`/`when` are very provider specific and the query structs are a bit less provider specific, through more so in the gnews branch then here

What I think would be better is to replace `from`/`when` with `max_age_days: Option<usize>` and then in the code which encodes the query parameters to the appropriate conversion specific to the query type :thinking:

rustonaut (approved):
lgtm

x3ro:
bors merge

PR #432, opened 2022-06-02
[Merged by Bors] - TY-2880 reacted sources repository

finiteprods:
**Summary**

adds `SourceReactedRepository`, a repository of source domains of documents that have been (dis)liked.

### DIFF
--- a/discovery_engine/lib/src/domain/models/source_reacted.dart
+++ b/discovery_engine/lib/src/domain/models/source_reacted.dart
@@ -1,0 +1,38 @@
+// Copyright 2022 Xayn AG
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, version 3.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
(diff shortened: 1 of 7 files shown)


discovery_engine/lib/src/domain/models/source_reacted.dart
```diff
@@ -0,0 +1,38 @@
...
+@HiveType(typeId: sourceReactedTypeId)
+class SourceReacted {
+  @HiveField(0)
+  final Source source;
+  @HiveField(1)
+  int weight;
+  @HiveField(2)
+  DateTime timestamp;
+  @HiveField(3)
+  final bool liked;
```
Robert-Steiner, line 33: ❔ Does `liked = false` mean that the source has been disliked? If so, we could also use an enum instead of a bool here. It would be similar to the `UserReaction` enum we use in the `Document` model.

finiteprods: that's right. I did think about about using `UserReaction` but it appears for the sources ranker, they're only interested in explicit likes and dislikes, so I thought a `bool` would suffice. would you still prefer a separate enum instead?

Robert-Steiner: I'm also fine with a bool but maybe we should add a comment that like = false means dislike

Robert-Steiner (approved):

discovery_engine/lib/src/domain/models/source_reacted.dart
finiteprods: sure will add a docstring

finiteprods:
bors merge

PR #433, opened 2022-06-02
TY-2895 rebase TY-2670 and follow ups onto main

rustonaut:
I'm not sure how this best should be reviewed as it's a re-based and as such you can't point it against a branch for the changes.
E.g. the only thing this branch can point against is `main` but it SHOULD NOT be merged into main and it means there is no differences between changes in context of rebase and other changes.

One think which should be looked at is the `bing` module as it had to be non-trivially changed when rebasing.

**References:**

- [TY-2895](https://xainag.atlassian.net/browse/TY-2895)


acrrd:
This should not be based on `main`.

rustonaut:
To avoid future confusion:

> This should not be based on main.

It must it is re-basing the `TY-2670` branch onto the head of main to include the latest improvements from main.

The reason it has a PR instead of force pushing to the branch is, that there are non trivial changes from resolving rebase conflicts. Mainly to the `bing` module. So a reviewer should review that modules changes.

Robert-Steiner:
> The reason it has a PR instead of force pushing to the branch is, that there are non trivial changes from resolving rebase conflicts. Mainly to the `bing` module (in `providers`). So a reviewer should review that modules changes.

I looked into the bing module + engine. looks good to me 👍

rustonaut:
ok then I merge the follow ups based on this into this and then force push it to the `TY-2670-implement-gnews` branch.

discovery_engine_core/providers/src/bing.rs
```diff
@@ -17,66 +17,48 @@
...
 }
 
-impl Query for TrendingQuery<'_> {
-    fn setup_url(&self, url: &mut Url) -> Result<(), Error> {
-        url.path_segments_mut()
-            .map_err(|_| Error::InvalidUrlBase(None))?
-            .push(URL_SUFFIX);
-
-        let query = &mut url.query_pairs_mut();
+/// Bing implementation for tending topics.
```
finiteprods, line 28:
```suggestion
/// Bing implementation for trending topics.
```

rustonaut:
As everything is on TY-26770-implement-gnews I will close this here.

PR #434, opened 2022-06-06
[Merged by Bors] - TY-2904 set in parameter for gnews

rustonaut:
Set `in` parameter for gnews/search to `title,description,content`.

**References:**

- [story TY-2810](https://xainag.atlassian.net/browse/TY-2810)
- [issue TY-2904](https://xainag.atlassian.net/browse/TY-2904)

**Bases on:**

- #433

PR #435, opened 2022-06-06
TY-2905 partially unify topic query parameter

rustonaut:
**References:**

- [story TY-2810](https://xainag.atlassian.net/browse/TY-2810)
- [issue TY-2905](https://xainag.atlassian.net/browse/TY-2905)

**Bases on:**

- #434


acrrd:
I think the only thing we should do here is to set 'breacking-news' for gnews when asking for the headlines.

rustonaut:
Replaced with simpler implementation:  #437

PR #450, opened 2022-06-15
[Merged by Bors] - TY-2924 repo methods used for cleaning up reacted sources

finiteprods:
**Summary**

adds 2 additional methods to the reacted sources repository to be used later for purging old sources.

### DIFF
--- a/discovery_engine/lib/src/infrastructure/repository/hive_source_reacted_repo.dart
+++ b/discovery_engine/lib/src/infrastructure/repository/hive_source_reacted_repo.dart
@@ -34,6 +34,27 @@ class HiveSourceReactedRepository implements SourceReactedRepository {
       box.values.where((source) => source.liked == like).toList();
 
   @override
+  Future<SourceReacted?> fetchMinWeightLiked() async {
+    final likes = box.values.where((source) => source.liked);
+    if (likes.isEmpty) return null;
+
+    return likes.reduce(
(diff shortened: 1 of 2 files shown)


discovery_engine/lib/src/infrastructure/repository/hive_source_reacted_repo.dart
```diff
@@ -33,6 +33,27 @@ class HiveSourceReactedRepository implements SourceReactedRepository {
   Future<List<SourceReacted>> fetchByReaction(bool like) async =>
       box.values.where((source) => source.liked == like).toList();
 
+  @override
+  Future<SourceReacted?> fetchMinWeightLiked() async {
```
janpetschexain, line 37: ❓ since the name is also used for the reduced variable, wdyt about naming the function in the same way? this would also better align the naming with the other new function `fetchOldestDisliked`.
```suggestion
  Future<SourceReacted?> fetchLightestLiked() async {
```

janpetschexain (approved):
lgtm, just a naming suggestion

discovery_engine/lib/src/infrastructure/repository/hive_source_reacted_repo.dart
finiteprods: it would indeed, sure.

finiteprods:
bors merge

PR #457, opened 2022-06-21
[Merged by Bors] - TY-2922 search and headlines use different market rank mappings

rustonaut:
(through both still use the same values for now).

### DIFF
--- a/discovery_engine_core/providers/src/filter.rs
+++ b/discovery_engine_core/providers/src/filter.rs
@@ -54,7 +54,7 @@ pub struct Market {
 }
 
 impl Market {
-    /// Returns the default news quality rank limit for given country for search news.
+    /// Returns the default quality rank limit for news.
     pub fn news_quality_rank_limit(&self) -> Option<usize> {
         #[allow(clippy::match_same_arms)]
         Some(match &*self.country_code {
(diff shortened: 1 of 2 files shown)


discovery_engine_core/providers/src/filter.rs
```diff
@@ -54,7 +54,7 @@ pub struct Market {
 }
 
 impl Market {
-    /// Returns the default news quality rank limit for given country.
+    /// Returns the default news quality rank limit for given country for search news.
```
finiteprods, line 57:
```suggestion
    /// Returns the default quality rank limit for news.
```

discovery_engine_core/providers/src/filter.rs
```diff
@@ -72,6 +72,25 @@ impl Market {
             _ => return None,
         })
     }
+
+    /// Returns the default news quality rank limit for given country for latest headlines.
```
finiteprods, line 76:
```suggestion
    /// Returns the default quality rank limit for latest headlines.
```

finiteprods (approved):
included a minor suggestion for shortening the docstrings a bit otherwise looks good.

rustonaut:
bors merge

PR #458, opened 2022-06-22
[Merged by Bors] - TY-2939 accept log path in config

rustonaut:
Allow configuring a log file into which logs will be written (formatted as one json blob per line).

As it was missing this also allows setting debug levels through env variables (see comments,
normally we set `DISCOVERY_ENGINE_LOG` to only set logging for cates/targets starting with
`xayn_` but the full `RUST_LOG` functionality is still available).

We need to be able to set the log level as some of the logging added in the follow up PR uses
the debug log level as it can become quite verbose.

**References:**

- [issue TY-2939](https://xainag.atlassian.net/browse/TY-2939)
- [story TY-2938](https://xainag.atlassian.net/browse/TY-2938)

### DIFF
--- a/discovery_engine_core/bindings/src/tracing.rs
+++ b/discovery_engine_core/bindings/src/tracing.rs
@@ -12,7 +12,7 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-//! Setup tracing on different platform.
+//! Setup tracing on different platforms.
 
 use std::{fs::OpenOptions, io, path::Path, sync::Once};
 
--- a/discovery_engine_core/bindings/src/tracing.rs
+++ b/discovery_engine_core/bindings/src/tracing.rs
@@ -39,8 +39,11 @@ fn init_tracing_once(log_file: Option<&Path>) {
     let android_logging;
     cfg_if::cfg_if! {
         if #[cfg(target_os = "android")] {
-            //TODO error
-            android_logging = tracing_android::layer("xayn_discovery_engine").ok()
+            android_logging = tracing_android::layer("xayn_discovery_engine")
+                .map_err(|err| {
+                    delayed_errors.push(Box::new(err) as _);
+                })
+                .ok()
         } else {
             // workaround to not have to specify a alternative type per hand
             android_logging = false.then(|| tracing_subscriber::fmt::layer())
@@ -63,7 +66,6 @@ fn init_tracing_once(log_file: Option<&Path>) {
         .transpose()
         .map_err(|err| {
             delayed_errors.push(Box::new(err) as _);
-            ()
         })
         .ok();
 
--- a/discovery_engine_core/bindings/src/tracing.rs
+++ b/discovery_engine_core/bindings/src/tracing.rs
@@ -67,11 +67,11 @@ fn init_tracing_once(log_file: Option<&Path>) {
         .init();
 
     for error in delayed_errors {
-        tracing::error!(error=%error, "logging setup failed");
+        tracing::error!(%error, "logging setup failed");
     }
 }
 
-/// Crates an [`EnvFilter`] based on the env of `RUST_LOG`.
+/// Creates an [`EnvFilter`] based on the env of `RUST_LOG`.
 ///
 /// If `RUST_LOG` is not set/empty `DISCOVERY_ENGINE_LOG` will
 /// be used instead in the form of `info,xayn_=${DISCOVERY_ENGINE_LOG}`,
--- a/discovery_engine_core/bindings/src/tracing.rs
+++ b/discovery_engine_core/bindings/src/tracing.rs
@@ -16,7 +16,8 @@
 
 use std::{fs::OpenOptions, io, path::Path, sync::Once};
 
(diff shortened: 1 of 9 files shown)


discovery_engine_core/bindings/src/tracing.rs
```diff
@@ -14,37 +14,105 @@
...
 
+    let stdout_log = tracing_subscriber::fmt::layer();
+
+    let android_logging;
     cfg_if::cfg_if! {
         if #[cfg(target_os = "android")] {
-            use tracing_subscriber::layer::SubscriberExt;
-            let layer = tracing_android::layer("xayn_discovery_engine").ok();
-            subscriber.with(layer).init();
+            android_logging = tracing_android::layer("xayn_discovery_engine").ok()
```
rustonaut, line 39: It makes no sense to add this error to delayed errors as:

- we most likely won't see the log in that case anyway
- it can only fail if the hard coded tag string  is changed to contain '\0'

discovery_engine_core/bindings/src/tracing.rs
```diff
@@ -14,37 +14,105 @@
...
         } else {
-            subscriber.init();
+            // workaround to not have to specify a alternative type per hand
+            android_logging = false.then(|| tracing_subscriber::fmt::layer())
```
rustonaut, line 42: the type inference around `Layered` and `Subscriber` is a bit iffy, so it's simpler to just use this small trick here to not need to specify a type hint. For the same reason I didn't move this or the file_log creation code into a standalone function.

discovery_engine_core/bindings/src/tracing.rs
```diff
@@ -14,37 +14,105 @@
...
     }
+}
 
-    log_panic();
+/// Crates an [`EnvFilter`] based on the env of `RUST_LOG`.
+///
+/// If `RUST_LOG` is not set/empty `DISCOVERY_ENGINE_LOG` will
+/// be used instead in the form of `info,xayn_=${DISCOVERY_ENGINE_LOG}`,
+/// if that fails `info` is used.
+///
```
rustonaut, line 81: Team Blue doesn't know about the syntax and concepts of `EnvFilter` and normally only want to set log levels less then `info` for `xayn_` crates.  At the same time  for us for debugging being able to enable `debug` or even `trace` messages for `tract`, `tokio` etc is quite use full for us.

As such I decided to use `RUST_LOG` and if it isn't set fallback to using `DISCOVERY_ENGINE_LOG` and only setting it for targets starting with `xayn_`.

This means if we use `RUST_LOG` it behaves as we expect and if Team Blue uses `DISCOVERY_ENGINE_LOG` it behaves like they expect.

discovery_engine_core/bindings/src/tracing.rs
```diff
@@ -14,37 +14,105 @@
...
-    let subscriber = tracing_subscriber::fmt()
-        .with_max_level(LevelFilter::INFO)
-        .finish();
+fn init_tracing_once(log_file: Option<&Path>) {
+    let mut delayed_errors = Vec::new();
 
+    let stdout_log = tracing_subscriber::fmt::layer();
+
+    #[allow(clippy::needless_late_init)]
+    let android_logging;
```
janpetschexain, line 37: 🔧 the `clippy::needless_late_init` can be avoided by moving the binding inside the macro. (i can't mark it as a suggestion because of the multiline diff)
```rust
    cfg_if::cfg_if! {
        if #[cfg(target_os = "android")] {
            let android_logging = tracing_android::layer("xayn_discovery_engine").ok();
        } else {
            // workaround to not have to specify a alternative type per hand
            let android_logging = false.then(tracing_subscriber::fmt::layer);
        }
    };
```

discovery_engine_core/bindings/src/tracing.rs
```diff
@@ -14,37 +14,105 @@
...
+
+    tracing_subscriber::registry()
+        .with(stdout_log)
+        .with(android_logging)
+        .with(file_log)
+        .with(filter)
+        .init();
+
+    for error in delayed_errors {
+        tracing::error!(error=%error, "logging setup failed");
```
janpetschexain, line 73: 🎨 the local variable can be used directly without a field name in case of same names
```suggestion
        tracing::error!(%error, "logging setup failed");
```

discovery_engine_core/bindings/src/tracing.rs
```diff
@@ -14,37 +14,105 @@
...
     }
+}
 
-    log_panic();
+/// Crates an [`EnvFilter`] based on the env of `RUST_LOG`.
```
janpetschexain, line 77: typo
```suggestion
/// Creates an [`EnvFilter`] based on the env of `RUST_LOG`.
```

janpetschexain (commented):
looks good, just some small issues.

the dartdoc issue in the ci seems to be unrelated?

acrrd:
I don't think the `DISCOVERY_ENGINE_LOG` part is needed for us. I'm not sure blue can set it in a way that we can use. Having that options exposed as and env var instead of a proper options in the configuration is very misleading and hidden to blue. 
What we can simply do is to use DEBUG if we have the file path and info otherwise.

finiteprods (approved):
lgtm!

janpetschexain:
bors merge

PR #461, opened 2022-06-24
Lucas' experiment w/ Dart code generation

x3ro:


x3ro:
I think it'd be mostly cleaning up the code (~1-2h of effort, I suppose), and I also wanted to get an opinion from @biernacki on whether this seems like a good idea :D

discovery_engine/lib/src/generators/multiplier_generator.dart
```diff
@@ -0,0 +1,191 @@
...
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+import 'package:analyzer/dart/element/element.dart';
+import 'package:build/build.dart';
+import 'package:source_gen/source_gen.dart';
+
+import 'annotations.dart';
+import 'class_visitor.dart';
+
+class MultiplierGenerator extends GeneratorForAnnotation<MapEvent> implements ElementVisitor<Element> {
```
janpetschexain, line 22: can this implementation be simplified such that we only override those methods which currently do more than a print statement?

it also looks like the [lib authors](https://pub.dev/documentation/analyzer/latest/dart_element_element/ElementVisitor-class.html) didn't intent to implement `ElementVisitor` directly, but one of its subclasses, but i guess we could easily rewrite this to eg implement `SimpleElementVisitor` with some overrides instead.

janpetschexain (commented):
this looks very nice indeed! i definitely like the idea of code generation over manual additions.

discovery_engine/lib/src/generators/multiplier_generator.dart
x3ro: Yep, I just didn't want to invest anymore time than necessary until I know we're going to use this. I would employ a similar approach as for the `class_visitor`, but I didn't know about that yet when I created this class :she

x3ro:
Cool, I'll clean this up and get it into a reviewable state then :)

discovery_engine/pubspec.yaml
```diff
@@ -27,6 +27,9 @@ dependencies:
       url: https://github.com/xaynetwork/xayn_async_bindgen.git
       path: async_bindgen_dart_utils
       ref: a299c69aacb1f017c0f062a83203b5f6225c462f
+  source_gen: ^1.2.2
+  build: ^2.3.0
+  analyzer: ^3.4.1
```
janpetschexain, line 32: ❓ could we also use the latest `4.2.0` or is this conflicting with other deps or blue?
```suggestion
  analyzer: ^4.2.0
```

discovery_engine/build.yaml
```diff
@@ -5,3 +5,20 @@ targets:
         options:
           any_map: true
           explicit_to_json: true
+
+      xayn_discovery_engine|property_multiply:
+        enabled: true
+#        generate_for:
+#          - lib/src/api/events/engine_events.dart
```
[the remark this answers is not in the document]
x3ro: ah yes, this is a leftover and can be removed

discovery_engine/pubspec.yaml
x3ro: I'll give it a try. I think analyzer was already being used, and I just used the same version 🤔 but I may be mistaken

discovery_engine/lib/src/generators/visitors/const_member_name_visitor.dart
```diff
@@ -0,0 +1,27 @@
...
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+import 'package:analyzer/dart/element/element.dart';
+import 'package:analyzer/dart/element/visitor.dart';
+
+class ConstMemberNameVisitor extends SimpleElementVisitor<void> {
+  var constMemberNames = List<String>.empty(growable: true);
```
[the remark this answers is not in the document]
biernacki: it is a fixed type, it's infered from `List<String>.empty()`. But we could change this to something like:

```dart
final constMemberNames = <String>[];
```
We can use `final` because we don't reassign the variable, we just mutate it.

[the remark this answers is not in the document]
janpetschexain: `final` is even better 👍

discovery_engine/pubspec.yaml
x3ro: Yep seems like this doesn't work because of `freezed`:

```
Because freezed 1.1.1 depends on analyzer ^3.0.0 and no versions of freezed match >1.1.1
  <2.0.0, freezed ^1.1.1 requires analyzer ^3.0.0.
So, because xayn_discovery_engine depends on both analyzer ^4.2.0 and freezed ^1.1.1,
  version solving failed.
```

I wouldn't want to upgrade freezed as part of this PR as well, so it'll have to remain outdated for the time being 😅

janpetschexain: iirc `freezed` is tied to that version because of blue, so we can't do much here. thx for giving it a try 👌

discovery_engine/build.yaml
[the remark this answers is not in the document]
x3ro: Forgot to commit and push that change yesterday 😅 Done now.

janpetschexain (approved):
lgtm 👍

biernacki (approved):
lgtm, great addition 💪

PR #464, opened 2022-06-27
[Merged by Bors] - TY-2971 impl fetch history

Robert-Steiner:
Ticket:

- [TY-2971]

Summary:

- moved `SqliteStorage` into separate file 
- added sql code for creating the `Document`, `HistoricDocument`, `NewsResource` and `NewscatcherData` tables
- added `fetch_history` to `Storage` trait and implemented the method on the `SqliteStorage` struct
- added a build script for [triggering recompilation on migration changes](https://docs.rs/sqlx/0.6.0/sqlx/macro.migrate.html#triggering-recompilation-on-migration-changes)

[TY-2971]: https://xainag.atlassian.net/browse/TY-2971?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine_core/core/src/document.rs
+++ b/discovery_engine_core/core/src/document.rs
@@ -54,6 +54,12 @@ impl Id {
     pub(crate) fn new() -> Self {
         Self(Uuid::new_v4())
     }
+
+    #[cfg(feature = "storage")]
+    #[cfg(test)]
+    pub(crate) fn as_uuid(&self) -> &Uuid {
+        &self.0
+    }
 }
 
 impl From<Uuid> for Id {
--- a/discovery_engine_core/core/src/document.rs
+++ b/discovery_engine_core/core/src/document.rs
@@ -55,8 +55,7 @@ impl Id {
         Self(Uuid::new_v4())
     }
 
-    #[cfg(feature = "storage")]
-    #[cfg(test)]
+    #[cfg(all(feature = "storage", test))]
     pub(crate) fn as_uuid(&self) -> &Uuid {
         &self.0
     }
--- a/discovery_engine_core/core/src/storage/migrations/20220627125903_init.sql
+++ b/discovery_engine_core/core/src/storage/migrations/20220627125903_init.sql
@@ -1,0 +1,52 @@
+--  Copyright 2022 Xayn AG
+--
+--  This program is free software: you can redistribute it and/or modify
+--  it under the terms of the GNU Affero General Public License as
+--  published by the Free Software Foundation, version 3.
+--
+--  This program is distributed in the hope that it will be useful,
+--  but WITHOUT ANY WARRANTY; without even the implied warranty of
+--  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+--  GNU Affero General Public License for more details.
+--
+--  You should have received a copy of the GNU Affero General Public License
+--  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+CREATE TABLE IF NOT EXISTS Document (
+    id BLOB NOT NULL
+        PRIMARY KEY
+);
+
+CREATE TABLE IF NOT EXISTS HistoricDocument (
+    document BLOB NOT NULL
+        PRIMARY KEY
+        REFERENCES Document(id) ON DELETE CASCADE
+);
+
+CREATE TABLE IF NOT EXISTS NewsResource (
+    document BLOB NOT NULL
+        PRIMARY KEY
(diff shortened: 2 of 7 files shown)


discovery_engine_core/core/src/document.rs
```diff
@@ -54,6 +54,12 @@ impl Id {
     pub(crate) fn new() -> Self {
         Self(Uuid::new_v4())
     }
+
+    #[cfg(feature = "storage")]
+    #[cfg(test)]
```
janpetschexain, line 59: :art: the attributes could be logically combined
```suggestion
    #[cfg(all(feature = "storage", test))]
```

discovery_engine_core/core/src/storage/sqlite.rs
```diff
@@ -0,0 +1,196 @@
...
+            title: String,
+        }
+
+        let mut con = self
+            .pool
+            .acquire()
+            .await
+            .map_err(Error::AcquireConnection)?;
+
+        sqlx::query_as::<_, _HistoricDocument>("SELECT nr.document, nr.url, nr.snippet, nr.title FROM HistoricDocument AS hd, NewsResource AS nr ON hd.document = nr.document;")
```
janpetschexain, line 98: ❓ the query tries to fetch the historic documents, but the sql string selects the news resource fields (with the same id), is this on purpose? i guess this is due to the way our sql tables are defined?

discovery_engine_core/core/src/storage/sqlite.rs
```diff
@@ -0,0 +1,196 @@
...
+        let storage = SqliteStorage::connect("sqlite::memory:").await.unwrap();
+        storage.init_database().await.unwrap();
+        let history = storage.fetch_history().await.unwrap();
+        assert!(history.is_empty());
+
+        let docs = create_historic_documents(10);
+        insert_historic_documents(&storage, docs.clone()).await;
+
+        let history = storage.fetch_history().await.unwrap();
+        assert_eq!(docs.len(), history.len());
```
janpetschexain, line 188: :art: i'd revert the argument order to match the usual `assert_eq!(actual, expected)`. similarly for the iterated assertions below.
```suggestion
        assert_eq!(history.len(), docs.len());
```

discovery_engine_core/core/src/storage/migrations/20220627125903_init.sql
```diff
@@ -0,0 +1,52 @@
...
+    topic TEXT NOT NULL,
+    url TEXT NOT NULL,
+    image TEXT,
+    -- format should be RFC3339/ISO8601 & sqlite compliant
+    datePublished TEXT NOT NULL,
+    -- implied by url, but allows us to easier implement
+    -- things like pruning history when excluding a source
+    source TEXT NOT NULL,
+    -- compound format <2-letter-lang><2-letter-state>
+    -- should be same as market primary key
```
janpetschexain, line 41: ❓ which market primary key is this referring to? there seems to be no primary key of this name in any of the tables.

janpetschexain (commented):
looks good, mainly some questions for my understanding

discovery_engine_core/core/src/storage/migrations/20220627125903_init.sql
Robert-Steiner: You are right. the table does not exit yet but it will be added later. the table is called `EnabledMarket` in the sql schema document.
Would it be better to remove the comments for now?

discovery_engine_core/core/src/storage/sqlite.rs
Robert-Steiner: yes exactly the `HistoricDocument` only has an id column. the idea is if we want to add/remove data then we only have to adjust the sql statement but not the tables. for example: let's say we also want to return the embedding then we just change the statement to 
`SELECT e.embedding, nr.document, nr.url, nr.snippet, nr.title FROM HistoricDocument AS hd, NewsResource AS nr, Embedding AS e ON hd.document = nr.document, hd.document = e.document;`

discovery_engine_core/core/src/storage/migrations/20220627125903_init.sql
janpetschexain: no, if it's added soon anyways, then just leave it there.

Robert-Steiner:
bors merge

Robert-Steiner:
bors merge

PR #465, opened 2022-06-28
[Merged by Bors] - TY-2963 refactor engine constructors

janpetschexain:
**References**

- [TY-2963]
- followed by #466

**Summary**

we need to have write access to the stack data when we read the configs in the `from_config()` constructor:
- deserialization was scattered across the constructors and is now concentrated in the `from_config()` constructor
- the `from_stack_data()` constructor now takes a map instead of a closure for the stack data and is renamed to `new()`, this removes redundancy between the three constructors `new()`/`from_state()`/`from_stack_data()` where the former two have been removed

[TY-2963]: https://xainag.atlassian.net/browse/TY-2963?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine_core/core/src/engine.rs
+++ b/discovery_engine_core/core/src/engine.rs
@@ -70,7 +70,7 @@ use crate::{
     ranker::Ranker,
     stack::{
         self,
-        exploration,
+        exploration::Stack as Exploration,
         filters::{
             filter_semantically,
             ArticleFilter,
@@ -141,7 +141,7 @@ pub struct Engine<R> {
     endpoint_config: EndpointConfig,
     core_config: CoreConfig,
     stacks: RwLock<HashMap<StackId, Stack>>,
-    exploration_stack: exploration::Stack,
+    exploration_stack: Exploration,
     ranker: R,
     request_after: usize,
     #[cfg(feature = "storage")]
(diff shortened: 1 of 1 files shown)


discovery_engine_core/core/src/engine.rs
```diff
@@ -1002,50 +934,58 @@ impl XaynAiEngine {
...
+
 #[derive(Serialize, Deserialize)]
 struct RankerState(Vec<u8>);
 
+impl RankerState {
+    fn deserialize<'a>(
+        &self,
+        builder: Builder<'a, AveragePooler>,
+    ) -> Result<Builder<'a, AveragePooler>, Error> {
+        builder.with_serialized_state(&self.0).map_err(Into::into)
```
biernacki, line 985: just a question for my benefit, could this be written like below and would it be more/less clear in your opinion? 

```rust
Ok(builder.with_serialized_state(&self.0)?)
```

biernacki (approved):
I very much like the changes 👍

discovery_engine_core/core/src/engine.rs
janpetschexain: it could very well be written in the suggested way. the `?` operator internally uses some error mapping to return the desired error type.

regarding readability, i personally prefer the sequential functional style, but this is very subjective of course. that means, for me it's clearer to read the line as "use the builder, then call some serialized state method, then do some error mapping, then return". with the `Ok()` wrapper i have to step back and forth while reading the line, which makes it more difficult imo, but again this is subjective.

there are some unofficial design patterns [like this page](https://rust-unofficial.github.io/patterns/intro.html), but i didn't find if the is an idiomatic way for this kind of error handling.

PR #466, opened 2022-06-28
[Merged by Bors] - TY-2963 stack mab config

janpetschexain:
**References**

- [TY-2963]
- requires #465
- followed by #467

**Summary**

- move `BoxedOps` into the `ops` module and implement the `Ops` trait for it
- assign MAB parameters to the respective stack data if they are present in the config

[TY-2963]: https://xainag.atlassian.net/browse/TY-2963?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #467, opened 2022-06-28
[Merged by Bors] - TY-2964 exploration stack config

janpetschexain:
**References**

- [TY-2964]
- requires #466

**Summary**

- move the exploration stack config to the config module
- read and pass exploration stack config from json or fall back to default values

[TY-2964]: https://xainag.atlassian.net/browse/TY-2964?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine_core/core/src/config.rs
+++ b/discovery_engine_core/core/src/config.rs
@@ -156,6 +156,24 @@ impl Default for CoreConfig {
     }
 }
 
+/// Configurations for the exploration stack.
+#[derive(Debug)]
+pub(crate) struct ExplorationConfig {
+    /// The number of candidates.
+    pub(crate) number_of_candidates: usize,
+    /// The maximum number of documents to keep.
+    pub(crate) max_selected_docs: usize,
+}
+
+impl Default for ExplorationConfig {
+    fn default() -> Self {
+        Self {
+            number_of_candidates: 40,
+            max_selected_docs: 20,
+        }
+    }
+}
+
 /// Reads the DE configurations from json and sets defaults for missing fields (if possible).
 pub(crate) fn de_config_from_json(json: &str) -> Figment {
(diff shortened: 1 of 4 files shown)


discovery_engine_core/core/src/config.rs
```diff
@@ -164,11 +185,16 @@ pub(crate) fn de_config_from_json(json: &str) -> Figment {
         .join(Serialized::defaults(CoiSystemConfig::default()))
         .join(Serialized::default("core", CoreConfig::default()))
         .join(Serialized::default("endpoint", EndpointConfig::default()))
+        .join(Serialized::default(
+            &Exploration::id().to_string(),
+            ExplorationConfig::default(),
+        ))
```
finiteprods, line 191: :grey_question:  not so familiar with `figment` and not related to your change, but just wondering why (unlike for the other configs) there is no key here for `CoiSystemConfig`?

finiteprods (approved):
looks good :+1:

discovery_engine_core/core/src/config.rs
janpetschexain: this is due to the way the `CoiSystemConfig` is layed out. it contains to fields `coi` and `kpe` which themselves hold configs. when reading the via `Figment`s, the whole json string is treated a the readable object and the two keys are mapped to the subconfigs of the `CoiSystemConfig`. this is also why it uses `Serialized::defaults(value)` instead of the `Serialized::default(key, value)` (mind the trailing "s").
we could also key the `CoiSystemConfig`, but then we would just need another nesting in the json config, which we avoid by the current handling.

janpetschexain:
bors merge

PR #468, opened 2022-06-28
TY-2970 Modify the release CI to publish on private pubdev

x3ro:
[TY-2970]


.github/actions/setup-job-docker/action.yml
```diff
@@ -39,7 +47,25 @@ runs:
...
     - name: dart deps
       shell: bash
       if: ${{ inputs.dart != 'false' }}
       run: just dart-deps
+
+    - name: bindgen dart deps
+      shell: bash
+      working-directory: async_bindgen
+      if: ${{ inputs.bindgen_dart != 'false' }}
+      run: just dart-deps
```
rustonaut, line 71: This duplicates the previous step, probably not what we want.

justfile
```diff
@@ -27,14 +27,17 @@ default:
...
 # Gets/updates dart deps
-dart-deps:
+_dart-deps $WORKSPACE:
     #!/usr/bin/env sh
     set -eux
-    cd "$DART_WORKSPACE";
-    dart pub get
-    cd example
+    cd "$WORKSPACE";
     dart pub get
```
rustonaut, line 34: We can change this into a one liner: `cd "$WORKSPACE" && dart pub get`

justfile
```diff
@@ -79,14 +76,17 @@ rust-fmt:
...
+_dart-analyze $WORKSPACE:
+    cd "$WORKSPACE"; \
+    dart analyze --fatal-infos
+
 # Checks dart code, fails on info on CI
 dart-check: dart-build
-    cd "$DART_WORKSPACE"; \
-    dart analyze --fatal-infos
+    @{{just_executable()}} _dart-analyze "$DART_WORKSPACE"
+    @{{just_executable()}} _dart-analyze "$BINDGEN_DART_WORKSPACE"
```
rustonaut, line 86: :wrench: ? maybe if we are at it we should add the dart example workspace

discovery_engine_core/Cargo.toml
```diff
@@ -1,19 +1,18 @@
...
 
 [patch.crates-io.async-bindgen]
-git = "https://github.com/xaynetwork/xayn_async_bindgen.git"
-branch = "main"
+path = "./async-bindgen/async-bindgen"
 
 [patch.crates-io.async-bindgen-gen-dart]
-git = "https://github.com/xaynetwork/xayn_async_bindgen.git"
-branch = "main"
+path = "./async-bindgen/async-bindgen-gen-dart"
```
rustonaut, line 15: we should just change the dependencies in the package to same workspace path dependencies, the reason this are patch dependencies is because we originally considered publishing the async-bindgen tools

discovery_engine_flutter/example/ios/Runner/Info.plist
```diff
@@ -43,5 +43,7 @@
 	</array>
 	<key>UIViewControllerBasedStatusBarAppearance</key>
 	<true/>
+	<key>CADisableMinimumFrameDurationOnPhone</key>
+	<true/>
```
rustonaut, line 47: This prevents the example app to be limited to 60Hz, it seems that scrolling on a 120Hz iPhone can feel laggy without that option(?).

It either way only affects the example.

My guess is it was auto-added by some tool.

.github/actions/setup-job-docker/action.yml
rustonaut: I removed it I also removed the copy of the async-bindgen binary (which doesn't work like that anymore, it's now handled implicitly by rust-cach)

rustonaut:
Ok it should now work, I tested it by building the flutter example with dependencies from the private repo.

The changes from Lucas are either approved by me or I fixed them in follow up commits. Someone still needs to review my follow up commits.

finiteprods (approved):
lgtm apart from a few typos

discovery_engine_core/bindings/.gitignore
```diff
@@ -1,4 +1,4 @@
 # Omit generated files
-# DELETE_AFTER_THIS_IN_RELEASE
+#NOTE if we want to publish this to crates.io auto generated fiels need some special care
```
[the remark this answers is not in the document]
rustonaut: The build process would need to work without someone calling a `just` file.

So we need to make sure any generated files are included in the package, which they currently are not due to .gitignore.

For this we need to use `package.include` or `package.exclude` (which like .pubignore _fully_ replaces .gitignore wrt. package publishing).

Then we probably would also strip out the build.rs script, because when we ship it as a package there is no point in ad-hoc generating a c header which isn't used anyway (through we can bundle in the C header generated by CI if we want to).

We also probably then want to not push a Cargo.toml with the no longer needed build dependencies...

Anyway all of this is:
- way out of scope of what you should place into a comment in a .gitignore file
- irrelevant as hopefully we switched to the message passing API long before we consider seriously publishing this on crates.io, in which all the things are different anyway, preferably gone.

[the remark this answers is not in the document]
rustonaut: I updated the comment.

rustonaut:
bors merge

PR #469, opened 2022-06-29
[Merged by Bors] - TY-2988 Events for setting excluded and trusted sources in batches

biernacki:
**References**
- [TY-2986]
- [TY-2988]

**Summary**
- added events for setting/overriding trusted and excluded sources
- added unit test to cover writing, detecting duplicates, and overriding sources 

The usecase the blue team wants to cover is to be able to set multiple trusted and excluded sources in one go (the settings screen), so that the user doesn't have to wait for processing multiple events (one event per source).

[TY-2988]: https://xainag.atlassian.net/browse/TY-2988?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[TY-2986]: https://xainag.atlassian.net/browse/TY-2986?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


discovery_engine/lib/src/domain/feed_manager.dart
```diff
@@ -163,6 +164,41 @@ class FeedManager {
...
+
+    if (duplicates.isNotEmpty) {
+      return EngineEvent.setSourcesRequestFailed(duplicates);
+    }
+
+    final history = await _docRepo.fetchHistory();
+    await _sourcePreferenceRepository.clear();
+    await _sourcePreferenceRepository.saveAll(entries);
+    await _engine.setTrustedSources(history, trustedSources);
+    await _engine.setExcludedSources(history, excludedSources);
```
Robert-Steiner, line 194: 📌 The solution is better than the current one, but I think we can make it a little bit better (but not in this pr because it probably require some more changes). 
two things that concern me:
- when we call `setTrustedSources` and `setExcludedSources` we still update the stack two times
- we call `setTrustedSources` and `setExcludedSources` (and update the stack) even though only one or none (depends on whether blue only calls it when the sources have changed) of them has changed

Robert-Steiner (approved):

discovery_engine/lib/src/api/events/engine_events.dart
```diff
@@ -185,6 +185,21 @@ class EngineEvent with _$EngineEvent {
...
+  @Implements<FeedEngineEvent>()
+  const factory EngineEvent.setSourcesRequestSucceeded({
+    required Set<Source> trustedSources,
+    required Set<Source> excludedSources,
+  }) = SetSourcesRequestSucceeded;
+
+  /// Event created as a success response to SetSourcesRequested event.
+  /// Returns a set of duplicates [Source]s back to the client.
+  @Implements<FeedEngineEvent>()
+  const factory EngineEvent.setSourcesRequestFailed(
```
frankpepermans, line 199: Can this operation then only fail if we have duplicates?
What if it fails for other scenarios?

frankpepermans: Ah nevermind, then it's an EngineExceptionRaised :)
Saw it below, so ignore this

discovery_engine/lib/src/domain/feed_manager.dart
biernacki: Good point 👍 I've added a value equality check for sets based on implementation found in flutter's sdk, so now we call the engine only when the current set and the new one are different (added a test too). Please check if that covers the second case.

Robert-Steiner: > Please check if that covers the second case.

Nice! yes it does 👍

Robert-Steiner (approved):

biernacki:
bors merge

PR #472, opened 2022-06-29
[Merged by Bors] - TY-2976 db code for feed 

Robert-Steiner:
Ticket

- [TY-2976]

[TY-2976]: https://xainag.atlassian.net/browse/TY-2976?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine_core/core/src/engine.rs
+++ b/discovery_engine_core/core/src/engine.rs
@@ -170,7 +170,6 @@ where
         sources: &[WeightedSource],
         mut stack_data: HashMap<StackId, StackData>,
         stack_ops: Vec<BoxedOps>,
-        client: Arc<Client>,
         #[cfg(feature = "storage")] storage: BoxedStorage,
         providers: Providers,
     ) -> Result<Self, Error> {
--- a/discovery_engine_core/core/src/engine.rs
+++ b/discovery_engine_core/core/src/engine.rs
@@ -94,6 +94,13 @@ use crate::{
     },
 };
 
+#[cfg(feature = "storage")]
+type BoxedStorage = Box<
+    dyn Storage<StorageError = sqlite::StorageError, FeedScopeError = sqlite::FeedScopeError>
+        + Send
+        + Sync,
+>;
+
 /// Discovery engine errors.
 #[derive(Error, Debug, Display)]
 pub enum Error {
@@ -135,7 +142,7 @@ pub enum Error {
 
     #[cfg(feature = "storage")]
     /// Storage error: {0}.
-    Storage(#[from] sqlite::Error),
+    Storage(#[from] sqlite::StorageError),
 
     /// Provider error: {0}
     ProviderError(#[source] xayn_discovery_engine_providers::Error),
@@ -151,7 +158,7 @@ pub struct Engine<R> {
     request_after: usize,
     #[cfg(feature = "storage")]
     #[allow(dead_code)]
-    storage: Box<dyn Storage<StorageError = sqlite::Error> + Send + Sync>,
+    storage: BoxedStorage,
     providers: Providers,
 }
 
@@ -173,9 +180,8 @@ where
         sources: &[WeightedSource],
         mut stack_data: HashMap<StackId, StackData>,
         stack_ops: Vec<BoxedOps>,
-        #[cfg(feature = "storage")] storage: Box<
-            dyn Storage<StorageError = sqlite::Error> + Send + Sync,
-        >,
+        client: Arc<Client>,
+        #[cfg(feature = "storage")] storage: BoxedStorage,
         providers: Providers,
     ) -> Result<Self, Error> {
         let stacks = stack_ops
--- a/discovery_engine_core/core/src/engine.rs
+++ b/discovery_engine_core/core/src/engine.rs
@@ -52,10 +52,7 @@ use xayn_discovery_engine_providers::{
 use xayn_discovery_engine_tokenizer::{AccentChars, CaseChars};
 
 #[cfg(feature = "storage")]
-use crate::storage::{
-    sqlite::{self, SqliteStorage},
-    Storage,
-};
+use crate::storage::{self, sqlite::SqliteStorage, BoxedStorage, Storage};
 use crate::{
     config::{de_config_from_json, CoreConfig, EndpointConfig, ExplorationConfig, InitConfig},
     document::{
@@ -93,13 +90,6 @@ use crate::{
         TrustedNews,
     },
 };
-
-#[cfg(feature = "storage")]
-type BoxedStorage = Box<
-    dyn Storage<StorageError = sqlite::StorageError, FeedScopeError = sqlite::FeedScopeError>
-        + Send
-        + Sync,
->;
 
 /// Discovery engine errors.
 #[derive(Error, Debug, Display)]
@@ -142,7 +132,7 @@ pub enum Error {
 
     #[cfg(feature = "storage")]
     /// Storage error: {0}.
-    Storage(#[from] sqlite::StorageError),
+    Storage(#[from] storage::Error),
 
     /// Provider error: {0}
     ProviderError(#[source] xayn_discovery_engine_providers::Error),
--- a/discovery_engine_core/core/src/storage.rs
+++ b/discovery_engine_core/core/src/storage.rs
@@ -68,7 +68,7 @@ pub mod models {
     }
 
     pub struct ApiDocumentView {
-        pub document: document::Id,
+        pub document_id: document::Id,
         pub news_resource: NewsResource,
         pub newscatcher_data: NewscatcherData,
         pub user_reacted: Option<UserReaction>,
@@ -103,6 +103,7 @@ pub mod models {
         /// The domain of the article's source, e.g. `example.com`. Not a valid URL.
         pub source: String,
 
+        /// The market of news.
         pub market: Market,
     }
 
--- a/discovery_engine_core/core/src/storage.rs
+++ b/discovery_engine_core/core/src/storage.rs
@@ -42,7 +42,7 @@ pub(crate) trait Storage {
 
 #[async_trait]
 pub(crate) trait FeedScope {
-    async fn close_document(&self, document: document::Id) -> Result<(), Error>;
+    async fn close_document(&self, document: &document::Id) -> Result<(), Error>;
 
     async fn clear(&self) -> Result<(), Error>;
 
--- a/discovery_engine_core/core/src/storage.rs
+++ b/discovery_engine_core/core/src/storage.rs
@@ -38,12 +38,11 @@ pub trait Storage: FeedScope {
 pub trait FeedScope {
     type FeedScopeError;
 
-    async fn close_feed_document(&self, document: document::Id)
-        -> Result<(), Self::FeedScopeError>;
+    async fn close_document(&self, document: document::Id) -> Result<(), Self::FeedScopeError>;
 
-    async fn clear_feed(&self) -> Result<(), Self::FeedScopeError>;
+    async fn clear(&self) -> Result<(), Self::FeedScopeError>;
 
-    async fn fetch_feed(&self) -> Result<Vec<ApiDocumentView>, Self::FeedScopeError>;
+    async fn fetch(&self) -> Result<Vec<ApiDocumentView>, Self::FeedScopeError>;
 
     // helper function. will be replaced later by move_from_stacks_to_feed
     async fn store_documents(
--- a/discovery_engine_core/core/src/storage.rs
+++ b/discovery_engine_core/core/src/storage.rs
@@ -14,16 +14,101 @@
 
 use async_trait::async_trait;
 
-use crate::document::HistoricDocument;
+use crate::document::{self, HistoricDocument};
+
+use self::models::ApiDocumentView;
 
 pub mod sqlite;
 
 #[async_trait]
-pub trait Storage {
+pub trait Storage: FeedScope {
     type StorageError;
 
     async fn init_database(&self) -> Result<(), <Self as Storage>::StorageError>;
 
     async fn fetch_history(&self)
         -> Result<Vec<HistoricDocument>, <Self as Storage>::StorageError>;
+
+    fn feed(
+        &self,
+    ) -> &(dyn FeedScope<FeedScopeError = <Self as FeedScope>::FeedScopeError> + Send + Sync);
+}
+
+#[async_trait]
+pub trait FeedScope {
+    type FeedScopeError;
+
+    async fn close_feed_document(&self, document: document::Id)
+        -> Result<(), Self::FeedScopeError>;
+
+    async fn clear_feed(&self) -> Result<(), Self::FeedScopeError>;
+
+    async fn fetch_feed(&self) -> Result<Vec<ApiDocumentView>, Self::FeedScopeError>;
+
+    // helper function. will be replaced later by move_from_stacks_to_feed
+    async fn store_documents(
+        &self,
+        documents: &[document::Document],
+    ) -> Result<(), Self::FeedScopeError>;
+}
+
+pub mod models {
+
+    use chrono::NaiveDateTime;
+    use url::Url;
+    use xayn_discovery_engine_ai::Embedding;
+    use xayn_discovery_engine_providers::Market;
+
+    use crate::document::{self, UserReaction};
+
+    pub struct NewDocument {
+        pub id: document::Id,
+        pub news_resource: NewsResource,
+        pub newscatcher: NewscatcherData,
+        pub embedding: Embedding,
+    }
+
+    pub struct ApiDocumentView {
+        pub document: document::Id,
+        pub news_resource: NewsResource,
+        pub newscatcher_data: NewscatcherData,
+        pub user_reacted: Option<UserReaction>,
+        // //FIXME I don't think this is helpful as multiple documents in the vec can have the same value for this!
+        pub in_batch_index: usize,
+    }
+
+    /// Represents a news that is delivered by an external content API.
+    #[derive(Debug, Clone)]
+    pub struct NewsResource {
+        /// Title of the resource.
+        pub title: String,
+
+        /// Snippet of the resource.
+        pub snippet: String,
+
+        /// Main topic of the publisher.
+        pub topic: String,
+
+        /// Url to reach the resource.
+        pub url: Url,
+
+        /// Image attached to the news.
+        pub image: Option<Url>,
+
+        /// Publishing date.
+        //FIXME it's NativeDateTime in the current codebase but we can't compare
+        //      NativeDateTimes across different markets, but we do! So this needs to be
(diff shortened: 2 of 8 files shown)


discovery_engine_core/core/src/storage/sqlite.rs
```diff
@@ -84,113 +92,301 @@ impl Storage for SqliteStorage {
...
+
+    async fn store_documents(
+        &self,
+        documents: &[document::Document],
+    ) -> Result<(), Self::FeedScopeError> {
+        if documents.is_empty() {
+            return Ok(());
+        }
+
+        // const BIND_LIMIT: usize = 32766;
```
Robert-Steiner, line 242: Do we need such check or is it unrealistic that we will ever have so many documents?

discovery_engine_core/core/src/storage.rs
```diff
@@ -14,16 +14,101 @@
...
+        pub url: Url,
+
+        /// Image attached to the news.
+        pub image: Option<Url>,
+
+        /// Publishing date.
+        //FIXME it's NativeDateTime in the current codebase but we can't compare
+        //      NativeDateTimes across different markets, but we do! So this needs to be
+        //      at least a Utc DateTime.
+        pub date_published: NaiveDateTime,
```
Robert-Steiner, line 102: To keep things simpler I will keep the NaiveDateTime for now and change it in a later pr.

discovery_engine_core/core/src/storage.rs
```diff
@@ -14,16 +14,101 @@
...
+
+        /// The domain of the article's source, e.g. `example.com`. Not a valid URL.
+        pub source: String,
+
+        pub market: Market,
+    }
+
+    pub struct NewscatcherData {
+        pub domain_rank: usize,
```
Robert-Steiner, line 111: sqlx doesn't support `usize`. shall we change it to `u32`?

discovery_engine_core/core/src/storage/sqlite.rs
acrrd: We are not cleaning anything so it is currently possible. Is this limit only for 32bit bindings or every binding?
Anyway, panic is not acceptable in this case, we can take the first `BIND_LIMIT` documents and discard the others.

acrrd:
I would say that the associated type are quite a pain. If we use a boxed generic error for the low level errors we can remove them.
I'm not sure we are interested in matching them, and if we are the implementation can return them as logical error instead.

discovery_engine_core/core/src/storage.rs
```diff
@@ -14,16 +14,101 @@
...
+    fn feed(
+        &self,
+    ) -> &(dyn FeedScope<FeedScopeError = <Self as FeedScope>::FeedScopeError> + Send + Sync);
+}
+
+#[async_trait]
+pub trait FeedScope {
+    type FeedScopeError;
+
+    async fn close_feed_document(&self, document: document::Id)
```
acrrd, line 41: This are all operations on the Feed trait, we don't have to repeat feed in each of them

discovery_engine_core/core/src/storage/sqlite.rs
```diff
@@ -15,67 +15,75 @@
...
-    Fetch(#[source] sqlx::Error),
+    /// Sql error: {0}
+    Sql(#[source] sqlx::Error),
 
     /// Failed to covert type: {0}
     TypeConversion(#[source] GenericError),
 }
 
+#[derive(Error, Debug, Display)]
+pub enum StorageError {
```
Robert-Steiner, line 53: @rustonaut 
I can see the benefit of splitting the errors for different traits. For example if we have one big error enum, we would have to match against all error variants, even though some may never occur.

```
pub enum StorageError {
    /// Database error: {0}
    Database(#[from] DatabaseError),
    History1(...),
    History2(...),
    Feed1(...),
    Feed2(...),
}

match storage.fetch_history() {
	Database =>
	History1 => 
	History2 => 
	Feed1 & Feed2 will never occur but have to handle them 
}
```

but yes, the question is whether we want to put any other checks into the storage implementation at all (like checking uuid, etc.)

discovery_engine_core/core/src/storage/migrations/20220628114905_feed.sql
```diff
@@ -0,0 +1,54 @@
...
+
+CREATE INDEX IF NOT EXISTS idx_presentation_ordering_sort
+  ON PresentationOrdering(timestamp, inBatchIndex);
+
+CREATE TABLE IF NOT EXISTS UserReaction (
+    document BLOB NOT NULL
+        PRIMARY KEY
+        REFERENCES HistoricDocument(document) ON DELETE CASCADE,
+
+    userReaction INTEGER NOT NULL DEFAULT 0
```
acrrd, line 53: Why do we want a default here?

discovery_engine_core/core/src/storage/migrations/20220628114905_feed.sql
```diff
@@ -0,0 +1,54 @@
...
+    -- index in the batch of document which where
+    -- presented to the app (user) at the same time
+    inBatchIndex INTEGER NOT NULL
+);
+
+CREATE INDEX IF NOT EXISTS idx_presentation_ordering_sort
+  ON PresentationOrdering(timestamp, inBatchIndex);
+
+CREATE TABLE IF NOT EXISTS UserReaction (
+    document BLOB NOT NULL
```
acrrd, line 49: maybe `document_id`? It could be a serialized document otherwise

discovery_engine_core/core/src/storage/migrations/20220628114905_feed.sql
```diff
@@ -0,0 +1,54 @@
...
+);
+
+CREATE TRIGGER IF NOT EXISTS feed_history_sync
+    BEFORE INSERT
+    ON FeedDocument
+    BEGIN
+      INSERT INTO HistoricDocument(document)
+        VALUES (new.document)
+        ON CONFLICT DO NOTHING;
+    END;
```
acrrd, line 28: Will this much code to do this?
I would prefer to have all the database logic in one 'place' and not split between rust and sql

discovery_engine_core/core/src/storage.rs
```diff
@@ -14,16 +14,101 @@
...
-use crate::document::HistoricDocument;
+use crate::document::{self, HistoricDocument};
+
+use self::models::ApiDocumentView;
 
 pub mod sqlite;
 
 #[async_trait]
-pub trait Storage {
+pub trait Storage: FeedScope {
```
acrrd, line 24: Why we want this if we have the method `feed`?

discovery_engine_core/core/src/storage/sqlite.rs
Robert-Steiner: > We are not cleaning anything so it is currently possible. Is this limit only for 32bit bindings or every binding?

looks like for every binding. but the version is important SQLite prior to 3.32.0 only allowed [999](https://docs.rs/sqlx-core/latest/sqlx_core/query_builder/struct.QueryBuilder.html#note-database-specific-limits)

discovery_engine_core/core/src/storage.rs
Robert-Steiner: I had to add this in order to write `FeedScope<FeedScopeError = <Self as FeedScope>::FeedScopeError>`

discovery_engine_core/core/src/storage.rs
```diff
@@ -14,16 +14,101 @@
...
+        pub embedding: Embedding,
+    }
+
+    pub struct ApiDocumentView {
+        pub document: document::Id,
+        pub news_resource: NewsResource,
+        pub newscatcher_data: NewscatcherData,
+        pub user_reacted: Option<UserReaction>,
+        // //FIXME I don't think this is helpful as multiple documents in the vec can have the same value for this!
+        pub in_batch_index: usize,
```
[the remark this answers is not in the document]
rustonaut: yes

discovery_engine_core/core/src/storage/migrations/20220628114905_feed.sql
Robert-Steiner: renamed it to `documentId`

discovery_engine_core/core/src/storage/sqlite.rs
```diff
@@ -15,182 +15,448 @@
...
+            .await
+            .map_err(|err| storage::Error::Database(err.into()))?;
+
+        // insert data into NewscatcherData table
+        let mut query_builder =
+            QueryBuilder::new("INSERT INTO NewscatcherData (documentId, domainRank, score) ");
+        query_builder.push_values(documents.clone(), |mut stm, doc| {
+            stm.push_bind(doc.id.as_uuid())
+                .push_bind(doc.resource.rank as u32)
+                .push_bind(doc.resource.score.unwrap_or_default());
```
Robert-Steiner, line 297: Can we use `unwrap_or_default` here? in the NewscatcherData table we define `score` as not null so we need a value or we allow it to be null.

discovery_engine_core/core/src/storage/sqlite.rs
```diff
@@ -15,182 +15,448 @@
...
+            .execute(&mut tx)
+            .await
+            .map_err(|err| storage::Error::Database(err.into()))?;
+
+        // insert data into NewscatcherData table
+        let mut query_builder =
+            QueryBuilder::new("INSERT INTO NewscatcherData (documentId, domainRank, score) ");
+        query_builder.push_values(documents.clone(), |mut stm, doc| {
+            stm.push_bind(doc.id.as_uuid())
+                .push_bind(doc.resource.rank as u32)
```
Robert-Steiner, line 296: in `document::NewsResource`the type of rank is `u64` should we cast here or should we change the type in the NewscatcherData table to be `BIGINT`?

discovery_engine_core/core/src/storage/sqlite.rs
rustonaut: Note that this is the bind limit, not db size limit. 

I.e. in this context it's the max number of documents we can add to the feed at once.

We can ignore it here I think, even if we are on a older sqlite version.


(but shouldn't panic)

discovery_engine_core/core/src/storage/sqlite.rs
```diff
@@ -84,113 +92,301 @@ impl Storage for SqliteStorage {
...
+        let mut query_builder = QueryBuilder::new("INSERT INTO HistoricDocument (document) ");
+        query_builder.push_values(documents.iter(), |mut stm, doc| {
+            stm.push_bind(doc.id.as_uuid());
+        });
+        query_builder
+            .build()
+            .persistent(false)
+            .execute(&mut tx)
+            .await
+            .map_err(DatabaseError::Sql)?;
```
rustonaut: Not needed, there is a trigger which does this automatically for you.

The only reason there is no such trigger for Document is because the design long term plans to move
documents from stacks to the feed instead.

discovery_engine_core/core/src/storage/sqlite.rs
rustonaut: no that table should have a nullable value, the score only applies to newscatcher documents from search not from headlines.

discovery_engine_core/core/src/storage/sqlite.rs
rustonaut: I don't think it makes ever sense to exhaustively match error variants, I mean we generally hardly match errors as most of them are bugs anyway.

discovery_engine_core/core/src/storage/sqlite.rs
```diff
@@ -84,113 +92,301 @@ impl Storage for SqliteStorage {
...
             })
-            .collect()
+        })
+        .collect()
+    }
+
+    async fn store_documents(
+        &self,
+        documents: &[document::Document],
+    ) -> Result<(), Self::FeedScopeError> {
```
rustonaut: We probably should have a sqlite specific helper function which given a transaction and a list of documents will  insert them into `Documents`, `NewsResource` and `NewscatcherData`. So that we then can use it here and in the `ActiveDocumentScope`.

discovery_engine_core/core/src/storage/migrations/20220628114905_feed.sql
rustonaut: I would prefer triggers for this use case tbh. it's faster and makes things more reliable.

Especially if we e.g. move documents from the stack to the feed.

discovery_engine_core/core/src/storage/sqlite.rs
```diff
@@ -15,67 +15,75 @@
...
         let opt = SqliteConnectOptions::from_str(uri)
-            .map_err(Error::Init)?
+            .map_err(DatabaseError::Sql)?
             .create_if_missing(true);
 
         let pool = SqlitePoolOptions::new()
             .connect_with(opt)
             .await
-            .map_err(Error::Init)?;
+            .map_err(DatabaseError::Sql)?;
```
rustonaut, line 57: do we need to enable foreign key constraints or is that a default for `SqliteConnectionOptions` (it isn't a default for sqlite3 by itself).

Robert-Steiner: it is enabled by [default](https://docs.rs/sqlx/latest/sqlx/sqlite/struct.SqliteConnectOptions.html#method.foreign_keys) but good point i forgot to check that

discovery_engine_core/core/src/engine.rs
```diff
@@ -96,6 +96,13 @@ use crate::{
     },
 };
 
+#[cfg(feature = "storage")]
+type BoxedStorage = Box<
+    dyn Storage<StorageError = sqlite::StorageError, FeedScopeError = sqlite::FeedScopeError>
+        + Send
+        + Sync,
+>;
```
rustonaut: I have thought about it some more.

To avoid verbosity/accidental complexity we probably would want to use roughly the same error type (or same type per scope) for all backends, e.g. sqlite and postgres. Naturally that won't work for the error variant for the sqlite error but lets ignore that for a moment.

The reason I'm saying this is because if we don't do so we need to be generic over the error in any place we want to potentially match over it... which would be quite annoying.

If we use the same error type(s) anyway we could just not make the traits generic over their error type. I mean the trait is only used internally so why not.

I think any db specific error we expect as part of the domain logic we should convert to it's own error variant, any error left is a but we anyway can only report back in a way which is in the end "somehow" reported to us (i.e. report to team blue which then logs an error we then might be reported to us).

Which means we could just box any db specific error not handled by db specific code.

discovery_engine_core/core/src/storage/migrations/20220628114905_feed.sql
rustonaut: I'm not strictly against no-triggers, it's just having triggers for this specific kind of use-case is quite nice as it means you don't have to handle the history explicitly.

If we enable foreign constraint checking in sqlite history being there is also enforce through the fk-constraint on the feeds document id so it's not too important. But if we don't enable fk-constraint checking I would strongly recommend it.

discovery_engine_core/core/src/storage/sqlite.rs
Robert-Steiner: I implemented it so that we only take the first (BIND_LIMIT / number of fields for the largest tuple) documents

discovery_engine_core/core/src/storage/migrations/20220628114905_feed.sql
Robert-Steiner: From my point of view, I don't have a strong preference in either way

discovery_engine_core/core/src/storage/sqlite.rs
Robert-Steiner: made score nullable

discovery_engine_core/core/src/storage/sqlite.rs
Robert-Steiner: Does anyone have an opinion on this or should we leave it like that for now?

discovery_engine_core/core/src/storage/sqlite.rs
Robert-Steiner: moved it into a helper function

discovery_engine_core/core/src/engine.rs
Robert-Steiner: The associated error type has been removed. We now have a simple storage error type/enum with currently just `Database` as the only variant for both traits. Any database error will be converted into a generic error. I think github didn't pick up the latest changes. please have a look again

discovery_engine_core/core/src/storage/migrations/20220628114905_feed.sql
rustonaut: I'm not strictly against no-triggers, it's just having triggers for this specific kind of use-case is quite nice as it means you don't have to handle the history explicitly.

If we enable foreign constraint checking in sqlite history being there is also enforce through the fk-constraint on the feeds document id so it's not too important. But if we don't enable fk-constraint checking I would strongly recommend it. 

EDIT: I wrote this already yesterday and send it but now it's unsend again???

discovery_engine_core/core/src/storage/sqlite.rs
rustonaut: right, INTEGER is i64 not u64 and BIGINT in sqlite is also i64 so even if you use BIGINT it _sliently_ case any number >=2**63 to a REAL which is fun when we fetch it... :unamused: 

So I guess for any u64 we have three choices:
- not use it
- bit cast to i64 when inserting and back to u64 when fetching
- proper cast to i64 if it fails either fail or clamp it

In this case rank is the domain rank, theoretically it's at most 30?bit because there are just that many domains but if it is over time, it could easily be many times the number of domain, but then 16 times is still only 35bit. Also if it's rank is very very bad or very very bad +1 doesn't really matter so I would go with a clamping cast from u64 to i64 for this case.

discovery_engine_core/core/src/storage.rs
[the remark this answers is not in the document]
rustonaut: After remembering that SQL databases tend to have i64 but not u64 I would always go with u32 here :wink:

discovery_engine_core/core/src/storage/migrations/20220628114905_feed.sql
rustonaut: then lets just keep the trigger removed, makes it easier for anyone not familiar with sql(ite)/triggers, I guess

discovery_engine_core/core/src/storage/sqlite.rs
Robert-Steiner: so something like:
```rust
i64::try_from(u64_number).unwrap_or(i64::MAX);
u64::try_from(i64_number).unwrap_or(u64::MIN);
```
?

PR #474, opened 2022-06-30
[Merged by Bors] - Fix for dartdoc issue

biernacki:
Since the `dartdoc` command is now [part of `dart` cli](https://dart.dev/tools/dart-doc), we can use it without installing `dartdoc` as global dependency.

Because of the switch some cli options are only available via `dartdoc_options.yaml` [config file](https://github.com/dart-lang/dartdoc#dartdoc_optionsyaml).

### DIFF
--- a/.github/workflows/_reusable.dart.yml
+++ b/.github/workflows/_reusable.dart.yml
@@ -72,9 +72,5 @@ jobs:
           dart: true
           rust: true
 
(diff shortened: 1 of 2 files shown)


x3ro (approved):
LGTM 🚀 nice!

biernacki:
bors merge

PR #485, opened 2022-07-07
TY-2920 remove ranker facade

janpetschexain:
**References**

- [TY-2920]

**Summary**

*AI crate*

- `ai::ranker::context` moved to `ai::coi::context`
- `State` and its related `Ranker` methods moved from `ai::ranker::system` to `ai::coi::state`
- `Ranker::rank()` and its tests moved from `ai::ranker::system` to `CoiSystem::rank()` in `ai::coi::system`
- `ai::ranker::document` moved to `ai::document`
- `ai::embedding::mod` and `ai::embedding::utils` moved to `ai::embedding`
- the logic of the remaining methods from `Ranker` moved directly into the respective `Engine` methods
- remove the `Ranker` facade and builder in `ai::ranker::system` and `ai::ranker::public`
- some visibilities changed as things become part of the public interface due to the ranker removal

*DE crate*

- `Document` trait impls moved from `de::ranker` to `de::document`
- replace the `ranker` system with the individual `smbert` & `coi` & `kpe` systems in the `Engine`, update the `Engine` methods with these systems and use the moved logic of the `Ranker` methods respectively
- remove the `XaynAiEngine` type alias and move `XaynAiEngine::from_config()` to `Engine::from_config()`
- rename `StackState` & `RankerState` & `State` to `SerializedStackState` & `SerializedCoiSystemState` & `SerializedState` to better differentiate the serialized states from the nonserialized states
- update the `Engine` tests: we could continue to use `mock!` for the systems, but this requires all referenced arguments to be `'static`, which proves quite difficult since we now have to mock structs instead of traits. the recommended workaround from the `mockall` crate is to have different arguments/methods for `#[cfg(not(test))]` & `#[cfg(test)]` which fractures the code quite a lot. instead there is a shared test `ENGINE` now, which only initializes the systems once (to have the same time saving effect as the old no-op mocked systems) and which can be configured wrt the state. this also removes some code duplication in the tests.
- remove the `Ranker` facade in `de::ranker`

*Misc*

- `Pipeline::from(config)` moved to `Config::build()` to build the pipelines directly from the config for `SMBert` and `KPE`
- remove unused dependencies
- replace `lazy_static` with `once_cell` in `providers` because the former is unmaintained and the latter is used as a dependency in our other crates as well


discovery_engine_core/core/src/engine.rs
```diff
@@ -210,6 +223,124 @@ where
...
+        } else {
+            (HashMap::default(), CoiSystemState::default())
+        };
+        for id in stack_ops.iter().map(Ops::id).chain(once(Exploration::id())) {
+            if let Ok(alpha) = de_config.extract_inner::<f32>(&format!("stacks.{id}.alpha")) {
+                stack_data.entry(id).or_default().alpha = alpha;
+            }
+            if let Ok(beta) = de_config.extract_inner::<f32>(&format!("stacks.{id}.beta")) {
+                stack_data.entry(id).or_default().beta = beta;
+            }
```
rustonaut, line 315: :thinking:  Isn't that wrong?

I mean it's outside of the scope of the PR and even if it is wrong shouldn't be fixed in this PR but shouldn't config `stacks.{id}.alpha` only be the default alpha value. Instead here it will always reset it to this value every time the engine is created overwriting whatever value was deserialized (on line 305).
 
I.e. shouldn't it be `.or_insert(alpha)` ?

discovery_engine_core/core/src/engine.rs
```diff
@@ -1065,230 +1114,290 @@ enum SearchBy<'a> {
...
+    async fn init_engine<'a, I, F>(
+        stack_ops: I,
+        update_after_cleanse: bool,
+    ) -> MappedMutexGuard<'a, Engine>
+    where
+        I: IntoIterator<Item = F> + Send,
+        <I as IntoIterator>::IntoIter: Send,
+        F: FnOnce(&EndpointConfig, &Providers) -> BoxedOps,
+    {
+        let engine = async move {
```
rustonaut, line 1157: :art: 
```suggestion
        let create_engine = async move {
```

it's very easy to end up with the misconception that this will create a engine every time,
e.g. through a `.await?` at the end of the async block.

So renaming it to make clear the the variable contains a lazy constructor of the engine
is a good idea I think.

discovery_engine_core/core/src/engine.rs
```diff
@@ -1065,230 +1114,290 @@ enum SearchBy<'a> {
...
+            // be the same as when it's initialized for the first time after the app is downloaded.
+            let engine = Engine::from_config(config, None, &[], &[]).await.unwrap();
+
+            Mutex::new((server, engine))
         };
-        let endpoint_config = EndpointConfig::default()
-            .with_init_config(config.clone())
-            .await;
-        let providers = Arc::new(Providers::new(&config.clone().into()).unwrap());
+        let mut engine = MutexGuard::map(ENGINE.get_or_init(engine).await.lock().await, |engine| {
```
rustonaut, line 1197: :thinking: I'm not sure I like this "reuse engine but clear all state approach" compared to a "create anew but explicitly reuse some immutable parts approach" but I guess it's simpler here so I will approve it anyway.

janpetschexain: i'll try and see if i can change it into the suggested direction.

discovery_engine_core/core/src/engine.rs
janpetschexain: the current implementation mirrors the previous behavior (from before the config was changed in #466 and the constructors were refactored in #465), in `Engine::from_config()` we have that
- if no `state` is present, then a default stack data is created (as in the old `Engine::new()`)
- if some `state` is present, then this is taken (as in the old `Engine::from_state()`)
- if we have a value present in the `de_config`, then this is by design meant to overwrite any existing or default value, which happens in the marked lines. this design can of course be changed if needed.

rustonaut: > if we have a value present in the de_config, then this is by design meant to overwrite any existing or default value, which happens in the marked lines. this design can of course be changed if needed.

I didn't know about this ;)

discovery_engine_core/core/src/engine.rs
rustonaut: my complain was mainly wrt. the engine as a whole then how we change the stacks, as it's prone to forgetting adding clearing for some data we might add in the future.

but like I said it's not too much of a problem so it's fine either way, I guess

rustonaut (approved):
lgtm

discovery_engine_core/core/src/engine.rs
```diff
@@ -874,185 +1049,60 @@ fn documentify_articles(
...
 
 #[derive(Serialize, Deserialize)]
-struct State {
+struct SerializedState {
     /// The serialized engine state.
-    engine: StackState,
+    stacks: SerializedStackState,
     /// The serialized ranker state.
-    ranker: RankerState,
+    coi: SerializedCoiSystemState,
```
finiteprods, line 1096: :grey_question: do the associated doc strings need to be updated? e.g. the serialized stack state

discovery_engine_core/core/src/engine.rs
```diff
@@ -874,185 +1049,60 @@ fn documentify_articles(
...
-        state: &'a [u8],
-        builder: Builder<'a, AveragePooler>,
-    ) -> Result<(HashMap<StackId, StackData>, Builder<'a, AveragePooler>), Error> {
+impl SerializedState {
+    fn deserialize(state: &[u8]) -> Result<(HashMap<StackId, StackData>, CoiSystemState), Error> {
         let state = bincode::deserialize::<Self>(state).map_err(Error::Deserialization)?;
-        let stack_data = state.engine.deserialize()?;
-        let builder = state.ranker.deserialize(builder)?;
+        let stack_data = state.stacks.deserialize()?;
+        let ranker_state = state.coi.deserialize()?;
```
finiteprods, line 1103: similar to previous comment, should this be e.g. `coi_system_state`?

finiteprods (approved):
looks good, just some minor comments included.

janpetschexain:
bors merge

PR #494, opened 2022-07-14
[Merged by Bors] - Add `#[allow(clippy::unused_async)]` where needed.

rustonaut:
This is done so that the current master branch passes checks locally.

Currently I have this commit cherry picked in 2 feature branches. 

It's not clear how the lint triggers locally as it only should have been in rust 1.62 (but we use 1.61) and it doesn't trigger on CI, but we will sooner or later switch to 1.62 anyway :shrug:


rustonaut:
bors merge

PR #496, opened 2022-07-18
[Merged by Bors] - Use env bash instead of env sh.

rustonaut:
Due to a funny coincidence the errors caused by this
kinda "evened" out their effects in the tests.


rustonaut:
bors merge

rustonaut:
bors merge

PR #497, opened 2022-07-18
[Merged by Bors] - TY-2970 allow locale example testing of published packages

rustonaut:
This add a  helper function to the justfile which overrides the dependencies of the flutter package to use specific published packages instead of the local checkout.

This is useful for testing changes to the CI.

The helper function is hidden by default as it adds changes to the local repo which should never be committed.

**Based on:**
- #496

### DIFF
--- a/justfile
+++ b/justfile
@@ -286,6 +286,21 @@ download-assets:
 check-android-so:
     {{justfile_directory()}}/.github/scripts/check_android_so.sh "$FLUTTER_WORKSPACE"/android/src/main/jniLibs/
 
+_override-flutter-self-deps $BUILD_ID:
+    #!/usr/bin/env bash
+    set -eux
+    cd "$FLUTTER_EXAMPLE_WORKSPACE"
+
+    SED_CMD="sed"
+    if [[ "{{os()}}" == "macos" ]]; then
+        SED_CMD="gsed"
+    fi
+
+    # This will add changes to your repo which should never be committed.
+    $SED_CMD -i s/dependency_overrides/x_dependency_overrides/ ./pubspec.yaml
+    $SED_CMD -i s/change.me.to.commit.ref/${BUILD_ID}/ ./pubspec.yaml
+
+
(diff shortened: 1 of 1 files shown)


justfile
```diff
@@ -286,8 +286,23 @@ download-assets:
...
+    set -eux
+    cd "$FLUTTER_EXAMPLE_WORKSPACE"
+
+    SED_CMD="sed"
+    if [[ "{{os()}}" == "macos" ]]; then
+        SED_CMD="gsed"
+    fi
+
+    # This will add changes to your repo which should never be committed.
+    $SED_CMD -i s/dependency_overrides/x_dependency_overrides/ ./pubspec.yaml
```
janpetschexain, line 300: ❓ does the `x_` just modifies the name such that flutter won't recognize this item as the `dependency_overrides`? if so, should we give it a more readable name, maybe sth lik `DISABLED_dependency_overrides`?

justfile
```diff
@@ -286,8 +286,23 @@ download-assets:
...
+    $SED_CMD -i s/change.me.to.commit.ref/${BUILD_ID}/ ./pubspec.yaml
+
+
 _dart-publish $WORKSPACE:
-    #!/usr/bin/env sh
+    #!/usr/bin/env bash
```
janpetschexain, line 305: :art: this pr should be rebased onto the latest `main` before merging

justfile
rustonaut: yes it just makes the entry unrecognizable for dart, it's the same as we do in the release script.

rustonaut: this is also not disabling dependency overrides per-se, just making them unrecognizable for dart.

In the end it's an ugly hack, but doing it properly isn't worth the time needed to make it work.

This is not something you should ever commit to the repo, it's just temporary for debugging purpose. Hence why I also made the job start with a `_`.

janpetschexain (approved):
lgtm

rustonaut:
bors merge

PR #498, opened 2022-07-19
TY-3019 Message Passing API rust part.

rustonaut:


discovery_engine_core/core/src/message_api.rs
```diff
@@ -0,0 +1,232 @@
...
+
+            let message = system.shutdown().await;
+            Some(Package {
+                message,
+                cookie: MagicCookie::generate_cookie(),
+            })
+        });
+
+        let res = run_engine_fut
+            .catch_unwind()
```
rustonaut, line 111: This line won't compile in this scatch, basically assert unwind safe/catch unwind are fundamentally broken with futures in most situations :cry:

rustonaut:
to to pivoting probably doesn't matter anymore, and the branch is still around and tagged.

PR #501, opened 2022-07-19
[Merged by Bors] - Change versioning schema and other minor improvements.

rustonaut:
- use timestamp based versions
- use the branch name in the metadata.
- strips bad characters from the branch name
- echo version
- give the version replacement marker a more telling name

### DIFF
--- a/justfile
+++ b/justfile
@@ -316,16 +316,14 @@ _dart-publish $WORKSPACE:
     # Dependency overrides are not allowed in published dart packages
     $SED_CMD -i s/dependency_overrides/HACK_hide_dependency_overrides/ ./pubspec.yaml
 
-    # Make it easier to figure out which builds came from `main`, and order the builds
-    # by Github's "run ID". Note that this ID is reset if you change the name of the
-    # workflow.
-    BUILD_ID="${GITHUB_RUN_ID:-missingRunId}.$(git rev-parse HEAD)"
+    # Add semver _metadata_ in the form of "<branch-name>.<commit-hash>.<github-run-id>".
+    # - If we are in detached head mode `HEAD` will be used instead of the branch name.
+    # - Non `[0-9A-Za-z-]` characters in the branch name will be replaced with `-`
+    # - If we have no run id "missingRunId" is used instead.
+    BUILD_ID="$(git rev-parse --abbrev-ref HEAD | sed s/[^0-9a-zA-Z-]/-/g ).$(git rev-parse HEAD).${GITHUB_RUN_ID:-missingRunId}"
     echo "Build Id: $BUILD_ID"
-    if [[ "$(git rev-parse --abbrev-ref HEAD)" == "main" ]]; then
-        BUILD_ID="main.${BUILD_ID}"
-    fi
 
-    $SED_CMD -i s/change.me.to.commit.ref/${BUILD_ID}/ ./pubspec.yaml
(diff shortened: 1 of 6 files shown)


rustonaut:
It is already annoying long, so I don't think it makes a difference wrt. to using it.

But it makes a difference wrt. maintenance, e.g. deleting old packages from JFrog is much easier if you can easily group them.

rustonaut:
Like that last build ID is: `main.2697734334.febf041700b2042fd130f8658a37d12e9e4c49f3`

acrrd:
Why we need both commit ref and build number?

I would lie to not have too much manual maintenance on jfrog, silver is working to a script to automatically delete all the artifacts that are not main.xxxx after a period of time. We can have a version number without the + part for release build that we don't have to delete.

justfile
```diff
@@ -301,16 +301,14 @@ _dart-publish $WORKSPACE:
...
-    # Make it easier to figure out which builds came from `main`, and order the builds
-    # by Github's "run ID". Note that this ID is reset if you change the name of the
-    # workflow.
-    BUILD_ID="${GITHUB_RUN_ID:-missingRunId}.$(git rev-parse HEAD)"
+    # Add semver _metadata_ in the form of "<branch-name>.<commit-hash>.<github-run-id>".
+    # - If we are in detached head mode `HEAD` will be used instead of the branch name.
+    # - Non `[0-9A-Za-z-]` characters in the branch name will be replaced with `-`
+    # - If we have no run id "missingRunId" is used instead.
+    BUILD_ID="$(git rev-parse --abbrev-ref HEAD | sed s/[^0-9a-zA-Z-]/-/g ).$(git rev-parse HEAD).${GITHUB_RUN_ID:-missingRunId}"
     echo "Build Id: $BUILD_ID"
```
acrrd, line 309: Would be better to print the whole version that blue has to put in the pubspec.yaml

rustonaut:
But I just had a idea, why don't we use "timestamp" semver.

E.g. we replace the whole `0.1.0+marker...` with a version like e.g. `202207201345.0.0+branch-name` where team blue can skip "+branch-name" if they want to, or keep it better know which branch they are one.

This is quite simple to do and scales well I would say.

If we want to we could also add the commit ref somewhere in the package for the case we ever need it.

> Why we need both commit ref and build number?

We can have multiple builds for the same branch/commit so we need the build number. And having the commit ref somewhere is nice and generally more usefull then the build id.

rustonaut:
I changed the versioning schema versions are now like e.g.:

`0.220720143433.0+TY-2972-minor-improvements`

- We use `0.<timestamp>` so that we can easier change later on to proper versioning if needed
- The timestamp is year (2 digit), month (2), day (2), hour (24h, 2 digit), minute (2), seconds (2)
   -  While this doesn't fit in a i32/u32 it still can be properly represented by a IEEE754 float, and as such
      should always work with all dart tooling.
- The branch name is sanitized (as needed for semver metadata): Any non [0-9A-Za-z-] character is replaced with `-`.
- If for whatever reason we build on detached head mode the git has is used instead of the branch name

This means all build have distinct, but semver increasing numbers. The only way this could go wrong is if two builds
for two different branches finish at exactly the same second and no `+branch-name` is used when specifying the dependency :shrug:.

acrrd (approved):

rustonaut:
bors merge

PR #502, opened 2022-07-20
[Merged by Bors] - TY-3018 change used sql join style

rustonaut:
- use explicit `JOIN` instead of the `,` alias
- add the `ON` per-join instead of relying on sqlite handling compound `ON` nicely

Reason: We will have some `LEFT JOIN`s in the future at which point the previous stile stops working nicely.

### DIFF
--- a/discovery_engine_core/core/src/storage/sqlite.rs
+++ b/discovery_engine_core/core/src/storage/sqlite.rs
@@ -243,8 +243,8 @@ impl Storage for SqliteStorage {
             "SELECT
                 nr.documentId, nr.title, nr.snippet, nr.url
             FROM
-                HistoricDocument AS hd, NewsResource AS nr
-            ON hd.documentId = nr.documentId;",
+                HistoricDocument AS hd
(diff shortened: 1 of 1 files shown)


discovery_engine_core/bindings/Cargo.toml
```diff
@@ -5,7 +5,7 @@ license = "AGPL-3.0-only"
 edition = "2021"
 
 [dependencies]
-async-bindgen = { path="../async-bindgen/async-bindgen" }
+async-bindgen = { path = "../async-bindgen/async-bindgen" }
```
rustonaut, line 8: Should be included in the fix CI PR, so we will have to rebase/remove it then.

biernacki (approved):
wow the bot really likes me :)

I needed to read a bit about SQL to understand that change, but it looks definitely better 👍

discovery_engine_core/bindings/Cargo.toml
rustonaut: done :rocket:

rustonaut:
bors merge

PR #503, opened 2022-07-20
[Merged by Bors] - More strict justfile

acrrd:
Now the `rust-check` task can pass even if `cargo clippy` returns an error. This is because we are using `;` as a command separator and we didn't set `-e` as a bash option.

This PR set `-euxo pipefail` for the main shell and all the shell that we use to make it fail early.

### DIFF
--- a/discovery_engine_core/bindings/Cargo.toml
+++ b/discovery_engine_core/bindings/Cargo.toml
@@ -5,7 +5,7 @@ license = "AGPL-3.0-only"
 edition = "2021"
 
 [dependencies]
-async-bindgen = { path="../async-bindgen/async-bindgen" }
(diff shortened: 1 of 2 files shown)


rustonaut (approved):
lgtm,

as a side note `set -euxo pipefail` work just fine, through then making
`-o` a separate argument is a question of preference so idc

acrrd:
> lgtm,
> 
> as a side note `set -euxo pipefail` work just fine, through then making `-o` a separate argument is a question of preference so idc

I descided for the splited -o because I think that when we set the shell having `["-euxo", "pipefail"]` is less clear then having it at as a separate options. I kept the same style for all other instances.

acrrd:
bors merge

PR #505, opened 2022-07-21
[Merged by Bors] - TY-2978 move active search logic to rust (0): activeSearchRequested

janpetschexain:
**References**

- [TY-2978]
- followed by #506

**Summary**

- add `cfgFeatureStorage` constant as a pseudo feature flag in dart
- store new searches in `Engine::search_by_query()` & `Engine::search_by_topic()` in rust
- remove corresponding db calls in `SearchManager.activeSearchRequested()` in dart
- prepare `ActiveSearch` cleanup

[TY-2978]: https://xainag.atlassian.net/browse/TY-2978?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine/lib/discovery_engine.dart
+++ b/discovery_engine/lib/discovery_engine.dart
@@ -31,4 +31,4 @@ export 'package:xayn_discovery_engine/src/worker/common/exceptions.dart';
 void discoveryEngineInitLogger(Logger logger) => initLogger(logger);
 Logger get discoveryEngineLogger => logger;
 
-const kCfgFeatureStorage = false;
+const cfgFeatureStorage = false;
--- a/discovery_engine/lib/discovery_engine.dart
+++ b/discovery_engine/lib/discovery_engine.dart
@@ -30,3 +30,5 @@ export 'package:xayn_discovery_engine/src/worker/common/exceptions.dart';
 
 void discoveryEngineInitLogger(Logger logger) => initLogger(logger);
 Logger get discoveryEngineLogger => logger;
+
+const kCfgFeatureStorage = false;
--- a/discovery_engine_core/core/src/engine.rs
+++ b/discovery_engine_core/core/src/engine.rs
@@ -587,9 +587,37 @@ impl Engine {
         if query.trim().is_empty() {
             return Err(Error::InvalidTerm);
         }
+
+        #[cfg(feature = "storage")]
+        if let Ok((search, _)) = self.storage.search().fetch().await {
+            if search.search_by != storage::models::SearchBy::Query || search.search_term != query {
+                return Err(storage::Error::OpenSearch.into());
+            }
+        }
+
         let filter = &Filter::default().add_keyword(&query);
-        self.active_search(SearchBy::Query(filter), page, page_size)
-            .await
+        let documents = self
+            .active_search(SearchBy::Query(filter), page, page_size)
+            .await?;
+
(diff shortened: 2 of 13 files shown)


discovery_engine_core/core/src/engine.rs
```diff
@@ -587,9 +587,37 @@ impl Engine {
         if query.trim().is_empty() {
             return Err(Error::InvalidTerm);
         }
+
+        #[cfg(feature = "storage")]
+        if let Ok((search, _)) = self.storage.search().fetch().await {
+            if search.search_by != storage::models::SearchBy::Query || search.search_term != query {
```
finiteprods, line 593: :grey_question: how about paging, does that need to be checked too?
similar question for `search_by_topic`

finiteprods (approved):
some minor questions otherwise looks good

discovery_engine/lib/discovery_engine.dart
```diff
@@ -30,3 +30,5 @@ export 'package:xayn_discovery_engine/src/worker/common/exceptions.dart';
 
 void discoveryEngineInitLogger(Logger logger) => initLogger(logger);
 Logger get discoveryEngineLogger => logger;
+
+const kCfgFeatureStorage = false;
```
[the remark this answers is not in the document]
janpetschexain: it stands for `konstant`/`constant`. i'll just remove it, after all it's a bad naming practice for global constants in dart which predates the time of statically typed constants.

discovery_engine_core/core/src/engine.rs
janpetschexain: actually i should fail if there is any search already running (as in the dart code), not just on those conditions. i'm not sure why i wrote it that way, i'll update it accordingly. good catch after all 👍

rustonaut: :pushpin: this is currently silently dropping all errors, not just the "no search was open".

Also I personally prefer it if in places where we expect that things are potentially not given we use a `Option` result.

But I can't decide if it applies here.

If this being called with an already open search is a "unexpected situation " (e.g. bug or illegal input) I would say it should be an error. But with it's own error variant, so that we can check on the variant nicely.

But it looks more as if this is somewhat expected in some situations where the query didn't change, in which case I would say it's not falling into the "unexpected" category so I maybe we should change fetch to return a option.

Either way it needs some change in the future.

I used :pushpin:  because I think it's best to do so in a follow up PR so that we can merge pr 0-5 tomorrow to reduce friction (e.g. my PR is based on one of your feature branches as it needs embedding).

janpetschexain: > this is currently silently dropping all errors, not just the "no search was open".

i took the changes from the following pr which introduces the `Error::NoSearch` variant and changed the return values. now it returns all other errors properly (i think this is also what i had in mind with the original impl).

> But it looks more as if this is somewhat expected in some situations where the query didn't change, in which case I would say it's not falling into the "unexpected" category so I maybe we should change fetch to return a option.

i tend to disagree, even if we create a new query with the very same query term, then it might still return different results, depending on the underlying provider. hence we should treat every new query as distinct queries and not reuse cached data.

janpetschexain:
bors merge

PR #508, opened 2022-07-21
[Merged by Bors] - TY-3079 Dockerize web service

biernacki:
**Summary:**
- added `Dockerfile` for building image
- added just recipe
- added dedicated github actions workflow

**References:**
[TY-3079]

[TY-3079]: https://xainag.atlassian.net/browse/TY-3079?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/web-service.yml
+++ b/.github/workflows/web-service.yml
@@ -76,10 +76,9 @@ jobs:
         run: |
           tar -xvf web-service.tar
           cd ./web-service
-          TAG="$(git rev-parse --short HEAD)"
-          if [[ "${GITHUB_REF#refs/heads/}" != "" ]]; then
-            #TAG="$(echo "branch-${GITHUB_REF#refs/heads/}" | tr '[:upper:]' '[:lower:]')"
-            TAG="branch-${GITHUB_REF#refs/heads/}"
+          TAG="branch-$(git rev-parse --abbrev-ref HEAD)"
+          if [[ "${TAG}" == "branch-HEAD" ]]; then
+            TAG="$(git rev-parse --short HEAD)"
           fi
           IMAGE_NAME="xaynetci/xayn_discovery_web_service:$TAG"
           docker build -t "${IMAGE_NAME}" .
--- a/.github/workflows/web-service.yml
+++ b/.github/workflows/web-service.yml
@@ -1,0 +1,106 @@
+name: Web Service Release
+
+on:
+  workflow_dispatch:
+  push:
+
+permissions:
+  contents: read
+
+env:
+  DENY_WARNINGS: false
+
+jobs:
+  rust-build:
+    runs-on: hetzner-pm
+    container:
+      image: xaynetci/yellow:v3
+    timeout-minutes: 10
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # v2.4.0
+
+      - name: Setup job
+        uses: ./.github/actions/setup-job-docker
+        with:
+          rust: true
+
+      - name: Build artifacts
+        run: just build-web-service
+
+      - name: Download assets
+        run: just download-assets
+
+      - name: Prepare temp directory
+        working-directory: ${{ runner.temp }}
+        run: |
+          rm -rf ./web-service
+          mkdir -p web-service/assets
+
+      - name: Create archive
+        working-directory: ${{ runner.temp }}
+        run: |
+          cp ${GITHUB_WORKSPACE}/${{ env.RUST_WORKSPACE }}/target/release/web-api ./web-service/web-api
+          cp ${GITHUB_WORKSPACE}/${{ env.RUST_WORKSPACE }}/web-api/.env ./web-service/.env
+          cp ${GITHUB_WORKSPACE}/${{ env.RUST_WORKSPACE }}/web-api/Dockerfile ./web-service/Dockerfile
+          cp ${GITHUB_WORKSPACE}/${{ env.FLUTTER_WORKSPACE }}/example/assets/smbert_v0001/smbert-quantized.onnx ./web-service/assets/model.onnx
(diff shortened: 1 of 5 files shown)


.github/workflows/web-service.yml
```diff
@@ -0,0 +1,106 @@
...
+          IMAGE_NAME="xaynetci/xayn_discovery_web_service:$TAG"
+          docker build -t "${IMAGE_NAME}" .
+          echo "IMAGE_NAME=${IMAGE_NAME}" >> $GITHUB_ENV
+          echo "TAG=${TAG}" >> $GITHUB_ENV
+
+      - name: Login to Docker Hub
+        uses: docker/login-action@v1
+        with:
+          username: "${{ secrets.DOCKERHUB_USERNAME }}"
+          password: "${{ secrets.DOCKERHUB_TOKEN }}"
```
biernacki, line 93: @jorge-cr-13  could you add these 2 secrets to our repo? I don't have admin access here, and the workflow fails at this stage

.github/workflows/web-service.yml
```diff
@@ -0,0 +1,106 @@
...
+
+      - name: Create docker image
+        run: |
+          tar -xvf web-service.tar
+          cd ./web-service
+          TAG="$(git rev-parse --short HEAD)"
+          if [[ "${GITHUB_REF#refs/heads/}" != "" ]]; then
+            #TAG="$(echo "branch-${GITHUB_REF#refs/heads/}" | tr '[:upper:]' '[:lower:]')"
+            TAG="branch-${GITHUB_REF#refs/heads/}"
+          fi
```
rustonaut, line 83: :question: I'm not sure why you originally used trim.

You should be able to get the branch name using `git rev-parse --abbrev-ref HEAD`, which will be `HEAD` if there is no branch name. 

So following should be a bit more clean:

```suggestion
          TAG="branch-$( git rev-parse --abbrev-ref HEAD)"
          if [[ "${TAG}" == "branch-HEAD" ]]; then
            TAG="$(git rev-parse --short HEAD)"
          fi
```

The difference between branch and commit hash is also normally very clear, so if we don't want to manually test on the tag starting with `branch-` we probably could just drop the `branch`.

.github/workflows/web-service.yml
jorge-cr-13: I added both secrets, let me know if they work. :)

biernacki: thank you :)

biernacki: yes, it works

.github/workflows/web-service.yml
biernacki: oh, that was something that I just copied from prev solution, your version is better, thanks! :)

.github/workflows/web-service.yml
```diff
@@ -0,0 +1,104 @@
...
+        run: docker push "${IMAGE_NAME}"
+
+  #     - name: Trigger Kubernetes Web deployment
+  #       id: k8s-deployment
+  #       if: steps.docker-push.outcome == 'success'
+  #       env:
+  #         GHA_TOKEN: ${{ secrets.GHA_TOKEN }}
+  #         GHA_ACTOR: ${{ github.actor }}
+  #       run: |
+  #         curl -v -XPOST https://api.github.com/repos/xaynetwork/xayn_infrastructure/dispatches -H "Authorization: token ${GHA_TOKEN}" -H "Accept: application/vnd.github.v3+json" -H "Content-Type: application/json" -d "{\"event_type\":\"remote_poker-deployment\",\"client_payload\":{\"actor\":\"${GHA_ACTOR}\",\"tag\":\"${TAG}\"}}"
```
biernacki, line 104: @jorge-cr-13 not about this part, should I remove it, or leave it for you to update (maybe even in this PR?), wdyt?

discovery_engine_core/web-api/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM debian:bookworm-slim
+
+WORKDIR /service
+
+COPY ./ ./
+
+RUN chmod +x ./web-api
+VOLUME /service/assets/data.json
```
biernacki, line 8: @jorge-cr-13 is it ok for you if we provide the json file with data this way? so we can just restart the container with updated data instead of bundling it with the image

rustonaut (approved):
lgtm

.github/workflows/web-service.yml
jorge-cr-13: I don't think there is any POD running with this container at the moment, right? If we want to run PODs with the image, then it would be good to leave these comments in case we want to update the `tag` manually. If we want to implement it I think it is better to adapt it in another PR after we implemented it from our side.

discovery_engine_core/web-api/Dockerfile
jorge-cr-13: This `data.json` for doing a backup of the state of the container?

jorge-cr-13: Actually, I don't understand what is the plan of this `data.json` file. Should we store it somewhere so everytime the container has been restored it needs to find this file and restore it from here?

.github/workflows/web-service.yml
biernacki: I'll remove it then from this one

discovery_engine_core/web-api/Dockerfile
biernacki: the `data.json` file is source of internal state for the web service. At the startup we read the contents of this file, and calculate some additional state from that. So this file is read only. It's needed so that the api can return something. The reason it was made a volume it's to have a way of updating content of this file without the need to rebuild the image.

jorge-cr-13: Ok.. then yeah, I don't see a problem with using the file as a `data.json`.

jorge-cr-13 (approved):

biernacki:
bors merge

PR #507, opened 2022-07-21
[Merged by Bors] - TY-2978 move active search logic to rust (2): nextActiveSearchBatchRequested

janpetschexain:
**References**

- [TY-2978]
- requires #506
- followed by #509

**Summary**

- add `Engine::search_next_batch()` & corresponding ffi
- use new logic & remove corresponding db calls in `SearchManager.nextActiveSearchBatchRequested()` in dart

[TY-2978]: https://xainag.atlassian.net/browse/TY-2978?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #509, opened 2022-07-21
[Merged by Bors] - TY-2978 move active search logic to rust (3): restoreActiveSearchRequested

janpetschexain:
**References**

- [TY-2978]
- requires #507
- followed by #510

**Summary**

- add native bytes conversion for `Embedding`
- add `Embedding` db migration & store the embedding in `SqliteStorage::store_new_documents()` & query it for `QueriedApiDocumentView`
- add `Engine::searched()` & corresponding ffi
- use new logic & remove corresponding db calls in `SearchManager.restoreActiveSearchRequested()` in dart

[TY-2978]: https://xainag.atlassian.net/browse/TY-2978?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine/lib/src/domain/engine/engine.dart
+++ b/discovery_engine/lib/src/domain/engine/engine.dart
@@ -102,7 +102,7 @@ abstract class Engine {
   Future<List<DocumentWithActiveData>> searchNextBatch();
 
   /// Restores the current active search, ordered by their global rank (timestamp & local rank).
-  Future<List<DocumentWithActiveData>> searched();
+  Future<List<DocumentWithActiveData>> restoreSearch();
 
   /// Gets the current active search mode and term.
   Future<ActiveSearch> searchedBy();
--- a/discovery_engine/lib/src/domain/engine/engine.dart
+++ b/discovery_engine/lib/src/domain/engine/engine.dart
@@ -101,6 +101,9 @@ abstract class Engine {
   /// Gets the next batch of the current active search.
   Future<List<DocumentWithActiveData>> searchNextBatch();
 
+  /// Restores the current active search, ordered by their global rank (timestamp & local rank).
+  Future<List<DocumentWithActiveData>> searched();
+
   /// Gets the current active search mode and term.
   Future<ActiveSearch> searchedBy();
 
--- a/discovery_engine_core/ai/bert/src/pooler.rs
+++ b/discovery_engine_core/ai/bert/src/pooler.rs
@@ -12,7 +12,10 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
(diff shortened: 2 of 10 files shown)


discovery_engine_core/ai/bert/src/pooler.rs
```diff
@@ -41,6 +44,24 @@ impl<const N: usize> From<[f32; N]> for Embedding<Ix1> {
     }
 }
 
+impl Embedding<Ix1> {
+    /// Converts from values in logical order to bytes in native byte order.
+    pub fn to_ne_bytes(&self) -> Vec<u8> {
+        self.iter().flat_map(|value| value.to_ne_bytes()).collect()
+    }
```
rustonaut, line 51: If we it this way we could use a explicit byte order, like e.g. little endian which is the native endianess of probably all platforms we do support.

This would make it simpler to avoid subtle bugs if we ever add some "sync across devices" feature involving.

janpetschexain: i'd prefer to handle it this way, but wouldn't this break db migration? if there is a system which uses big endian as native endianess, then migrating and reading the byte blob as little endian will make a mess.

discovery_engine/lib/src/domain/engine/engine.dart
```diff
@@ -101,6 +101,9 @@ abstract class Engine {
   /// Gets the next batch of the current active search.
   Future<List<DocumentWithActiveData>> searchNextBatch();
 
+  /// Restores the current active search, ordered by their global rank (timestamp & local rank).
+  Future<List<DocumentWithActiveData>> searched();
```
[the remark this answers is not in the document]
janpetschexain: imo `restore_` is quite a bad naming, but i'll change it for consistency.

discovery_engine_core/ai/bert/src/pooler.rs
rustonaut: > if there is a system which uses big endian as native 

I'm not sure if there is even a single phone currently in existence (which still is supported by us) for which this is true.

To quote [android docs](https://developer.android.com/ndk/guides/abis):

>  Android is always little-endian.

Also x86 is always little-endian. AFIK this is also true for iOs, Windows and macOs (since switching to Intel).

Arm is bi-endian but we basically compile for little endian arm (it's a bit more complex then this, but for this context we should be able to treat it as little endian).

So I'm pretty sure big endian systems do not matters for us currently.

I think likely will never matter for us, too. But if it ever does start to matter using a explicit byte order here is much less prone to bugs.

Lastly the hive->sqlite migration should go through the FFI and should use `f32` there.

rustonaut (approved):
lgtm

janpetschexain:
bors merge

PR #512, opened 2022-07-25
[Merged by Bors] - TB-4079 Fix date wrong time offset

carminetoriello:
What has been changed ?

 - When creating the `DateTime` object from the `fromMicrosecondsSinceEpoch` method, set the `isUtc` flag to true

Why it has been changed ?

 - The date received from the newsCatcher seems to be already in UTC, therefore when creating the `DateTime` object for it we have to consider that by setting the isUtc flag to true.


rustonaut (approved):
lgtm

PR #513, opened 2022-07-25
[Merged by Bors] - TY-3082 More flexible json structure

biernacki:
**Summary:**
- changed `Article` type to be more flexible, but still required to contain at least `id` and `description` fields

**References:**
- [TY-3082]

[TY-3082]: https://xainag.atlassian.net/browse/TY-3082?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine_core/web-api/src/models.rs
+++ b/discovery_engine_core/web-api/src/models.rs
@@ -61,6 +61,12 @@ impl AiDocument for Document {
 /// Represents an article that is stored and loaded from local json file.
 pub(crate) type Article = HashMap<String, serde_json::Value>;
 
+impl From<Document> for Article {
+    fn from(doc: Document) -> Self {
+        doc.article
+    }
+}
+
 /// Represents user interaction request body.
 #[derive(Debug, Clone, Serialize, Deserialize)]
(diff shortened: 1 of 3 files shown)


discovery_engine_core/web-api/src/models.rs
```diff
@@ -59,18 +59,11 @@ impl AiDocument for Document {
 }
 
 /// Represents an article that is stored and loaded from local json file.
-#[derive(Debug, Clone, Serialize, Deserialize)]
-pub(crate) struct Article {
-    pub(crate) id: Uuid,
-    pub(crate) description: String,
-}
+pub(crate) type Article = HashMap<String, serde_json::Value>;
```
janpetschexain, line 62: :art: if space/performance is an issue, then we could also use an enum instead of strings as keys for the hashmap. another nice side effect is that typos in the keys would be caught at compile time instead of at runtime.

janpetschexain (approved):
lgtm

PR #516, opened 2022-07-26
[Merged by Bors] - TY-3018-fk-violation-error-helper

rustonaut:
Adds two helper:

- `From<sqlx::Error>` for `storage::Error` so that we don't have to write `.await.map(|err| Error::Database(err.into()))?` all the time
- a `.fk_violation_is_invalid_document()` extension on `Result<V, sqlx::Errror>` which will convert the (potential) error to a `storage::Error` like using `into()` but if there is a fk violation error it will produce an `Error::InvalidDocumentId` variant instead of an `Error::Database`.

### DIFF
--- a/discovery_engine_core/core/src/storage/sqlite.rs
+++ b/discovery_engine_core/core/src/storage/sqlite.rs
@@ -12,14 +12,17 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-use std::str::FromStr;
+use std::{borrow::Cow, str::FromStr};
 
 use async_trait::async_trait;
 use chrono::{NaiveDateTime, Utc};
 use num_traits::FromPrimitive;
 use sqlx::{
     sqlite::{Sqlite, SqliteConnectOptions, SqlitePoolOptions},
-    FromRow, Pool, QueryBuilder, Transaction,
+    FromRow,
+    Pool,
+    QueryBuilder,
+    Transaction,
 };
 use url::Url;
 use xayn_discovery_engine_providers::Market;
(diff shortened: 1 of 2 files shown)


discovery_engine_core/core/src/storage/sqlite.rs
```diff
@@ -49,6 +49,29 @@ use crate::{
 // Sqlite bind limit
 const BIND_LIMIT: usize = 32766;
 
+trait SqlxSqliteErrorExt<V> {
+    /// Use this on the result of an sqlite query which might have failed a foreign
+    /// key constraint if a illegal document if was passed in.
```
Robert-Steiner, line 54:
```suggestion
    /// Use this on the result of a sqlite query which might have failed a foreign
    /// key constraint when an illegal document was passed in.
```

discovery_engine_core/core/src/storage/sqlite.rs
```diff
@@ -785,4 +808,39 @@ mod tests {
...
+        let res = sqlx::query("INSERT INTO Bar(x) VALUES (?);")
+            .bind(10)
+            .execute(&storage.pool)
+            .await
+            .fk_violation_is_invalid_document_id(document_id);
+
+        assert!(matches!(res, Err(Error::InvalidDocumentId(_))));
+        if let Err(Error::InvalidDocumentId(id)) = res {
+            assert_eq!(id, document_id);
+        }
```
Robert-Steiner, line 837: 🎨 
```suggestion
        assert!(matches!(res, Err(Error::InvalidDocumentId(id)) if id == document_id));
```

Robert-Steiner (approved):
lgtm

rustonaut:
bors merge

PR #517, opened 2022-07-26
[Merged by Bors] - TY-2979 move feed logic to rust (0): nextFeedBatch

janpetschexain:
**References**

- [TY-2979]
- followed by #518

**Summary**

- add `Engine::feed_next_batch()` & corresponding ffi
- use new logic & remove corresponding db calls in `FeedManager.nextFeedBatch()` in dart

[TY-2979]: https://xainag.atlassian.net/browse/TY-2979?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine/lib/src/domain/feed_manager.dart
+++ b/discovery_engine/lib/src/domain/feed_manager.dart
@@ -12,6 +12,8 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
+import 'package:xayn_discovery_engine/discovery_engine.dart'
+    show cfgFeatureStorage;
 import 'package:xayn_discovery_engine/src/api/events/client_events.dart'
     show FeedClientEvent;
 import 'package:xayn_discovery_engine/src/api/events/engine_events.dart'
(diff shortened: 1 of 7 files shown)


discovery_engine/lib/src/domain/feed_manager.dart
```diff
@@ -109,6 +111,31 @@ class FeedManager {
 
   /// Obtain the next batch of feed documents and persist to repositories.
   Future<EngineEvent> nextFeedBatch() async {
+    if (cfgFeatureStorage) {
+      final sources = await _sourceReactedRepo.fetchAll();
+      final List<DocumentWithActiveData> docs;
```
biernacki, line 116: might be better to mark it as `late final List<DocumentWithActiveData> docs;`, but if the analyzer doesn't complain, then maybe that's fine too 🤔

biernacki (approved):
lgtm, just one small suggestion, but optional if the analyzer is happy with the current implementation

discovery_engine/lib/src/domain/feed_manager.dart
janpetschexain: if i understood this correctly, then `late final` does the check for initialization at runtime, whereas just declaring the variable (as `final`, and to have it available in the outer scope) and then assigning it in every if-branch (or returning early) does the check at compile time.

janpetschexain:
bors merge

PR #518, opened 2022-07-26
TY-2979 move feed logic to rust (1): restoreFeed

janpetschexain:
**References**

- [TY-2979]
- requires #517

**Summary**

- add `Engine::restore_feed()` & corresponding ffi
- use new logic & remove corresponding db calls in `FeedManager.restoreFeed()` in dart


discovery_engine_core/core/src/engine.rs
```diff
@@ -491,6 +491,24 @@ impl Engine {
         Ok(documents)
     }
 
+    /// Restores the feed documents, ordered by their global rank (timestamp & local rank).
+    // TODO: rename methods to `fed()` and adjust events & docs accordingly after DB migration
```
Robert-Steiner, line 495:
```suggestion
    // TODO: rename methods to `feed()` and adjust events & docs accordingly after DB migration
```

Robert-Steiner (approved):

discovery_engine_core/core/src/engine.rs
janpetschexain: "fed" as the past of "to feed" (to return the previously fed documents, ie the documents of the feed so far), similarly like in "searched" for the active search. i see that this can lead to some confusion, we probably have to visit the `restore_` naming issue later again.

janpetschexain:
bors merge

PR #520, opened 2022-07-27
DEVOPS-959_Move_jfrog_action_WIP

APB-s:
- The action was first implemented in **xayn_infrastructure**

- The idea is to have an action running once a week and delete by default the old files (29 days +).

- This action is meant to be used with manual parameter to delete specific Artifactory 

- Tasks [DEVOPS-959](https://xainag.atlassian.net/jira/software/projects/DEVOPS/boards/65?selectedIssue=DEVOPS-959)


.github/workflows/scheduled_cleanup.yaml
```diff
@@ -0,0 +1,38 @@
...
+     name: install jfrog cli
+     runs-on: ubuntu-latest
+     steps:
+       - uses: actions/checkout@v2
+       - uses: jfrog/setup-jfrog-cli@v2
+         env:
+           JF_ENV_DEVOPS: ${{secrets.JF_SECRET_ENV_DEVOPS}}
+       - run: |
+           if [ -n ${{env.artifactory_name}} ]; then
+           sed -i '/15/ i ${{env.artifactory_name}}' .github/scripts/artifactory.spec
```
acrrd, line 31: What's 15 and ddddd in the next sed?

.github/scripts/artifactory.spec
```diff
@@ -0,0 +1,23 @@
...
+    "aql": {
+      "items.find": {
+        "repo": {"$eq":"dart.yellow.private"},
+        "path": {"$match":"*"},
+        "name": {"$match":"*"},
+        "$or": [
+          {
+            "$and": [
+              {
+                "created": { "$before":"number_days_limit" }
```
APB-s, line 13: Number_days_limit is a value where all the artifacts older than this value will be deleted.

.github/workflows/scheduled_cleanup.yaml
```diff
@@ -0,0 +1,35 @@
...
+  jfrog_packages_cleanup:
+     name: install jfrog cli
+     runs-on: ubuntu-latest
+     steps:
+       - uses: actions/checkout@v2
+       - uses: jfrog/setup-jfrog-cli@v2
+         env:
+           JF_ENV_DEVOPS: ${{secrets.JF_SECRET_ENV_DEVOPS}}
+       - run: |
+           if [ -n ${{inputs.artifactory_name}} ]; then
```
APB-s, line 27: if the user has enter an artifactory name then the line `"name": {"$match":"*"},` will be replaced by:
`"name": {"$match":"${{inputs.artifactory_name}}"},

this is possible thanks to:
`"name":.*` This will replace "name": and all of the rest of the line.

.github/workflows/scheduled_cleanup.yaml
```diff
@@ -0,0 +1,35 @@
+name: Scheduled Cleanup repo
+
+on:
+  push:
+  schedule:
+    - cron: '30 00 * * sun'
```
APB-s, line 6: This action is planned to be running once a week. 00:30 on sunday

.github/scripts/artifactory.spec
```diff
@@ -0,0 +1,23 @@
+{
+"files": [
+  {
+    "aql": {
+      "items.find": {
+        "repo": {"$eq":"dart.yellow.private"},
```
acrrd, line 6: We need to change the name to the new one

.github/workflows/scheduled_cleanup.yaml
acrrd: I don't think this is good enough for our case. We want to delete all the artifacts that have been built from a branch given in input. With this match I need to specify exactly the name and version

.github/scripts/artifactory.spec
APB-s: Done

acrrd (approved):

PR #521, opened 2022-07-27
[Merged by Bors] - TY-3018 added support for stack id to db.

rustonaut:
- Adds a `StackDocument` and `Stack` table
  - The `Stack` table is needed as there is no nice way to add a foreign key constraint in
    retrospective to the `StackDocument.stackId` column
  - but we do not add any fields from the "stacks in db instead of serialized state task",
    we can add them easily later
- Adds code to keep the `Stack`s in the db in sync with the stacks we have in code.
  - :pushpin: This is a MVP solution using a hard coded list of ids which is prone to be forgotten
    when changing  stacks we can improve on this when we fully use the `Stack` table.
  - :pushpin: We also should  decide what happens with documents from stacks we no longer have,
    for now they will forget from which stack they where, but this means they are non differentiable
    from documents from the search.
- Updated `FeedScope` to set the stack for documents coming from the feed (and in turn from the stack)


discovery_engine_core/core/src/storage.rs
```diff
@@ -65,7 +70,11 @@ pub(crate) trait FeedScope {
     async fn fetch(&self) -> Result<Vec<ApiDocumentView>, Error>;
 
     // helper function. will be replaced later by move_from_stacks_to_feed
-    async fn store_documents(&self, documents: &[NewDocument]) -> Result<(), Error>;
+    async fn store_documents(
+        &self,
+        documents: &[NewDocument],
+        stack_ids: &HashMap<document::Id, stack::Id>,
```
rustonaut, line 76: In future when the stacks use the db directly this will be a bit nicer as we will always insert only a batch of documents for the same stack.

discovery_engine_core/core/src/storage/sqlite.rs
```diff
@@ -342,13 +369,14 @@ impl FeedScope for SqliteStorage {
...
                 nr.datePublished, nr.source, nr.market, nc.domainRank, nc.score,
-                ur.userReaction, po.inBatchIndex, em.embedding
+                ur.userReaction, po.inBatchIndex, em.embedding, st.stackId
             FROM FeedDocument           AS fd
             JOIN NewsResource           AS nr   USING(documentId)
             JOIN NewscatcherData        AS nc   USING(documentId)
             JOIN PresentationOrdering   AS po   USING(documentId)
             JOIN UserReaction           AS ur   USING(documentId)
             JOIN Embedding              AS em   USING(documentId)
+            JOIN StackDocument          As st   USING(documentId)
```
rustonaut, line 379: here we always have a stack so a `JOIN` is ok

discovery_engine_core/core/src/storage/sqlite.rs
```diff
@@ -489,7 +536,7 @@ impl SearchScope for SqliteStorage {
             .push(
                 "sd.documentId, nr.title, nr.snippet, nr.topic, nr.url, nr.image,
                 nr.datePublished, nr.source, nr.market, nc.domainRank, nc.score,
-                ur.userReaction, po.inBatchIndex, em.embedding
+                ur.userReaction, po.inBatchIndex, em.embedding, null as stackId
```
rustonaut, line 539: here we never have a stack so no need to `JOIN` on the stack, still the mapping requires a `stackId` column to be included

discovery_engine_core/core/src/storage/sqlite.rs
```diff
@@ -581,7 +628,7 @@ impl SearchScope for SqliteStorage {
             "SELECT
                 hd.documentId, nr.title, nr.snippet, nr.topic, nr.url, nr.image,
                 nr.datePublished, nr.source, nr.market, nc.domainRank, nc.score,
-                ur.userReaction, po.inBatchIndex, em.embedding
+                ur.userReaction, po.inBatchIndex, em.embedding, null as stackId
```
rustonaut, line 631: here we never have a stack so no need to `LEFT JOIN` on the stack, still the mapping requires a `stackId` column to be included

discovery_engine_core/core/src/storage/migrations/20220727110259_stack_id.sql
```diff
@@ -0,0 +1,30 @@
...
+--
+--  This program is distributed in the hope that it will be useful,
+--  but WITHOUT ANY WARRANTY; without even the implied warranty of
+--  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+--  GNU Affero General Public License for more details.
+--
+--  You should have received a copy of the GNU Affero General Public License
+--  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+CREATE TABLE Stack(
```
janpetschexain, line 15: ❓ why doesn't this table and the `StackDocument` table need an `IF NOT EXISTS`?

discovery_engine_core/core/src/storage/sqlite.rs
```diff
@@ -91,7 +96,40 @@ impl SqliteStorage {
...
+            stack::exploration::Stack::id(),
+        ];
+
+        let mut query_builder = QueryBuilder::new(String::new());
+        query_builder
+            .push("INSERT INTO Stack (stackId) ")
+            .push_values(expected_ids, |mut stm, id| {
+                stm.push_bind(id);
+            })
+            .push("ON CONFLICT DO NOTHING")
```
janpetschexain, line 113: :art: space after the closing bracket of `VALUES`, and statement termination
```suggestion
            .push(" ON CONFLICT DO NOTHING;")
```

discovery_engine_core/core/src/storage/sqlite.rs
```diff
@@ -91,7 +96,40 @@ impl SqliteStorage {
...
+            .build()
+            .persistent(false)
+            .execute(&mut *tx)
+            .await
+            .map_err(|err| Error::Database(err.into()))?;
+
+        query_builder
+            .reset()
+            .push("DELETE FROM Stack WHERE stackId not in ")
```
janpetschexain, line 122: ❓ shouldn't key words be upper case?
```suggestion
            .push("DELETE FROM Stack WHERE stackId NOT IN ")
```

discovery_engine_core/core/src/storage/sqlite/utils.rs
```diff
@@ -0,0 +1,41 @@
...
+    where
+        I: IntoIterator,
+        I::Item: 'args + Encode<'args, DB> + Send + Type<DB>,
+    {
+        let mut separated = self.push("(").separated(", ");
+        for value in values {
+            separated.push_bind(value);
+        }
+        separated.push_unseparated(")");
+        drop(separated);
```
janpetschexain, line 38: 🔧 the explicit drop is not required
```suggestion

```

discovery_engine_core/core/src/storage/sqlite/utils.rs
```diff
@@ -0,0 +1,41 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+use sqlx::{Database, Encode, QueryBuilder, Type};
+
+pub(super) trait SqlxPushTupleExt<'args, DB: Database> {
```
janpetschexain, line 17: ❓ since this is a `sqlx` utitlity and not a `sqlite` utility, should we put this module somewhere else? maybe as `crate::storage::sqlx` or sth similar?

discovery_engine_core/core/src/storage/sqlite.rs
```diff
@@ -91,7 +96,40 @@ impl SqliteStorage {
         Ok(Self { pool })
     }
 
-    async fn begin_tx(&self) -> Result<Transaction<'_, Sqlite>, Error> {
+    async fn setup_keep_stacks_in_sync(tx: &mut Transaction<'_, Sqlite>) -> Result<(), Error> {
```
janpetschexain, line 99: :art: maybe this?
```suggestion
    async fn setup_stacks_sync(tx: &mut Transaction<'_, Sqlite>) -> Result<(), Error> {
```

discovery_engine_core/core/src/storage.rs
janpetschexain, line 76: ❓ can't we instead add a `stack_id` field to `NewDocument`? this seems less work than to collect the doc-to-stack-id mapping.

discovery_engine_core/core/src/storage/sqlite.rs
```diff
@@ -91,7 +96,40 @@ impl SqliteStorage {
...
+        let expected_ids = &[
```
janpetschexain, line 100: 🔧 since all values are constant, we can also make the slice constant
```suggestion
        const EXPECTED_IDS: &[stack::Id] = &[
```

discovery_engine_core/core/src/storage/sqlite.rs
```diff
@@ -489,7 +548,7 @@ impl SearchScope for SqliteStorage {
             .push(
                 "sd.documentId, nr.title, nr.snippet, nr.topic, nr.url, nr.image,
                 nr.datePublished, nr.source, nr.market, nc.domainRank, nc.score,
-                ur.userReaction, po.inBatchIndex, em.embedding
+                ur.userReaction, po.inBatchIndex, em.embedding, null as stackId
```
Robert-Steiner, line 551: 🎨 
```suggestion
                ur.userReaction, po.inBatchIndex, em.embedding, NULL AS stackId
```

discovery_engine_core/core/src/storage/sqlite.rs
```diff
@@ -581,7 +640,7 @@ impl SearchScope for SqliteStorage {
             "SELECT
                 hd.documentId, nr.title, nr.snippet, nr.topic, nr.url, nr.image,
                 nr.datePublished, nr.source, nr.market, nc.domainRank, nc.score,
-                ur.userReaction, po.inBatchIndex, em.embedding
+                ur.userReaction, po.inBatchIndex, em.embedding, null as stackId
```
Robert-Steiner, line 643: 🎨 
```suggestion
                ur.userReaction, po.inBatchIndex, em.embedding, NULL AS stackId
```

discovery_engine_core/core/src/storage/migrations/20220727110259_stack_id.sql
rustonaut: good point they need it as much as any other table in our migrations :sweat_smile: 

I will add it here but we might want to discuss this in slack.

Mainly do we want a migration which is run twice to:
- error
- not error, but use the tables of the first run, which might not be the same if you change a migration

discovery_engine_core/core/src/storage/sqlite.rs
rustonaut: :rocket: 

the statement termination btw. is fully optional and it's not uncommon for query builders to expect you to not include it

through it might be useful in case you forget a `reset` on a query builder. But I'm not sure if, as I haven't tried out what happens if you place multiple statements in the same query.

discovery_engine_core/core/src/storage/sqlite.rs
rustonaut: :rocket:  yes, stylistically that would be better

discovery_engine_core/core/src/storage/sqlite/utils.rs
rustonaut: I considered it, but thought that given that we only have sqlite for not it is for now easier to place it into the sqlite `utils` module and move it when needed.

But thinking about it more it probably is better to move it now to avoid updating imports.

discovery_engine_core/core/src/storage/sqlite.rs
rustonaut: In the best case this will  be turned to exactly the same code as I'm using and in the worst case would turn it into a `static` which permanently would increase the runtime memory.

There is AFIK no drawback of having static stack values in "cold" code.

discovery_engine_core/core/src/storage.rs
rustonaut: I wanted to avoid this as `NewDocument` is currently not meant to be specific to the sub-system which added it to the document (i.e. sub-system  == feed, search, etc. we don't have terminology for this yet).

And the stack id and how it's handled is quite specific to the feed.

Furthermore when we have the full database soon in sqlite we will add new documents to a specific stack, i.e. all their stack ids need to be the same.  For which I would like to use `.stack().store_new_documents(StackId, &[NewDocument])` as this most reliable represents the mechanism involved.

And for the current situation using the hash map is the simplest solution which touches the least amount of unrelated code.

janpetschexain (approved):
lgtm

Robert-Steiner (approved):

rustonaut:
bors merge

PR #523, opened 2022-07-28
[Merged by Bors] - TY-3018 cleanup

rustonaut:
- cleans up error handling
- Fix incorrect `NoRowFound` error handling.

  A `NoRowFound` error can only be returned by methods which require a  row to be returned. E.g. by `fetch_one` but not `fetch_all` or   `fetch_optional`.
- Removed premature usage of `.persitent(false)`

  In general this should only be set in cases where a "dynamic"  query is build (i.e. using `push_{values,tuple}`) is used. Even   in some cases where `push_{values,tuple}` is used it might make  sense to not set it to false (potentially conditionally).

  Every non-dynamic  use of `.persistent(false)` is at best a premature optimization we should avoid for now AFIK.
- Remove `QueryBuilder` misuse

  Only use `QueryBuilder` if a query needs to be dynamically build,  e.g. `push_values` or `push_tuple` is used or conditionally `push`  is called. Using it with a static query has no benefits at all.
 - Move `SqlxSqliteErrorExt` to sqlite/utils.rs.
 - Added `.on_row_not_found` `Result<V, sqlx::Error>` extension.

**Based On:**
- #521

### DIFF
--- a/discovery_engine_core/core/src/storage/sqlite.rs
+++ b/discovery_engine_core/core/src/storage/sqlite.rs
@@ -82,14 +82,9 @@ pub(crate) struct SqliteStorage {
 
 impl SqliteStorage {
     pub(crate) async fn connect(uri: &str) -> Result<Self, Error> {
-        let opt = SqliteConnectOptions::from_str(uri)
-            .map_err(|err| Error::Database(err.into()))?
-            .create_if_missing(true);
+        let opt = SqliteConnectOptions::from_str(uri)?.create_if_missing(true);
 
-        let pool = SqlitePoolOptions::new()
-            .connect_with(opt)
-            .await
-            .map_err(|err| Error::Database(err.into()))?;
+        let pool = SqlitePoolOptions::new().connect_with(opt).await?;
 
         Ok(Self { pool })
     }
@@ -112,8 +107,7 @@ impl SqliteStorage {
             .build()
             .persistent(false)
             .execute(&mut *tx)
-            .await
-            .map_err(|err| Error::Database(err.into()))?;
+            .await?;
 
         query_builder
             .reset()
@@ -122,20 +116,8 @@ impl SqliteStorage {
             .build()
             .persistent(false)
             .execute(tx)
-            .await
-            .map_err(|err| Error::Database(err.into()))?;
+            .await?;
         Ok(())
-    }
-
-    async fn begin_tx(&self) -> Result<Transaction<'static, Sqlite>, Error> {
-        self.pool
-            .begin()
-            .await
-            .map_err(|err| Error::Database(err.into()))
-    }
-
-    async fn commit_tx(tx: Transaction<'_, Sqlite>) -> Result<(), Error> {
-        tx.commit().await.map_err(|err| Error::Database(err.into()))
     }
 
     async fn store_new_documents(
@@ -163,8 +145,7 @@ impl SqliteStorage {
                 .build()
                 .persistent(false)
                 .execute(&mut *tx)
-                .await
-                .map_err(|err| Error::Database(err.into()))?;
+                .await?;
 
             // insert id into HistoricDocument table
             query_builder
@@ -176,8 +157,7 @@ impl SqliteStorage {
                 .build()
                 .persistent(false)
                 .execute(&mut *tx)
-                .await
-                .map_err(|err| Error::Database(err.into()))?;
+                .await?;
 
             // insert data into NewsResource table
             query_builder
@@ -201,8 +181,7 @@ impl SqliteStorage {
             .build()
             .persistent(false)
             .execute(&mut *tx)
-            .await
-            .map_err(|err| Error::Database(err.into()))?;
+            .await?;
 
             // insert data into NewscatcherData table
             query_builder
@@ -218,8 +197,7 @@ impl SqliteStorage {
                 .build()
                 .persistent(false)
                 .execute(&mut *tx)
-                .await
-                .map_err(|err| Error::Database(err.into()))?;
+                .await?;
 
             // insert data into UserReaction table
             query_builder
@@ -232,8 +210,7 @@ impl SqliteStorage {
                 .build()
                 .persistent(false)
                 .execute(&mut *tx)
-                .await
-                .map_err(|err| Error::Database(err.into()))?;
+                .await?;
 
             // insert data into PresentationOrdering table
             query_builder
@@ -249,8 +226,7 @@ impl SqliteStorage {
                 .build()
                 .persistent(false)
                 .execute(&mut *tx)
-                .await
-                .map_err(|err| Error::Database(err.into()))?;
+                .await?;
 
             // insert data into Embedding table
             query_builder
@@ -262,8 +238,7 @@ impl SqliteStorage {
                 .build()
                 .persistent(false)
                 .execute(&mut *tx)
-                .await
-                .map_err(|err| Error::Database(err.into()))?;
+                .await?;
         }
 
         Ok(())
@@ -288,8 +263,7 @@ impl SqliteStorage {
                 .build()
                 .persistent(false)
                 .execute(&mut *tx)
-                .await
-                .map_err(|err| Error::Database(err.into()))?
+                .await?
                 .rows_affected()
                 > 0;
         }
@@ -306,14 +280,14 @@ impl Storage for SqliteStorage {
             .await
             .map_err(|err| Error::Database(err.into()))?;
 
-        let mut tx = self.begin_tx().await?;
+        let mut tx = self.pool.begin().await?;
         Self::setup_stacks_sync(&mut tx).await?;
-        Self::commit_tx(tx).await?;
+        tx.commit().await?;
         Ok(())
     }
 
     async fn fetch_history(&self) -> Result<Vec<HistoricDocument>, Error> {
-        let mut tx = self.begin_tx().await?;
+        let mut tx = self.pool.begin().await?;
 
         let documents = sqlx::query_as::<_, QueriedHistoricDocument>(
             "SELECT
@@ -332,7 +306,7 @@ impl Storage for SqliteStorage {
             }
         })?;
 
-        Self::commit_tx(tx).await?;
+        tx.commit().await?;
 
         documents.into_iter().map(TryInto::try_into).collect()
     }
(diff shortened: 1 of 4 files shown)


discovery_engine_core/core/src/storage/sqlite/utils.rs
```diff
@@ -39,3 +43,75 @@ where
...
+pub(crate) trait SqlxSqliteErrorExt<V> {
+    /// Use this on the result of a sqlite query which might have failed a foreign
+    /// key constraint when an illegal document was passed in.
+    ///
+    /// For example it can be used in `.user_reacted` to handle it being
+    /// called with a random document id.
+    ///
+    /// If there is an error and it is a fk violation an appropriate error variant is
+    /// used instead of the default generic database error.
+    fn fk_violation_is_invalid_document_id(self, id: document::Id) -> Result<V, Error>;
```
janpetschexain, line 56: ❓ should we also use the `on_` naming since this method acts on an error as well?
```suggestion
    fn on_fk_violation_is_invalid_document_id(self, id: document::Id) -> Result<V, Error>;
```

discovery_engine_core/core/src/storage/sqlite/utils.rs
```diff
@@ -39,3 +43,75 @@ where
         self
     }
 }
+
+pub(crate) trait SqlxSqliteErrorExt<V> {
```
janpetschexain, line 47: :art: if we use `T` as the generic parameter, then the name is consistent with the `Result<T, E>` definition of `std`.
```suggestion
pub(crate) trait SqlxSqliteErrorExt<T> {
```

discovery_engine_core/core/src/storage/sqlite/utils.rs
```diff
@@ -39,3 +43,75 @@ where
...
+
+        sqlx::query("CREATE TABLE Bar(x INTEGER PRIMARY KEY REFERENCES Foo(x));")
+            .execute(&storage.pool)
+            .await
+            .unwrap();
+
+        let document_id = document::Id::new();
+
+        let res = sqlx::query("INSERT INTO Bar(x) VALUES (?);")
+            .bind(10)
```
janpetschexain, line 103: ❓ not sure if i understand this correctly, shouldn't this bind the `document_id`?

discovery_engine_core/core/src/storage/sqlite.rs
```diff
@@ -437,10 +375,10 @@ impl FeedScope for SqliteStorage {
             .build()
             .persistent(false)
             .execute(&mut tx)
-            .await
-            .map_err(|err| Error::Database(err.into()))?;
+            .await?;
 
-        Self::commit_tx(tx).await
+        tx.commit().await?;
+        Ok(())
```
janpetschexain, line 381: :art: this could be handled as a single expression
```suggestion
        tx.commit().await.map_err(Into::into)
```

discovery_engine_core/core/src/storage/sqlite.rs
```diff
@@ -451,102 +389,79 @@ impl SearchScope for SqliteStorage {
...
             .bind(search.paging.next_page)
             .execute(&mut tx)
-            .await
-            .map_err(|err| Error::Database(err.into()))?;
+            .await?;
 
         if documents.is_empty() {
-            return Self::commit_tx(tx).await;
+            tx.commit().await?;
+            return Ok(());
```
janpetschexain, line 406: :art: this could be handled as a single statement
```suggestion
            return tx.commit().await.map_err(Into::into);
```

discovery_engine_core/core/src/storage/sqlite.rs
```diff
@@ -451,102 +389,79 @@ impl SearchScope for SqliteStorage {
...
         };
 
         SqliteStorage::store_new_documents(&mut tx, documents).await?;
-        query_builder
-            .reset()
-            .push("SearchDocument (documentId) ")
+
+        QueryBuilder::new("INSERT INTO SearchDocument (documentId) ")
```
janpetschexain, line 411: 🔧 somewhat unrelated to the current change, i just noticed that this should be wrapped into a
```rust
for documents in documents.chunks(BIND_LIMIT) {}
```
loop. even though this case is quite unlikely, it's better to be safe. similarly in other methods where we can't guarantee that the `BIND_LIMIT` is not exceeded.

discovery_engine_core/core/src/storage/sqlite.rs
```diff
@@ -451,102 +389,79 @@ impl SearchScope for SqliteStorage {
...
             .build()
             .persistent(false)
             .execute(&mut tx)
-            .await
-            .map_err(|err| Error::Database(err.into()))?;
+            .await?;
 
-        Self::commit_tx(tx).await
+        tx.commit().await?;
+        Ok(())
```
janpetschexain, line 421: :art: this could be handled in a single expression
```suggestion
        tx.commit().await.map_err(Into::into)
```

discovery_engine_core/core/src/storage/sqlite.rs
```diff
@@ -451,102 +389,79 @@ impl SearchScope for SqliteStorage {
...
+        sqlx::query("UPDATE Search SET pageNumber = ? WHERE rowid = 1;")
             .bind(page_number)
             .execute(&mut tx)
-            .await
-            .map_err(|err| Error::Database(err.into()))?;
+            .await?;
 
-        Self::commit_tx(tx).await
+        tx.commit().await?;
+        Ok(())
```
janpetschexain, line 448: :art: this could be handled in a single expression
```suggestion
        tx.commit().await.map_err(Into::into)
```

discovery_engine_core/core/src/storage/sqlite.rs
```diff
@@ -954,6 +829,32 @@ mod tests {
...
+            .search()
+            .store_new_search(&new_search, &[])
+            .await
+            .unwrap();
+
+        let search = storage.search().fetch().await.unwrap();
+        assert!(search.1.is_empty());
+
+        let a_search_was_closed = storage.search().clear().await.unwrap();
+        assert!(a_search_was_closed);
```
janpetschexain, line 855: :art:
```suggestion
        assert!(storage.search().fetch().await.unwrap().1.is_empty());
        assert!(storage.search().clear().await.unwrap());
```

discovery_engine_core/core/src/storage/sqlite/utils.rs
rustonaut: In sqlite who only know that a FK-violation happened but not which fields it touched.

Wrt. this test I just test what the method checks, out of context of everything else. While I could rename `x` to `document_id` it could lead to some misunderstanding of what the method does.

This is a bit less misleading I hope  with how I changed the method in context of this review.

As a side note: This is different in postgres where you can get additional information and can name the check/fk/unique constraints. So in case of postgres we could have some more reliable/automatic conversion. This is (one of) the reasons why this extension is sqlite specific (and why I called it `SqlxSqliteErrorExt`).

discovery_engine_core/core/src/storage/sqlite.rs
rustonaut: Through it doesn't convey the same meaning:

- "commit and handle the error"; "At the end of function return a empty ok result."
- "commit and return the result from the commit from the function while mapping the error"

The only reason why we can use `map_err` is because it happens that `commit` has the same return type as the function and it happens that the type only has one valid value.

discovery_engine_core/core/src/storage/sqlite.rs
rustonaut: IMHO I mentioned it before but I still think we should not bother with handling `BIND_LIMIT` it doesn't add any value for our use cases as in no situations will we have a SQL queries which inserts that many documents for the same user (like we currently add at most 10 in practice, and this is very unlikely to ever be over 100).  The bind limit does even by far out class the number of documents we ever want to have in the database at any point in time AFIK. 

This means that basically any situation where we have more then `BIND_LIMIT` documents (or other entities) is basically a unexpected bug-like situation where having an error should be ok.

The think is even if we do use chunking by `BIND_LIMIT`, it is very maintenance error prone as it is trivial to forget to update the divider used and in turn still run into an error because it runs out of possible binds.

This is also prone to cause a lot of maintenance/review friction over time (forgotten loop, forgotten updated divider). It also can any make any code using a `WHERE id not in <tuple>`  more complex for no practical benefit.

Either way I would say it is out of scope (:pushpin: ) of this PR as it would need to be changed in several places, not just here. And might entail further discussion.

discovery_engine_core/core/src/storage/sqlite.rs
rustonaut: I would say assigning it to a temporary variable conveys the meaning of what is changed better.

rustonaut: :rocket:  (changed it anyway)

discovery_engine_core/core/src/storage/sqlite.rs
rustonaut: (as a side note this is one of the kind of situations I was asking about with the poll on slack, where you could fix it in this PR but I generally think its better to add a :pushpin: , maybe a task with maybe even a estimate and do it in a follow up)

discovery_engine_core/core/src/storage/sqlite.rs
janpetschexain: nonetheless this distinction not really needed here, imo. the commit can be regarded as the final operation of the method, hence the result of the commit is to a certain extend also the result of the method, hence it's not just error mapping, but also returning the result of the commit as the result of the method.

discovery_engine_core/core/src/storage/sqlite.rs
janpetschexain: i'm ok either way, fully remove it or fully adapt it, but it should be handled uniformly across the code (in a different pr of course).

discovery_engine_core/core/src/storage/sqlite.rs
rustonaut: <details>
<summary><- Me reasoning that this is not the case and we should generally use `?` where applicable.</summary>

it's not the same, the result of commit has nothing to do with the return value of `commit` as     long as it's not an error, e.g. in every case where we don't return `Ok(())` we return a value created before calling `commit()`.

More important look at if not from your or mine POV but of the POV of someone new to rust.

`?` is what they learn should always be used for error handling, `map_err` only appears much later in context of "when automatic error conversion doesn't work" , and even there extension traits like `context` and similar are often promoted over `.map_err`. For someone new to rust `.map_err(Into::into)` is only really nicely readable if they:

1. fully understand error handling 
2. understand the trait system
3. understand that methods are also free functions, and free functions are also closures

Or, more likely, learn it as another specific pattern.

Additionally it (minimally) increases the size of PRs changing the return type and in turn increases the chance for merge conflicts (negligible).

Sure both points are small, but I really can't agree that a stile which is harder to read for beginners and (negligible) bloats PRs without any benefit is better (or I'm missing the benefit).
</details>

I'm pretty sure we had that discussion before and for me it's both trivial to read so I will change it so that we can move on with this PR.

janpetschexain (approved):

rustonaut:
bros merge

rustonaut:
bors merge

PR #524, opened 2022-07-28
[Merged by Bors] - TY-2979 move feed logic to rust (3): changeConfiguration

janpetschexain:
**References**

- [TY-2979]
- requires #519

**Summary**

- add `FeedConfig` and `SearchConfig` & add corresponding ffi in `InitConfig`
- align field access and function calls for `InitConfig` with its field declaration order for better readability
- add `Engine::configure()` & corresponding ffi
- remove `EventConfig` and use new logic in `SystemManager.changeConfiguration()` in dart
- update [confluence config docs](https://xainag.atlassian.net/wiki/spaces/M2D/pages/2476605445/Discovery+Engine+Configurations)

[TY-2979]: https://xainag.atlassian.net/browse/TY-2979?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine_core/bindings/src/types/init_config.rs
+++ b/discovery_engine_core/bindings/src/types/init_config.rs
@@ -19,6 +19,8 @@ use std::ptr::addr_of_mut;
 use xayn_discovery_engine_core::InitConfig;
 use xayn_discovery_engine_providers::Market;
 
+use super::primitives::FfiUsize;
+
 /// Returns a pointer to the `api_key` field of a configuration.
 ///
 /// # Safety
@@ -39,6 +41,32 @@ pub unsafe extern "C" fn init_config_place_of_api_key(place: *mut InitConfig) ->
 #[no_mangle]
 pub unsafe extern "C" fn init_config_place_of_api_base_url(place: *mut InitConfig) -> *mut String {
     unsafe { addr_of_mut!((*place).api_base_url) }
+}
+
+/// Returns a pointer to the `news_provider_path` field of a configuration.
+///
+/// # Safety
+///
+/// The pointer must point to a valid [`InitConfig`] memory object,
+/// it might be uninitialized.
+#[no_mangle]
+pub unsafe extern "C" fn init_config_place_of_news_provider_path(
+    place: *mut InitConfig,
+) -> *mut String {
+    unsafe { addr_of_mut!((*place).news_provider_path) }
+}
+
+/// Returns a pointer to the `headlines_provider_path` field of a configuration.
+///
+/// # Safety
+///
+/// The pointer must point to a valid [`InitConfig`] memory object,
+/// it might be uninitialized.
+#[no_mangle]
+pub unsafe extern "C" fn init_config_place_of_headlines_provider_path(
+    place: *mut InitConfig,
+) -> *mut String {
+    unsafe { addr_of_mut!((*place).headlines_provider_path) }
 }
 
 /// Returns a pointer to the `markets` field of a configuration.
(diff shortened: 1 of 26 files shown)


discovery_engine_core/core/src/engine.rs
```diff
@@ -272,6 +288,16 @@ impl Engine {
...
+        let feed_config = FeedConfig {
+            max_docs_per_batch: de_config
+                .extract_inner("feed.max_docs_per_batch")
+                .unwrap_or(config.max_docs_per_feed_batch as usize),
+        };
+        let search_config = SearchConfig {
+            max_docs_per_batch: de_config
+                .extract_inner("search.max_docs_per_batch")
+                .unwrap_or(config.max_docs_per_search_batch as usize),
+        };
```
rustonaut, line 300: :question: Is there any reason why we make it configurable twice?

- once through the configuration
- once through the deConfig

rustonaut, line 300: :wrench:  I would prefer if we don't repeat the (de-)serialization logic of the `{Feed,Search}Config`:

```suggestion
        let mut feed_config = FeedConfig {
            max_docs_per_batch: config.max_docs_per_feed_batch as usize
        };
        feed_config.merge(&de_config);

        let mut search_config = SearchConfig {
            max_docs_per_batch: config.max_docs_per_search_batch as usize,
        };
        search_config.merge(&de_config);
```

discovery_engine_core/bindings/src/types/init_config.rs
```diff
@@ -146,30 +174,30 @@ pub unsafe extern "C" fn init_config_place_of_kpe_classifier(
...
 /// The pointer must point to a valid [`InitConfig`] memory object,
 /// it might be uninitialized.
 #[no_mangle]
-pub unsafe extern "C" fn init_config_place_of_news_provider_path(
+pub unsafe extern "C" fn init_config_place_of_max_docs_per_feed_batch(
     place: *mut InitConfig,
-) -> *mut String {
-    unsafe { addr_of_mut!((*place).news_provider_path) }
+) -> *mut FfiUsize {
+    unsafe { addr_of_mut!((*place).max_docs_per_feed_batch) }.cast()
```
rustonaut, line 187: Idk. if I like the way we use FffiUsize here, it's I think a bit dangerous to lead to later-change introduced bugs.

But given the whole context (we will replace it with a message passing API, we can reuse the bounds checks, etc.) I agree that it's probably the best thing to do here in this PR.

discovery_engine_core/core/src/config.rs
```diff
@@ -193,9 +197,62 @@ impl Default for ExplorationConfig {
...
+impl FeedConfig {
+    /// Merges existent values from the DE configuration into this configuration.
+    pub(crate) fn merge(&mut self, de_config: &Figment) {
+        if let Ok(max_docs_per_batch) = de_config.extract_inner("feed.max_docs_per_batch") {
+            self.max_docs_per_batch = max_docs_per_batch;
+        }
+    }
+}
+
+impl Default for FeedConfig {
```
rustonaut, line 216: :wrench:  To make it clear the "default" value is semi arbitrary/doesn't need to be maintained/might be out of sync with the defaults the actual app uses.

```suggestion
#[cfg(test)]
impl Default for FeedConfig {
```

discovery_engine_core/core/src/config.rs
```diff
@@ -193,9 +197,62 @@ impl Default for ExplorationConfig {
...
+impl SearchConfig {
+    /// Merges existent values from the DE configuration into this configuration
+    pub(crate) fn merge(&mut self, de_config: &Figment) {
+        if let Ok(max_docs_per_batch) = de_config.extract_inner("search.max_docs_per_batch") {
+            self.max_docs_per_batch = max_docs_per_batch;
+        }
+    }
+}
+
+impl Default for SearchConfig {
```
rustonaut, line 240: :wrench:  (see above)

```suggestion
#[cfg(test)]
impl Default for SearchConfig {
```

discovery_engine_core/core/src/engine.rs
janpetschexain: it would be somewhat unexpected if this is not configurable via the `de_config`, since this config is supposed to allow to override every all configurations.

i even thought about removing these two config values from the `InitConfig`, but this should be discussed beforehand, and we have to check if blue uses these at all (as far as i can tell they have hardcoded constant values, so we could just make them internal defaults and just keep the `de_config` override).

acrrd: No particularly strong opinion but I'm more to remove it from the `de_config`, none of the values in the `InitConfig` are configurable using `de_config`, moreover we allows to change these values at runtime like market and excluded sources so I'm more to keep consistency with them and the rest of `InitConfig`.

rustonaut (approved):
lgmt

janpetschexain:
bors merge

PR #527, opened 2022-08-02
[Merged by Bors] - TY-3050 MLT (more like this) provider

finiteprods:
**Summary**

adds a provider to the MLT endpoint for fetching similar articles.

### DIFF
--- a/discovery_engine_core/providers/src/lib.rs
+++ b/discovery_engine_core/providers/src/lib.rs
@@ -71,13 +71,13 @@ pub use helpers::{
 use async_trait::async_trait;
 use url::Url;
 
-/// Abstraction over a provider for a news searching functionality.
+/// Provider for news search functionality.
 #[async_trait]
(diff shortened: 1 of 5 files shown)


discovery_engine_core/providers/src/lib.rs
```diff
@@ -101,6 +104,15 @@ pub trait TrendingTopicsProvider: Send + Sync {
     ) -> Result<Vec<TrendingTopic>, Error>;
 }
 
+/// Abstraction over a provider for similar news.
```
janpetschexain, line 107: :art: the naming is somewhat redundant, because this is a trait after all. i saw that the doc naming is also used for the other traits, maybe it's worth updating their docs as well.
```suggestion
/// A provider for similar news.
```

finiteprods: thanks, I cut out the unnecessary words from all the provider docstrings.

finiteprods:
bors merge

PR #528, opened 2022-08-02
[Merged by Bors] - TY-3054 impl db code for source preferences

Robert-Steiner:
Ticket:
- [TY-3054]

[TY-3054]: https://xainag.atlassian.net/browse/TY-3054?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine_core/core/src/storage/migrations/20220801152246_source_preference.sql
+++ b/discovery_engine_core/core/src/storage/migrations/20220801152246_source_preference.sql
@@ -1,0 +1,19 @@
+--  Copyright 2022 Xayn AG
+--
+--  This program is free software: you can redistribute it and/or modify
+--  it under the terms of the GNU Affero General Public License as
+--  published by the Free Software Foundation, version 3.
+--
+--  This program is distributed in the hope that it will be useful,
+--  but WITHOUT ANY WARRANTY; without even the implied warranty of
+--  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+--  GNU Affero General Public License for more details.
+--
+--  You should have received a copy of the GNU Affero General Public License
+--  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+CREATE TABLE IF NOT EXISTS SourcePreference (
+    source TEXT NOT NULL PRIMARY KEY,
+    -- 1 = preferred, 2 = excluded
+    preference INTEGER NOT NULL
+);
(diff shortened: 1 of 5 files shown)


discovery_engine_core/core/src/storage/migrations/20220801152246_source_preference.sql
```diff
@@ -0,0 +1,19 @@
...
+--  but WITHOUT ANY WARRANTY; without even the implied warranty of
+--  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+--  GNU Affero General Public License for more details.
+--
+--  You should have received a copy of the GNU Affero General Public License
+--  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+CREATE TABLE IF NOT EXISTS SourcePreference (
+    source TEXT NOT NULL PRIMARY KEY,
+    -- 1 = preferred, 2 = excluded
```
Robert-Steiner, line 17: Do we want to keep the name `preferred` or would it be better to use `trusted`?

finiteprods: I'd say `trusted` as that's how we have been calling it

discovery_engine_core/core/src/storage/sqlite.rs
```diff
@@ -1129,4 +1215,60 @@ mod tests {
...
+            .unwrap();
+
+        let trusted_db = storage.source_preference().fetch_trusted().await.unwrap();
+        assert_eq!(trusted_db, trusted_sources);
+        let excluded_db = storage.source_preference().fetch_excluded().await.unwrap();
+        assert_eq!(excluded_db, excluded_sources);
+
+        // set the excluded source and so far unknown source as new trusted sources
+        // excluded sources should be empty
+        // trusted sources should return {"d", "c"}
```
finiteprods, line 1252: :art:  would help to clarify a bit as in:
```suggestion
        // set the excluded source "c" and so far unknown source "d" as new trusted sources
        // excluded sources should be empty
        // trusted sources should return {"d", "c"}
```
:grey_question:  so when you set an excluded source as trusted, it stops being excluded. how about if you set a trusted source as excluded, is it worth testing this as well?

finiteprods (approved):
minor suggestion in the test otherwise looks good

discovery_engine_core/core/src/storage/migrations/20220801152246_source_preference.sql
Robert-Steiner: i have changed it to `trusted`

discovery_engine_core/core/src/storage/sqlite.rs
Robert-Steiner: good point 👍 i have added a test for this as well

Robert-Steiner:
bors merge

PR #530, opened 2022-08-02
[Merged by Bors] - TY-3066 port dart logic to rust

Robert-Steiner:
Ticket:
- [TY-3066]

Summary:

- implemented the  `set_trusted_sources`, `trusted_sources`, `set_excluded_sources`, `excluded_sources`, `add_trusted_sources`, `remove_trusted_sources`, `add_excluded_sources` and `remove_excluded_sources` method

based on #528 

[TY-3066]: https://xainag.atlassian.net/browse/TY-3066?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine_core/core/src/engine.rs
+++ b/discovery_engine_core/core/src/engine.rs
@@ -1160,7 +1160,16 @@ impl Engine {
         }
     }
 
-    /// Updates the trusted sources.
+    async fn filter_excluded_sources_for_all_stacks(&mut self, sources: &HashSet<String>) {
+        let mut stacks = self.stacks.write().await;
+        for stack in stacks.values_mut() {
+            stack.prune_by_excluded_sources(sources);
+        }
+        drop(stacks); // guard
+        self.exploration_stack.prune_by_excluded_sources(sources);
+    }
+
+    /// Sets a new list of trusted sources.
     pub async fn set_trusted_sources(
         &mut self,
         history: &[HistoricDocument],
@@ -1172,7 +1181,7 @@ impl Engine {
             .await
     }
 
-    /// Sets a new list of excluded sources
+    /// Sets a new list of excluded sources.
     pub async fn set_excluded_sources(
         &mut self,
         history: &[HistoricDocument],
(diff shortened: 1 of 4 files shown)


rustonaut (commented):
:question: There is nothing inherently wrong with it but it does approach the migration different then we do with the rest of the code.

Here it adds another alternative FFI for the with/without storage case.

But in the rest of the code base we use the same FFI but then:

1. in dart based on `cfgFeatureStorage` change how we call it (e.g. if `true` pass in a empty history as we don't have one dart anymore)

2. in rust  use the passed history, or fetch one from the db depending on `cfg(feature="storage")`

Will there be a followup PR where `set_trusted_sources` is called instead of `set_trusted_sources_with_history` depending on `cfgFeatureStorage`?

Robert-Steiner:
> ❓ There is nothing inherently wrong with it but it does approach the migration different then we do with the rest of the code.
> 
> Here it adds another alternative FFI for the with/without storage case.
> 
> But in the rest of the code base we use the same FFI but then:
> 
>     1. in dart based on `cfgFeatureStorage` change how we call it (e.g. if `true` pass in a empty history as we don't have one dart anymore)
> 
>     2. in rust  use the passed history, or fetch one from the db depending on `cfg(feature="storage")`
> 
> 
> Will there be a followup PR where `set_trusted_sources` is called instead of `set_trusted_sources_with_history` depending on `cfgFeatureStorage`?

Makes sense to just pass the empty history when we rust it with the storage feature 👍 I adjusted the code :)

Robert-Steiner:
I have added the add/remove trusted/excluded source methods as well because team blue still uses them. I'm not so happy with the split of the set trusted/excluded sources methods. In dart we have just [one method](https://github.com/xaynetwork/xayn_discovery_engine/blob/5c185a9123a821dbc4f45e6d050b814c10be8d2c/discovery_engine/lib/src/domain/feed_manager.dart#L227) and that makes the code a bit simpler. Also, if we split the methods we can't check if there are any duplicates in the provided trusted/excluded sources. That check would still remain in dart. Do you think it would make sense to also have just one `set` method in rust?

rustonaut:
> Also, if we split the methods we can't check if there are any duplicates in the provided trusted/excluded sources.

If we have two methods I don't think we need to check this as setting a source to trusted which before had been excluded will just override that state in the db.

> In dart we have just one method

I think we either should have also two methods in dart or also just one method in rust.

In the end I would say what is most appropriate depends on how it's handled in the app UX wise.

If they are handled in the UX distinctly then having separate dart methods and rust methods would be a bit easier/more efficient.

If they are handled in the same context then having a compound method is better (but then we probably also should accept a `Map<Source, excluded|trusted>`.

In rust we also might need to take different actions depending on what changed, so if we always set both we might need to add additional logic to check what changed.

**But this is on a relatively "cold" path and given the design in the current dart API  and that it is used like that by the current app I would say we  should mirror that in rust for now.** We can come back to changing it just after switching to the message passing API.

rustonaut:
@Robert-Steiner are you pushed all code, this doesn't contain any dart code changes :thinking:

Robert-Steiner:
> @Robert-Steiner are you pushed all code, this doesn't contain any dart code changes (anymore) 🤔

the dart changes are in  #538 :)

discovery_engine_core/core/src/engine.rs
```diff
@@ -1109,18 +1118,236 @@ impl Engine {
...
+    #[cfg_attr(not(feature = "storage"), allow(unused_variables))]
+    pub async fn set_sources(
+        &mut self,
+        sources: &[WeightedSource],
+        excluded: Vec<String>,
+        trusted: Vec<String>,
+    ) -> Result<(), Error> {
+        #[cfg(feature = "storage")]
+        {
+            let trusted_set = excluded.iter().cloned().collect::<HashSet<_>>();
```
rustonaut, line 1137:
```suggestion
            let trusted_set = trusted.iter().cloned().collect::<HashSet<_>>();
```

rustonaut (approved):
lgtm (besides one copy past error which needs to be fixed  :wrench:)

discovery_engine_core/core/src/engine.rs
Robert-Steiner: Oh boy! I checked it several times but I missed this one 
Nice catch!

Robert-Steiner:
bors merge

PR #532, opened 2022-08-02
[Merged by Bors] - TY-2983 db code for time spent

rustonaut:
**References:**

- [TY-2983]

**Based On:**

- #515 

[TY-2983]: https://xainag.atlassian.net/browse/TY-2983?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine_core/ai/bert/src/pooler.rs
+++ b/discovery_engine_core/ai/bert/src/pooler.rs
@@ -51,13 +51,17 @@ impl<const N: usize> From<[f32; N]> for Embedding<Ix1> {
     }
 }
 
+#[derive(Clone, Debug, Error)]
+#[error("bytes do not represent a valid embedding")]
+pub struct MalformedBytesEmbedding;
+
 impl TryFrom<Vec<u8>> for Embedding<Ix1> {
-    type Error = Vec<u8>;
+    type Error = MalformedBytesEmbedding;
 
     /// Converts from bytes in little endianness to values in standard order.
     fn try_from(bytes: Vec<u8>) -> Result<Self, Self::Error> {
         if bytes.len() % size_of::<f32>() != 0 {
-            return Err(bytes);
+            return Err(MalformedBytesEmbedding);
         }
 
         let floats = bytes
--- a/discovery_engine_core/ai/bert/src/pooler.rs
+++ b/discovery_engine_core/ai/bert/src/pooler.rs
@@ -51,8 +51,8 @@ impl<const N: usize> From<[f32; N]> for Embedding<Ix1> {
     }
 }
 
-#[derive(Clone, Debug, Error)]
-#[error("bytes do not represent a valid embedding")]
+#[derive(Clone, Debug, Display, Error)]
+/// Bytes do not represent a valid embedding.
 pub struct MalformedBytesEmbedding;
 
 impl TryFrom<Vec<u8>> for Embedding<Ix1> {
--- a/discovery_engine_core/core/src/storage.rs
+++ b/discovery_engine_core/core/src/storage.rs
@@ -12,19 +12,19 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-use std::collections::{HashMap, HashSet};
+use std::{collections::{HashMap, HashSet}, time::Duration};
 
 use async_trait::async_trait;
 use displaydoc::Display;
 use thiserror::Error;
-use xayn_discovery_engine_ai::GenericError;
+use xayn_discovery_engine_ai::{GenericError, MalformedBytesEmbedding};
 
 use crate::{
-    document::{self, HistoricDocument, UserReaction},
+    document::{self, HistoricDocument, UserReaction, ViewMode},
     stack,
 };
 
-use self::models::{ApiDocumentView, NewDocument, Search};
+use self::models::{ApiDocumentView, NewDocument, Search, TimeSpendDocumentView};
 
 pub mod sqlite;
 mod utils;
@@ -46,6 +46,12 @@ pub enum Error {
 impl From<sqlx::Error> for Error {
     fn from(generic: sqlx::Error) -> Self {
         Error::Database(generic.into())
+    }
+}
+
+impl From<MalformedBytesEmbedding> for Error {
+    fn from(err: MalformedBytesEmbedding) -> Self {
+        Error::Database(Box::new(err))
     }
 }
 
@@ -113,6 +119,13 @@ pub(crate) trait FeedbackScope {
         document: document::Id,
         reaction: UserReaction,
     ) -> Result<ApiDocumentView, Error>;
+
+    async fn update_time_spend(
+        &self,
+        document: document::Id,
+        view_mode: ViewMode,
+        view_time: Duration,
+    ) -> Result<TimeSpendDocumentView, Error>;
 }
 
 #[async_trait]
@@ -136,6 +149,8 @@ pub(crate) trait SourcePreferenceScope {
 }
 
 pub mod models {
+    use std::time::Duration;
+
     use chrono::NaiveDateTime;
     use url::Url;
     use xayn_discovery_engine_ai::Embedding;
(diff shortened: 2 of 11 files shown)


discovery_engine_core/ai/bert/src/pooler.rs
```diff
@@ -51,13 +51,17 @@ impl<const N: usize> From<[f32; N]> for Embedding<Ix1> {
     }
 }
 
+#[derive(Clone, Debug, Error)]
+#[error("bytes do not represent a valid embedding")]
+pub struct MalformedBytesEmbedding;
```
janpetschexain, line 56: 🔧 since we already use `displaydoc::Display` in this crate, we can use it here as well
```suggestion
#[derive(Clone, Debug, Display, Error)]
/// Bytes do not represent a valid embedding
pub struct MalformedBytesEmbedding;
```

discovery_engine_core/core/src/storage.rs
```diff
@@ -111,6 +117,13 @@ pub(crate) trait FeedbackScope {
         document: document::Id,
         reaction: UserReaction,
     ) -> Result<ApiDocumentView, Error>;
+
+    async fn update_time_spend(
```
janpetschexain, line 121: 🔧 past time, as in the engine
```suggestion
    async fn update_time_spent(
```

discovery_engine_core/core/src/storage.rs
```diff
@@ -286,4 +301,11 @@ pub mod models {
         pub(crate) size: u32,
         pub(crate) next_page: u32,
     }
+
+    #[derive(Debug, PartialEq)]
+    pub(crate) struct TimeSpendDocumentView {
```
janpetschexain, line 306: :wrench: past time, as in the engine
```suggestion
    pub(crate) struct TimeSpentDocumentView {
```

discovery_engine_core/core/src/storage/migrations/20220727161140_view_time.sql
```diff
@@ -0,0 +1,26 @@
...
+--
+--  This program is distributed in the hope that it will be useful,
+--  but WITHOUT ANY WARRANTY; without even the implied warranty of
+--  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+--  GNU Affero General Public License for more details.
+--
+--  You should have received a copy of the GNU Affero General Public License
+--  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+CREATE TABLE IF NOT EXISTS ViewTimes(
```
janpetschexain, line 15: ❓ the other tables in the database use singular case for the names, should this be done here as well?
```suggestion
CREATE TABLE IF NOT EXISTS ViewTime(
```

discovery_engine_core/core/src/storage/sqlite.rs
```diff
@@ -759,6 +758,82 @@ impl FeedbackScope for SqliteStorage {
...
         Ok(document)
     }
+
+    async fn update_time_spend(
+        &self,
+        document: document::Id,
+        view_mode: ViewMode,
+        duration: Duration,
+    ) -> Result<TimeSpendDocumentView, Error> {
+        const MS_PER_DAY: i64 = 1000 * 60 * 60 * 24 * 7;
```
janpetschexain, line 768: 🔧 this value represents milliseconds per week, but not per day

discovery_engine_core/core/src/storage/sqlite.rs
```diff
@@ -759,6 +758,82 @@ impl FeedbackScope for SqliteStorage {
...
+        let view_time_ms = i64::try_from(duration.as_millis())
+            .ok()
+            .unwrap_or(MS_PER_DAY);
+        let view_time_ms = min(view_time_ms, MS_PER_DAY);
+
+        let mut tx = self.pool.begin().await?;
+
+        sqlx::query(
+            "INSERT INTO ViewTimes(documentId, viewMode, viewTimeMs) VALUES (?, ?, ?)
+                ON CONFLICT DO UPDATE SET viewTimeMs =  viewTimeMs + excluded.viewTimeMs;",
```
janpetschexain, line 781: :art: 
```suggestion
                ON CONFLICT DO UPDATE SET viewTimeMs = viewTimeMs + excluded.viewTimeMs;",
```

discovery_engine_core/core/src/storage/sqlite.rs
```diff
@@ -759,6 +758,82 @@ impl FeedbackScope for SqliteStorage {
...
+            embedding,
+            aggregated_view_time,
+            user_reaction,
+        }: QueryTimeSpendDocumentView,
+    ) -> Result<Self, Self::Error> {
+        Ok(TimeSpendDocumentView {
+            smbert_embedding: embedding.try_into()?,
+            last_reaction: user_reaction_from_db(user_reaction)?,
+            aggregated_view_time: Duration::from_millis(
+                u64::try_from(aggregated_view_time).unwrap_or(0),
```
janpetschexain, line 833: ❓ should we make a distinction here and map negative values to `0` and too large values to `u64::MAX`/`MS_PER_DAY`?

discovery_engine_core/core/src/storage/migrations/20220727161140_view_time.sql
rustonaut: It's plural because it has multiple entries per document, while each other has one entry per document.

discovery_engine_core/core/src/storage/sqlite.rs
rustonaut: oh yes I changed it back and force, I will change the value

discovery_engine_core/core/src/storage/sqlite.rs
rustonaut: :rocket: For this use case `u32` is enough so I

- limited the view time to u32::MAX 
  - both when getting it from the app
  - and when upsert increasing it in SQL
- use `u32` in struct

:pushpin: sql not supporting unsigned integers is annoying, in different places we use different strategies to work around it. In all cases the out of bound handling falls into the  "extremely unlikely edge cases which likely can only appear when some bugs are involved" category. At the same time we don't really want our app to ever lock up, because due to some bug (from us, or team blue, or system time glitch) there are some bad stats in the system we might not even care too much about. I have to think about it a bit more, maybe we should add some `u63` type, idk. :thinking:

discovery_engine_core/core/src/storage/sqlite.rs
```diff
@@ -757,6 +756,73 @@ impl FeedbackScope for SqliteStorage {
...
         Ok(document)
     }
+
+    async fn update_time_spent(
+        &self,
+        document: document::Id,
+        view_mode: ViewMode,
+        duration: Duration,
+    ) -> Result<TimeSpentDocumentView, Error> {
+        let view_time_ms = u32::try_from(duration.as_millis()).ok().unwrap_or(u32::MAX);
```
janpetschexain, line 766: 🔧 this can be unwrapped directly, no need to convert it to an option first
```suggestion
        let view_time_ms = u32::try_from(duration.as_millis()).unwrap_or(u32::MAX);
```

discovery_engine_core/core/src/storage/sqlite.rs
```diff
@@ -757,6 +756,73 @@ impl FeedbackScope for SqliteStorage {
...
+
+        let mut tx = self.pool.begin().await?;
+
+        sqlx::query(
+            "INSERT INTO ViewTimes(documentId, viewMode, viewTimeMs) VALUES (?, ?, ?)
+                ON CONFLICT DO UPDATE SET viewTimeMs = viewTimeMs + excluded.viewTimeMs;",
```
janpetschexain, line 772: ❓ if `viewTimeMs` already is (close to) an `u32::MAX`, doesn't this overflow/error when adding the `excluded.viewTimeMs`? it seems that we need a minimum here as well
```suggestion
                ON CONFLICT DO UPDATE SET viewTimeMs = min(viewTimeMs + excluded.viewTimeMs, 4294967295);",
```

rustonaut: as the internal representation goes up to `i64` this doesn't really, but adding it doesn't hurt either I guess

janpetschexain (approved):
lgtm

rustonaut:
bors merge

PR #534, opened 2022-08-02
[Merged by Bors] - Fix: Allow release builds 24h a day.

rustonaut:
Due to a bug in the date formatting string release builds could only be done from 10:00:00 to 23:59:59 :shrug: 

If fixed that and also added some more quoting (which isn't needed but better) and used `/g` in `sed` which 
also shouldn't be needed but is a bit more robust.


Robert-Steiner (approved):

rustonaut:
bors merge

PR #535, opened 2022-08-03
[Merged by Bors] - TY-3087 install rust targets via just

janpetschexain:
**References**

- [TY-3087]

**Summary**

- use full target triples for `cargo ndk`
- add rust targets to `just rust-deps`

[TY-3087]: https://xainag.atlassian.net/browse/TY-3087?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.env
+++ b/.env
@@ -12,7 +12,7 @@ FLUTTER_EXAMPLE_WORKSPACE=discovery_engine_flutter/example
 DART_VERSION=2.17.3
 FLUTTER_VERSION=3.0.2
 
-ANDROID_TARGETS="arm64-v8a x86_64 x86"
+ANDROID_TARGETS="aarch64-linux-android x86_64-linux-android i686-linux-android"
 ANDROID_PLATFORM_VERSION="21"
 
 IOS_TARGETS="aarch64-apple-ios x86_64-apple-ios"
--- a/justfile
+++ b/justfile
@@ -46,7 +46,15 @@ flutter-deps:
 
 # Fetches rust dependencies
 rust-deps:
-    cd "$RUST_WORKSPACE"; \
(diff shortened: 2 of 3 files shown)


.env
```diff
@@ -12,7 +12,7 @@ FLUTTER_EXAMPLE_WORKSPACE=discovery_engine_flutter/example
 DART_VERSION=2.17.3
 FLUTTER_VERSION=3.0.2
 
-ANDROID_TARGETS="arm64-v8a x86_64 x86"
+ANDROID_TARGETS="aarch64-linux-android x86_64-linux-android i686-linux-android"
```
janpetschexain, line 15: our github actions also contain the `armv7-linux-androideabi` target. we should either add it here as well or remove it there, what's the current situation?

acrrd: We usually don't want to build for it locally while we want it on the CI.

janpetschexain: ah ok, then i'll leave it as is 👌

justfile
```diff
@@ -46,7 +46,15 @@ flutter-deps:
 
 # Fetches rust dependencies
 rust-deps:
-    cd "$RUST_WORKSPACE"; \
+    #!/usr/bin/env bash
+    set -eux -o pipefail
+    cd "$RUST_WORKSPACE"
+    for TARGET in $ANDROID_TARGETS; do
+        rustup target add $TARGET
+    done
```
acrrd, line 54: :art: We could have a `if os() == "macos"`

acrrd (approved):

janpetschexain:
bors merge

PR #536, opened 2022-08-04
[Merged by Bors] - TY-3052 set dart cfgFeatureStorage based on how rust has been compiled

rustonaut:
This just does what the title says, it doesn't setup the justfile  to allow running tests
with the feature enabled and doesn't fix any of the tests failing with it enabled.

**References:**

- TY-3052

### DIFF
--- a/discovery_engine_core/bindings/src/lib.rs
+++ b/discovery_engine_core/bindings/src/lib.rs
@@ -40,6 +40,12 @@ use std::path::Path;
 use itertools::Itertools;
 use xayn_discovery_engine_core::Engine;
 
+#[allow(unsafe_code)]
+#[no_mangle]
+pub extern "C" fn rust_cfg_storage() -> u8 {
+    u8::from(cfg!(feature = "storage"))
+}
+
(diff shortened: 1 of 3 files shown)


discovery_engine_core/bindings/src/lib.rs
```diff
@@ -40,6 +40,12 @@ use std::path::Path;
 use itertools::Itertools;
 use xayn_discovery_engine_core::Engine;
 
+#[allow(unsafe_code)]
+#[no_mangle]
+pub extern "C" fn rust_cfg_storage() -> u8 {
```
janpetschexain, line 45: :art: what do you think about the following naming? it's implied that this comes from rust via the ffi anyways. then it would be aligned with the naming we already use in dart and rust.
```suggestion
pub extern "C" fn cfg_feature_storage() -> u8 {
```

janpetschexain (approved):
lgtm, just one naming suggestion

acrrd (approved):

rustonaut:
bors merge

PR #539, opened 2022-08-08
[Merged by Bors] - TO-2999 add path for db to config

rustonaut:
This does following:

- expose the application data we already have in the configuration in dart to rust
- use that dir for the db (file name `db.sqlite`)
- add another `use_in_memory_db` option to `InitConfig`
  - this option is for integration tests
  - to reduce complexity this option is always there no matter the compiler feature
  

**References:**

- [TO-2999]

[TO-2999]: https://xainag.atlassian.net/browse/TO-2999?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine_core/bindings/src/types/init_config.rs
+++ b/discovery_engine_core/bindings/src/types/init_config.rs
@@ -224,6 +224,17 @@ pub unsafe extern "C" fn init_config_place_of_data_dir(place: *mut InitConfig) -
     unsafe { addr_of_mut!((*place).data_dir) }
 }
 
+/// Returns a pointer to the `use_in_memory_db` field of a configuration.
+///
+/// # Safety
+///
+/// The pointer must point to a valid [`InitConfig`] memory object,
+/// it might be uninitialized.
+#[no_mangle]
+pub unsafe extern "C" fn init_config_place_of_use_in_memory_db(place: *mut InitConfig) -> *mut u8 {
+    unsafe { addr_of_mut!((*place).use_in_memory_db).cast() }
+}
+
 /// Returns a pointer to the `log_file` field of a configuration.
 ///
 /// # Safety
--- a/discovery_engine_core/bindings/src/types/init_config.rs
+++ b/discovery_engine_core/bindings/src/types/init_config.rs
@@ -213,6 +213,17 @@ pub unsafe extern "C" fn init_config_place_of_de_config(
     unsafe { addr_of_mut!((*place).de_config) }
 }
 
+/// Returns a pointer to the `data_dir` field of a configuration.
+///
+/// # Safety
+///
+/// The pointer must point to a valid [`InitConfig`] memory object,
+/// it might be uninitialized.
+#[no_mangle]
+pub unsafe extern "C" fn init_config_place_of_data_dir(place: *mut InitConfig) -> *mut String {
+    unsafe { addr_of_mut!((*place).data_dir) }
+}
+
 /// Returns a pointer to the `log_file` field of a configuration.
 ///
 /// # Safety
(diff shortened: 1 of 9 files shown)


discovery_engine_core/bindings/src/types/init_config.rs
```diff
@@ -213,6 +213,28 @@ pub unsafe extern "C" fn init_config_place_of_de_config(
...
+/// Returns a pointer to the `use_in_memory_db` field of a configuration.
+///
+/// # Safety
+///
+/// The pointer must point to a valid [`InitConfig`] memory object,
+/// it might be uninitialized.
+#[no_mangle]
+pub unsafe extern "C" fn init_config_place_of_use_in_memory_db(place: *mut InitConfig) -> *mut u8 {
+    unsafe { addr_of_mut!((*place).use_in_memory_db).cast() }
+}
```
rustonaut, line 236: On some platforms dart ffigen doesn't handle `bool` correctly :shrug: 

This has been somewhat improve in newer versions (I'm not sure if they are released yet) but in consideration that it will soon all be removed it makes little sense to spent additional time and trying to fix it again.

discovery_engine_core/core/src/engine.rs
```diff
@@ -294,6 +294,10 @@ impl Engine {
         let search_config = SearchConfig {
             max_docs_per_batch: config.max_docs_per_search_batch as usize,
         };
+        #[cfg(feature = "storage")]
+        let data_dir = std::path::PathBuf::from(&config.data_dir);
+        #[cfg(feature = "storage")]
+        let use_in_memory_db = config.use_in_memory_db;
```
rustonaut, line 300: Config is consumed below.

Generally I think there is a lot of potential to cleanup code around `new`, `from_config` the structure of the context etc. but this is outside of the scope of this PR.

discovery_engine_core/core/src/engine.rs
```diff
@@ -1462,6 +1472,8 @@ pub(crate) mod tests {
                     .unwrap_or(u32::MAX),
                 de_config: None,
                 log_file: None,
+                data_dir: "tmp_test_data_dir".into(),
+                use_in_memory_db: true,
```
rustonaut, line 1476: At some point we probably need to setup some testing with a `TempDir` `data_dir` and proper teardown/setup but that's outside of the scope of this PR but I might need to do it for TY-3061.

discovery_engine_core/core/src/engine.rs
janpetschexain: i agree, it's also better to finish the db migration first and after the dead code is removed, we can then check what's left to simplify.

discovery_engine_core/core/src/config.rs
```diff
@@ -65,6 +65,10 @@ pub struct InitConfig {
     pub de_config: Option<String>,
     /// Log file path.
     pub log_file: Option<String>,
+    /// Directory in which user data should be stored.
+    pub data_dir: String,
+    /// Use a in-memory db instead of a db in the `data_dir`
+    pub use_in_memory_db: bool,
```
janpetschexain, line 71: ❓ wdyt about merging both fields into an `Option<String>` which either implies in-memory db if `None` or a db directory if `Some(data_dir)`?

rustonaut: I had considered this, but in the end this is the generic application data directory which is used for more things then just the database.  Using `Option<String>` would work currently as we only use it from rust for the db. But IMHO it would sooner or later brake with us moving more parts to rust. Which is why I kept it separate.

janpetschexain (approved):

rustonaut:
bors merge

PR #544, opened 2022-08-10
[Merged by Bors] - Change workflow to build with a data.json from workflow input

joecks:


discovery_engine_core/web-api/Dockerfile
```diff
@@ -5,6 +5,5 @@ WORKDIR /service
 COPY ./ ./
 
 RUN chmod +x ./web-api
-VOLUME /service/assets/data.json
```
biernacki, line 8: why remove the volume? you don't have to use it, but it gives additional possibility to change data without rebuilding the container

.github/workflows/web-service.yml
```diff
@@ -75,10 +84,7 @@ jobs:
         run: |
           tar -xvf web-service.tar
           cd ./web-service
-          TAG="branch-$(git rev-parse --abbrev-ref HEAD)"
-          if [[ "${TAG}" == "branch-HEAD" ]]; then
-            TAG="$(git rev-parse --short HEAD)"
-          fi
+          TAG=${{ github.event.inputs.tag }}
```
biernacki, line 87: instead of requiring the tag to be provided, could we just use it when it's provided, and fallback on the branch name?

discovery_engine_core/web-api/Dockerfile
joecks: In fact it was pretty tricky to use the volume in our kubernetis setup - or maybe I and Antonie were too stupid to do it 
Now I am not sure how this behaves when you have both Volume and the file but I can test it.

.github/workflows/web-service.yml
joecks: yes we can sure.

biernacki:
for me the changes are fine, but we can't use an image build with this pipeline until we have postgres instance in place, also I would wait to @jorge-cr-13 to check this too

jorge-cr-13:
From my perspective it looks good.

biernacki (approved):

.github/workflows/web-service.yml
```diff
@@ -2,12 +2,20 @@ name: Web Service Release
 
 on:
   workflow_dispatch:
+    inputs:
+      data_json_url:
+        required: true
+        description: A json file url with the data needed for the service (i.e. pastebin.pl)
+      tag:
+        required: true
```
biernacki, line 10: wait, this is still required, please change that to `false`

biernacki (changes requested):

biernacki (approved):

joecks:
bors merge

.github/workflows/web-service.yml
janpetschexain: ❓ after the latest push this is in there again, does this need to be resolved, @biernacki ?

.github/workflows/web-service.yml
```diff
@@ -2,12 +2,20 @@ name: Web Service Release
...
+        description: a tag that will be used for identifying this version (xaynetci/xayn_discovery_web_service:$TAG)
+
 
 permissions:
   contents: read
 
 env:
-  DENY_WARNINGS: false
+  DENY_WARNINGS: false
```
janpetschexain, line 18: 🔧 pls don't add needless whitespace
```suggestion
  DENY_WARNINGS: false
```

joecks:
bors merge

PR #550, opened 2022-08-14
use single loop with triangle numbers pattern

frankpepermans:
An attempt to improve `semantics::normalized_distance`.

- [x] Swap out `cartesian_product` for `triangular_product` (This avoids overhead loops, as we only want items where `i < j`)
- [x] Use lookup tables for heavy computations, like `matmul` and `exp`
- [x] Use rayon to parallelize the lookup tables
- [x] Refactor calculations to return early when possible

The above results in a net 66% avg. gain on my desktop, results may vary on mobile devices, but theoretically, non-ARMv7 droids should also see significant gain.


discovery_engine_core/core/src/stack/filters/semantic.rs
```diff
@@ -16,40 +16,39 @@ use std::collections::BTreeMap;
...
 /// Computes the condensed cosine similarity matrix of the documents' embeddings.
 fn condensed_cosine_similarity(documents: &[Document]) -> Vec<f32> {
-    pairwise_cosine_similarity(
-        documents
-            .iter()
-            .map(|document| document.smbert_embedding.view()),
-    )
-    .indexed_iter()
-    .filter_map(|((i, j), &similarity)| (i < j).then(|| similarity))
+    triangular_product(documents, |doc_a: &Document, doc_b: &Document, _i, _j| {
```
frankpepermans, line 27: Tried to refactor this, but the test fails, the original code is a bit complicated, would love to better understand it

discovery_engine_core/core/src/stack/filters/semantic.rs
```diff
@@ -16,40 +16,39 @@ use std::collections::BTreeMap;
...
+        let v_a = doc_a.smbert_embedding.view();
+        let v_b = doc_b.smbert_embedding.view();
+        let ni = l2_norm(v_a);
+        let nj = l2_norm(v_b);
+
+        if ni > 0. && nj > 0. {
+            return (true, (v_a.dot(&v_b) / ni / nj).clamp(-1., 1.));
```
frankpepermans, line 34: I assume the Rusty way here would be returning an Option instead?

janpetschexain:
the timeout seems not to be a ci issue, it consistently times out for some of the changed tests, even though they run fine locally. if i had to take a wild guess, i'd say this might be due to the `rayon` parallelism which for some reason starves the ci.

frankpepermans:
@janpetschexain Think it's fixed, added `-- --test-threads=1` for `cargo test` in the just file

acrrd:
I'm working to add benchmark to the code base so we can evaluate this. Unfortunatly for the moment just running them consecutive time leads to +-10% variation in the speed so it will be difficult to have a good understanding if we don't hav better speed up then that.

I would like to split this PR in:
* iterate only on the triangular matrix
* cache exp for days
* reducing allocation
* using multi-thread

For reducing allocation, it seems good if the code does not become too complicated. We can also try to use a custom allocator that can improve the performance here.
For the multi-thread part I'm a bit more sceptical because I'm afraid that will we will starve other requests if this goes on the web api.

frankpepermans:
+1 for benchmarks, we'll need to focus mostly on a server solution, one we can do stress tests, we already know that performance can still be improved, and there we should be able to have threads

frankpepermans:
@acrrd things have changed a bit meanwhile, since we're targeting server only, so we keep this open, or just await benchmarks still?

frankpepermans:
Ok, can always revisit this when the bench is there.

From what I recall, the biggest gains were in the other changes, not so much in the triangle loop change

PR #552, opened 2022-08-22
[Merged by Bors] - TO-3254_change-document-id-to-string

biernacki:
**Summary:**
- added non-empty check for provided article id
- added better non-empty check for user id

**References:**
https://xainag.atlassian.net/browse/TO-3254

### DIFF
--- a/discovery_engine_core/web-api/src/db.rs
+++ b/discovery_engine_core/web-api/src/db.rs
@@ -17,10 +17,9 @@ use std::{collections::HashMap, fs::File, path::PathBuf, sync::Arc};
 use tokio::sync::RwLock;
 use xayn_discovery_engine_ai::{CoiSystem, CoiSystemConfig, CoiSystemState};
 use xayn_discovery_engine_bert::{AveragePooler, SMBert, SMBertConfig};
-use xayn_discovery_engine_core::document::Id;
 use xayn_discovery_engine_tokenizer::{AccentChars, CaseChars};
 
-use crate::models::{Article, Document, UserId};
+use crate::models::{Article, Document, ProviderId, UserId};
 
 pub(crate) type Db = Arc<AppState>;
 
@@ -28,16 +27,16 @@ pub(crate) type Db = Arc<AppState>;
 pub(crate) struct AppState {
     pub(crate) smbert: SMBert,
     pub(crate) coi: CoiSystem,
-    pub(crate) documents_by_id: HashMap<Id, Document>,
+    pub(crate) documents_by_provider_id: HashMap<ProviderId, Document>,
     pub(crate) documents: Vec<Document>,
     pub(crate) user_interests: RwLock<HashMap<UserId, CoiSystemState>>,
 }
 
 impl AppState {
-    fn new(documents_by_id: HashMap<Id, Document>, smbert: SMBert) -> Self {
-        let documents = documents_by_id.clone().into_values().collect();
+    fn new(documents_by_provider_id: HashMap<ProviderId, Document>, smbert: SMBert) -> Self {
+        let documents = documents_by_provider_id.clone().into_values().collect();
         Self {
-            documents_by_id,
+            documents_by_provider_id,
             documents,
             smbert,
             coi: CoiSystemConfig::default().build(),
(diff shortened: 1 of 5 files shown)


discovery_engine_core/web-api/src/db.rs
```diff
@@ -69,21 +68,27 @@ pub(crate) fn init_db(config: &InitConfig) -> Result<Db, Box<dyn std::error::Err
...
                 .as_str()
-                .expect("The 'id' field needs to be represented as String")
-                .try_into()
-                .expect("The 'id' field needs to be a valid UUID");
+                .expect("The article's 'id' field needs to be represented as String")
+                .trim()
+                .to_string();
+
+            assert!(
+                !article_id.is_empty(),
```
Robert-Steiner, line 80: 🔧  The Stroy description also notes that the string must not contain zero bytes. We could check it with `article_id.contains('\u{0000}')`
same fro the UserID

discovery_engine_core/web-api/src/models.rs
```diff
@@ -79,27 +83,30 @@ impl From<Document> for Article {
...
 /// Unique identifier for the user.
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash, Display)]
 pub(crate) struct UserId(String);
 
 impl UserId {
-    fn new(user_id: &str) -> Result<Self, Error> {
-        if user_id.is_empty() {
-            Err(Error::EmptyUserId)
+    fn new(value: &str) -> Result<Self, Error> {
+        let value = value.trim();
```
Robert-Steiner, line 95: I just tested it and it looks like that the string that wrap returns is url encoded :(
that means `trim` won't help here 

e.g.
```
curl --request POST \
  --url http://0.0.0.0:3000/user/%20/interaction \
  --header 'Content-Type: application/json' \
  --data '{
  "document_id": "0251ae9f73ec12f4d3eced9c4dc9ccc8"
}'
```
is a valid request with the user_id `%20`

discovery_engine_core/web-api/src/models.rs
```diff
@@ -79,27 +83,30 @@ impl From<Document> for Article {
 /// Represents user interaction request body.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub(crate) struct InteractionRequestBody {
-    pub(crate) document_id: Id,
+    pub(crate) document_id: String,
```
Robert-Steiner, line 86: ❔ Maybe we should also validating the `document_id` in the `handle_user_interaction` function. wdyt?

biernacki: it is used to lookup a value from `documents_by_id` HashMap, so it will return 404 if it won't find a value under provided key. Since we are validating article id when constructing this HashMap I would assume we don't have to add additional validation, but maybe I'm missing something?

discovery_engine_core/web-api/src/models.rs
biernacki: ah, good catch! I've added this to the validations, could you check again?

discovery_engine_core/web-api/src/models.rs
Robert-Steiner: Makes sense 👍 then we don't need to check it

discovery_engine_core/web-api/src/models.rs
Robert-Steiner: That works but now something like `0.0.0.0:3000/user/E%20W` is not valid anymore.

what about decoding first with `urlencoding::decode` and after that check if it the value is empty or contains a zero byte  

```rust 
        let value = urlencoding::decode(value).map_err(Error::Utf8Error)?;

        let value = value.trim();
        if value.is_empty() {
            Err(Error::UserIdEmpty)
        } else if value.contains('\u{0000}') {
            Err(Error::UserIdContainsZeroByte)
        } else {
            Ok(Self(value.to_string()))
        }
```

biernacki: yup, you're right :) I've changed it per your recommendation

Robert-Steiner (approved):

biernacki:
bors merge

PR #553, opened 2022-08-22
[Merged by Bors] - TO-3259 fix dart/jfrog problems

rustonaut:
This does a number of small changes:

- switch from `dart.yellow.private` to `dart.xayn.private`
- change versioning schema to `0.1.0+<branch-name>.<timestamp>`
  - we use the timestamp instead of the git hash as this creates a proper sorting of the packages by "branch,version"
  - note that this relies on non-documented standard conflicting implementation details of JFrog/dart pub, while it will probably not brake it could, especially if you depend on a version which had been deleted in the repo :sob: 
  - but it's for now the best we can do
 - slightly changes how the justfile does the CI publishing (to allow easier debugging locally and makes sure the same timestamp is used across all packages)

**References:**
 
 - TO-3259

### DIFF
--- a/.github/actions/setup-job-docker/action.yml
+++ b/.github/actions/setup-job-docker/action.yml
@@ -50,7 +50,7 @@ runs:
           echo "JFROG access token missing"
           exit 1
         fi
-        echo "${JFROG_TOKEN}" | dart pub token add "https://xayn.jfrog.io/artifactory/api/pub/dart.yellow.private"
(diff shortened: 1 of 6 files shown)


rustonaut:
Works: https://github.com/xaynetwork/xayn_discovery_engine/actions/runs/2903743433

biernacki (approved):
lgtm!

rustonaut:
bors merge

PR #555, opened 2022-08-23
[Merged by Bors] - TO-3061 allow integration tests with storage feature enabled

rustonaut:
Some tests are skipped (if the storage feature is enabled) they fall roughly into 3 categories:
- tests which are tightly coupled to hive, but where we need similar (potentially storage generic tests) in dart
- tests which have some hive coupling but it can be fixed (anything using `load/saveEngineState`)
- tests which are broken with storage enabled due to a bug

All of them will be fixed in follow up PRs. Through I did fix some smaller bugs directly in this PR.

The `justfile` is changed to allow running tests with storage enabled, I went a bit back and force between different approaches and decided to do so by introducing a `XAYN_DE_FEATURES` environment variable. This means in difference to before running test/check will _either_ run it all without the storage feature enabled or all with the feature enabled. In CI rust-test/check is run once with the storage feature and once without. While this isn't perfect other approaches I tried turned out much worse (worse usability, more complex code, hardly any benefits).

**References:**

- [TO-3061]

[TO-3061]: https://xainag.atlassian.net/browse/TO-3061?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine_core/core/src/engine.rs
+++ b/discovery_engine_core/core/src/engine.rs
@@ -764,9 +764,7 @@ impl Engine {
         );
         if let UserReaction::Positive = reaction {
             #[cfg(feature = "storage")]
-            let history = Some(self.storage.fetch_history().await?);
-            #[cfg(feature = "storage")]
-            let history = &history;
+            let history = &self.storage.fetch_history().await?.into();
             if let Some(history) = history {
                 update_stacks(
                     &mut stacks,
--- a/discovery_engine_core/core/src/engine.rs
+++ b/discovery_engine_core/core/src/engine.rs
@@ -763,6 +763,10 @@ impl Engine {
             &self.state.user_interests,
         );
         if let UserReaction::Positive = reaction {
+            #[cfg(feature = "storage")]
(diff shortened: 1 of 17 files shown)


discovery_engine_core/core/src/engine.rs
```diff
@@ -763,6 +763,10 @@ impl Engine {
             &self.state.user_interests,
         );
         if let UserReaction::Positive = reaction {
+            #[cfg(feature = "storage")]
+            let history = Some(self.storage.fetch_history().await?);
+            #[cfg(feature = "storage")]
+            let history = &history;
```
janpetschexain, line 769: :art: this could be condensed into one line
```suggestion
            #[cfg(feature = "storage")]
            let history = &self.storage.fetch_history().await?.into();
```

janpetschexain (approved):
lgtm

rustonaut:
bors merge

rustonaut:
bors merge

PR #560, opened 2022-08-23
[Merged by Bors] - TO-3242 integration tests for feed

rustonaut:
**Based on:**

- #556

PR #563, opened 2022-08-25
[Merged by Bors] - TO-3243 integration tests search

rustonaut:
**Based On:**

- #560


discovery_engine/test/feed_manager_test.dart
```diff
@@ -60,151 +60,148 @@ Future<void> main() async {
...
+      expect(feed[1].documentId, equals(doc2.documentId));
+      // doc0 has the later timestamp
+      expect(
+        feed[2].documentId,
+        equals(engine.feedDocuments[0].document.documentId),
+      );
+      // doc3 is excluded since it is inactive
+    });
+    // ignore: require_trailing_commas
+  }, skip: cfgFeatureStorage);
```
rustonaut, line 206: The only real change in this file is the two lines above, the rest is formatting.

Originally I didn't add the `// ignore: require_trailing_commas`  but this means any time you add a skipp to a group the whole group is indented adding a lot of changes to PRs.

So this was changed to using  `// ignore: require_trailing_commas` in such situations, but one this place where I didn't use it was already merged.

discovery_engine_core/core/src/storage/sqlite.rs
```diff
@@ -220,6 +220,28 @@ impl SqliteStorage {
...
+                stm.push_bind(&doc.id);
+            })
+            .build()
+            .persistent(false)
+            .execute(tx)
+            .await?;
+
+        Ok(())
+    }
+
```
rustonaut, line 244: Code was de-duplicated the SQL is the same but the `documents.is_empty()` check was added.

discovery_engine_core/core/src/storage/sqlite.rs
```diff
@@ -220,6 +220,28 @@ impl SqliteStorage {
...
+        documents: &[NewDocument],
+    ) -> Result<(), Error> {
+        if documents.is_empty() {
+            return Ok(());
+        }
+
+        SqliteStorage::store_new_documents(tx, documents).await?;
+
+        QueryBuilder::new("INSERT INTO SearchDocument (documentId) ")
+            .push_values(documents, |mut stm, doc| {
```
finiteprods, line 234: :grey_question: `stm` appears a lot in code like this in this file. what does it actually stand for?

discovery_engine/test/integration/search_test.dart
```diff
@@ -0,0 +1,240 @@
...
+    });
+
+    tearDown(() async {
+      await engine.dispose();
+      await server.close();
+      await Directory(data.applicationDirectoryPath).delete(recursive: true);
+    });
+
+    test('a new search can be opened', () async {
+      final result = expectEvent<ActiveSearchRequestSucceeded>(
```
finiteprods, line 62: :art: perhaps worth checking the latest re naming of "active search" - I seem to recall people in favour of dropping the "active". but this would prob be a separate pr

discovery_engine/test/integration/search_test.dart
```diff
@@ -0,0 +1,240 @@
...
+
+      expect(
+        expectEvent<ActiveSearchRequestFailed>(
+          await engine.requestQuerySearch(searchTerm2),
+        ).reason,
+        equals(SearchFailureReason.openActiveSearch),
+      );
+    });
+
+    test('we can request more results until we ran out of results', () async {
```
finiteprods, line 91:
```suggestion
    test('we can request more results until we run out of results', () async {
```

discovery_engine/test/integration/search_test.dart
```diff
@@ -0,0 +1,240 @@
...
+
+      expect(result.items, isNotEmpty);
+      expect(result.search.searchBy, SearchBy.query);
+      expect(result.search.searchTerm, searchTerm);
+
+      final nextResult = expectEvent<NextActiveSearchBatchRequestSucceeded>(
+        await engine.requestNextActiveSearchBatch(),
+      );
+
+      expect(nextResult.items, isNot(equals(result.items)));
```
finiteprods, line 104: :art: is the `equals` (here and similarly in other places) necessary?

discovery_engine/test/integration/search_test.dart
```diff
@@ -0,0 +1,240 @@
...
+        ).reason,
+        SearchFailureReason.noActiveSearch,
+      );
+
+      expect(
+        expectEvent<ActiveSearchClosedFailed>(
+          await engine.closeActiveSearch(),
+        ).reason,
+        SearchFailureReason.noActiveSearch,
+      );
```
finiteprods, line 149: :thought_balloon: I guess returning `ActiveSearchClosedSucceeded` could also arguably be valid in this case... either way, it is good this case is tested.

finiteprods (commented):
looks very good, mostly minor comments and a few questions.

discovery_engine/test/integration/search_test.dart
rustonaut: This is why the file is  `search_test.dart`. Through `ActiveSearchRequestSucceeded` is auto generated by freezed and part of the team blue public API. We probably won't rename it until switching to the message passing API. But even if we rename it earlier it has nothing to do with this PR.

discovery_engine/test/integration/utils/local_newsapi_server.dart
```diff
@@ -101,7 +109,7 @@ class LocalNewsApiServer {
 
   set replyWith(ReplyWith flag) => _replyWith = flag;
 
-  set snFile(String filename) => _snFile = filename;
+  set snFile(String filename) => _snFiles = [filename];
```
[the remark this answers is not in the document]
rustonaut: It's the setter for `snFile` not a setter for `snFiles` I didn't change it to keep compatibility with any tests setting a single file.

discovery_engine/test/integration/utils/local_newsapi_server.dart
```diff
@@ -42,7 +42,12 @@ enum ReplyWith {
 class LocalNewsApiServer {
   final HttpServer _server;
   ReplyWith _replyWith = ReplyWith.data;
-  String _snFile = 'climate-change.json';
+  int _snCounter = 0;
+  List<String> _snFiles = [
+    'climate-change.json',
+    'climate-change2.json',
+    'empty.json'
+  ];
```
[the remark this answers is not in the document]
rustonaut: `sn` stands for search news and is badly named but as it's an ad-hock test utility no one bothered to rename it and then go through all the tests using it and rename them, too.

As a side note ` 'climate-change.json'` is also used by latest news in some places as far as I remember so renaming it to `search-news.json` might not be the best idea. Generally we probably should rework the testing infrastructure at some point (:pushpin: ) but we probably want to wait until after switching to the message passing API and then do it in rust.

discovery_engine/test/integration/search_test.dart
rustonaut: I had proposed that in the past. But it was decided that it should fail in this case (and similar cases).

discovery_engine/test/integration/search_test.dart
rustonaut: it is not needed, as they did overload `isNot` :rocket:

discovery_engine/test/integration/search_test.dart
```diff
@@ -0,0 +1,240 @@
...
+          await engine.getActiveSearchTerm(),
+        ).searchTerm,
+        equals(searchTerm),
+      );
+
+      final nextResult = expectEvent<NextActiveSearchBatchRequestSucceeded>(
+        await engine.requestNextActiveSearchBatch(),
+      );
+
+      expect(nextResult.items, isNot(equals(result.items)));
```
finiteprods, line 210:
```suggestion
      expect(nextResult.items, isNot(result.items));
```

finiteprods (approved):
lgtm

discovery_engine/test/integration/utils/local_newsapi_server.dart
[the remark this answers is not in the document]
finiteprods: cool, thanks for renaming and agree with the general comment re postponing a more comprehensive overhaul till after message passing api in place

rustonaut:
bors merge

PR #441, opened 2022-06-09
[Merged by Bors] - TY-2881 update reacted sources repo on reaction

finiteprods:
**Summary**

- previously #432

adds logic to the handling of user reaction events to update the reacted sources repo where appropriate

PR #564, opened 2022-08-25
TO-3074 rust database logic for source reactions

finiteprods:


discovery_engine_core/core/src/storage.rs
```diff
@@ -151,6 +155,17 @@ pub(crate) trait SourcePreferenceScope {
     async fn fetch_excluded(&self) -> Result<HashSet<String>, Error>;
 }
 
+#[async_trait]
+pub(crate) trait SourceReactionScope {
+    async fn fetch_source_reaction(&self, source: &str) -> Result<Option<bool>, Error>;
```
biernacki, line 160: ❔ why not fetch whole source with `weight` and `liked`? we don't plan to use it anywhere?

discovery_engine_core/core/src/storage/sqlite.rs
```diff
@@ -855,6 +860,14 @@ impl FeedbackScope for SqliteStorage {
 
         Ok(view)
     }
+
+    async fn update_source_reaction(&self, source: &str, liked: bool) -> Result<(), Error> {
+        match self.fetch_source_reaction(source).await? {
+            None => self.store_source_reaction(source, liked).await,
+            Some(reaction) if reaction == liked => self.update_source_weight(source).await,
+            _ => self.delete_source_reaction(source).await,
```
biernacki, line 868: ❔ why are we deleting the source info if the `reaction` is different then `liked` param? how can we then change it from like to dislike?

discovery_engine_core/core/src/storage/sqlite.rs
```diff
@@ -918,6 +931,85 @@ impl SourcePreferenceScope for SqliteStorage {
...
+
+    async fn update_source_weight(&self, source: &str) -> Result<(), Error> {
+        let last_updated = Utc::now().naive_utc();
+
+        let mut tx = self.pool.begin().await?;
+
+        sqlx::query(
+            "UPDATE SourceReaction
+            SET lastUpdated = ?,
+                weight = CASE WHEN liked = 1 THEN (weight + 1) END
```
biernacki, line 989: so we only increment when user likes and do nothing to the `weight` in case of a dislike?

discovery_engine_core/core/src/storage/sqlite.rs
```diff
@@ -918,6 +931,85 @@ impl SourcePreferenceScope for SqliteStorage {
...
+        .fetch_optional(&mut tx)
+        .await?;
+
+        tx.commit().await?;
+
+        Ok(reaction.map(|r| r.liked))
+    }
+
+    async fn store_source_reaction(&self, source: &str, liked: bool) -> Result<(), Error> {
+        let weight = if liked { 1 } else { -1 };
```
biernacki, line 963: what if we call `store_source_reaction` multiple times for the same source? we always reset the weight?

discovery_engine_core/core/src/storage.rs
finiteprods: there will probably be a "fetch all" or "fetch best source" in a later PR so the `weight` in particular will be used then!

discovery_engine_core/core/src/storage/sqlite.rs
finiteprods: right, it's not so intuitive but that was the heuristic originally implemented in the python implementation, see [this comment](https://github.com/xaynetwork/xayn_discovery_engine/pull/441#discussion_r895509780). to change to a dislike, the user would then have to dislike the source again.

discovery_engine_core/core/src/storage/sqlite.rs
finiteprods: right, `weight` is incremented in the case of liked, untouched in the case of disliked.

discovery_engine_core/core/src/storage/sqlite.rs
biernacki: ah, you're right on both :) it's a bit counterintuitive and also how it's implemented currently

discovery_engine_core/core/src/storage/sqlite.rs
biernacki: right now it's not possible because first we check if we have a source with `fetch_source_reaction` and then later call `store_source_reaction` if the former returns `None`, but then maybe we should rename it to `create_source_reaction` (or add a line of documentation) to indicate that it should be only used for that and not for persisting source reaction (so creating and/or updating), wdyt?

finiteprods: I agree the name was confusing, now changed as suggested! if you did call this multiple times for the same source, it should be a sql error (will check this when tests added).

biernacki: ah, true, because it's an `INSERT` operation, sorry, I'm not so fluent in SQL

biernacki (approved):
👍

finiteprods:
bors merge

PR #567, opened 2022-08-25
[Merged by Bors] - Don't retry on error when getting more articles

acrrd:
We already have retry policy in the provider crate.
This function should only handle the logic to ask for more articles in case we don't have enough.


janpetschexain (approved):

rustonaut (approved):
lgtm

PR #570, opened 2022-08-29
[Merged by Bors] - TO-3278 split coi & kps configs

janpetschexain:
**References**

- [TO-3278]
- followed by #571

**Summary**

- split the `coi::config::Config as CoiSystemConfig` into `coi::config::Config as CoiConfig` & `kps::config::Config as KpsConfig` (kps for key phrase selection instead of kpe for key phrase extraction, because the latter is actually a subsystem/presystem of the former)
- use the splitted configs in the `CoiSystem` & the `Engine`
- update the [confluence config docs](https://xainag.atlassian.net/wiki/spaces/M2D/pages/2476605445/Discovery+Engine+Configurations)

[TO-3278]: https://xainag.atlassian.net/browse/TO-3278?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine_core/ai/ai/src/kps/config.rs
+++ b/discovery_engine_core/ai/ai/src/kps/config.rs
@@ -1,0 +1,119 @@
+// Copyright 2022 Xayn AG
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, version 3.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+use std::{cmp::Ordering, time::Duration};
+
+use displaydoc::Display;
+use serde::{Deserialize, Serialize};
+use thiserror::Error;
+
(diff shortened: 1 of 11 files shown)


discovery_engine_core/ai/ai/src/kps/config.rs
```diff
@@ -0,0 +1,119 @@
...
+use thiserror::Error;
+
+use crate::utils::{nan_safe_f32_cmp_desc, serde_duration_as_days, SECONDS_PER_DAY_U64};
+
+/// Configurations of the kps system.
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+#[must_use]
+pub struct Config {
+    #[serde(with = "serde_duration_as_days")]
+    horizon: Duration,
```
biernacki, line 28: ❓ Config for Coi also has this `horizon` setting, won't it get confusing which one to use? Or maybe we need to have this configured separately?

biernacki (commented):
looks good 👍 I just have one question to resolve

discovery_engine_core/ai/ai/src/kps/config.rs
janpetschexain: good point, the `horizon` is used in the `coi::context` as well as `kps::key_phrase`. i removed it from the `KpsConfig` such that there is a single source of truth in the `CoiConfig`. as the `KpsSystem` requires the `CoiSystem` anyways, it can also access its config and doesn't need to duplicate the value.

biernacki (approved):
lgtm 👍

janpetschexain:
bors merge

PR #571, opened 2022-08-29
TO-3284 split coi & kps states

janpetschexain:
**References**

- [TO-3284]
- requires #570
- followed by #572

**Summary**

- move `Engine` state serde logic from `core::engine` & `ai::coi::state` to `core::state`
- increment `STATE_VERSION` & adapt state serde to splitted states & add deserialization for previous state version
- use splitted states in the `Engine` & `web-api`

[TO-3284]: https://xainag.atlassian.net/browse/TO-3284?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


discovery_engine_core/core/src/state.rs
```diff
@@ -0,0 +1,110 @@
...
+                bincode::deserialize(&bytes[1..]).map_err(Error::Deserialization)
+            }
+            Some(version) => Err(GenericError::from(format!(
+                "Unsupported serialized data. Found version {} expected {}",
+                *version, STATE_VERSION,
+            ))
+            .into()),
+            None => Err(GenericError::from("Empty serialized data").into()),
+        }
+        // Serialized data could be the partially unversioned data we had before
```
finiteprods, line 72: :grey_question: can you clarify what you mean by the _partially_ unversioned data?

the case where it's a `SerializedState` but where the containing `SerializedCoiSystemState` possibly is versioned?

discovery_engine_core/core/src/state.rs
```diff
@@ -0,0 +1,110 @@
...
+
+                            bincode::deserialize::<CoiSystemState>(&state.coi.0[1..])
+                                .map(|coi| (stacks, coi.user_interests, coi.key_phrases))
+                        }
+                        // Serialized data could be the unversioned data we had before
+                        _ => bincode::deserialize::<UserInterests>(bytes)
+                            .map(|user_interests| (stacks, user_interests, KeyPhrases::default())),
+                    }
+                })
+                .map_err(|_| error)
```
finiteprods, line 107: :thought_balloon: just wondering if it is worth refining the error for some cases e.g. where the `stacks` part was deserialized but not the `coi` part, rather than just returning the error from the level up. perhaps not worth the trouble, just a thought.

finiteprods (dismissed):
lgtm

discovery_engine_core/core/src/state.rs
janpetschexain: exactly, in the current `main` only the `CoiSystemState` is versioned, but the `StackState` is not.

discovery_engine_core/core/src/state.rs
janpetschexain: i'd rather skip this. the code does as much as the current `main` provides. also we will split up state serialization soon anyways, once we persist state more granular in the db.

janpetschexain:
bors merge

PR #575, opened 2022-08-30
[Merged by Bors] - TO-3249 fix update stacks error counting

janpetschexain:
**References**

- [TO-3249]

**Summary**

- fix error counting for needy stacks in case of multiple markets
- fix error counting for updated stacks including the exploration stack

[TO-3249]: https://xainag.atlassian.net/browse/TO-3249?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine_core/core/src/engine.rs
+++ b/discovery_engine_core/core/src/engine.rs
@@ -1539,8 +1539,11 @@ async fn update_stacks(
         };
     }
 
-    // return early if all needy stacks failed for all markets
-    if all_documents.is_empty() && errors.len() >= needy_stacks.len() {
+    // return early if all needy-ready stacks failed for all markets
+    if all_documents.is_empty()
+        && !errors.is_empty()
+        && errors.len() >= needy_stacks.len() - not_ready_stacks.len()
+    {
         return Err(Error::Errors(errors.into_values().flatten().collect()));
     }
     errors.clear();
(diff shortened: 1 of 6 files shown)


discovery_engine_core/core/src/engine.rs
```diff
@@ -1615,9 +1623,9 @@ async fn update_stacks<'a>(
         }
     }
 
-    // only return an error if all stacks that were ready to get new items failed
-    if !errors.is_empty() && errors.len() >= ready_stacks {
-        Err(Error::Errors(errors))
+    // only return an error if all stacks (including exploration) failed
+    if errors.len() > stacks.len() {
```
rustonaut, line 1627: :art:??

This line is quite easy to misunderstand/confuse.

I'm wondering if following would be better but I'm really not sure about it.

```suggestion
    if errors.len() >= stacks.len() + 1 {
```

rustonaut (approved):
lgtm, maybe one minor stile change, but I'm not sure about it

discovery_engine_core/core/src/engine.rs
janpetschexain: i implemented it before like you suggested, but then clippy complained. i'll change it back and ignore clippy for this one.

janpetschexain:
bors merge

PR #577, opened 2022-08-31
[Merged by Bors] - TO-3110 [1] rework storage init

rustonaut:
- move sqlite initialization code into `sqlite/setup.rs`

- Collapsed create storage and init db into one
(we can't use the storage without having called
init db).

- We now path an optional file path instead of an
  uri to the storage system, none is used to indicate
  a in memory (or otherwise temporary) db should
  be used.

- If we can't open the database file or can't apply
  migrations on an existing database file we overwrite
  it with a new db.

- When creating the storage instance we return an
  indicator weather or not it's on a new db, a
  existing db or an overwritten db.
  
- Rename `use_in_memory_db` to `use_ephemeral_db` as 
  ephemeral are not necessary in memory for all kinds of potential
  storage implementations.
  
**Based On:**

- #573

### DIFF
--- a/discovery_engine/lib/src/ffi/types/init_config.dart
+++ b/discovery_engine/lib/src/ffi/types/init_config.dart
@@ -54,7 +54,7 @@ class InitConfigFfi with EquatableMixin {
   final String? deConfig;
   final String? logFile;
   final String dataDir;
-  final bool useInEphemeralDb;
+  final bool useEphemeralDb;
 
   @override
   List<Object?> get props => [
@@ -76,7 +76,7 @@ class InitConfigFfi with EquatableMixin {
         deConfig,
         logFile,
         dataDir,
-        useInEphemeralDb,
+        useEphemeralDb,
       ];
 
   factory InitConfigFfi(
@@ -105,7 +105,7 @@ class InitConfigFfi with EquatableMixin {
         deConfig: deConfig,
         logFile: configuration.logFile,
         dataDir: configuration.applicationDirectoryPath,
-        useInEphemeralDb: configuration.useInEphemeralDb,
+        useEphemeralDb: configuration.useEphemeralDb,
       );
 
   InitConfigFfi.fromParts({
@@ -125,7 +125,7 @@ class InitConfigFfi with EquatableMixin {
     required this.maxDocsPerFeedBatch,
     required this.maxDocsPerSearchBatch,
     required this.dataDir,
-    required this.useInEphemeralDb,
+    required this.useEphemeralDb,
     this.deConfig,
     this.logFile,
   });
@@ -163,7 +163,7 @@ class InitConfigFfi with EquatableMixin {
     dataDir.writeNative(
       ffi.init_config_place_of_data_dir(place),
     );
-    useInEphemeralDb
+    useEphemeralDb
         .writeNative(ffi.init_config_place_of_use_ephemeral_db(place));
   }
 
@@ -213,7 +213,7 @@ class InitConfigFfi with EquatableMixin {
       dataDir: StringFfi.readNative(
         ffi.init_config_place_of_data_dir(config),
       ),
-      useInEphemeralDb: BoolFfi.readNative(
+      useEphemeralDb: BoolFfi.readNative(
         ffi.init_config_place_of_use_ephemeral_db(config),
       ),
     );
--- a/discovery_engine/lib/src/ffi/types/init_config.dart
+++ b/discovery_engine/lib/src/ffi/types/init_config.dart
@@ -54,7 +54,7 @@ class InitConfigFfi with EquatableMixin {
   final String? deConfig;
   final String? logFile;
   final String dataDir;
-  final bool useInMemoryDb;
+  final bool useInEphemeralDb;
 
   @override
   List<Object?> get props => [
@@ -76,7 +76,7 @@ class InitConfigFfi with EquatableMixin {
         deConfig,
         logFile,
         dataDir,
-        useInMemoryDb,
+        useInEphemeralDb,
       ];
 
   factory InitConfigFfi(
@@ -105,7 +105,7 @@ class InitConfigFfi with EquatableMixin {
         deConfig: deConfig,
         logFile: configuration.logFile,
         dataDir: configuration.applicationDirectoryPath,
-        useInMemoryDb: configuration.useInMemoryDb,
+        useInEphemeralDb: configuration.useInEphemeralDb,
       );
 
   InitConfigFfi.fromParts({
@@ -125,7 +125,7 @@ class InitConfigFfi with EquatableMixin {
     required this.maxDocsPerFeedBatch,
     required this.maxDocsPerSearchBatch,
     required this.dataDir,
-    required this.useInMemoryDb,
+    required this.useInEphemeralDb,
     this.deConfig,
     this.logFile,
   });
@@ -163,7 +163,8 @@ class InitConfigFfi with EquatableMixin {
     dataDir.writeNative(
       ffi.init_config_place_of_data_dir(place),
     );
-    useInMemoryDb.writeNative(ffi.init_config_place_of_use_ephemeral_db(place));
+    useInEphemeralDb
+        .writeNative(ffi.init_config_place_of_use_ephemeral_db(place));
   }
 
   @visibleForTesting
@@ -212,7 +213,7 @@ class InitConfigFfi with EquatableMixin {
       dataDir: StringFfi.readNative(
         ffi.init_config_place_of_data_dir(config),
       ),
-      useInMemoryDb: BoolFfi.readNative(
+      useInEphemeralDb: BoolFfi.readNative(
         ffi.init_config_place_of_use_ephemeral_db(config),
       ),
     );
--- a/discovery_engine/lib/src/ffi/types/init_config.dart
+++ b/discovery_engine/lib/src/ffi/types/init_config.dart
@@ -163,7 +163,7 @@ class InitConfigFfi with EquatableMixin {
     dataDir.writeNative(
       ffi.init_config_place_of_data_dir(place),
     );
-    useInMemoryDb.writeNative(ffi.init_config_place_of_use_in_memory_db(place));
+    useInMemoryDb.writeNative(ffi.init_config_place_of_use_ephemeral_db(place));
   }
 
   @visibleForTesting
@@ -213,7 +213,7 @@ class InitConfigFfi with EquatableMixin {
         ffi.init_config_place_of_data_dir(config),
       ),
       useInMemoryDb: BoolFfi.readNative(
-        ffi.init_config_place_of_use_in_memory_db(config),
+        ffi.init_config_place_of_use_ephemeral_db(config),
       ),
     );
   }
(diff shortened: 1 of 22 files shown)


discovery_engine_core/core/src/storage/sqlite.rs
```diff
@@ -1168,49 +1126,6 @@ mod tests {
...
-            stack::ops::trusted::TrustedNews::id(),
-            stack::exploration::Stack::id(),
-        ];
-
-        assert_eq!(
-            ids.into_iter().collect::<HashSet<_>>(),
-            expected_ids.into_iter().collect::<HashSet<_>>()
-        );
-    }
-
```
rustonaut, line 1213: Moved to sqlite/setup.rs

discovery_engine_core/core/src/utils.rs
```diff
@@ -0,0 +1,110 @@
...
+
+    fn extract_error(self, push_to: &mut Vec<E>) -> Option<T> {
+        match self {
+            Ok(val) => Some(val),
+            Err(err) => {
+                push_to.push(err);
+                None
+            }
+        }
+    }
```
rustonaut, line 75: I cycled through using `log_error`, `extract_first_error` and `extract_error` when writing this PR, in the end I ended up only using `extract_error` but I thing the other ones are helpful, too so I kept them. (e.g. `foo().log_error().ok()` for ignoring but logging errors is pretty neat).

discovery_engine_core/core/src/utils.rs
```diff
@@ -0,0 +1,110 @@
...
+pub(crate) trait MiscErrorExt<T, E> {
+    // Logs the contained error if there is one.
+    fn log_error(self) -> Self
+    where
+        E: Display;
+
+    /// Like [`std::result::Result::ok()`] but if `slot` is `None` the error is moved into `slot`.
+    fn extract_first_error(self, slot: &mut Option<E>) -> Option<T>;
+
+    /// Like [`std::result::Result::ok()`] the error is moved into the passed in vec (using `push`).
```
janpetschexain, line 40: :art: typo
```suggestion
    /// Like [`std::result::Result::ok()`] and the error is moved into the passed in vec (using `push`).
```

discovery_engine/lib/src/ffi/types/init_config.dart
```diff
@@ -163,7 +163,7 @@ class InitConfigFfi with EquatableMixin {
     dataDir.writeNative(
       ffi.init_config_place_of_data_dir(place),
     );
-    useInMemoryDb.writeNative(ffi.init_config_place_of_use_in_memory_db(place));
+    useInMemoryDb.writeNative(ffi.init_config_place_of_use_ephemeral_db(place));
```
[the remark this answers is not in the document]
rustonaut: :rocket: 

wait this is already updated in rust, how :thinking: :shrug:

discovery_engine_core/core/src/storage/sqlite.rs
```diff
@@ -351,16 +314,15 @@ impl SqliteStorage {
...
 impl Storage for SqliteStorage {
-    async fn init_database(&self) -> Result<(), Error> {
-        sqlx::migrate!("src/storage/migrations")
-            .run(&self.pool)
+    async fn init_storage_system(
+        file_path: Option<String>,
+    ) -> Result<(BoxedStorage, InitDbHint), Error>
+    where
+        Self: Sized,
+    {
```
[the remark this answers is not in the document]
rustonaut: :rocket: just copied over from the `Storage` trait it implements, where it is needed to allow trait-objects.

discovery_engine_core/core/src/storage/sqlite/setup.rs
```diff
@@ -0,0 +1,331 @@
...
+        .execute(&mut tx)
+        .await?;
+    query_builder
+        .reset()
+        .push("DELETE FROM Stack WHERE stackId NOT IN ")
+        .push_tuple(expected_ids)
+        .build()
+        .persistent(false)
+        .execute(&mut tx)
+        .await?;
```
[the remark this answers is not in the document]
rustonaut: :rocket:  If the array would be big enough then depending on how rustc/llvm would handle it it could make a relevant difference. But yes, here it's at best a pre-mature optimization probably not even that.

discovery_engine/lib/src/ffi/types/init_config.dart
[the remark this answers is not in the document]
biernacki: `useInMemoryDb` => `useEphemeralDb`

discovery_engine_core/core/src/config.rs
```diff
@@ -100,7 +100,7 @@ pub struct InitConfig {
     /// Directory in which user data should be stored.
     pub data_dir: String,
     /// Use a in-memory db instead of a db in the `data_dir`
```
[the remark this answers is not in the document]
biernacki: `Use a in-memory db instead.. -> Use an ephemeral db instead..`

discovery_engine_core/core/src/engine.rs
```diff
@@ -321,15 +321,15 @@ impl Engine {
...
-            storage.init_database().await?;
-            Box::new(storage)
+            let db_file_path = (!config.use_ephemeral_db).then(|| {
+                PathBuf::from(&config.data_dir).join("db.sqlite")
+                        .into_os_string()
+                        .into_string()
+                        .unwrap(/*can't fail as we only join rust strings*/)
+            });
+            let (storage, _init_hint) = SqliteStorage::init_storage_system(db_file_path).await?;
+            //FIXME pass the hint to dart for deciding if migrations are needed
```
biernacki, line 331: do we have a task to track this?

discovery_engine_core/core/src/storage/sqlite.rs
```diff
@@ -55,59 +48,29 @@ use crate::{
...
 #[derive(Clone)]
 pub(crate) struct SqliteStorage {
+    /// Path to the database file.
+    ///
+    /// `None` in case of a ephemeral (i.e. in memory) db.
+    ///
+    /// We didn't use PathBuf as it must be valid string.
+    //TODO[pmk] use file_path for app triggered db reset
+    #[allow(unused)]
+    file_path: Option<String>,
```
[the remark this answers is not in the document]
rustonaut: The reason why it's a `String` are:

1. `Storage::init_storage_system` takes a db identifier which might not be a path (the renaming of it's parameter from `file_path` to `db_identifier` was in the fixup I forgot to push).
1. it's more complicated to concat with `PathBuf` as you want to concat to the file name, not add a path suffix
1.  `SqliteConnectOptions::from_str()` is a bit special doing some "magic" in some cases. But I checked it and it does only "magic" for `:memory:` so we can use it only for `:memory:`.

I changed it to `PathBuf` as that is fine, too.

---

To elaborate on `from_str` vs. using the builder methods:

```rust
SqliteConnectOptions::from_str("sqlite::memory:")
```

is roughly equivalent to 

```rust
let seqno = IN_MEMORY_DB_SEQ.fetch_add(1, Ordering::Relaxed);
SqliteConnectionOptions::new()
    .filename(format!("file:sqlx-in-memory-{}", seqno))
    .memory(true) //this actually doesn't exist
    .shared_cache(true)
```

it's not equivalent to 

```
SqliteConnectionOptions::new()
```

Be aware that the behavior of `from_str` is what you need/want for `Pool<Sqlite>`.

So the sqlx uri is not just a sqlite uri prefixed with `sqlite:` (sqlite uri as defined here <https://www.sqlite.org/uri.html>).

discovery_engine_core/core/src/engine.rs
rustonaut: yes the task from which this is the first (of multiple, likely 3) PRs :wink: 

Split because this is the first.

discovery_engine_core/core/src/storage/sqlite/setup.rs
```diff
@@ -0,0 +1,331 @@
...
+    // either the database was corrupted or we messed up the migrations,
+    // as the app must continue working so we start from scratch
+    //FIXME we could make a backup so that if we push a bad update the
+    //      data could be restored with a follow-up update. But for now
+    //      this is fully out of scope of what we have resources for.
+    delete_db_files(file_path.as_deref()).await?;
+
+    init_storage_system_once(file_path)
+        .await
+        .map(|storage| (storage, InitDbHint::DbOverwrittenDueToErrors(first_error)))
```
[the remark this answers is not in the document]
biernacki: I agree with Jan here, I think it would be more readable if it was moved into the last march arm

discovery_engine/lib/src/ffi/types/init_config.dart
```diff
@@ -54,7 +54,7 @@ class InitConfigFfi with EquatableMixin {
   final String? deConfig;
   final String? logFile;
   final String dataDir;
-  final bool useInMemoryDb;
+  final bool useInEphemeralDb;
```
biernacki, line 57:
```suggestion
  final bool useEphemeralDb;
```

discovery_engine/lib/src/ffi/types/init_config.dart
```diff
@@ -76,7 +76,7 @@ class InitConfigFfi with EquatableMixin {
         deConfig,
         logFile,
         dataDir,
-        useInMemoryDb,
+        useInEphemeralDb,
```
biernacki, line 79:
```suggestion
        useEphemeralDb,
```

biernacki: and also below

discovery_engine/test/ffi/types/init_config_test.dart
```diff
@@ -40,7 +40,7 @@ void main() {
       maxDocsPerFeedBatch: 2,
       maxDocsPerSearchBatch: 20,
       dataDir: 'foo/bar',
-      useInMemoryDb: false,
+      useInEphemeralDb: false,
```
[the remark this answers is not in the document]
biernacki: and below as well

discovery_engine_core/core/src/config.rs
```diff
@@ -103,8 +103,8 @@ pub struct InitConfig {
     pub log_file: Option<String>,
     /// Directory in which user data should be stored.
     pub data_dir: String,
-    /// Use a in-memory db instead of a db in the `data_dir`
-    pub use_in_memory_db: bool,
+    /// Use a ephemeral db instead of a db in the `data_dir`
```
biernacki, line 106:
```suggestion
    /// Use an ephemeral db instead of a db in the `data_dir`
```

biernacki (commented):
search and replace for `useInEphemeralDb` => `useEphemeralDb`, otherwise it looks good

rustonaut:
@biernacki the `useInEphemeralDb `=> `useEphemeralDb` where already there when I requested a re-review :thinking: 

The reason I missed it originally was because I had forgotten to remove some globs from my project wide string search exclude (it's normally `.git/**,*.g.part,*.g.dart,*.freezed.dart,*.ffigen.dart,target/**`). :laughing: 

Through I did miss a clippy hint, which is now fixed (with the spelling). I just removed the unused field. We can re-add it when we need it.

biernacki:
I think you might need to push some changes, as I can still see the `useInEphemeralDb` version

biernacki (approved):
nice work 🚀

rustonaut:
bors merge

PR #578, opened 2022-08-31
[Merged by Bors] - TO-3086 better type for date_published

janpetschexain:
**References**

- [TO-3086]

**Summary**

- replace `NaiveDateTime` with `DateTime<Utc>` for `date_published`
- adjust ffi & tests

[TO-3086]: https://xainag.atlassian.net/browse/TO-3086?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine_core/bindings/src/types/date_time.rs
+++ b/discovery_engine_core/bindings/src/types/date_time.rs
@@ -14,17 +14,20 @@
 
 //! FFI functions for handling date time fields.
 
-use chrono::NaiveDateTime;
+use chrono::{DateTime, TimeZone, Utc};
+
+/// cbindgen:ignore
+pub(super) type DateTimeUtc = DateTime<Utc>;
 
 const NANOS_PER_MICRO: i64 = 1_000;
 const MICROS_PER_SECOND: i64 = 1_000_000;
 
-/// [`chrono::naive::MAX_DATETIME`] in micros
+/// [`DateTimeUtc::MAX_UTC`] in micros
 const MAX_MICRO_SECONDS: i64 = 8_210_298_412_799_999_999;
-/// [`chrono::naive::MIN_DATETIME`] in micros
+/// [`DateTimeUtc::MIN_UTC`] in micros
 const MIN_MICRO_SECONDS: i64 = -8_334_632_851_200_000_000;
 
-/// Creates a rust `NaiveDateTime` at given memory address.
+/// Creates a rust `DateTimeUtc` at given memory address.
 ///
 /// Returns `1` if it succeeded `0` else wise.
 ///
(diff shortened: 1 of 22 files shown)


discovery_engine_core/providers/src/newscatcher.rs
```diff
@@ -261,54 +261,56 @@ pub struct Article {
     /// While Newscatcher claims to have some sort of timezone support in their
     /// [API][<https://docs.newscatcherapi.com/api-docs/endpoints/search-news>] (via the
     /// `published_date_precision` attribute), in practice they do not seem to be supplying any
-    /// sort of timezone information. As a result, we provide NaiveDateTime for now.
+    /// sort of timezone information. As a result, we provide DateTime<Utc> for now.
```
janpetschexain, line 264: does anybody know how a valid newscatcher `published_date` looks like if `published_date_precision` is not `timezone unknown`? it would be nice have deserialization also if `published_date_precision is `full` or `date`, because currently this causes the deserialization of the whole `Article` to fail. see also https://docs.newscatcherapi.com/api-docs/endpoints/search-news.

janpetschexain: @acrrd

acrrd: does not work when `published_date_precision` is `full`? Usually all articles have it

acrrd: https://demo.newscatcherapi.com/ for `full` it looks the same.
I don't know how we got so many `timezone_unknow` in our test data

janpetschexain: ok, then it will work for `full` and `timezone unknown` and for both cases we just continue to assume UTC 👍 i guess `date` is an edge case which can be ignored.

discovery_engine_core/bindings/src/types/date_time.rs
```diff
@@ -33,100 +36,106 @@ const MIN_MICRO_SECONDS: i64 = -8_334_632_851_200_000_000;
...
-        unsafe { init_naive_date_time_at(place, MAX_MICRO_SECONDS) };
-        let truncated_max = NaiveDateTime::MAX.with_nanosecond(999_999_000).unwrap();
-        assert_eq!(*place, truncated_max);
-
-        let micros = unsafe { get_naive_date_time_micros_since_epoch(&NaiveDateTime::MAX) };
-        unsafe { init_naive_date_time_at(place, micros) };
-        assert_eq!(*place, truncated_max);
+        let mut place = MaybeUninit::uninit();
+        unsafe { init_date_time_utc_at(place.as_mut_ptr(), MAX_MICRO_SECONDS) };
+        let place = unsafe { place.assume_init() };
```
finiteprods, line 102: :grey_question: so here (and similarly below) presumably you could have followed the same pattern as before but you wanted to emphasize that `place` should be passed here uninitialized?

finiteprods (approved):
looks good :+1:

discovery_engine_core/bindings/src/types/date_time.rs
janpetschexain: i would even argue that the tests were written wrongly before. they worked because the overwritten parts in `init_date_time_utc_at()` etc don't have an explicit `Drop` implementation. but as the docs of `init_date_time_utc_at()` correctly state, they should only be called on uninitialized memory (to avoid such pitfalls).

janpetschexain:
bors merge

janpetschexain:
bors cancel

janpetschexain:
bors merge

discovery_engine_core/bindings/src/types/date_time.rs
finiteprods: cool, thanks for clarifying

PR #580, opened 2022-09-01
[Merged by Bors] - ET-3026 endpoints path configs

janpetschexain:
**References**

- [TO-3026]

**Summary**

- change the provider `Config` to be available to each provider individually
- align provider names
- update engine/ffi/dart configs

[TO-3026]: https://xainag.atlassian.net/browse/TO-3026?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


discovery_engine/example/lib/example.dart
```diff
@@ -28,8 +28,6 @@ Future<void> runExample() async {
     applicationDirectoryPath: appDirPath,
     feedMarkets: {const FeedMarket(langCode: 'de', countryCode: 'DE')},
     manifest: manifest,
-    headlinesProviderPath: '/newscatcher/v1/latest-headlines',
-    newsProviderPath: '/newscatcher/v1/search-news',
```
biernacki, line 32: ❔ I don't see any information about removing this in the task or story descriptions, was it decided somewhere but not documented? Is there a plan to introduce other way for the app to configure the endpoints/providers? I think it would be nice to leave some form on control over this to the app to be able to do a/b test (that is if this is still our priority)

discovery_engine_core/core/src/engine.rs
```diff
@@ -1759,9 +1763,7 @@ pub(crate) mod tests {
                 .set_body_string(include_str!("../test-fixtures/newscatcher/duplicates.json"));
 
             Mock::given(method("POST"))
-                .and(path_regex(
-                    "/newscatcher/headlines-endpoint-name|/newscatcher/v2/trusted-sources|newscatcher/news-endpoint-name",
-                ))
```
biernacki, line 1764: ❔ could you remind me what was the reason for this regex?

biernacki (commented):
Code looks good, cleaner 👍 I left just a couple of questions to help me better understand the reason to drop the parameters that the app provides.

discovery_engine/example/lib/example.dart
janpetschexain: i removed them as those paths moved into their own config struct. currently there is no way to set them from the outside, but we can easily arange for this via the `de_config` JSON config. should this be implemented now or only when needed later on @acrrd ?

discovery_engine_core/core/src/engine.rs
janpetschexain: we used routes in the tests which return a mocked fixed result. at some point additional routes were used in the tests, but they don't actually matter since we always return the same mocked result anyways. hence i simplified this to avoid to update outdated routes, resp. to expose the internal routes from the providers crate.

discovery_engine/example/lib/example.dart
acrrd: We still need to accept them from outside, but it should be optional. This should not use the `de_config`  but dart struct we have now.

discovery_engine/test/integration/utils/local_newsapi_server.dart
```diff
@@ -69,14 +69,14 @@ class LocalNewsApiServer {
...
+            case '/newscatcher/v2/search':
             case '/_mlt':
               await _replyWithData(
                 request,
                 _snFiles.asMap()[_snCounter++] ?? 'search-news-empty.json',
               );
               break;
-            case '/newscatcher/v1/latest-headlines':
+            case '/newscatcher/v2/latest_headlines':
             case '/newscatcher/v2/trusted-sources':
```
biernacki, line 80: I think we should stick to either snake_case or kebab-case for the endpoint names, but it's probably something to configure on the proxy level as well, right?

discovery_engine_core/core/src/config.rs
```diff
@@ -124,11 +128,6 @@ pub struct EndpointConfig {
     /// The maximum age of a news article, in days, after which we no longer
     /// want to display them.
     pub max_article_age_days: usize,
-    /// The timeout after which a provider aborts a request.
-    #[serde(with = "serde_duration_as_milliseconds")]
-    pub timeout: Duration,
-    /// The number of retries in case of a timeout.
-    pub retry: usize,
```
biernacki, line 131: why we're removing these options?

biernacki (approved):

discovery_engine/test/integration/utils/local_newsapi_server.dart
janpetschexain: yes, the proxy routes have to be changed first. this is somewhat covered by the other tasks of the story.

discovery_engine_core/core/src/config.rs
janpetschexain: they are just moved to the provider config. the procedure to configure them from the app side via the JSON `de_config` remains the same.

janpetschexain:
bors merge

PR #582, opened 2022-09-05
[Merged by Bors] - TO-3110 [2] pass dummy dart migration data to rust

rustonaut:
Create a skeleton to passing dart migration data to rust, using them when initializing the db and cleaning up the dart db(s) afterward.

**Based On:**

- #577

### DIFF
--- a/discovery_engine/lib/src/infrastructure/migration.dart
+++ b/discovery_engine/lib/src/infrastructure/migration.dart
@@ -75,8 +75,7 @@ class DartMigrationData {
       engineState: await engineStateRepository.load(),
       documents: await documentRepository.fetchAll(),
       activeDocumentData: activeDocumentDataRepository.box.toMap().map(
-            // ignore: avoid_annotating_with_dynamic
-            (dynamic key, data) =>
+            (Object? key, data) =>
                 MapEntry(DocumentId.fromString(key as String), data),
           ),
       reactedSources: await sourceReactedRepository.fetchAll(),
--- a/discovery_engine/lib/src/infrastructure/migration.dart
+++ b/discovery_engine/lib/src/infrastructure/migration.dart
@@ -1,0 +1,98 @@
+// Copyright 2022 Xayn AG
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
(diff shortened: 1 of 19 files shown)


rustonaut:
@biernacki I added an additional commit which extracts the dart migration code into it's own module, sorry for that late change.

(It also now has the `DartMigrationData` class no longer under ffi, which stylistically wasn't that grate/uniform with the rest of the code.)

discovery_engine/lib/src/ffi/types/init_config.dart
```diff
@@ -163,7 +163,7 @@ class InitConfigFfi with EquatableMixin {
     dataDir.writeNative(
       ffi.init_config_place_of_data_dir(place),
     );
-    useInMemoryDb.writeNative(ffi.init_config_place_of_use_in_memory_db(place));
+    useInMemoryDb.writeNative(ffi.init_config_place_of_use_ephemeral_db(place));
```
biernacki, line 166: I think this needs a rebase :)

discovery_engine/lib/src/infrastructure/migration.dart
```diff
@@ -0,0 +1,98 @@
...
+        sourceReactedRepository.box.isEmpty) {
+      return null;
+    }
+
+    return DartMigrationData(
+      engineState: await engineStateRepository.load(),
+      documents: await documentRepository.fetchAll(),
+      activeDocumentData: activeDocumentDataRepository.box.toMap().map(
+            // ignore: avoid_annotating_with_dynamic
+            (dynamic key, data) =>
```
biernacki, line 79: Maybe you can try without dynamic
```suggestion
            (Object key, data) =>
```

biernacki (commented):
Looks good, just needs a rebase :)

discovery_engine/lib/src/infrastructure/migration.dart
rustonaut: sadly that doesn't work :sweat:

biernacki: maybe try `Object?` but if that doesn't work either, then just leave it as is

biernacki (approved):
lgtm

rustonaut:
bors merge

PR #584, opened 2022-09-07
[Merged by Bors] - TO-3075 rustify source reaction logic

finiteprods:
**Summary**

- previously #564 

ports the dart logic for source reactions to rust

- when `storage` feature is on, source reactions are fetched from the rust db.
- refactors various engine methods to exclude no longer needed `sources` argument.
- removes no longer needed source reaction repository calls on dart side.

### DIFF
--- a/discovery_engine/lib/src/domain/engine/engine.dart
+++ b/discovery_engine/lib/src/domain/engine/engine.dart
@@ -77,7 +77,6 @@ abstract class Engine {
 
   /// Changes the excluded and trusted sources.
   Future<void> setSources(
-    List<SourceReacted> sources,
     Set<Source> excluded,
     Set<Source> trusted,
   );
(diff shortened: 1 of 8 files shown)


rustonaut (commented):
mostly looks good but there is one problem not with this PR  but with the merged predecessor PR

The thing is:

1. weight implies liked, so from this point alone we might want to maybe not store `liked` in the table
2. I think the case for not storing `liked` get stronger with the fact that `liked` is a computed property based on the last user reaction on a document with given source
3. The line `weight = CASE WHEN liked = 1 THEN (weight + 1) END` is slightly problematic
    - we don't need to duplicate the only-update-if-liked logic in dart and sql, duplicating logic in this way makes code harder to change later
    - if  liked is false then it will try to assign `null` to `weight` which will  happen to fail due to `weight` being `not null`.  This isn't grate as its a misleading error and there is no reason to fail here.
    - Like mentioned above I also think we shouldn't store `liked` here.
    - I would change it in one of two ways:
      1. just change it to `weight = weight + 1`
      2. change it to `weight = weight + ?` and accept the difference as `i32` as additional parameter in `update_source_weight`, which would fully remove any logic-coupling to "liked" from the storage (which is why I prefer this solution)

finiteprods:
> All things related to the previous PR could go into a follow up PR, I guess.

sure, will open a follow up pr for those things then.

rustonaut (approved):
lgtm

finiteprods:
bors merge

finiteprods:
bors merge

PR #585, opened 2022-09-07
[Merged by Bors] - TO-3105 migrate engine state

rustonaut:
Added engine state to the dart/hive->rust/sqlite migration.

PR #586, opened 2022-09-07
[Merged by Bors] - TO-3107 migrate source preferences

rustonaut:
**Based On:**
- #585

### DIFF
--- a/discovery_engine_core/core/src/storage/sqlite/dart_migrations.rs
+++ b/discovery_engine_core/core/src/storage/sqlite/dart_migrations.rs
@@ -14,6 +14,8 @@
 
 //! Module for handling dart->rust/sqltie migrations
 
+use itertools::Itertools;
+
 use crate::{
     storage::{Error, Storage},
     DartMigrationData,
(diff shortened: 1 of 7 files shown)


discovery_engine_core/core/src/storage/sqlite/dart_migrations.rs
```diff
@@ -29,13 +31,25 @@ pub(super) async fn store_migration_data(
...
 
     if let Some(engine_state) = &data.engine_state {
         storage.state().store(engine_state).await?;
     }
 
+    storage
+        .source_preference()
+        .set_trusted(&data.trusted_sources.iter().map_into().collect())
+        .await
+        .unwrap();
```
biernacki, line 45: is there a chance for `panic` here? if not maybe a comment why is it safe could be useful here

discovery_engine_core/core/src/storage/sqlite/dart_migrations.rs
```diff
@@ -29,13 +31,25 @@ pub(super) async fn store_migration_data(
...
+
+    storage
+        .source_preference()
+        .set_excluded(&data.excluded_sources.iter().map_into().collect())
+        .await
+        .unwrap();
```
biernacki, line 51: same as above

biernacki (commented):
a few questions, the rest looks good!

biernacki (approved):

PR #588, opened 2022-09-08
[Merged by Bors] - Revert split of coi and kps

acrrd:
There is a problem with the migration that fails and does not seem to be an easy to understand why is happening.


rustonaut (approved):
lgtm

through I didn't look too closely into it, but then it's auto generate reverts :shrug:

acrrd:
bors merge

PR #589, opened 2022-09-08
Keep symbols for mlt

acrrd:
When using the mlt endpoint we don't need to strip the symbols. Not doing it will return more results in some case.

### DIFF
--- a/discovery_engine_core/ai/ai/src/coi/key_phrase.rs
+++ b/discovery_engine_core/ai/ai/src/coi/key_phrase.rs
@@ -26,7 +26,7 @@ use itertools::izip;
 use ndarray::{s, Array1, Array2, ArrayBase, Axis, Data, Ix, Ix2};
 use rayon::iter::{IntoParallelIterator, ParallelIterator};
 use serde::{Deserialize, Serialize};
-use xayn_discovery_engine_providers::{clean_query, Market};
+use xayn_discovery_engine_providers::Market;
 
 use crate::{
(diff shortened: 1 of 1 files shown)


discovery_engine_core/ai/ai/src/coi/key_phrase.rs
```diff
@@ -213,7 +213,6 @@ fn unify(
     let mut candidates = candidates
         .iter()
         .filter_map(|candidate| {
-            let candidate = clean_query(candidate);
```
rustonaut, line 216: shouldn't we make this conditional based on weather the mlt endpoint is used?

acrrd: We have no idea if that is used or not here. There is no plan to revert the use of mlt.
We can always do `clean_query` when generating the filter and that is probably the correct place where to do it.

rustonaut:
@acrrd  not sure if you didn't notice or if you had other things to do, but there are still some tests failing in `key_phrase.rs` due to the change.

They mainly test indirectly if the cleaning works so I guess some of the failing ones could be just removed.

Also I assume we did make sure to properly validate/sanitize input in the mlt endpoint.(?)

acrrd:
I'm waiting for the kps split to be merged back since it modifies something here. Also we still need to escape single e double quote. So I think I'll close this and have another task.

acrrd:
We need more understanding of what symbols are ok and what are not for this endpoint.

PR #591, opened 2022-09-12
TO-3317 Fix for engine state deserialization

biernacki:
**References**:
- [TO-3317]

**Summary**:
- after merging this change https://github.com/xaynetwork/xayn_discovery_engine/pull/578 old serialized engine state couldn't be deserialized without errors because of type change `NaiveDateTime` => `DateTime<Utc>`
- added a custom deserializer which doesn't return error when deserializer has parsing issues
- added test to cover this scenario

[TO-3317]: https://xainag.atlassian.net/browse/TO-3317?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine_core/core/src/document.rs
+++ b/discovery_engine_core/core/src/document.rs
@@ -197,19 +197,19 @@ impl NewsResource {
     }
 }
 
-/// NaiveDateTime tolerant deserialization of `DateTime<Utc>`.
+/// `NaiveDateTime` tolerant deserialization of `DateTime<Utc>`.
 fn deserialize_date_time_custom<'de, D>(deserializer: D) -> Result<DateTime<Utc>, D::Error>
 where
     D: Deserializer<'de>,
 {
-    const FORMAT: &'static str = "%Y-%m-%d %H:%M:%S";
+    const FORMAT: &str = "%Y-%m-%d %H:%M:%S";
 
     let s = String::deserialize(deserializer)?;
     Utc.datetime_from_str(&s, FORMAT)
         // .or(Ok(DateTime::<Utc>::MIN_UTC))
         // .map_err(|err| serde::de::Error::custom(format!("err: {}, s: {}", err, s)))
         .or_else(|_| {
-            let val = NaiveDateTime::parse_from_str(&s, FORMAT).unwrap_or(NaiveDateTime::default());
+            let val = NaiveDateTime::parse_from_str(&s, FORMAT).unwrap_or_default();
             let val = DateTime::<Utc>::from_utc(val, Utc);
             Ok(val)
         })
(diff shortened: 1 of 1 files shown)


discovery_engine_core/core/src/document.rs
```diff
@@ -196,6 +197,20 @@ impl NewsResource {
     }
 }
 
+/// Deserialization of `DateTime<Utc>` with fallback.
+fn deserialize_date_time_with_fallback<'de, D>(deserializer: D) -> Result<DateTime<Utc>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    // const FORMAT: &str = "%Y-%m-%d %H:%M:%S";
+    const FORMAT: &str = "%Y-%m-%dT%H:%M:%S";
```
biernacki, line 206: this is the format which works for `NaiveDateTime` serialized by bincode, and then later deserialized into `String`

discovery_engine_core/core/src/document.rs
```diff
@@ -196,6 +197,20 @@ impl NewsResource {
...
+    let s = String::deserialize(deserializer).unwrap_or_default();
+
+    DateTime::parse_from_rfc3339(&s)
+        .map(|dt| dt.with_timezone(&Utc))
+        .or_else(|_| Ok(Utc.datetime_from_str(&s, FORMAT).unwrap_or_default()))
```
rustonaut, line 211: :question:  What kind of fallback is wanted?

Due to the `unwrap_or_default` at l.207 and l.211 it will fallback to unix-epoch 0 if:

- it can't deserialize a string because of any reason including field is missing, malformed utf8 and not unicode text
- it is neither in the `NativeDateTime` nor in the `DateTime<Utc>` serialization fomrat

What we need is only "if rfc3339 parsing fails parse with `FORMAT` instead".

But then I would say failing state deserialization because of `date_published` is never something we want so this _is_ a good solution even through it does more then just fixing the problem we have.

:wrench:  Through it would be a bit nicer if we would log any error which doesn't come from failing to parse `DateTime::parse_from_rfc3339(&s)` i.e. all the `unwrap_or_default()` cases. We added an extension
trait a while ago which adds a `log_error()` functionality to result. So you can use `.log_error().unwrap_or_default()` instead. (it's in `core/src/utils.rs`; Currently this module is `cfg(feature="storage")` but that is only to avoid unused warnings so  you can just move the cfg to the unused members of `core/src/uitls.rs` or similar).

biernacki:
Closing in favor of migration script during deserialization, which will be introduced as a new PR

PR #593, opened 2022-09-13
[Merged by Bors] - Add draft of openapi

acrrd:
Add specification of our api using the openapi format.
The current specification does not completly match what is currently implemented but document the api that we want for our customers.

### DIFF
--- a/discovery_engine_core/web-api/openapi.yaml
+++ b/discovery_engine_core/web-api/openapi.yaml
@@ -23,7 +23,8 @@ paths:
       summary: Get personalized document for the user
       description: |-
         Returns a map from `document_id` to the score of that document for the
-        specified `user_id`.
+        specified `user_id`. The score is a value between 0 and 1 where an higher value
+        means that the document matches more the preferences of the user.
       parameters:
         - name: user_id
           in: path
--- a/discovery_engine_core/web-api/openapi.yaml
+++ b/discovery_engine_core/web-api/openapi.yaml
@@ -119,7 +119,7 @@ components:
       example: "document id"
     UserId:
       type: string
-      description: A document id can be any UTF-8 string that does not contain the null byte.
+      description: A user id can be any non-empty UTF-8 string that does not contain the null byte.
       example: "user id"
     PersonalizedDocumentData:
       type: object
--- a/discovery_engine_core/web-api/openapi.yaml
+++ b/discovery_engine_core/web-api/openapi.yaml
@@ -17,7 +17,7 @@ paths:
     get:
       tags:
         - personalization
-      summary: Get personalized document for the user
+      summary: Get personalized documents for the user
       description: |-
         Returns a map from `document_id` to the score of that document for the
         specified `user_id`. The score is a value between 0 and 1 where a higher value
--- a/discovery_engine_core/web-api/openapi.yaml
+++ b/discovery_engine_core/web-api/openapi.yaml
@@ -13,7 +13,7 @@ tags:
   - name: ingestion
     description: Add documents to the system
 paths:
-  /users/{user_id}/documents/personalized:
+  /users/{user_id}/personalized_documents:
     get:
       tags:
         - personalization
@@ -57,7 +57,7 @@ paths:
               schema:
                 $ref: '#/components/schemas/PersonalizedDocumentsError'
 
-  /users/{user_id}/documents/interact:
+  /users/{user_id}/interactactions:
     patch:
       tags:
         - personalization
--- a/discovery_engine_core/web-api/openapi.yaml
+++ b/discovery_engine_core/web-api/openapi.yaml
@@ -4,11 +4,30 @@ info:
   version: 1.0.0-rc1
   description: |-
     # Overview
-    The API is divided 
+    The personalization system is divided into two parts: one to manage documents and one to manage users.
+
+    ## Documents management
+    For this system, a document is anything that has an id, a snippet, and an arbitrary set of properties.
+    The id uniquely identifies the document in the system.
+    The system uses the snippet to create a document representation that can be matched against user interests.
+
+    It is possible to attach any properties to a document; the system will not use them but will be returned
+    together with the document when asking for personalized documents for a user.
+    They could help store information like a title and the URL where it is possible to find the document.
+    Each document can have its own set of properties.
+
+    The API offers an endpoint that allows uploading the documents to the system.
+    After a successful upload, it will be possible to use the document in other endpoints.
+
+    ## User management
+    The system identifies a user with only an id.
+    There are two main endpoints: one that allows registering an interaction between a user and a document (a user clicked or liked a document),
+    and one to retrieve a personalized set of documents for the user.
+    The system needs a few user interactions with documents before it can generate a personalized list.
   
     # Auth
     To authenticate with the API two bearer tokens will be provided. One will work for the
-    personalization endpoints and the other one will allow documents to be added to the system.
+    user endpoints and the other one to manage the documents.
     
 tags:
   - name: users
--- a/discovery_engine_core/web-api/openapi.yaml
+++ b/discovery_engine_core/web-api/openapi.yaml
@@ -3,6 +3,9 @@ info:
   title: Personalization API
   version: 1.0.0-rc1
   description: |-
+    # Overview
+    The API is divided 
+  
     # Auth
     To authenticate with the API two bearer tokens will be provided. One will work for the
     personalization endpoints and the other one will allow documents to be added to the system.
@@ -121,6 +124,10 @@ components:
     DocumentProperties:
       description: Arbitrary properties that can be attached to a document.
       type: object
+      example:
+        title: "News title"
+        image_url: "http://example.com/news_image.jpg"
+        link: "http://example.com/this_news"
     UserId:
       description: A user id can be any non-empty UTF-8 string that does not contain the null byte.
       type: string
--- a/discovery_engine_core/web-api/openapi.yaml
+++ b/discovery_engine_core/web-api/openapi.yaml
@@ -56,7 +56,7 @@ paths:
         - personalization
       summary: Add interaction between an user and a document
       description: |-
-        The interaction is used to provide personalized documents to the user
+        The interaction is used to provide personalized documents to the user.
       parameters:
         - name: user_id
           in: path
--- a/discovery_engine_core/web-api/openapi.yaml
+++ b/discovery_engine_core/web-api/openapi.yaml
@@ -141,14 +141,24 @@ components:
       properties:
         kind:
           type: string
+    UserInteractionType:
+      type: string
+      enum: [positive]
+    UserInteractionData:
+      type: object
+      properties:
+        id:
+          $ref: '#/components/schemas/DocumentId'
+        type:
+          $ref: '#/components/schemas/UserInteractionType'
     UserInteractionRequest:
       type: object
-      required: [documents_id]
+      required: [documents]
       properties:
-        documents_id:
+        documents:
           type: array
           items:
-            $ref: '#/components/schemas/DocumentId'
+            $ref: '#/components/schemas/UserInteractionData'
     IngestedDocument:
       type: object
       required: [id, snippet, published_date]
--- a/discovery_engine_core/web-api/openapi.yaml
+++ b/discovery_engine_core/web-api/openapi.yaml
@@ -8,7 +8,7 @@ info:
     
     # Auth
     To authenticate with the API two bearer tokens will be provided. One will work for the
-    personalization endpoints and the other one will allow to add documents to the system.
+    personalization endpoints and the other one will allow documents to be added to the system.
     
 tags:
   - name: personalization
--- a/discovery_engine_core/web-api/openapi.yaml
+++ b/discovery_engine_core/web-api/openapi.yaml
@@ -24,7 +24,7 @@ paths:
       description: |-
         Returns a map from `document_id` to the score of that document for the
         specified `user_id`. The score is a value between 0 and 1 where a higher value
-        means that the document matches more the preferences of the user.
+        means that the document matches the preferences of the user better.
       parameters:
         - name: user_id
           in: path
--- a/discovery_engine_core/web-api/openapi.yaml
+++ b/discovery_engine_core/web-api/openapi.yaml
@@ -3,8 +3,7 @@ info:
   title: Personalization API
   version: 1.0.0-rc1
   description: |-
-    The endpoints can be contacted at `t<name>.xaynet.dev/v1` or `t<name>.xaynet.dev/v1.0`.
-    The `v1` path will always point to the latest `v1.x`.
+    The endpoints can be contacted at `t<name>.xaynet.dev/v1`.
     
     # Auth
     To authenticate with the API two bearer tokens will be provided. One will work for the
@@ -35,25 +34,33 @@ paths:
         - name: count
           in: query
           description: Maximum number of personalized documents to return
+          required: false
           schema:
             type: integer
             format: int32
             minimum: 1
             maximum: 100
+            default: 10
       responses:
         '200':
           description: successful operation
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/PersonalizedDocuments'
+                $ref: '#/components/schemas/PersonalizedDocumentsResponse'
         '400':
-          description: invalid user id
+          description: invalid user id.
         '404':
-          description: user not found or the user does not have enough interactions
+          description: user not found.
+        '422':
+          description: impossible to create a personalized list for the user.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/PersonalizedDocumentsError'
 
-  /users/{user_id}/documents/interacted:
-    post:
+  /users/{user_id}/documents/interact:
+    patch:
       tags:
         - personalization
       summary: Add interaction between a user and a document
@@ -70,14 +77,12 @@ paths:
         content:
           application/json:
             schema:
-              $ref: '#/components/schemas/UserInteraction'
+              $ref: '#/components/schemas/UserInteractionRequest'
       responses:
         '204':
           description: successful operation
         '400':
           description: invalid request. User or document id is invalid
-        '404':
-          description: user not found
         
   /documents/:
     post:
@@ -102,40 +107,53 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/IngestionResponseFail'
-  
+                $ref: '#/components/schemas/IngestionError'
+
 components:
   schemas:
     DocumentId:
       type: string
-      example: "document_id_1"
+      description: A document id can be any UTF-8 string that does not contain the null byte.
+      example: "document id"
     UserId:
       type: string
-      example: "user_id_1"
+      description: A document id can be any UTF-8 string that does not contain the null byte.
+      example: "user id"
     PersonalizedDocumentData:
       type: object
+      required: [id, score]
       properties:
+        id:
+          $ref: '#/components/schemas/DocumentId'
         score:
           type: number
-      example:
-        score: 0.872
-    PersonalizedDocuments:
+    PersonalizedDocumentsResponse:
       type: object
-      additionalProperties:
-        type: object
-        $ref: '#/components/schemas/PersonalizedDocumentData'
-      example:
-        - "document_id_1":
-          - score: 0.872
-    UserInteraction:
-      type: object
+      required: [documents]
       properties:
-        document_id:
-          $ref: '#/components/schemas/DocumentId'
+        documents:
+          type: array
+          items:
+            $ref: '#/components/schemas/PersonalizedDocumentData'
+    PersonalizedDocumentsError:
+      type: object
+      required: [kind]
+      properties:
+        kind:
+          type: string
+    UserInteractionRequest:
+      type: object
+      required: [documents_id]
+      properties:
+        documents_id:
+          type: array
+          items:
+            $ref: '#/components/schemas/DocumentId'
     IngestedDocument:
       type: object
+      required: [id, snippet, published_date]
       properties:
-        document_id:
+        id:
           $ref: '#/components/schemas/DocumentId'
         snippet:
           type: string
@@ -145,13 +163,15 @@ components:
           format: date-time
     IngestionRequest:
       type: object
+      required: [documents]
       properties:
         documents:
           type: array
           items:
             $ref: '#/components/schemas/IngestedDocument'
-    IngestionResponseFail:
+    IngestionError:
       type: object
+      required: [documents]
       properties:
         documents:
           type: array
--- a/discovery_engine_core/web-api/openapi.yaml
+++ b/discovery_engine_core/web-api/openapi.yaml
@@ -143,15 +143,13 @@ components:
         published_date:
           type: string
           format: date-time
-    IngestedDocuments:
-      type: array
-      items:
-        $ref: '#/components/schemas/IngestedDocument'
     IngestionRequest:
       type: object
       properties:
         documents:
-          $ref: '#/components/schemas/IngestedDocuments'
+          type: array
+          items:
+            $ref: '#/components/schemas/IngestedDocument'
     IngestionResponseFail:
       type: object
       properties:
--- a/discovery_engine_core/web-api/openapi.yaml
+++ b/discovery_engine_core/web-api/openapi.yaml
@@ -55,7 +55,7 @@ paths:
     post:
       tags:
         - personalization
-      summary: Add interaction between an user and a document
+      summary: Add interaction between a user and a document
       description: |-
         The interaction is used to provide personalized documents to the user.
       parameters:
--- a/discovery_engine_core/web-api/openapi.yaml
+++ b/discovery_engine_core/web-api/openapi.yaml
@@ -176,16 +176,13 @@ components:
           enum: [invalid_user_id, invalid_document_id]
     IngestedDocument:
       type: object
-      required: [id, snippet, published_date]
+      required: [id, snippet]
       properties:
         id:
           $ref: '#/components/schemas/DocumentId'
         snippet:
-          type: string
           description: Text that will be used to match the document against the user interests.
-        published_date:
           type: string
-          format: date-time
         properties:
           $ref: '#/components/schemas/DocumentProperties'
     IngestionRequest:
--- a/discovery_engine_core/web-api/openapi.yaml
+++ b/discovery_engine_core/web-api/openapi.yaml
@@ -85,7 +85,7 @@ paths:
       summary: Add documents to the system
       description: |-
         Add documents to the system. The system will create a representation of the document
-        that will be used to match it against the preference of a user.
+        that will be used to match it against the preferences of a user.
       requestBody:
         content:
           application/json:
--- a/discovery_engine_core/web-api/openapi.yaml
+++ b/discovery_engine_core/web-api/openapi.yaml
@@ -137,7 +137,7 @@ components:
           $ref: '#/components/schemas/DocumentId'
         snippet:
           type: string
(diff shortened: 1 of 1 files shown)


discovery_engine_core/web-api/openapi.yaml
```diff
@@ -0,0 +1,152 @@
...
+        '404':
+          description: user not found or the user does not have enough interactions
+
+  /users/{user_id}/documents/interacted:
+    post:
+      tags:
+        - personalization
+      summary: Add interaction between an user and a document
+      description: |-
+        The interaction is used to provide personalized documents to the user
```
PanicButtonPressed, line 59:
```suggestion
        The interaction is used to provide personalized documents to the user.
```

discovery_engine_core/web-api/openapi.yaml
```diff
@@ -0,0 +1,152 @@
...
+        document_id:
+          $ref: '#/components/schemas/DocumentId'
+    IngestedDocument:
+      type: object
+      properties:
+        document_id:
+          $ref: '#/components/schemas/DocumentId'
+        snippet:
+          type: string
+          description: Text that that will be used to match the document against the user interests
```
PanicButtonPressed, line 140:
```suggestion
          description: Text that will be used to match the document against the user interests
```

discovery_engine_core/web-api/openapi.yaml
```diff
@@ -0,0 +1,152 @@
...
+  - name: ingestion
+    description: Add documents to the system
+paths:
+  /users/{user_id}/documents/personalized:
+    get:
+      tags:
+        - personalization
+      summary: Get personalized document for the user
+      description: |-
+        Returns a map from `document_id` to the score of that document for the
```
PanicButtonPressed, line 25: Should we probably be more specific about the `score`, e.g. saying its related to the ranking?

acrrd: I added some more information here.

PanicButtonPressed: Thanks

PanicButtonPressed (approved):
LGTM!

discovery_engine_core/web-api/openapi.yaml
```diff
@@ -0,0 +1,153 @@
...
+info:
+  title: Personalization API
+  version: 1.0.0-rc1
+  description: |-
+    The endpoints can be contacted at `t<name>.xaynet.dev/v1` or `t<name>.xaynet.dev/v1.0`.
+    The `v1` path will always point to the latest `v1.x`.
+    
+    # Auth
+    To authenticate with the API two bearer tokens will be provided. One will work for the
+    personalization endpoints and the other one will allow to add documents to the system.
```
finiteprods, line 11:
```suggestion
    personalization endpoints and the other one will allow documents to be added to the system.
```

discovery_engine_core/web-api/openapi.yaml
```diff
@@ -0,0 +1,153 @@
...
+    description: Add documents to the system
+paths:
+  /users/{user_id}/documents/personalized:
+    get:
+      tags:
+        - personalization
+      summary: Get personalized document for the user
+      description: |-
+        Returns a map from `document_id` to the score of that document for the
+        specified `user_id`. The score is a value between 0 and 1 where an higher value
```
finiteprods, line 26:
```suggestion
        specified `user_id`. The score is a value between 0 and 1 where a higher value
```

discovery_engine_core/web-api/openapi.yaml
```diff
@@ -0,0 +1,153 @@
...
+        means that the document matches more the preferences of the user.
```
finiteprods, line 27:
```suggestion
        means that the document matches the preferences of the user better.
```

discovery_engine_core/web-api/openapi.yaml
```diff
@@ -0,0 +1,153 @@
...
+        '400':
+          description: invalid user id
+        '404':
+          description: user not found or the user does not have enough interactions
+
+  /users/{user_id}/documents/interacted:
+    post:
+      tags:
+        - personalization
+      summary: Add interaction between an user and a document
```
finiteprods, line 58:
```suggestion
      summary: Add interaction between a user and a document
```

discovery_engine_core/web-api/openapi.yaml
```diff
@@ -0,0 +1,153 @@
...
+          description: user not found
+        
+  /documents/:
+    post:
+      tags:
+        - ingestion
+      summary: Add documents to the system
+      description: |-
+        Add documents to the system. The system will create a representation of the document
+        that will be used to match it against the preference of a user.
```
finiteprods, line 88:
```suggestion
        that will be used to match it against the preferences of a user.
```

finiteprods (commented):
:art: just some englishisms

discovery_engine_core/web-api/openapi.yaml
```diff
@@ -0,0 +1,153 @@
...
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/PersonalizedDocuments'
+        '400':
+          description: invalid user id
+        '404':
+          description: user not found or the user does not have enough interactions
+
+  /users/{user_id}/documents/interacted:
```
joecks, line 54: Because you can only report a single document this should be probably named 
`users/{user_id}/document/interacted`

Also much simpler would be: 

GET `users/{user_id}/document/{documentId}/interacted`

discovery_engine_core/web-api/openapi.yaml
```diff
@@ -0,0 +1,153 @@
...
+    To authenticate with the API two bearer tokens will be provided. One will work for the
+    personalization endpoints and the other one will allow to add documents to the system.
+    
+tags:
+  - name: personalization
+    description: Interaction of the user and personalized document
+  - name: ingestion
+    description: Add documents to the system
+paths:
+  /users/{user_id}/documents/personalized:
```
joecks, line 19: I don't really understand this namesapce tbh: 

`userid/documents` 

From the customers perspective documents and personalized is is basically the name so we can shorten this
to

`userID/personalizedDocuments` or just `documents`

discovery_engine_core/web-api/openapi.yaml
acrrd: It will be easier to extend in the future if the user wants to report more then one documents with one call time.
Even you report only one document the collections is still `documents`, especially in the case where you put the `{documentId}` in the path.

discovery_engine_core/web-api/openapi.yaml
acrrd: Documents for the user are what they load on the system and are the basic items that we can match to the users.
`users/{id}/documents` could be later use to return which documents the user interacted with or something similar.

discovery_engine_core/web-api/openapi.yaml
```diff
@@ -0,0 +1,153 @@
...
+        '200':
+          description: successful operation
+        '400':
+          description: invalid request
+        '422':
+          description: all or some of the documents were not successfully uploaded
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/UserInteraction'
```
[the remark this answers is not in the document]
acrrd: This was an error. Fixed with the correct type that we return.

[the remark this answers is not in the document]
joecks: But you also have to use the type here.

discovery_engine_core/web-api/openapi.yaml
```diff
@@ -0,0 +1,160 @@
...
+          application/json:
+            schema:
+              $ref: '#/components/schemas/UserInteraction'
+      responses:
+        '204':
+          description: successful operation
+        '400':
+          description: invalid request. User or document id is invalid
+        '404':
+          description: user not found
```
biernacki, line 79: maybe 404 should be if we don't find a document with specified `document_id`? User interests won't necessarily exist when we first call the endpoint

discovery_engine_core/web-api/openapi.yaml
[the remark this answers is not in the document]
biernacki:
```suggestion
                $ref: '#/components/schemas/IngestionResponseFail'
```

discovery_engine_core/web-api/openapi.yaml
```diff
@@ -0,0 +1,160 @@
...
+    UserId:
+      type: string
+      example: "user_id_1"
+    PersonalizedDocumentData:
+      type: object
+      properties:
+        score:
+          type: number
+      example:
+        - score: 0.872
```
Robert-Steiner, line 120: in the editor the example is `example: List [ OrderedMap { "score": 0.872 } ]` but shound't it be just `example: OrderedMap { "score": 0.872 }`? if that is the case we just need to remove the hyphen

discovery_engine_core/web-api/openapi.yaml
```diff
@@ -0,0 +1,160 @@
...
+    PersonalizedDocuments:
+      type: object
```
Robert-Steiner, line 122: souldn't that be an array where the items are an object with an id + the data
also the example might be updated because currently it is `example: List [ OrderedMap { "document_id_1": List [ OrderedMap { "score": 0.872 } ] } ]`

frankpepermans:
Same remark as I had on Slack

If the personalized feed returns just a key-value pair of DocumentID and score then:

- when this api is invoked from a client (presumably JS in a browser page), then this JS would also still need to do an additional query to fetch title / image / url of that Document, in order to render/display it.

We can however almost assume that this is rather impractical, since most news agencies run a classic database + web backend combo, and do not expose things like "get-document-by-id" public-facing.

- when this api is invoked as part of a page request on the server, it then adds substantial time-to-render, i.e. the round trip to our api to fetch these IDs and scores, might take unexpectedly long (100ms+ range) => which adds to the server-time required before it can present the HTML to the user.

So for now it's fine, we just need a better view of how they would implement it exactly, but as a more generic approach (and to have a fully client-side JS script for example they can just embed), we'd need the personalized call to return not just DocumentID, but also image / link / title etc...

This also affects the intake of course, as we'd then also store a bit more data.

discovery_engine_core/web-api/openapi.yaml
```diff
@@ -0,0 +1,160 @@
...
+          description: user not found or the user does not have enough interactions
+
+  /users/{user_id}/documents/interacted:
+    post:
+      tags:
+        - personalization
+      summary: Add interaction between a user and a document
+      description: |-
+        The interaction is used to provide personalized documents to the user.
+      parameters:
```
frankpepermans, line 61: Also add the document_id param here

rustonaut:
:wrench:  One additional thing is that with openapi everything is not-required by default, which IMHO is insanity as both client and server will implicitly require fields but anyway we can't change that.

This means to comply with the openapi spec we need to:

- make sure all required fields of objects as required
- make sure all required parameters are marked as required

discovery_engine_core/web-api/openapi.yaml
```diff
@@ -0,0 +1,160 @@
+openapi: 3.0.3
+info:
+  title: Personalization API
+  version: 1.0.0-rc1
+  description: |-
+    The endpoints can be contacted at `t<name>.xaynet.dev/v1` or `t<name>.xaynet.dev/v1.0`.
+    The `v1` path will always point to the latest `v1.x`.
```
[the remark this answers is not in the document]
acrrd: `t` for `tenant` but could also be different. I think it could be useful for us to have some kind of template in case we want to automatize some part of the domain.

For the second part or we do it or we don't. No one want to point to and endpoint that has no guaranteed of existing.

discovery_engine_core/web-api/openapi.yaml
```diff
@@ -0,0 +1,160 @@
+openapi: 3.0.3
+info:
+  title: Personalization API
+  version: 1.0.0-rc1
+  description: |-
+    The endpoints can be contacted at `t<name>.xaynet.dev/v1` or `t<name>.xaynet.dev/v1.0`.
```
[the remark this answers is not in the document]
acrrd: `t` for `tenant` but could also be different. I think it could be useful for us to have some kind of template in case we want to automatize some part of the domain.

discovery_engine_core/web-api/openapi.yaml
```diff
@@ -0,0 +1,160 @@
...
+    The `v1` path will always point to the latest `v1.x`.
+    
+    # Auth
+    To authenticate with the API two bearer tokens will be provided. One will work for the
+    personalization endpoints and the other one will allow documents to be added to the system.
```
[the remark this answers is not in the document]
acrrd: Thanks, I'll add that

discovery_engine_core/web-api/openapi.yaml
```diff
@@ -0,0 +1,160 @@
...
+          required: true
+          schema:
+            $ref: '#/components/schemas/UserId'
+        - name: count
+          in: query
+          description: Maximum number of personalized documents to return
+          type: integer
+          format: int32
+          minimum: 1
+          maximum: 100
```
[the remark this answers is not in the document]
acrrd: I used the editor to write this, not sure how I didn't notice. Thanks

discovery_engine_core/web-api/openapi.yaml
```diff
@@ -0,0 +1,160 @@
...
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/UserInteraction'
+  
+components:
+  schemas:
+    DocumentId:
+      type: string
+      example: "document_id_1"
```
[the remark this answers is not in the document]
acrrd: This does not want to be a format, is mainly to have the examples automatically have this value instead of `string`. 
Would `document_id` be better?

discovery_engine_core/web-api/openapi.yaml
```diff
@@ -0,0 +1,160 @@
...
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/PersonalizedDocuments'
+        '400':
+          description: invalid user id
+        '404':
+          description: user not found or the user does not have enough interactions
+
+  /users/{user_id}/documents/interacted:
```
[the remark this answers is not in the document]
acrrd: For `/users/{user_id}/documents/{document_id}/interaction`, there was a comment from Simon on this.
I'm going to modify this to accept multiple document_id.
I'll modify to `PATCH`, I wasn't aware that "If the Request-URI does not point to an existing resource, the serve MAY create a new resource. 

I also change to `interaction`. Should we do the same for `personalized`?

discovery_engine_core/web-api/openapi.yaml
```diff
@@ -0,0 +1,160 @@
...
+        score:
+          type: number
+      example:
+        - score: 0.872
+    PersonalizedDocuments:
+      type: object
+      additionalProperties:
+        type: object
+        $ref: '#/components/schemas/PersonalizedDocumentData'
+      example:
```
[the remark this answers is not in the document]
acrrd: They use it themself in the example they have for v3.0. I agree that is better to keep it till is supported. It will not be much work to change them.

discovery_engine_core/web-api/openapi.yaml
acrrd: I'm not sure why the fact that is a response should imply that is an object, but is a good idea. I'll go with
```js
{
  documents: [
    {
       id: "document_id",
       score: 0.1
     }
  ]
}
```
The list is not necessary ordered by score.

discovery_engine_core/web-api/openapi.yaml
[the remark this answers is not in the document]
acrrd: I left only `v1`

discovery_engine_core/web-api/openapi.yaml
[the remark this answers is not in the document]
acrrd: We don't plan to generate server from this. Also the auth will be handled by the nginx in front and not by the service directly

discovery_engine_core/web-api/openapi.yaml
acrrd: the document id is not part of the URI so I think is not correct to return 404 in that case.
Yes actually 404 does not make sense here, if the user does not exists we implicitly create one.

frankpepermans: Idd, but it's in the post, so then a 500 would make sense if it doesn't exist

acrrd:
> Same remark as I had on Slack

We don't have more information then when we designed this system. We design this to be as simple as possible and to work as a back-end system that is not exposed directly to the users. We knew from the beginning that an integration system was need to allow queries directly from Js if that would have been a requirement. We should have, as it was requested, learned more about what is possible at infrastructure and start from there.
Let's try to not panic for problem that are not there, also we know that we can easily solve them. The integration system can heavily cache data about the documents to reduce the latency or we can store some opaque data to return to the user.

acrrd:
I modified the interaction to allow to specify more documents and the kind of interaction (only positive for now)

discovery_engine_core/web-api/openapi.yaml
```diff
@@ -0,0 +1,189 @@
...
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/UserInteractionRequest'
+      responses:
+        '204':
+          description: successful operation
+        '400':
+          description: invalid request. User or document id is invalid
```
biernacki, line 85: maybe we should return more information here, like what document ids were invalid/not found, wdyt?

discovery_engine_core/web-api/openapi.yaml
```diff
@@ -0,0 +1,189 @@
...
+        
+  /documents/:
```
biernacki, line 87:
```suggestion
  /documents:
```

Robert-Steiner (commented):
just some doc string improvements 
rest looks good for me

discovery_engine_core/web-api/openapi.yaml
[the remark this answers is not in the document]
acrrd: Actually I decided to remove this part completely. Is not that relevant at the moment, we still need to decide which domain to use and I don't want to commit to information what we might change later and are not directly relevant for the api.

discovery_engine_core/web-api/openapi.yaml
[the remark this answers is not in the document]
acrrd: Removed.

discovery_engine_core/web-api/openapi.yaml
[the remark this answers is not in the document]
rustonaut: > The specification allows to have AND and OR for authentication.

not really 

> A declaration of which security mechanisms can be used for this operation. The list of values includes **alternative security requirement objects that can be used. Only one of the security requirement objects need to be satisfied** to authorize a request. To make security optional, an empty security requirement ({}) can be included in the array. This definition overrides any declared top-level [security](https://spec.openapis.org/oas/latest.html#oasSecurity). To remove a top-level security declaration, an empty array can be used.

Through you can could maybe define a custom security requirement object which is a combination of multiple requirements, but as far as I can tell the spec has no direct support for that (through if you use oauth2 you can require multiple scopes, but that's a different mechanism more or less only available for oath).

---

> I'm not sure I we really want to specify this here because we will be limited to authentication schema that the specification support and the service will not handle it.

If we change the auth schema it's anyway a braking change weather we define the auth as part of the schema or in the documentation.

All IANA registered http auth schemas (e.g. bearer token) are spec supported, so is a api-key in both header and query as well as extended support for mutualTLS and especially oauth2 and openIdConnect. (Through spec supported doesn't necessary mean generator supported). IMHO we shouldn't use any other auth schemas. But if we must we can just define it our self without too much problems (through it might or might not be spec conform dependent on what we speak about), or remove the security hints at this point in time.

The benefit from having the auth defined is that you can directly play around with the endpoints in the swagger editor as developer to get a feel/understanding for the API. 

Either way writing out the security schema in the spec which was already required isn't a braking change so we can always add it later.

---

> We don't plan to generate server from this. 

Through our clients might do so, and they might want to play around with the api through the swagger editor or similar OpenAPI support tooling.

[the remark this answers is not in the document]
acrrd: https://swagger.io/docs/specification/authentication/ under "Using Multiple Authentication Types"

discovery_engine_core/web-api/openapi.yaml
```diff
@@ -0,0 +1,198 @@
...
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/IngestionRequest'
+      responses:
+        '204':
+          description: successful operation
+        '400':
+          description: invalid request
+        '422':
```
acrrd, line 107: Should this be a 500 instead? Most probably is not a problem with the data that the user sent but something broke in the smbert model.
@rustonaut @biernacki @Robert-Steiner

discovery_engine_core/web-api/openapi.yaml
```diff
@@ -0,0 +1,189 @@
...
+        '404':
+          description: user not found.
+        '422':
+          description: impossible to create a personalized list for the user.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/PersonalizedDocumentsError'
+
+  /users/{user_id}/documents/interact:
```
rustonaut, line 60: :question: Can't we not just keep with REST naming/structuring conventions and use nouns and put the document id in the path?

It's a update for an interaction with an user specific document. Which both are resources/entities of which there is more then one.

Reading the current path from a REST POV implies that either there is a single interaction for all documents or a document with the id `interact`. Given that document ids are client chosen there could be a random string id which happens to be `"interact"` which could become a problem with other hypothetical features which are user->document specific we might have in the future.

discovery_engine_core/web-api/openapi.yaml
rustonaut: I mean it could be a problem with there data, like empty or single letter strings for snippets or similar.

If it's a bug that we thought there is a problem with a document it's still 422, as the bug is in the classification of mal-formedness not the reporting/handling of malformed documents.

If our models messes up it would be a bug and as such a 500 error code, but in this case we don't need to document it as 500 is pretty much implicitly always possible. 

Also I think I have seen some  request libraries which handle 500 errors different then 400 errors making conditional control-flow based on 500 errors hard. As such we should avoid them as long as we don't have a panic or error we believe is "bug-like" wrt. the full system (e.g. the database being unavailable is bug-like wrt. the system even through it's not a bug per-see)

discovery_engine_core/web-api/openapi.yaml
rustonaut: There are multiple reasons:

1. legacy reason: various security guidelines requires json response bodies to be objects due to some security vulnerability wrt. JS and old browsers. This doesn't apply anymore.
2. convention, due to 1. and 3. it had become a de-facto convention for many developers
3. and the main reason: extend-ability. We can always add fields to an object if we need to return additional information in the future making it a non-braking change. We can't do so with an array. This makes the "always return objects" approach much less prone to accidental braking changes.

discovery_engine_core/web-api/openapi.yaml
Robert-Steiner: > Most probably is not a problem with the data that the user sent but something broke in the smbert model.
If the problem is not the user data but something else in our service, we should return 500.

But as @rustonaut wrote, I'm not sure it's worth mentioning here

discovery_engine_core/web-api/openapi.yaml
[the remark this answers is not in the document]
rustonaut: oh, nice I misread that :sweat:

discovery_engine_core/web-api/openapi.yaml
rustonaut: > the document id is not part of the URI so I think is not correct to return 404 in that case.

through IMHO it should be part of the URI

discovery_engine_core/web-api/openapi.yaml
acrrd: As already said we can upload more the one document in the request.

We can have  `/users/{user_id}/interactions` instead.
And for the preferences we can have `/users/{user_id}/personalizedDocuments` as previously suggested by Simon.

I'm also ok with ditching the compliance with REST since it does not really cope well with non CRUD operations.

discovery_engine_core/web-api/openapi.yaml
acrrd: An empty strings will return a 400 we are returning that only if the embedding or the push to elastic search fails. I'll change it to 500

acrrd: I don't think that an smbert error is really expected.

discovery_engine_core/web-api/openapi.yaml
```diff
@@ -0,0 +1,160 @@
...
+        '200':
+          description: successful operation
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/PersonalizedDocuments'
+        '400':
+          description: invalid user id
+        '404':
+          description: user not found or the user does not have enough interactions
```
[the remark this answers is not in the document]
rustonaut: > It is what we were doing for the demo but I think is extremely misleading for the customer because they could think that is already personalized.

We can add a field `"personalized": bool` to the returned object (like `{ "personalized": false, "documents": [...] }`). (or similar)

While it still can be confusing, we probably want to avoid unnecessary requests round trips especially when a person first time opens the app (to avoid a first "this is sluggish" impression).  So we might need to continue having it.

discovery_engine_core/web-api/openapi.yaml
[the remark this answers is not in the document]
rustonaut: yes :+1:

discovery_engine_core/web-api/openapi.yaml
rustonaut: Sorry I didn't saw the previous comment as it didn't show up in the diff view and was collapsed in the non-diff few.

Yes I guess the path you proposed here are nice.

REST can work nice if you just formulate things in the right way, but it doesn't work well at all with batching, like updating interactions for multiple documents at once.

discovery_engine_core/web-api/openapi.yaml
rustonaut: What if the string is non empty but only contains character we remove when normalizing?

I think a good idea could be deserialization failing => 400 no documents are  processed; post-deserialization validation failing => 422 only the specific document fails.

As we probably still want to ingest other documents even if one or two of them are bad I would say empty strings should be post-deserialization validation and yield 422 :thinking:

discovery_engine_core/web-api/openapi.yaml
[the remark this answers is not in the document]
rustonaut: Yes, or  `"document id"` developers tend to not see text with white-space as id :joy:.

discovery_engine_core/web-api/openapi.yaml
rustonaut: Also internal server errors (500) are basically bugs. Besides basic robustness you normally don't write code which expects a bug. If we expect it to fail with some input then it's not a bug anymore and in turn not 500. If we don't expect it to fail its not necessary document or document that we return a list of failed documents.

biernacki: if that is the case then 500 works for me

discovery_engine_core/web-api/openapi.yaml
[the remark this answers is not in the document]
acrrd: I will leave it out for the moment. We can add it later without be a break change as you said

discovery_engine_core/web-api/openapi.yaml
acrrd: There is no normalization and there are no known inputs that will make smbert fail. If the tokenization or the run of the model fail is a bug in our code or in the libraries we are running.

PR #595, opened 2022-09-14
To 3274 store coi directly

biernacki:


discovery_engine_core/web-api/src/migrations/20220908090717_coi.sql
```diff
@@ -12,7 +12,12 @@
 --  You should have received a copy of the GNU Affero General Public License
 --  along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-CREATE TABLE IF NOT EXISTS user_state (
-    id TEXT NOT NULL PRIMARY KEY,
-    state BYTEA NOT NULL
+CREATE TABLE IF NOT EXISTS center_of_interest (
+    coi_id UUID NOT NULL PRIMARY KEY,
+    user_id TEXT NOT NULL,
```
rustonaut, line 17: :wrench:  Shouldn't this have  be a foreign key constraint? 

Preferable with `on delete cascade` so that when we delete the user it always automatically deletes all cois of the user.

biernacki: it's not a foreign key, because we don't store users as separate entities, the `user_id` here is just a key to filter the CoIs. Do we need to store users in a separate table? @acrrd

rustonaut: from daily: no user table.

:wrench: We need  an index on `user_id` to make sure filtering by `user_id` is efficient.

biernacki: added, please check :)

discovery_engine_core/web-api/src/storage.rs
```diff
@@ -44,37 +58,115 @@ impl UserState {
...
+                .execute(&mut *tx)
+                .await?;
+        }
+
+        for cois in negative_cois.chunks(BIND_LIMIT / 5) {
+            query_builder
+                .reset()
+                .push("center_of_interest (coi_id, user_id, is_positive, embedding, last_view) ")
+                .push_values(cois, |mut stm, coi| {
+                    let timestamp: DateTime<Utc> = coi.last_view.into();
```
rustonaut, line 152: :pushpin:  Given that we now anyway go through `DateTime<Utc>` it could make sense to use it as type for `last_view` in the future.

discovery_engine_core/web-api/src/storage.rs
```diff
@@ -83,16 +175,22 @@ impl UserState {
...
 
 #[derive(FromRow)]
-struct QueriedState {
-    state: Vec<u8>,
+struct QueriedCoi {
+    coi_id: Uuid,
+    is_positive: bool,
+    embedding: Vec<f32>,
+    view_count: i32,
+    view_time_ms: i32,
```
rustonaut, line 194:
```suggestion
    /// The count is a `u32` stored as `i32` in database
    view_count: i32,
    /// The time is a `u32` stored as `i32` in database
    view_time_ms: i32,
```

rustonaut (approved):

biernacki:
bors merge

PR #598, opened 2022-09-19
[Merged by Bors] - Add request_id to errors

acrrd:
Add field that we can use to return the generated request_id to be able to debug single requests

### DIFF
--- a/discovery_engine_core/web-api/openapi.yaml
+++ b/discovery_engine_core/web-api/openapi.yaml
@@ -145,6 +145,12 @@ components:
       description: A user id can be any non-empty UTF-8 string that does not contain the null byte.
       type: string
       example: "user id"
+    BaseError:
+      type: object
+      properties:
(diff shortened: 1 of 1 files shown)


discovery_engine_core/web-api/openapi.yaml
```diff
@@ -145,6 +145,12 @@ components:
       description: A user id can be any non-empty UTF-8 string that does not contain the null byte.
       type: string
       example: "user id"
+    BaseError:
+      type: object
```
Robert-Steiner, line 149: 🔧 if the id is not optional
```suggestion
      type: object
      required: [request_id]
```

acrrd: I didn't put it optional because I'm not sure we can always return it. Change it later from optional to required does not break the api.

Robert-Steiner (approved):

PR #599, opened 2022-09-19
[Merged by Bors] - By default enable storage for local builds/test/check.

rustonaut:
Still do not enable it for release builds.

Also still test for with and without the feature.

**Why?**

1. By now we should default to storage enabled by default for local testing.
2.  I recently ran to often into situations where I missed some merge caused problems
     which only where visible with `just check` creating additional friction.
     
**Why not check both?**

We do on CI. But locally the change->check/test loop is already WAY to long
doubling that time is IMHO just not acceptable at all.


biernacki (approved):
lgtm

rustonaut:
bors merge

PR #601, opened 2022-09-20
[Merged by Bors] - TO-3104 [2] migrate documents ffi parts only

rustonaut:
- ffi parts of migrating documents from hive in dart to sqlite in rust

**Based on:**

- #600

PR #600, opened 2022-09-20
[Merged by Bors] - TO-3104 [1] Split out migration of search state (wihout any documents).

rustonaut:
-  the current search but no documents
-  some small refactorings

PR #602, opened 2022-09-20
[Merged by Bors] - TO-3104 [3] migrate documents migration code

rustonaut:
- rust migration code for documents

**Based On:**

- #600 
- #601

### DIFF
--- a/discovery_engine_core/core/src/document.rs
+++ b/discovery_engine_core/core/src/document.rs
@@ -44,7 +44,9 @@ pub enum Error {
 }
 
 /// Unique identifier of the [`Document`].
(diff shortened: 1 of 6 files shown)


Robert-Steiner (approved):
lgtm 👍

rustonaut:
@Robert-Steiner I backported some bug fixes from later PRs  can you recheck the last commit I added?

Basically there where two problems:

1. migration failing if some of the arrays are empty as `push_values` doesn't automatically turn the query into a no-op, so no there are a bunch  `if !docs.is_empty()`.

2. I only  added reactions if the user did spend time on the document, but that isn't an option because we don't store view times for all document in the history... so I removed that.

Robert-Steiner (approved):

rustonaut:
bors merge

PR #604, opened 2022-09-20
[Merged by Bors] - TO-3292 adjust deep search to use mlt instead of kpe

finiteprods:
**Summary**

deep search still uses kpe on the search term, followed by a query to the news provider.
this adjusts the method to bypass kpe and use the similar news provider instead.


rustonaut (approved):
lgtm

finiteprods:
bors merge

PR #607, opened 2022-09-21
TO-3108 migrate source weights

rustonaut:
**based on:**

- #600
- #601
- #602


discovery_engine_core/core/src/storage/sqlite/dart_migrations.rs
```diff
@@ -392,6 +410,38 @@ mod tests {
             data.excluded_sources.into_iter().collect()
         );
 
+        let weight = storage
+            .source_reaction()
+            .fetch_source_reaction("source1")
+            .await
+            .unwrap();
+        assert_eq!(weight, Some(true));
+        // assert_eq!(weight, 1);
```
finiteprods, line 419: :thought_balloon: actually `fetch_source_reaction` could return the weight, and then it would be more useful here... (but not essential for this pr)

discovery_engine_core/core/src/storage/sqlite.rs
```diff
@@ -326,6 +326,35 @@ impl SqliteStorage {
...
+                count += 1;
+                query
+                    .push_bind(&item.source)
+                    .push_bind(item.weight)
+                    .push_bind(now);
+            })
+            .build()
+            // outside of migrations we always will call this with exactly one
+            // values tuple, there is no reason to not cache it in that case
+            .persistent(count == 1)
```
finiteprods, line 351: :+1:

finiteprods (commented):
just minor comments

discovery_engine_core/core/src/storage/sqlite/dart_migrations.rs
rustonaut: yes that was planed, it was just written in parallel with another PR by someone else so I originally planed to do so in a follow up PR. But I can now do it now. :+1:

discovery_engine_core/core/src/storage/sqlite.rs
```diff
@@ -872,13 +901,16 @@ impl FeedbackScope for SqliteStorage {
...
-                self.update_source_weight(source, liked.into()).await
+        match self.fetch_source_weight(source).await? {
+            0 => self.create_source_reaction(source, liked).await?,
+            weight if (weight > 0) == liked => {
+                if liked {
+                    self.update_source_weight(source, 1).await?;
+                }
             }
-            _ => self.delete_source_reaction(source).await,
+            _ => self.delete_source_reaction(source).await?,
```
rustonaut, line 911: :thought_balloon: While this is the correct business logic I wonder if we should change it.

Currently it's:

- if your reactions differ from the previous reaction reset it to no reaction (0)
- if the reaction is positive add 1 (first reaction is 1)
- if the reaction is negative do nothing, except for the first reaction where it's set to -1.

Wouldn't it be more useful to also count negative reactions?

And it setting reaction to 0/delete once it differs once really the right approach?

rustonaut: In the last commit I did unify `create_source_reaction` and `update_source_weight`.

finiteprods: >if the reaction is negative do nothing, except for the first reaction where it's set to -1.

:wrench:  it's not quite exactly do nothing - `lastUpdated` should still be refreshed

>Wouldn't it be more useful to also count negative reactions?
>
>And it setting reaction to 0/delete once it differs once really the right approach? 

I don't see a problem with tracking the negative reactions but perhaps it's a bit out of scope of this pr and we should consider it separately? Similarly with deleting a "reset reaction" - I'd prefer to keep the logic as is here, and later double-check with ex-pink re changes to the logic. whether it's the _right_ approach - I had a similar question when first implementing it but that was the heuristic: see [here](https://xainag.atlassian.net/wiki/spaces/M2D/pages/2389639169/News+Source+Ranking#Updating-the-Source-Database%3A) 2nd point.

rustonaut: >  it's not quite exactly do nothing - lastUpdated should still be refreshed

yes :+1: 

> Similarly with deleting a "reset reaction"

deleting on "reset" it's not new logic and its already in the code base before this PR and I think was this way in dart, too

but it does eliminate the `lastUpdated`, through we don't use the `lastUpdated` currently at all :thinking:

rustonaut: @finiteprods

finiteprods: >deleting on "reset" it's not new logic and its already in the code base

yes, correct - what I meant was we should keep this logic too, so all good.

>we don't use the `lastUpdated` currently at all

true. this was used in the python version to purge older sources but wasn't implemented here. I added a ticket [TO-3387](https://xainag.atlassian.net/browse/TO-3387) to reinvestigate this and other things mentioned above.

finiteprods (approved):
lgtm

rustonaut:
borse merge

rustonaut:
bors merge

rustonaut:
bors merge

PR #612, opened 2022-09-26
[Merged by Bors] - more general coi system

acrrd:
This PR:
* allows using an arbitrary type for the id in the coi system
* it removed `published_date` field from the trait `Document`. To do this now the CoI system does not handle the ranking anymore it only returns the score. Ranking is done in a util function that can be used in both the web-api and the engine. In the engine, we handle the case where we use the published date as default when we don't have enough coi.

The PR is divided in two commits, one for each point above.

### DIFF
--- a/discovery_engine_core/ai/ai/src/document.rs
+++ b/discovery_engine_core/ai/ai/src/document.rs
@@ -30,8 +30,10 @@ impl From<Uuid> for DocumentId {
 
 /// Common document properties.
 pub trait Document {
+    type Id: std::fmt::Display + std::fmt::Debug + Eq + std::hash::Hash;
+
     /// Gets the document id.
-    fn id(&self) -> DocumentId;
+    fn id(&self) -> Self::Id;
 
     /// Gets the `SMBert` embedding of the document.
     fn smbert_embedding(&self) -> &Embedding;
@@ -72,7 +74,9 @@ pub(crate) mod tests {
     }
 
     impl Document for TestDocument {
-        fn id(&self) -> DocumentId {
+        type Id = DocumentId;
+
+        fn id(&self) -> Self::Id {
             self.id
         }
 
--- a/discovery_engine_core/ai/ai/src/document.rs
+++ b/discovery_engine_core/ai/ai/src/document.rs
@@ -12,7 +12,6 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-use chrono::{DateTime, Utc};
 use derive_more::Display;
 use uuid::Uuid;
 
@@ -37,9 +36,6 @@ pub trait Document {
 
     /// Gets the `SMBert` embedding of the document.
     fn smbert_embedding(&self) -> &Embedding;
-
-    /// Gets the publishing date.
-    fn date_published(&self) -> DateTime<Utc>;
 }
 
 #[cfg(test)]
@@ -56,19 +52,13 @@ pub(crate) mod tests {
     pub(crate) struct TestDocument {
         pub(crate) id: DocumentId,
         pub(crate) smbert_embedding: Embedding,
-        pub(crate) date_published: DateTime<Utc>,
     }
 
     impl TestDocument {
-        pub(crate) fn new(
-            id: u128,
-            embedding: impl Into<Embedding>,
-            date_published: DateTime<Utc>,
-        ) -> Self {
+        pub(crate) fn new(id: u128, embedding: impl Into<Embedding>) -> Self {
             Self {
                 id: DocumentId::from_u128(id),
                 smbert_embedding: embedding.into(),
-                date_published,
             }
         }
     }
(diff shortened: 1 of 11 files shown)


discovery_engine_core/ai/ai/src/utils.rs
```diff
@@ -73,6 +75,20 @@ pub(crate) fn system_time_now() -> SystemTime {
...
 }
 
+/// Rank documents based on their score.
+///
+/// # Panics
+/// This function will panic if a score for a document is missing.
+pub fn rank<D, S>(documents: &mut [D], scores: &HashMap<D::Id, f32, S>)
+where
+    D: Document,
+    S: ::std::hash::BuildHasher,
```
Robert-Steiner, line 85: 🎨 
```suggestion
    S: std::hash::BuildHasher,
```

or we could also add `hash::BuildHasher` to the `use` above and just use `BuildHasher` here

discovery_engine_core/ai/ai/src/document.rs
```diff
@@ -30,14 +29,13 @@ impl From<Uuid> for DocumentId {
 
 /// Common document properties.
 pub trait Document {
+    type Id: std::fmt::Display + std::fmt::Debug + Eq + std::hash::Hash;
```
Robert-Steiner, line 32: 🎨 
```suggestion
    type Id: std::fmt::Display + Debug + Eq + Hash;
```
+ adding `use std::{fmt::Debug, hash::Hash};` above

discovery_engine_core/core/src/engine.rs
```diff
@@ -1426,6 +1429,39 @@ impl Engine {
...
+    } else {
+        documents.sort_unstable_by(default_ord);
+    }
+}
+
+fn rank_documents(coi: &CoiSystem, user_interests: &UserInterests, documents: &mut [Document]) {
+    rank(
+        coi,
+        user_interests,
+        |a, b| a.resource.date_published.cmp(&b.resource.date_published),
```
Robert-Steiner, line 1452: ❔ I think the should be reversed here, if I haven't overlooked anything.
as in the old rank function:
```rust
            documents
                .sort_unstable_by(|this, other| other.date_published().cmp(&this.date_published()));
```

Robert-Steiner (commented):
One question regarding the comparison of the dates + `import` cleanup (feel free to ignore them)

discovery_engine_core/ai/ai/src/utils.rs
acrrd: I'm becoming less of a fan of polluting the global namespace with things that are super-specific and local.

discovery_engine_core/core/src/engine.rs
acrrd: Yes, thanks. I added a test that checks that we are returning the documents in the correct order when we don't rank using the coi.

acrrd:
I also change `id()` to return `&Self::id()` to work with non copy types.

Robert-Steiner (approved):

discovery_engine_core/ai/ai/src/utils.rs
```diff
@@ -73,6 +75,20 @@ pub(crate) fn system_time_now() -> SystemTime {
...
+///
+/// # Panics
+/// This function will panic if a score for a document is missing.
+pub fn rank<D, S>(documents: &mut [D], scores: &HashMap<D::Id, f32, S>)
+where
+    D: Document,
+    S: std::hash::BuildHasher,
+{
+    documents.sort_unstable_by(|a, b| {
+        nan_safe_f32_cmp_desc(scores.get(a.id()).unwrap(), scores.get(b.id()).unwrap())
```
janpetschexain, line 88: 🔧 if we unwrap anyways, we could just use index notation
```suggestion
        nan_safe_f32_cmp_desc(&scores[a.id()], &scores[b.id()])
```

discovery_engine_core/ai/ai/src/utils.rs
```diff
@@ -73,6 +75,20 @@ pub(crate) fn system_time_now() -> SystemTime {
...
 }
 
+/// Rank documents based on their score.
+///
+/// # Panics
+/// This function will panic if a score for a document is missing.
+pub fn rank<D, S>(documents: &mut [D], scores: &HashMap<D::Id, f32, S>)
+where
+    D: Document,
+    S: std::hash::BuildHasher,
```
janpetschexain, line 85: ❓ what's the reason to not use the default for `S`?
```suggestion
pub fn rank<D>(documents: &mut [D], scores: &HashMap<D::Id, f32>)
where
    D: Document,
```

acrrd:
bors merge

acrrd:
bors cancel

acrrd:
bors merge

discovery_engine_core/ai/ai/src/utils.rs
acrrd: It will forbid to use other hashes, and clippy complaints.

discovery_engine_core/ai/ai/src/utils.rs
acrrd: It will only make more obscure that we can panic here and more difficult to look for functions that can panic. Clippy don't even worn that the panic docs is missing.

acrrd:
bors merge

PR #614, opened 2022-09-26
Store embedding with its own type

acrrd:
Avoid copying embedding to a vec by storing an embedding directly.


discovery_engine_core/web-api/src/bin/ingestion.rs
```diff
@@ -73,7 +73,7 @@ struct Article {
     published_date: DateTime<Utc>,
 
     #[serde(skip_deserializing)]
-    embedding: Vec<f32>,
+    embedding: Embedding,
```
biernacki, line 76: I don't know why I though `Embedding` isn't (de)serializable.. 🤦 anyways, it's a good change, but could we add it to #620 to avoid merge conflicts?

acrrd: Sure :)

acrrd: When we talked about it I also didn't think it was possible.

acrrd:
Will be handled in #620.

PR #615, opened 2022-09-26
TO-3325 save interacted documents

janpetschexain:
**References**

- [TO-3325]
- requires #616

**Summary**

- store interacted document and user ids in a table via a n:n relation

[TO-3325]: https://xainag.atlassian.net/browse/TO-3325?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine_core/web-api/src/handlers.rs
+++ b/discovery_engine_core/web-api/src/handlers.rs
@@ -52,7 +52,7 @@ pub(crate) async fn handle_user_interaction(
 ) -> Result<impl warp::Reply, Rejection> {
     if let Some(document) = db.documents_by_id.get(&body.document_id) {
         db.user_state
-            .update_positive_cois(&user_id, |positive_cois| {
+            .update_positive_cois(&body.document_id, &user_id, |positive_cois| {
                 db.coi
                     .log_positive_user_reaction(positive_cois, &document.smbert_embedding)
             })
--- a/discovery_engine_core/web-api/src/handlers.rs
+++ b/discovery_engine_core/web-api/src/handlers.rs
@@ -52,7 +52,7 @@ pub(crate) async fn handle_user_interaction(
 ) -> Result<impl warp::Reply, Rejection> {
     if let Some(document) = db.documents_by_id.get(&body.document_id) {
         db.user_state
-            .update_positive_cois(&body.document_id, &user_id, |positive_cois| {
+            .update_positive_cois(&document.id, &user_id, |positive_cois| {
                 db.coi
                     .log_positive_user_reaction(positive_cois, &document.smbert_embedding)
             })
--- a/discovery_engine_core/web-api/src/migrations/20220926145400_document.sql
+++ b/discovery_engine_core/web-api/src/migrations/20220926145400_document.sql
@@ -17,3 +17,6 @@ CREATE TABLE IF NOT EXISTS document (
     user_id TEXT NOT NULL,
     PRIMARY KEY (doc_id, user_id)
 );
+
+CREATE INDEX IF NOT EXISTS idx_doc_by_user_id
+    ON document(user_id);
--- a/discovery_engine_core/web-api/src/migrations/20220926145400_document.sql
+++ b/discovery_engine_core/web-api/src/migrations/20220926145400_document.sql
@@ -12,12 +12,12 @@
 --  You should have received a copy of the GNU Affero General Public License
 --  along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-CREATE TABLE IF NOT EXISTS document (
+CREATE TABLE IF NOT EXISTS interaction (
     doc_id TEXT NOT NULL,
     user_id TEXT NOT NULL,
-    last_view TIMESTAMPTZ NOT NULL DEFAULT Now(),
-    user_reaction SMALLINT NOT NULL DEFAULT 0,
-    PRIMARY KEY (doc_id, user_id)
+    time_stamp TIMESTAMPTZ NOT NULL DEFAULT Now(),
+    user_reaction SMALLINT NOT NULL,
+    PRIMARY KEY (doc_id, user_id, time_stamp)
 );
 
 CREATE INDEX IF NOT EXISTS idx_doc_by_user_id
--- a/discovery_engine_core/web-api/src/migrations/20220926145400_document.sql
+++ b/discovery_engine_core/web-api/src/migrations/20220926145400_document.sql
@@ -1,0 +1,19 @@
+--  Copyright 2022 Xayn AG
+--
+--  This program is free software: you can redistribute it and/or modify
+--  it under the terms of the GNU Affero General Public License as
+--  published by the Free Software Foundation, version 3.
+--
+--  This program is distributed in the hope that it will be useful,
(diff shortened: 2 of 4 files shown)


discovery_engine_core/web-api/src/migrations/20220926145400_document.sql
```diff
@@ -0,0 +1,24 @@
...
+--  GNU Affero General Public License for more details.
+--
+--  You should have received a copy of the GNU Affero General Public License
+--  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+CREATE TABLE IF NOT EXISTS document (
+    doc_id TEXT NOT NULL,
+    user_id TEXT NOT NULL,
+    last_view TIMESTAMPTZ NOT NULL DEFAULT Now(),
+    user_reaction SMALLINT NOT NULL DEFAULT 0,
```
biernacki, line 19: I wasn't aware that we have `UserReaction` enum, so we should create a ticket for using that too in `center_of_interest` table instead of a boolean, wdyt @acrrd?

acrrd: On elastic search? Yes please. Actually we need to define a new `UserReaction` in the web-api because the one we have has `neutral` that we don't want to expose at the moment.

@janpetschexain this also means that we cannot have a default for it.

discovery_engine_core/web-api/src/migrations/20220926145400_document.sql
```diff
@@ -0,0 +1,24 @@
...
+--  but WITHOUT ANY WARRANTY; without even the implied warranty of
+--  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+--  GNU Affero General Public License for more details.
+--
+--  You should have received a copy of the GNU Affero General Public License
+--  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+CREATE TABLE IF NOT EXISTS document (
+    doc_id TEXT NOT NULL,
+    user_id TEXT NOT NULL,
```
biernacki, line 17: the ticket states
> We keep an association between the user and the document on postgres.
does it mean that we wanted to add a `user` table and FK to other tables? @acrrd

discovery_engine_core/web-api/src/migrations/20220926145400_document.sql
```diff
@@ -0,0 +1,24 @@
...
+--
+--  This program is distributed in the hope that it will be useful,
+--  but WITHOUT ANY WARRANTY; without even the implied warranty of
+--  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+--  GNU Affero General Public License for more details.
+--
+--  You should have received a copy of the GNU Affero General Public License
+--  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+CREATE TABLE IF NOT EXISTS document (
```
biernacki, line 15: it's not really a document, maybe a `document_interaction(_log)` might be a better name? wdyt?

discovery_engine_core/web-api/src/migrations/20220926145400_document.sql
acrrd: We want `doc_id, user_id, timestamp` as the primary key otherwise we lose some interaction.

discovery_engine_core/web-api/src/migrations/20220926145400_document.sql
acrrd, line 15: This table does not contain documents, maybe `interaction` instead?

discovery_engine_core/web-api/src/handlers.rs
```diff
@@ -50,7 +50,7 @@ pub(crate) async fn handle_user_interaction(
 ) -> Result<impl warp::Reply, Rejection> {
     if let Some(document) = db.documents_by_id.get(&body.document_id) {
         db.user_state
-            .update_positive_cois(&user_id, |positive_cois| {
+            .update_positive_cois(&body.document_id, &user_id, |positive_cois| {
```
biernacki, line 53: 🎨  you could also use `document.id`

biernacki (commented):
looks nice 👍 just couple of questions and some suggestions

discovery_engine_core/web-api/src/migrations/20220926145400_document.sql
biernacki: >On elastic search? Yes please.
no, I was referring to a `center_of_interest` Postgres table, where we store kind of interaction as a boolean

janpetschexain: we could also use a boolean here for now and decide for some enum later on

discovery_engine_core/web-api/src/migrations/20220926145400_document.sql
acrrd: with don't have a table of user to reference

discovery_engine_core/web-api/src/migrations/20220926145400_document.sql
acrrd: ah sorry. There matters less. A coi can only be positive or negative I don't see much room for extensions there.

discovery_engine_core/web-api/src/migrations/20220926145400_document.sql
janpetschexain: i'm fine with any suggested name, i'll change it to `interaction` for now

discovery_engine_core/web-api/src/migrations/20220926145400_document.sql
biernacki: is postgres is storing boolean as a `0` and `1` internally and we can avoid migration if we change the type then this could work too.. right now we're not in production so it would be better to use a type that is more future-proof

biernacki: > A coi can only be positive or negative

ah, right, good point

discovery_engine_core/web-api/src/migrations/20220926145400_document.sql
biernacki: not yet

discovery_engine_core/web-api/src/migrations/20220926145400_document.sql
janpetschexain: - added `time_stamp` to the PK
- added `UserReaction` wrapper & removed default for `user_reaction`

discovery_engine_core/web-api/src/handlers.rs
janpetschexain: i rebased this onto #616 and applied the change, before this was still a uuid

discovery_engine_core/web-api/src/migrations/20220926145400_document.sql
acrrd: still, we cannot create an FK to a table that does not exist :)
for the future, I don't know. Maybe it will pop up.

PR #617, opened 2022-09-26
[Merged by Bors] - TO-3276 remove kpe assets

janpetschexain:
**References**

- [TO-3276]

**Summary**

- remove KPE assets from the asset manifest, the assets are still in S3 as they are needed for the KPE crate tests
- remove the KPE system from the engine, as it can't be initialized without the assets

[TO-3276]: https://xainag.atlassian.net/browse/TO-3276?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rustonaut (approved):
lgtm

rustonaut:
Might also need to be removed from S3 I don't know if that is planed or if it should stay there for now.

janpetschexain:
> Might also need to be removed from S3 I don't know if that is planed or if it should stay there for now.

iirc in one of the previous plannings we said to keep them for now. we also need them to run tests for the `kpe` crate, those tests need to be disabled if we remove the assets from the bucket completely.

janpetschexain:
bors merge

PR #618, opened 2022-09-27
TO-3318 return request id on error

rustonaut:
Adds a random per-request uuid to all log messages and return it in case of an error.

There will be an follow up PR to improve/fix some other aspects of the error handling.

With the refactoring needed for [TO-3305] this will be made available for all endpoints instead of just the ingestion endpoint.

**References:**
- [TO-3318]

[TO-3318]: https://xainag.atlassian.net/browse/TO-3318?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[TO-3305]: https://xainag.atlassian.net/browse/TO-3305?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine_core/web-api/src/bin/ingestion.rs
+++ b/discovery_engine_core/web-api/src/bin/ingestion.rs
@@ -177,6 +177,9 @@ macro_rules! handle4 {
             $handle(t1, t2, t3, t4)
                 .instrument(span)
                 .await
+                //FIXME handle our errors here, and then handle rejection separately
+                // by turning them into Reponses and then grabbing the status code
+                // and the text body from the response to create our response.
                 .or_else(|err| handle_rejection(err, id))
         }
     };
--- a/discovery_engine_core/web-api/src/bin/ingestion.rs
+++ b/discovery_engine_core/web-api/src/bin/ingestion.rs
@@ -357,7 +357,7 @@ async fn handle_rejection(err: Rejection) -> Result<impl Reply, Infallible> {
     // Requires the request to be wrapped into a
     // span with a `log.id` field.
     let log_id = Span::current()
-        .field("log.id")
+        .field("log_id")
         .map(|f| f.to_string())
         .unwrap_or_default();
 
@@ -379,9 +379,9 @@ fn trace_requests_with_random_id(request_info: warp::trace::Info<'_>) -> Span {
         path = %request_info.path(),
         version = ?request_info.version(),
         // Linked to rejection handling which will include
-        // the spans "log.id" field in error response bodies.
+        // the spans "log_id" field in error response bodies.
         //FIXME use v6/v7 once supported/standardized
-        log.id = %Uuid::new_v4(),
+        log_id = %Uuid::new_v4(),
     );
 
     //keep same wording as in `warp::trace::request()` for forward/backward compatibility
--- a/discovery_engine_core/web-api/src/bin/ingestion.rs
+++ b/discovery_engine_core/web-api/src/bin/ingestion.rs
@@ -24,7 +24,7 @@ use reqwest::{
 };
 use serde::{de, Deserialize, Deserializer, Serialize};
 use std::{collections::HashMap, convert::Infallible, env, path::PathBuf, sync::Arc};
-use tracing::{debug, error, info, span, Level, Span};
+use tracing::{debug, error, info, span, Instrument, Level, Span};
 use tracing_subscriber::fmt::format::FmtSpan;
 use uuid::Uuid;
 use warp::{self, hyper::StatusCode, reject::Reject, Filter, Rejection, Reply};
@@ -145,9 +145,7 @@ async fn main() -> Result<(), GenericError> {
     let client = Client::new();
     let model = init_model(&config)?;
 
-    let routes = post_add_data(config, model, client)
-        .recover(handle_rejection)
-        .with(warp::trace::trace(trace_requests_with_random_id));
+    let routes = post_add_data(config, model, client);
 
     warp::serve(routes).run(([127, 0, 0, 1], 3030)).await;
 
@@ -172,6 +170,18 @@ fn init_model(config: &Config) -> Result<Model, GenericError> {
     Ok(Arc::new(smbert))
 }
 
+macro_rules! handle4 {
+    ($handle:ident) => {
+        |t1, t2, t3, t4| async move {
+            let (span, id) = setup_request_span(stringify!($handle));
+            $handle(t1, t2, t3, t4)
+                .instrument(span)
+                .await
+                .or_else(|err| handle_rejection(err, id))
+        }
+    };
+}
+
 // POST /add_data
 fn post_add_data(
     config: Config,
@@ -185,7 +195,7 @@ fn post_add_data(
         .and(with_model(model))
         .and(with_config(config))
         .and(with_client(client))
-        .and_then(handle_add_data)
+        .and_then(handle4!(handle_add_data))
 }
 
 async fn handle_add_data(
@@ -193,7 +203,8 @@ async fn handle_add_data(
     model: Model,
     config: Config,
     client: Client,
-) -> Result<impl warp::Reply, Rejection> {
+    //FIXME use custom error type to allow better error handling
+) -> Result<Box<dyn Reply>, Rejection> {
     if body.documents.len() > config.max_documents_length {
         return Err(warp::reject::custom(TooManyDocumentsError));
     }
@@ -249,7 +260,7 @@ async fn handle_add_data(
             warp::reject::custom(ReceivingOpError)
         })?;
 
-    Ok(StatusCode::OK)
+    Ok(Box::new(StatusCode::OK))
 }
 
 fn with_config(config: Config) -> impl Filter<Extract = (Config,), Error = Infallible> + Clone {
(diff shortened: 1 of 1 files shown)


discovery_engine_core/web-api/src/bin/ingestion.rs
```diff
@@ -170,6 +170,18 @@ fn init_model(config: &Config) -> Result<Model, GenericError> {
...
+    ($handle:ident) => {
+        |t1, t2, t3, t4| async move {
+            let (span, id) = setup_request_span(stringify!($handle));
+            $handle(t1, t2, t3, t4)
+                .instrument(span)
+                .await
+                .or_else(|err| handle_rejection(err, id))
+        }
+    };
+}
```
rustonaut, line 183: This hackery is necessary, it also means that depending on the number of parameters we will also need handle1,2,3,4,5 etc.

The reason this is necessary is due to serve!! limitations of warp:

1. `warp::trace::*` can only access a small number of fixed information, not including any parameters, extensions and similar
2. `filter.recover` can only depend on the rejection, there is no way to inject the log id here
3. due to non-exposed pub traits in warp you can not use the abstraction they use to allow functions with various parameter, elimination of () etc.
4. due to limitations of the rust type inference you can't make this a function instead of a macro
5. due to limitations of warp you can't create wrappers or filter from scratch, you can only create them through existing combinators which is not use-full if the existing combinators don't provide what you need
6. once you have a "Rejection" error result (e.g. because of an extractor of a parameter) you can't go back to having a custom error due to limitations in warp
7. you can't properly access hyper;s extension system nor can you include tower layers into warp due to limitations in it's design
8. the return reply of all endpoint handles and `handle_rejections` must be `Box<dyn Reply>` as due to limitations of warp you can't have an either type (and the either type warp has internally is not exposed). Also you can't do nicely the boxing with a `map` on the handle/handle_reject calls due to rust limitations of impl trait.
9. There was an initial bug where any error from middle ware (e.g. wrong content type) got turned into internal server error.
10. :pushpin: Now this is fixed but there is an error that such errors return a string instead of json body
11. It is not trivial to fix this due to what `Rejection` doesn't expose to consumers of the library
11. :pushpin: there is a bug where additional path segments might be accepted due to missing `path::end()` which is really easy to forget
12. While we now can have the `log_id` we in turn can no longer access the path, http method or http version, I used the handlers function name as a work around.

acrrd: This seems horrible enough to make me reconsider the use of warp. Are we missing something?

discovery_engine_core/web-api/src/bin/ingestion.rs
```diff
@@ -354,8 +371,25 @@ async fn handle_rejection(err: Rejection) -> Result<impl Reply, Infallible> {
...
+
+fn setup_request_span(handler_name: &str) -> (Span, Uuid) {
+    //FIXME use v6/v7 once supported/standardized
+    let log_id = Uuid::new_v4();
+    let span = span!(
+        Level::INFO,
+        "request",
+        method = %handler_name,
+        log_id = %log_id,
```
acrrd, line 388: This should be `request_id`.

discovery_engine_core/web-api/src/bin/ingestion.rs
```diff
@@ -354,8 +371,25 @@ async fn handle_rejection(err: Rejection) -> Result<impl Reply, Infallible> {
...
+        log_id,
         errored_ids,
     });
 
-    Ok(warp::reply::with_status(json, code))
+    Ok(Box::new(warp::reply::with_status(json, code)))
+}
+
+fn setup_request_span(handler_name: &str) -> (Span, Uuid) {
+    //FIXME use v6/v7 once supported/standardized
```
acrrd, line 382: I don't think is particularly useful to have a fixme that we have no clue when it could be "fixed" and will bring none to minimal benefits. If we want something that has lex order we can use ulid and in any case that will not exempt us from having a timestamp in a more computer and human readable form.

discovery_engine_core/web-api/src/bin/ingestion.rs
```diff
@@ -354,8 +371,25 @@ async fn handle_rejection(err: Rejection) -> Result<impl Reply, Infallible> {
...
+    let log_id = Uuid::new_v4();
+    let span = span!(
+        Level::INFO,
+        "request",
+        method = %handler_name,
+        log_id = %log_id,
+    );
+
+    //keep same wording as in `warp::trace::request()` for forward/backward compatibility
```
acrrd, line 391: :grey_question: what does this means exactly?

acrrd (commented):
Why do we need the refactoring to make this available to other endpoints that cannot be done here? Seems that the macro and the setup function are easily sharable and we need to change the returns and the routes of the user API to sue the boxed reply.
TO-3318 cannot be considered closed with only the ingestion.

discovery_engine_core/web-api/src/bin/ingestion.rs
janpetschexain: i might have found a solution to this problem, pls have a look if it fits your needs for the follow up prs:
- the instrumentation can be done via the attribute macro, such that we don't have to manage the span manually
- the `handle_rejection()` can be combined via an `or_else()` with the previous filter, such that there is no need for argument-size-dependent macros (the success result type has to be wrapped in a 1-tuple to satisfy the trait bounds)
- the `log_id` is then recorded as a field to the request handler and passed to the rejection handler, overall it looks like:

```rust
// POST /add_data
#[instrument(
    skip_all,
    level = "info",
    name = "request",
    fields(method = "handle_add_data", log_id)
)]
fn post_add_data(
    config: Config,
    model: Model,
    client: Client,
) -> impl Filter<Extract = impl Reply, Error = Rejection> + Clone {
    let log_id = Uuid::new_v4();
    Span::current().record("log_id", log_id.to_string());
    debug!("received request");

    warp::path("add_data")
        .and(warp::post())
        .and(warp::body::content_length_limit(config.max_body_size))
        .and(warp::body::json())
        .and(with_model(model))
        .and(with_config(config))
        .and(with_client(client))
        .and_then(handle_add_data)
        .or_else(move |err| handle_rejection(err, log_id))
}
```
where
```rust
async fn handle_rejection(err: Rejection, log_id: Uuid) -> Result<(Box<dyn Reply>,), Rejection> {
    // ... body ...
    Ok((Box::new(warp::reply::with_status(json, code)),))
}
```

wdyt, @rustonaut & @acrrd ?

acrrd: This will add a bit of boilerplate in each endpoint but is nicer then the macro.

rustonaut: this does not work, we need a new uuid/spawn for each request this creates a single uuid/spawn when creating route mapping.

discovery_engine_core/web-api/src/bin/ingestion.rs
rustonaut: I was a bit worried about the character limit some logging systems randomly impose on logging messages in various ways and places, but I can change it.

discovery_engine_core/web-api/src/bin/ingestion.rs
rustonaut: A leftover from when I had it mirror warp::trace::request as much as possible.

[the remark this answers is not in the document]
rustonaut: it seems to not show up under that path in the logs, but it's the path which was used in the code :thinking: 

```rust
#[doc(hidden)]
pub use self::filters::{
```

ah, that explains it :upside_down_face:  (in warp/src/lib.rs)

rustonaut:
> Why do we need the refactoring to make this available to other endpoints that cannot be done here? 

Because currently the crate has ingestion and non ingestion written as two totally independent code bases.

You have:

- `src/bin/ingestion.rs` which doesn't depend on any `xayn-web-api`
- and `src/main.rs` which has no an tracing setup  (or at least until very recently didn't head) 
- no `src/lib.rs`

The refactoring will introduce a `lib.rs` and then make the ingestion and other endoint implementations depend on it using the same generic setup code for loading config and setting up tracing for both (through potentially different config structs). Same for utility functions like the macro.

Then we can have `main.rs` and `bin/ingestion.rs` depend on that lib.

Or not have `main.rs` and instead have `bin/personalization.rs`.

discovery_engine_core/web-api/src/bin/ingestion.rs
rustonaut: Like or else would need to  accept a `|err, uuid| {}` function. This it doesn't. So I thought to make it accept an `|(err, uuid)|` and make this the error type, but you can't go from having a `Rejection` to a custom error. There is no `map_err` and `or_else` is restricted to returning `IsReject` which can't be implemented by custom types it only allows `Infallible` instead of `Rejection` (it's sealed, also I think `pub` but not `pub` exported).

biernacki:
We added `src/lib.rs` in #620 => https://github.com/xaynetwork/xayn_discovery_engine/pull/620/files#diff-7d4e5d7022f44784472254ce860753f40f762c69ecf3449ba013c6ee60e6988c

discovery_engine_core/web-api/src/bin/ingestion.rs
rustonaut: There is also the problem error handling via `recover` or `or_else` requires you to know every specific error type which an rejection might be created from to allow an reliable consistent error return value.

Not hard for our code as we can just have a custom error we defined and use that and only turn that into `Rejection`, but a problem because we also have to know that about every error type of every middle ware. I.e. without the `handle` macro we turn "wrong content type" or even 404 errors into internal server errors :face_exhaling:

rustonaut: hAlso in case we continue to use warp:

`warp::path("add_data")` needs to be `warp::path("add_data").and(warp::path::end())` or `warp::path!("add_data")`.

Here `warp::path()` is a hidden reexport of `warp::path::path()` and not quite the same as `warp::path!()` which auto includes an `end` but doesn't allow combining it with known path parts (as we use it for user).

Not including `end` means it will match `/add_data` and any `/add_data/*`.

PR #622, opened 2022-09-27
[Merged by Bors] - Decode user id in path only in route

acrrd:
The entity `UserId` can be constructed from any string that is non-empty and does not contain the null byte.

This PR move the decoding of the URL parameter in the route definition.

### DIFF
--- a/discovery_engine_core/web-api/src/routes.rs
+++ b/discovery_engine_core/web-api/src/routes.rs
@@ -15,7 +15,11 @@
 use std::convert::Infallible;
 use warp::{self, Filter, Rejection, Reply};
 
-use crate::{db::Db, handlers, models::UserId};
+use crate::{
+    db::Db,
+    handlers,
+    models::{Error, UserId},
+};
 
 pub(crate) fn api_routes(db: Db) -> impl Filter<Extract = impl Reply, Error = Rejection> + Clone {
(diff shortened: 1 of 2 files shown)


Robert-Steiner (approved):
👍

biernacki (approved):
good change 👍

discovery_engine_core/web-api/src/routes.rs
```diff
@@ -53,7 +57,14 @@ fn delete_internal_state(db: Db) -> impl Filter<Extract = impl Reply, Error = Re
...
 // PATH /user/:user_id
 fn user_path() -> impl Filter<Extract = (UserId,), Error = Rejection> + Clone {
-    warp::path("user").and(warp::path::param::<UserId>())
+    warp::path("user")
+        .and(warp::path::param::<String>())
+        .and_then(|user_id: String| async move {
+            urlencoding::decode(&user_id)
+                .map_err(Error::UserIdUtf8Conversion)
+                .and_then(UserId::new)
+                .map_err(|_| warp::reject())
```
rustonaut, line 66: This produces a `404` instead of a `400`.

acrrd: I changed this to return the error from `UserId::new`. That can be handled correctly in the task we have for the errors.

PR #623, opened 2022-09-28
[Merged by Bors] - TO-3326 don't return interacted documents

janpetschexain:
**References**

- [TO-3326]

**Summary**

- fix index naming
- add getter for interacted doc ids in postgres
- add filter for excluded doc ids in elastic

[TO-3326]: https://xainag.atlassian.net/browse/TO-3326?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine_core/web-api/src/migrations/20220926145400_document.sql
+++ b/discovery_engine_core/web-api/src/migrations/20220926145400_document.sql
@@ -20,5 +20,5 @@ CREATE TABLE IF NOT EXISTS interaction (
     PRIMARY KEY (doc_id, user_id, time_stamp)
 );
 
-CREATE INDEX IF NOT EXISTS idx_doc_by_user_id
-    ON document(user_id);
+CREATE INDEX IF NOT EXISTS idx_interaction_by_user_id
+    ON interaction(user_id);
--- a/discovery_engine_core/web-api/src/storage.rs
+++ b/discovery_engine_core/web-api/src/storage.rs
@@ -181,7 +181,7 @@ impl UserState {
         let mut tx = self.pool.begin().await?;
 
         let documents = sqlx::query_as::<_, QueriedInteractedDocumentId>(
-            "SELECT doc_id
+            "SELECT DISTINCT doc_id
             FROM interaction
             WHERE user_id = $1;",
         )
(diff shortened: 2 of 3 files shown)


discovery_engine_core/web-api/src/migrations/20220926145400_document.sql
```diff
@@ -20,5 +20,5 @@ CREATE TABLE IF NOT EXISTS interaction (
     PRIMARY KEY (doc_id, user_id, time_stamp)
 );
 
-CREATE INDEX IF NOT EXISTS idx_doc_by_user_id
-    ON document(user_id);
+CREATE INDEX IF NOT EXISTS idx_interaction_by_user_id
+    ON interaction(user_id);
```
biernacki, line 24: yeah, I also noticed that after rebasing :)

discovery_engine_core/web-api/src/storage.rs
```diff
@@ -171,6 +172,27 @@ impl UserState {
...
+
+    #[allow(dead_code)]
+    pub(crate) async fn interacted_documents(
+        &self,
+        user_id: &UserId,
+    ) -> Result<Vec<DocumentId>, GenericError> {
+        let mut tx = self.pool.begin().await?;
+
+        let documents = sqlx::query_as::<_, QueriedInteractedDocumentId>(
+            "SELECT doc_id
```
biernacki, line 184: if I understand correctly `SELECT DISTINCT` will get rid off duplicates and our PK is (doc_id, user_id, time_stamp) which means in theory we could have same doc_id multiple times returned, yes?

biernacki (commented):
I reviewed it only because I was curios if there is already a fix for the index in interaction table.

discovery_engine_core/web-api/src/storage.rs
janpetschexain: good point, it shouldn't matter much if we feed duplicates to the filter in elastic, but it's definitely better to return unique values here, i'll change it.

discovery_engine_core/web-api/src/migrations/20220926145400_document.sql
janpetschexain: yeah, sorry for that, i must have messed this up during one of my rebases.

biernacki (approved):
lgtm

janpetschexain:
bors merge

PR #629, opened 2022-10-05
[Merged by Bors] - TO-3389 actix skeleton with TO-3318 integrated

rustonaut:
This creates an actix-web skeleton.

To avoid merge conflict problems it's in web-api2, after filling in all missing parts we can rename it.

To make sure everything works this has TO-3318 integrated into it, including parts which
hadn't yet been part of the warp specific PR.

We have following parts:

- a stub for loading the config which will be filled in in a follow up PR
- routing to an unimplemented endpoints for ingestion and personalization
- tracing middleware which generates adds a request id and trace logs receiving requests
  -  this will be replaced with tracing-actix-web in a follow up PR due to that crates optional opentelemetry integration feature(s)
- a error type similar to anyhow::Error but instead of being based on std::error::Error it also carries information about how to turn things into an response and has a request id field (we probably would have needed this in warp, too to avoid warps very brittle error handling)
- a way to inject the request id into the application error similar to axums error handling
  - initially I thought this can be done as a normal middleware, it turns out that doesn't work due to when errors of handlers are turned into responses
- a error to json wrapper middleware which wrapps text errors of other middleware/routing in our common error format
- a small Application abstraction to allow reuse routing setup, middleware setup, config loading, tracing init etc.
  - this might have to be de-abstracted in the future if personalization and ingestion diverge a lot but for now is fine and probably will be fine for a long time
- `bin/ingestion.rs`, `bin/personalization.rs` with most logic under `lib.rs`
- tracing setup copied from the app
   - should be fine tuned and combined with `tracing-actix-web` in a follow up PR
   
**References:**

- [TO-3389]
- [TO-3318]

[TO-3389]: https://xainag.atlassian.net/browse/TO-3389?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rustonaut:
This fails due to clippy errors on main in code I haven't touched. :thinking:

janpetschexain:
> This fails due to clippy errors on main in code I haven't touched. 🤔

the `tokio` version increased in the lock file, i'd assume that they changed sth which now triggers clippy on our end.

rustonaut:
thats probably what it is, anyway it's fixed here #630 .

discovery_engine_core/bindings/src/tracing.rs
```diff
@@ -58,7 +58,7 @@ fn init_tracing_once(log_file: Option<&Path>) {
         })
         .transpose()
         .map_err(|error| {
-            tracing::error!(%error, "logging setup failed");
+            eprintln!("Setup file logging failed: {}", error);
```
biernacki, line 61: why this change?

discovery_engine_core/web-api2/src/server.rs
```diff
@@ -0,0 +1,90 @@
...
+};
+
+use crate::{
+    config::{load_config, Config},
+    middleware::{json_error::wrap_non_json_errors, tracing::tracing_log_request},
+    tracing::init_tracing,
+};
+
+pub(crate) fn default_bind_address() -> SocketAddr {
+    SocketAddrV4::new(Ipv4Addr::new(127, 0, 0, 1), 8080).into()
```
biernacki, line 37: until now we were using `0.0.0.0`, and maybe we should have different default ports for each service, wdyt?

discovery_engine_core/web-api2/src/server.rs
```diff
@@ -0,0 +1,90 @@
...
+}
+
+pub trait Application {
+    type Config: Config;
+    type AppState: TryFrom<Self::Config> + Send + Sync + 'static;
+    fn configure(config: &mut ServiceConfig);
+}
+
+/// Simple program to greet a person
```
biernacki, line 46: hello copy 🍝 :)

discovery_engine_core/web-api2/src/server.rs
```diff
@@ -0,0 +1,90 @@
...
+#[derive(Parser, Debug, Serialize)]
+#[command(author, version, about)]
+struct CliArgs {
+    /// Name of the person to greet
```
biernacki, line 50: same here :)

discovery_engine_core/web-api2/src/tracing.rs
```diff
@@ -0,0 +1,75 @@
...
+            eprintln!("Setup file logging failed: {}", error);
+        })
+        .ok();
+
+    subscriber
+        .with(stdout_log)
+        //FIXME this doesn't seem to work correctly
+        // .with(sqlx_query_no_info)
+        .with(file_log)
+        //FIXME use env to determine logging level
```
biernacki, line 57: maybe add a task number, where this will be fixed

discovery_engine_core/web-api2/src/tracing.rs
```diff
@@ -0,0 +1,75 @@
...
+        })
+        .transpose()
+        .map_err(|error| {
+            eprintln!("Setup file logging failed: {}", error);
+        })
+        .ok();
+
+    subscriber
+        .with(stdout_log)
+        //FIXME this doesn't seem to work correctly
```
biernacki, line 54: do you have an idea why this doesn't work? do you plan to fix it with this PR or should we create a separate ticket?

discovery_engine_core/web-api2/src/middleware/tracing.rs
```diff
@@ -0,0 +1,67 @@
...
+
+pub(crate) fn tracing_log_request<S, B>(
+    request: ServiceRequest,
+    service: &S,
+) -> impl Future<Output = Result<ServiceResponse<B>, actix_web::Error>> + 'static
+where
+    S: Service<ServiceRequest, Response = ServiceResponse<B>, Error = actix_web::Error>,
+    S::Future: 'static,
+{
+    //FIXME try to use tracing-actix-web and the request id of it instead
```
biernacki, line 50: is this fix planned for this PR?

discovery_engine_core/web-api2/src/middleware/json_error.rs
```diff
@@ -0,0 +1,144 @@
...
+            .map_or(true, |content_type| {
+                let mime_bytes = mime::TEXT_PLAIN.as_ref().as_bytes();
+                let bytes = content_type.as_bytes();
+                match bytes.len().cmp(&mime_bytes.len()) {
+                    Ordering::Less => false,
+                    Ordering::Equal => bytes == mime_bytes,
+                    Ordering::Greater => {
+                        bytes.starts_with(mime_bytes) && bytes[mime_bytes.len()] == b';'
+                    }
+                }
```
biernacki, line 87: I think it would be easier to understand with either an explanation comment or specifying explicit types

discovery_engine_core/web-api2/src/error/json_error.rs
```diff
@@ -0,0 +1,70 @@
...
+            mime::APPLICATION_JSON.try_into_value().unwrap(/* MIME is guaranteed well formed */),
+        );
+        response.set_body(self.body)
+    }
+
+    pub(crate) fn into_response(self, status_code: StatusCode) -> HttpResponse {
+        HttpResponse::build(status_code)
+        .content_type(mime::APPLICATION_JSON)
+        .message_body(self.body)
+        .unwrap(/* can not fail as MIME is guaranteed well formed */)
```
biernacki, line 63: doesn't it lack indentation here?

discovery_engine_core/web-api2/src/ingestion.rs
```diff
@@ -0,0 +1,69 @@
...
+
+impl Config for IngestionConfig {
+    fn bind_address(&self) -> std::net::SocketAddr {
+        self.bind_to
+    }
+}
+
+//FIXME use actual body
+#[derive(Deserialize)]
+struct NewDocuments {}
```
biernacki, line 57: maybe we could use models from original `web-api` crate instead as a way to have some "continuation" in the code?

discovery_engine_core/web-api2/src/error/application.rs
```diff
@@ -0,0 +1,258 @@
...
+///   probably make this work, but it's not worth the effort)
+/// - can't use `#[get(..)]` and similar
+///
+/// ## Alternatives
+///
+/// - wrap the handles with a (proc)macro (benefit compatible with `#[get(...)]`)
+/// - use axum (but it's eco system is less mature and it will change over time)
+/// - only include the request id in a custom header (trivial to do)
+/// - land a PR in actix which add an additional `ResponseError::error_response_with_context`
+///   method with an auto-forwarding default implementation which can access middleware data
```
biernacki, line 145: not sure if this should be a comment in code, maybe a github comment on the PR itself

discovery_engine_core/web-api2/src/error/application.rs
```diff
@@ -0,0 +1,258 @@
...
+    pub(crate) functionality: &'static str,
+}
+
+derive_application_error!(Unimplemented => INTERNAL_SERVER_ERROR);
+
+/// Allows to argument errors with an request id by wrapping the endpoint handler.
+///
+/// # Explanation/Why
+///
+/// This is a bit complex as it can only work when wrapping the endpoint handler
```
biernacki, line 120: I thought the whole point for moving to actix was to make this less complex then in warp, maybe we should reconsider, or take a closer look at axum, wdyt? Axum is built on top of tower, and looking at it's popularity couldn't we assume it's mature enough?

biernacki (commented):
I like a lot of parts introduced in this PR, and it's a very nice effort overall 👍 but I also think it's too complex and should be split into smaller PR, ideally adding it to the original `web-api` crate step by step. 

Maybe it would be good to first adjust the original crate to make `models` and `handlers` warp agnostic, which would make a switch to other framework easier, wdyt? I can help with such transition.

We could also add more documentation to certain functions, especially related to error handling.

discovery_engine_core/web-api2/src/error/application.rs
biernacki: I'm not sure if it covers our needs, but I found a small crate for request id for tower/axum and it looks simple in both implementation and application https://github.com/imbolc/tower-request-id

rustonaut: > I'm not sure if it covers our needs,

There are a few docent crates which do add a request id the the header or similar, that is trivial.
E.g. the opentelemtry actix integration does that, too.

The tricky part is to put it into the error response in a ergonomic manner.

> I thought the whole point for moving to actix was to make this less complex then in warp

Most of the parts in this PR are thinks we probably also should do in warp, too. But which where not included in the previous warp specific PR.

So if to compare that PR you would need to compare it with the `WithRequestIdExt` extension trait only. Not the json error, wrapping, error types etc.

I also thought for a while I could just add a middle ware in actix-web, but that failed at runtime because of how axtic-web converts errors into responses *before* it is passed to any middleware even through the middleware accepts an `Result` :face_exhaling: (It's only for errors form other middle-ware and similar but *not* for e.g. routing errors, endpoint returning an error, failing to parse a request parameter or anything like that.)

I personally also prefer axum and changing this PR into a axum one is way less work then creating this PR (most code stays the same, except the startup & routing code and the extension trait is gone, too)

On reason against axum is that it's still rather unstable and we can expect a lot of small changes over time. Especially when GAT is relased there is a good chance not only for axum to majorly change, but also for tower... but probably that is also true for actix-web which probably would go and bump a v5.0 release so idk. :shrug:

discovery_engine_core/web-api2/src/server.rs
rustonaut: Yes, but I copied the address choice form web-api... 

Generally for testing,local debugging the best default is 127.0.0.x and 0.0.0.0 should be avoided.

But for production you always want 0.0.0.0.

But serde doesn't take additional options and and making the address cfg-depend based on the kind of builds (instead of the context it runs on) is a bit to error prone IMHO. 

Which is why I choose 127.0.0.1 with the expectancy to always explicitly set the address for production usage.

discovery_engine_core/web-api2/src/middleware/json_error.rs
rustonaut: yes it might be a bit of premature optimization, probably should just check equal `mime` or starts with `mime_with_trailing_semicolon`.

discovery_engine_core/web-api2/src/ingestion.rs
rustonaut: yes, but in follow up PRs

rustonaut:
> , ideally adding it to the original web-api crate step by step

I would prefer not to do so, that can get very easily very mess and also is a lot more work. Given that handler functions have nearly the same signature copying them over in separate PRs  (one per-endpoint) and adjusting their return type should be the easiest solution as far as I can tell.

discovery_engine_core/web-api2/src/middleware/tracing.rs
biernacki: so maybe let's remove a `FIXME` comment and make sure that we handle the `tracing-actix-web` in opentelemetry integration story https://xainag.atlassian.net/browse/ET-2952

discovery_engine_core/web-api2/src/error/application.rs
biernacki: >The tricky part is to put it into the error response in a ergonomic manner.

Ok, I think I get it now. But it seems that we could achieve such thing in a more readable/easy to understand way, by abstracting framework (wrap or actix) route handlers from the handlers that perform actual business logic, so that our handlers take needed inputs and return a `Result`, and the route handlers are responsible for translating inputs and mapping returned `Result`s for/from functions they are wrapping, smth like:

```rust
pub(crate) async fn user_interactions_route(
    request_id: RequestId,
    user_id: UserId,
    body: InteractionRequestBody,
    state: Arc<AppState>,
) -> Result<impl warp::Reply, Rejection> {
    let _ = handle_user_interactions(user_id, state)
        .await
       // error mapping with req_id
        .map_err(|_| warp::reject::custom(ExternalError(request_id)))?;
    Ok(StatusCode::NO_CONTENT)
}

pub(crate) async fn handle_user_interactions(
    user_id: UserId,
    state: Arc<AppState>,
) -> Result<(), CustomInternalError> {
    // implementation
}
```

discovery_engine_core/web-api2/src/tracing.rs
rustonaut: Yes I now do:

> /// Note that an unset default level (`None`) behaves like [`LevelFilter::OFF`] when the filter is
> /// used, but it could also be set explicitly which may be useful to distinguish (such as when
> /// merging multiple `Targets`).

discovery_engine_core/web-api2/src/error/json_error.rs
rustonaut: yes, somehow rustfmt is broken in if there is a /**/ block comment in a empty function parameter list in some situations :upside_down_face: 

worked around this by moving the comment in a new line above the unwrap

discovery_engine_core/web-api2/src/error/application.rs
rustonaut: But that needs to be done with every single endpoint in just the right way meaning it is not just very verbose but also prone to have bugs and just generally bad style.

So you would write a macro to make sure you don't have change bugs due to code repetition.

And at that point you have exactly what I had in the warp PR.

The benefit of the extension trait is it's code you only write once and then never touch and then reuse and the type system makes sure all complex parts are correct.

biernacki: For error mapping we could have one function to reuse everywhere and only repeat the `RequestId` extractor in handlers, it might be more verbose but also simpler to understand and maintain for others, like me ;) That said we can ofc go forward with your solution, but then it would be good if someone more experienced reviewed the code.

discovery_engine_core/web-api2/src/error/json_error.rs
biernacki: if you changed something please push it, so we can resolve comments :)

rustonaut: done :+1:

discovery_engine_core/web-api2/src/server.rs
```diff
@@ -0,0 +1,93 @@
...
+    SocketAddrV4::new(Ipv4Addr::new(127, 0, 0, 1), 8080).into()
+}
+
+pub trait Application {
+    type Config: Config;
+    type AppState: TryFrom<Self::Config> + Send + Sync + 'static;
+    fn configure(config: &mut ServiceConfig);
+}
+
+/// Server for running the web-api.
```
biernacki, line 46:
```suggestion
/// Cli arguments for the web-api server.
```

discovery_engine_core/web-api2/src/error/application.rs
rustonaut: Yes, but you still have to e.g. keep the parameter list in sync or else you might for example fail an endpoint at runtime because you only removed an extractor in the actual implementation but not the wrapper.

discovery_engine_core/web-api2/src/error/application.rs
```diff
@@ -0,0 +1,258 @@
...
+            _args: PhantomData,
+        }
+    }
+}
+
+impl<H, Args, B> Handler<<Args as WithRequestIdHint>::ExtendedArgs>
+    for HandlerWithIdInjection<H, Args>
+where
+    H: Handler<Args, Output = Result<B, Error>>,
+    HandlerWithIdInjectionFut<H::Future>: Future<Output = Result<B, Error>>,
```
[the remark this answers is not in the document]
rustonaut: Probably not needed, just a left over from going through multiple different trait bounds where at some point it was needed.

discovery_engine_core/web-api2/src/middleware/tracing.rs
```diff
@@ -0,0 +1,66 @@
...
+use serde::Serialize;
+use std::future::Future;
+
+use actix_web::{
+    dev::{Service, ServiceRequest, ServiceResponse},
+    HttpMessage,
+};
+
+use tracing::{trace, Instrument};
+use tracing_actix_web::root_span_macro::private::tracing::info_span;
```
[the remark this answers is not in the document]
rustonaut: yep, that was rust-analyzer picking a bad import for the right thing.

discovery_engine_core/web-api2/src/error/application.rs
rustonaut: @janpetschexain can you add your opinion?

discovery_engine_core/web-api2/src/error/application.rs
```diff
@@ -0,0 +1,258 @@
...
+            &self.error.encode_details(),
+        )
+        .into_response(self.error.status_code())
+    }
+}
+
+pub trait ApplicationError: std::error::Error + Send + Sync + 'static {
+    fn status_code(&self) -> StatusCode;
+    fn kind(&self) -> &str;
+    fn encode_details(&self) -> Value {
```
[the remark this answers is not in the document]
rustonaut: You are already encoding the details into a json object.

discovery_engine_core/web-api2/src/middleware/tracing.rs
```diff
@@ -0,0 +1,66 @@
...
+#[derive(Clone, Copy, Debug, derive_more::Display, Serialize)]
+#[serde(transparent)]
+pub(crate) struct RequestId(Uuid);
+
+impl RequestId {
+    pub(crate) fn generate() -> Self {
+        Self(Uuid::new_v4())
+    }
+
+    pub(crate) fn missing() -> Self {
```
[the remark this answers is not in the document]
rustonaut: :rocket: 

Through it's never used in a const context so not really any benefit to mark it as const given that it is internal only.

discovery_engine_core/web-api2/src/error/application.rs
```diff
@@ -0,0 +1,256 @@
...
+        Value::Null
+    }
+}
+
+/// Derives [`ApplicationError`] for given type using given http status code.
```
janpetschexain, line 85: :art:
```suggestion
/// Implements [`ApplicationError`] for given type using given http status code.
```

discovery_engine_core/web-api2/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+name = "web-api2"
+version = { workspace = true }
+edition = { workspace = true }
+rust-version = { workspace = true }
+description = "Wep Api server for the document personalization service."
+license = "AGPL-3.0-only"
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+actix-web = "4.2.1"
```
janpetschexain, line 11: :art: actix enables quite a lot default features, maybe we can enable only some required features to save space and time.

discovery_engine_core/web-api2/src/error/application.rs
janpetschexain: i'm tending more towards the midleware solution. even though it's more complex to read, it's easier extendable and less repetetive. i fully get the argument for easier readability of more verbose code though. maybe we can have a sketch/short documentation of the middlewares and endpoints somewhere at the root of the crate (or in another fitting place) to make it easier for new reviewers to follow the code?

rustonaut: Be aware that the middle wares do two things:

- wrap errors not from us  (e.g. serde)

- put a request Id in the context when setting up the log span

The code which injects the request id into our errors is just kinda a middleware:

```rust
web::resource("personalized_documents")
    .route(web::get().to(personalized_documents.error_with_request_id()))
```

So you could say it's a middleware around a specific handler insted one around many handlers/services, 
except that actix-web doesn't have something like `.layer()` or similar on handlers, so I had to put in a bit
of type magic to make it work.

rustonaut:
bors merge

rustonaut:
bors merge

PR #627, opened 2022-10-04
[Merged by Bors] - TO-3388 new format for interactions batch

janpetschexain:
**References**

- [TO-3388]
- requires #626
- followed by #632

**Summary**

- change `InteractionRequestBody` to hold a `Vec<InteractionData>`
- query elastic for all ids and fail fast before coi updates if an id is missing
- update cois for all interactions

[TO-3388]: https://xainag.atlassian.net/browse/TO-3388?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #632, opened 2022-10-05
[Merged by Bors] - ET-3339 compliance of endpoint errors

janpetschexain:
**References**

- [ET-3339]
- [ET-3342]
- requires #627

**Summary**

- align personalized documents endpoint error handling
- align user interactions endpoint error handling
- align user interactions type names

[ET-3339]: https://xainag.atlassian.net/browse/ET-3339?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ET-3342]: https://xainag.atlassian.net/browse/ET-3342?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine_core/web-api/openapi.yaml
+++ b/discovery_engine_core/web-api/openapi.yaml
@@ -149,7 +149,7 @@ components:
       type: object
       properties:
         request_id:
-          description: Request ID optionally generated from the service. It can be comunicated to xayn to help debugging.
+          description: Request ID optionally generated from the service. It can be communicated to xayn to help debugging.
           type: string
     PersonalizedDocumentData:
       type: object
@@ -177,6 +177,7 @@ components:
           properties:
             kind:
               type: string
+              enum: [not_enough_interactions]
     UserInteractionType:
       type: string
       enum: [positive]
--- a/discovery_engine_core/web-api/src/handlers.rs
+++ b/discovery_engine_core/web-api/src/handlers.rs
@@ -134,7 +134,6 @@ pub(crate) async fn handle_personalized_documents(
             error!("Error scoring documents: {error}");
             return Ok(Box::new(
                 PersonalizedDocumentsError::new(
-                    None,
                     PersonalizedDocumentsErrorKind::NotEnoughInteractions,
                 )
                 .to_reply(StatusCode::UNPROCESSABLE_ENTITY),
@@ -169,7 +168,7 @@ pub(crate) async fn handle_user_interactions(
         Err(error) => {
             error!("Error fetching documents: {error}");
             return Ok(Box::new(
-                UserInteractionError::new(None, UserInteractionErrorKind::InvalidDocumentId)
+                UserInteractionError::new(UserInteractionErrorKind::InvalidDocumentId)
                     .to_reply(StatusCode::BAD_REQUEST),
             ) as Box<dyn Reply>);
         }
@@ -178,7 +177,7 @@ pub(crate) async fn handle_user_interactions(
     if embeddings.len() < ids.iter().unique().count() {
         debug!("Document not found");
         return Ok(Box::new(
-            UserInteractionError::new(None, UserInteractionErrorKind::InvalidDocumentId)
+            UserInteractionError::new(UserInteractionErrorKind::InvalidDocumentId)
                 .to_reply(StatusCode::BAD_REQUEST),
         ));
     }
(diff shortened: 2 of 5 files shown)


discovery_engine_core/web-api/src/handlers.rs
```diff
@@ -122,57 +125,68 @@ pub(crate) async fn handle_personalized_documents(
     }
 
     if all_documents.is_empty() && !errors.is_empty() {
-        return Err(warp::reject::custom(ElasticOpError));
+        return Ok(Box::new(StatusCode::BAD_REQUEST));
```
rustonaut, line 128: Isn't that an internal server error, because accessing an internal subsystem of ours failed.

discovery_engine_core/web-api/openapi.yaml
```diff
@@ -177,6 +177,7 @@ components:
           properties:
             kind:
               type: string
+              enum: [not_enough_interactions]
```
rustonaut, line 180: :pushpin: Slightly off topic by why did we move to naming thinks in snake case which make to rust types, bother here and in the db. Shouldn't it be `NotEnoughInteractions`?

janpetschexain: i don't know, i just followed the naming style of the already existing enums of the spec. if we should change the name style, i guess it makes sense to have a follow up ticket for that.

discovery_engine_core/web-api/src/models.rs
```diff
@@ -49,6 +49,11 @@ pub(crate) enum Error {
 
 impl Reject for Error {}
 
+#[derive(Clone, Debug, Serialize)]
+struct BaseError {
```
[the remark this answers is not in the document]
janpetschexain: > I'm not sure it is worth it to have an #[serde(flatten)]-focused struct for a single fields.

in case that `BaseError` gets extended or changed, then flattening means less code changes.

>  I also don't understand why this field is added here given that request id handling was already implemented in a PR as well as the actix PR we decided to have instead. It will never be used anywhere.

the ticket was about making the implementation compliant with the specification, and the specification still has this and wasn't updated yet. i agree that it needs to be changed once the spec is updated to match the new approach, but again this is not the scope of the current ticket.

discovery_engine_core/web-api/src/handlers.rs
janpetschexain: the [openapi docs](https://spec.openapis.org/oas/v3.1.0#responses-object) is not quite clear to me in that regard:

> The documentation is not necessarily expected to cover all possible HTTP response codes because they may not be known in advance. However, documentation is expected to cover a successful operation response and any known errors.

i'd say that a 500 would be a known error in this case, but our specification doesn't list 500 as a known response, hence i mapped it like this. i guess we can make one of the following decisions here:
- map it as 400 like currently
- extend our spec to include 500 and map it as 500
- implicitely allow 500 as a response in our spec and map it as 500

i have no strong preferences, it just should be handled uniformly in our spec.

acrrd: This is 500 since it does not depends on the client's request. We didn't list 500 because it could always happen and will always be there. We can also add it to the spec in every endpoint but will not add much.

janpetschexain: changed accordingly.

discovery_engine_core/web-api/src/models.rs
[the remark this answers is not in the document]
rustonaut: > gets extended or changed 

But that's a premature optimization for a use case which probably will never happen.

> wasn't updated yet

It's not that request id is gone, it's just already implemented somewhere else and is handled by middle-ware instead of needing to be included into every single error type.

discovery_engine_core/web-api/openapi.yaml
rustonaut: So I guess we accidentally did change the naming style at some point? :sweat_smile:  

I just wanted to ask, because in the actix PR I added a small macro which basically "derives" the error type and it uses `stringify!($name)` and converting that to snake case requires using a proc-macro or unnecessary runtime conversions.

rustonaut (approved):
lgtm

discovery_engine_core/web-api/src/models.rs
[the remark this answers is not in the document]
janpetschexain: removed as discussed in the daily.

janpetschexain:
bors merge

PR #634, opened 2022-10-10
Generalize smenatic filter functions

acrrd:
This helps in writing tests and bench code without the need to construct mocked documents.


discovery_engine_core/ai/ai/src/embedding.rs
```diff
@@ -74,10 +74,8 @@ pub const COSINE_SIMILARITY_RANGE: RangeInclusive<f32> =
 /// [`Chain`]: std::iter::Chain
 pub fn pairwise_cosine_similarity<'a, I>(iter: I) -> Array2<f32>
 where
-    I: IntoIterator<Item = ArrayView1<'a, f32>>,
-    I::IntoIter: Clone,
+    I: Iterator<Item = ArrayView1<'a, f32>> + Clone,
```
janpetschexain, line 77: ❓ afaik `IntoIterator` is usually prefered over `Iterator` for arguments, because the former applies to iterators as well as collections, also it's a bit less verbose at the call site. what's the reason to change this?

i guess this also refers to the changes in `condensed_cosine_similarity()` and `condensed_date_distance()`, isn't it preferable to also require `IntoIterator` instead of `Iterator` there?

acrrd: I don't remember why I changed it. I'll try to use it in the branch where I try to add benchmark from which I took this PR and see if everything is ok.

janpetschexain (approved):
lgtm

acrrd:
bors merge

PR #637, opened 2022-10-11
[Merged by Bors] - ET-3376 use futures unordered

rustonaut:
Use `FuturesUnordered` instead of `join_all`.

I'm still not sure it makes any sense as it could easily be a neglible small performance reduction instead of a similar small increase and it either way is IMHO a pre-mature optimization. Just to put it into context `join_all` will internally use `FuturesUnordered` (through `FuturesOrdered`) iff and only iff there are "many" futures (> 30), else it will not use it as it's slower.

There is also the thing that tokio promotes using `spawn` when you join over multiple futures. E.g. tokio has some mechanism to auto-brake live locks but that same mechanism can cripple larger nested futures, `FuturesUnordered` somewhat works around it but there are limits in how much it can do so. I might be better to spawn the requests and then just use `join_all` for simplicity (simpler as in simpler for us).  One thing which speaks against doing this is that spawing futures requires them to be static, so we would need to sprinkle in some `Arc`s.

**References:**

- [ET-3376]

[ET-3376]: https://xainag.atlassian.net/browse/ET-3376?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine_core/web-api/src/handlers.rs
+++ b/discovery_engine_core/web-api/src/handlers.rs
@@ -14,7 +14,7 @@
 
 use std::{collections::HashMap, sync::Arc, time::Duration};
 
-use futures::{future::join_all, stream::FuturesUnordered, StreamExt};
+use futures::{stream::FuturesUnordered, StreamExt};
 use itertools::Itertools;
 use tracing::{debug, error, instrument};
 use warp::{hyper::StatusCode, reject::Reject, Rejection};
--- a/discovery_engine_core/web-api/src/handlers.rs
+++ b/discovery_engine_core/web-api/src/handlers.rs
@@ -14,7 +14,7 @@
 
 use std::{collections::HashMap, sync::Arc, time::Duration};
 
-use futures::future::join_all;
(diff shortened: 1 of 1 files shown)


janpetschexain (approved):
> I'm still not sure it makes any sense as it could easily be a neglible small performance reduction instead of a similar small increase and it either way is IMHO a pre-mature optimization. Just to put it into context join_all will internally use FuturesUnordered (through FuturesOrdered) iff and only iff there are "many" futures (> 30), else it will not use it as it's slower.

`FuturesOrdered` has to maintain indexing though, which `FuturesUnordered` doesn't have to in our case.

overall the performance probably doesn't differ much, imo it's more a stylistic question. collecting into a `FuturesUnordered` immediately tells what the purpose of the collection is and also the `while let` loop is cleaner than the `join_all` loop. as always this is subjective of course.

PR #639, opened 2022-10-11
[Merged by Bors] - ET-3384 japanese tokenizer

janpetschexain:
**References**

- [ET-3384]
- [ET-3464]

**Summary**

add a japanese pre-tokenizer to the bert tokenizer:
- the vocabulary kind is currently chosen as `IpaDic` (which is also the default of the `lindera` crate), but there is also the `UniDic`, [according to wikipedia](https://en.wikipedia.org/wiki/MeCab) the former seems to be the better choice
- a custom vocabulary could be provided in addition to the internal vocabulary, the format is a csv with the colums `<surface>,<part_of_speech>,<reading>` (without header, [see the docs](https://github.com/lindera-morphology/lindera#user-dictionary-example)) or a binary format. regardless of the used crate, providing custom vocabularies is not as easy as with latin bert as they all require some additional indices, weights and alphabet information (are we able to provide that?). we can still start using the internal vocabulary without a custom one.
- in the core engine the japanese pre-tokenizer can be enabled via the json `de_config`, it is disabled if the field is not set: `{ "smbert": { "japanese": "path/to/mecab" } }`
- in the web-api the japanese pre-tokenizer can be enabled via the env var `JAPANESE_MECAB`, it is disabled if the env var is not set: `JAPANESE_MECAB="path/to/mecab"`
- the `smbert_v0002` assets including the japanese assets are uploaded to the `xayn-yellow-bert` and `xayn_ai_staging_assets` buckets

if we have to extract the mecab again, this can be done via ([see here](https://github.com/lindera-morphology/lindera/tree/main/lindera-ipadic) for the mecab source):
```rust
lindera::builder::build_dictionary(
    lindera::DictionaryKind::IPADIC,
    "mecab-ipadic-2.7.0-20070801".as_ref(),
    "mecab".as_ref(),
)
.unwrap();
```

[ET-3384]: https://xainag.atlassian.net/browse/ET-3384?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[ET-3464]: https://xainag.atlassian.net/browse/ET-3464?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


discovery_engine_core/ai/bert/src/config.rs
```diff
@@ -43,6 +45,8 @@ pub enum ConfigError {
 pub struct Config<'a, K, P> {
     model_kind: PhantomData<K>,
     vocab: Box<dyn BufRead + Send + 'a>,
+    #[cfg(feature = "japanese")]
+    japanese: Option<PathBuf>,
```
acrrd, line 49: :question: Since is optional I would expect that we can set this with a `with_internal_japanese_vocab` method instead.
That will allow us to leave all the current call sites untouched and I think we can also remove the feature flag.
Any reason we need to take it here?

acrrd: Then we can also have a method to tell the builder that we need the pre-tokenizer other we could avoid running it.

janpetschexain: it's probably still better to keep the feature flag, otherwise the binary size and compilation time will increase. also the embeddings of non-japanese sequences will be slightly shifted, because of additional white space, which affects our existing tests as well.

acrrd: We have a story where we want to be able to load any bert model (that respect the current architecture) from a configuration file to avoid having to change the code and this will apply also to the Japanese models.
I quickly tried a build for android and it adds around ~70Mb to the lib so we need to keep it under the feature flag at the moment :(
Do you know if we can disable the internal dictionary? I'm guessing this is what is bloating the binary so much.

janpetschexain: i will check. maybe we can extract the internal dict, then we could load it as an asset without including it in the binary.

acrrd (approved):
@rustonaut This change the configuration of the ingestion. We need to take it into account when switching to actix.

janpetschexain:
borse merge

janpetschexain:
bors merge

PR #641, opened 2022-10-12
ET-3368 delete document endpoint api spec

rustonaut:
**References:**

- [ET-3368]


discovery_engine_core/web-api/openapi.yaml
```diff
@@ -127,6 +127,45 @@ paths:
...
               schema:
                 $ref: '#/components/schemas/IngestionError'
+    delete:
+      tags:
+        - documents
+      summary: Deletes all listed documents.
+      description: |-
+        Deletes all documents listed in the request body. The endpoint is
+        idempotent. I.e. If the list contains one or multiple non-existing
+        document it does not produce  an error.
```
Robert-Steiner, line 137:
```suggestion
        idempotent. I.e. if the list contains one or multiple non-existing
        documents, no error is produced.
```

discovery_engine_core/web-api/openapi.yaml
```diff
@@ -127,6 +127,45 @@ paths:
...
+          application/json:
+            schema:
+              $ref: '#/components/schemas/DeleteDocumentsRequest'
+      responses:
+        '204':
+          description: successful operation
+        '400':
+          description: invalid request
+
+  /documents/{document_id}:
```
Robert-Steiner, line 149: ❔ Only the endpoint above is mentioned in the ticket. was it later decided that we also need this endpoint?

discovery_engine_core/web-api/openapi.yaml
```diff
@@ -127,6 +127,45 @@ paths:
...
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/DeleteDocumentsRequest'
+      responses:
+        '204':
+          description: successful operation
+        '400':
+          description: invalid request
```
Robert-Steiner, line 147: ❔ In the ticket we say that we that in case of an error we return the id of the documents that could not be deleted. Has this requirement changed?

discovery_engine_core/web-api/openapi.yaml
```diff
@@ -127,6 +127,45 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/IngestionError'
+    delete:
+      tags:
+        - documents
+      summary: Deletes all listed documents.
+      description: |-
+        Deletes all documents listed in the request body. The endpoint is
```
Robert-Steiner, line 135: 🎨  to be consistent with the other summaries ands descriptions
```suggestion
      summary: Delete all listed documents.
      description: |-
        Delete all documents listed in the request body. The endpoint is
```

discovery_engine_core/web-api/openapi.yaml
```diff
@@ -127,6 +127,45 @@ paths:
...
+        '204':
+          description: successful operation
+        '400':
+          description: invalid request
+
+  /documents/{document_id}:
+    delete:
+      tags:
+        - documents
+      summary: Deletes the document from the system.
```
Robert-Steiner, line 153:
```suggestion
      summary: Delete the document from the system.
```

discovery_engine_core/web-api/openapi.yaml
rustonaut: Good point but the endpoint is also idempotent so:

- this doesn't apply if the document can't be deleted because it is already deleted
- it also doesn't really apply for general "everything went wrong and we deleted nothing" errors
- so it only applies if while we delete documents somewhere in between it failed and some
  documents are deleted and some not and we _know_ that some are deleted
  - but the deletion in postgres is transactional so this can't happen there
  - and the deletion in elastic search could (and should) be batched, too
  
So isn't it a situation where it always either returns all or none documents and in the exceedingly 
rare case where this isn't the case (we din't implement batched ec deletion and internet/ec crashed in between)
it would be okay to tell the user to just redo the full deletion due to idempotency.

rustonaut: daily: we don't need to return ids but deletion at elastic search must use batch mode

discovery_engine_core/web-api/openapi.yaml
rustonaut: historically we once only had `DELETE /documents/{document_id}` and then moved to batched delete as main functionality it's not clear if only having batched delete was intentional or accidental but as long as single delete is time wise cheap to implement it would be nice to have 

daily: we can have it

me: We might just implement it by using bulk/batch request with a single element.

Robert-Steiner (approved):
LGTM!
there is a merge conflict but i approve it already

rustonaut:
bors merge

PR #643, opened 2022-10-12
Add benchmarks

acrrd:
Add benchmark for semantic clustering, pairwise similarity and other functions needed for the ranking.
Unfortunately running the benchmark consecutively will lead to improvement or regression. These values can sometimes be big (~10%). 
Maybe we can later try tuning criterion values to have less variability.


frankpepermans:
LGTM

But, maybe it also makes sense to bench the `fn normalized_distance` in semantics, as this is actually the combined method for all calculations?

Currently, all `condensed_x` methods basically do a full-iter calculation of one specific thing, but they can be actually `inlined` to run just one one embedding/document, as I did in my perf branch test.

This then results in just one single cartesian/triangle loop which originates from `normalized_distance`, instead of creating many individual mapped/collected vecs.

Not saying we _should_ do that, but having a bench on the `normalized_distance` itself, allows to optimize those "step" methods in a different way, leaving in the end the `normalized_distance` bench as the target to improve.

frankpepermans:
To better explain,

What I was trying to do, is combine all into a single loop, which then can (or cannot, leave that to be decided) use `into_par_iter` or just `into_iter`: https://github.com/xaynetwork/xayn_discovery_engine/blob/exp_optimize_condensed_date_distance/discovery_engine_core/core/src/stack/filters/semantic.rs#L195

Compare this to main, where it instead runs multiple iterations in sequence, to finally combine them: https://github.com/xaynetwork/xayn_discovery_engine/blob/main/discovery_engine_core/core/src/stack/filters/semantic.rs#L175
(and each individual one loops using `cartesian_product` there)

acrrd:
Good point. I hoped I could avoid to mock documents :(

frankpepermans:
@acrrd I've slapped together this bench, maybe it helps

```
use chrono::DateTime;
use criterion::{black_box, criterion_group, BatchSize, Criterion};
use ndarray::arr1;
use rand::Rng;
use rand_distr::Uniform;
use url::Url;
use uuid::Uuid;
use xayn_discovery_engine_core::{
    document::{Document, NewsResource},
    stack::filters::semantic::{normalized_distance, SemanticFilterConfig},
};

fn create_documents(n: u32) -> Vec<Document> {
    let range = Uniform::new(-10000., 10000.);

    (0..n)
        .map(|_| {
            let smbert_embedding: Vec<f32> =
                rand::thread_rng().sample_iter(&range).take(128).collect();

            Document {
                id: Uuid::new_v4().into(),
                stack_id: Uuid::new_v4().into(),
                smbert_embedding: arr1(&smbert_embedding).into(),
                reaction: None,
                resource: NewsResource {
                    country: "gb".to_string(),
                    date_published: DateTime::from(
                        DateTime::parse_from_str(
                            "2022 Oct 14 12:00:00.000 +0000",
                            "%Y %b %d %H:%M:%S%.3f %z",
                        )
                        .unwrap(),
                    ),
                    image: None,
                    language: "en".to_string(),
                    rank: 0,
                    score: None,
                    snippet: "".to_string(),
                    source_domain: "".to_string(),
                    title: "".to_string(),
                    topic: "".to_string(),
                    url: Url::parse("http://localhost").unwrap(),
                },
            }
        })
        .collect()
}

fn bench_normalized_distance(c: &mut Criterion) {
    let count = [100, 500, 1000];
    let embedding_size = 128;

    count.iter().for_each(|&n| {
        let base_name = &format!("n{}_s{}", n, embedding_size);
        let documents = create_documents(n as u32);

        c.bench_function(&format!("normalized_distance_{}", base_name), |b| {
            b.iter_batched(
                || black_box((&documents, SemanticFilterConfig::default())),
                |(documents, config)| normalized_distance(documents, &config),
                BatchSize::SmallInput,
            );
        });
    });
}

criterion_group!(b_bench_normalized_distance, bench_normalized_distance);

fn main() {
    criterion::Criterion::default()
        .configure_from_args()
        .final_summary();

    b_bench_normalized_distance();
}
```

acrrd:
Thanks! I added `normalized_distance` and removed some of the condensed functions. Now we can fuse it and we will be able to see some effect. Unfortunately even if `normalize_distance` take some time the results are still +-2/3% every run :(

frankpepermans (approved):
lgtm, can we merge this?
benches are becoming more and more relevant as we move the real engine in the cloud

acrrd:
bors merge

acrrd:
bors cancel

acrrd:
bors merge

acrrd:
bors merge

PR #644, opened 2022-10-12
[Merged by Bors] - ET-3443 fix for embedding type in ElasticDocumentData

biernacki:
**Reference**:
- [ET-3443]

**Summary**:
- changed type for embedding in `ElasticDocumentData` from `Embedding` to `Vec<f32>` to fix (de)serialization structure that we store in Elastic Search and use for kNN search.

Before change `Embedding` was serialized into:
```
"embedding": {
  "v": 1,
  "dim":[128],
  "data": [/* embeddings vector */]
}
```
After the change:
```
"embedding": [/* embeddings vector */]
```

[ET-3443]: https://xainag.atlassian.net/browse/ET-3443?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine_core/web-api/src/bin/ingestion.rs
+++ b/discovery_engine_core/web-api/src/bin/ingestion.rs
@@ -185,11 +185,11 @@ fn post_documents(
         .and(with_model(model))
         .and(with_config(config))
         .and(with_client(client))
-        .and_then(handle_add_data)
+        .and_then(handle_post_documents)
 }
 
 #[instrument(skip(model, config, client))]
-async fn handle_add_data(
+async fn handle_post_documents(
     body: IngestionRequest,
     model: Model,
     config: Config,
--- a/discovery_engine_core/web-api/src/bin/ingestion.rs
+++ b/discovery_engine_core/web-api/src/bin/ingestion.rs
@@ -185,11 +185,11 @@ fn post_documents(
         .and(with_model(model))
         .and(with_config(config))
         .and(with_client(client))
-        .and_then(handle_post_documents)
+        .and_then(handle_add_data)
 }
 
 #[instrument(skip(model, config, client))]
-async fn handle_post_documents(
+async fn handle_add_data(
     body: IngestionRequest,
     model: Model,
     config: Config,
--- a/discovery_engine_core/web-api/src/elastic.rs
+++ b/discovery_engine_core/web-api/src/elastic.rs
@@ -13,6 +13,7 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 use itertools::Itertools;
+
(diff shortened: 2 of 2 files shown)


Robert-Steiner (approved):

discovery_engine_core/web-api/src/bin/ingestion.rs
```diff
@@ -211,7 +211,7 @@ async fn handle_add_data(
                 ElasticDocumentData {
                     snippet: document.snippet,
                     properties: document.properties,
-                    embedding,
+                    embedding: embedding.to_vec(),
```
janpetschexain, line 214: :wrench: `to_vec()` always clones the data, which can become quite expensive since we ha a lot of arrays. we can add a utility funciton in the `xayn_discovery_engine_bert::pooler` module in the `impl Embedding<Ix1>` block:
```rust
    pub fn into_vec(self) -> Vec<f32> {
        if self.is_standard_layout() {
            self.0.into_raw_vec()
        } else {
            self.to_vec()
        }
    }
```
this way it's only cloned if the data is not in standard layout, which usually should be the cold path.

janpetschexain (changes requested):

discovery_engine_core/web-api/src/elastic.rs
```diff
@@ -165,3 +166,23 @@ struct Hit<T> {
...
+pub(crate) mod serde_embedding_as_vec {
+    use ndarray::Array;
+    use serde::{Deserialize, Deserializer, Serialize, Serializer};
+    use xayn_discovery_engine_ai::Embedding;
+
+    pub(crate) fn serialize<S>(embedding: &Embedding, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        embedding.iter().collect::<Vec<_>>().serialize(serializer)
```
janpetschexain, line 179: i like that we can keep the `Embedding` wrapper in `ElasticDocumentData`:+1: the issue of additional allocations is still the same though, we could use the same solution proposed in https://github.com/xaynetwork/xayn_discovery_engine/pull/644#discussion_r994303689:
```suggestion
        embedding.into_vec().serialize(serializer)
```

biernacki: I used serializing sequence approach described here https://serde.rs/impl-serialize.html#serializing-a-sequence-or-map, please check again :)

janpetschexain (approved):
nice 👍

Robert-Steiner (approved):
nice solution 👍

biernacki:
bors merge

PR #635, opened 2022-10-10
[Merged by Bors] - ET-3371 document properties endpoint spec

janpetschexain:
**References**

- [ET-3371]

**Summary**

- add endpoint specifications for all document properties
- add endpoint specifications for a single document property

[ET-3371]: https://xainag.atlassian.net/browse/ET-3371?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #645, opened 2022-10-13
[Merged by Bors] - ET-3372 all document properties endpoints

janpetschexain:
**References**

- [ET-3372]
- #635
- followed by #646

**Summary**

- add GET, PUT and DELETE endpoints for all document properties

[ET-3372]: https://xainag.atlassian.net/browse/ET-3372?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine_core/web-api/src/elastic.rs
+++ b/discovery_engine_core/web-api/src/elastic.rs
@@ -134,6 +134,24 @@ impl ElasticState {
         Ok(())
     }
 
+    pub(crate) async fn delete_document_properties(&self, id: &DocumentId) -> Result<(), Error> {
+        // https://www.elastic.co/guide/en/elasticsearch/reference/8.4/docs-update.html
+        // don't delete the field, but put an empty map instead, similar to the ingestion service
+        let body = Some(json!({
+            "script": {
+                "source": "ctx._source.properties = params.properties",
+                "params": {
+                    "properties": DocumentProperties::new()
+                }
+            },
+            "_source": false
+        }));
+        self.query_elastic_search::<GenericResponse>(&format!("_update/{id}"), body)
+            .await?;
+
+        Ok(())
+    }
+
     async fn query_elastic_search<T>(&self, route: &str, body: Option<Value>) -> Result<T, Error>
     where
         T: DeserializeOwned,
--- a/discovery_engine_core/web-api/src/elastic.rs
+++ b/discovery_engine_core/web-api/src/elastic.rs
@@ -13,8 +13,11 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 use itertools::Itertools;
-use reqwest::Client;
-use serde::{Deserialize, Serialize};
+use reqwest::{
+    header::{HeaderValue, CONTENT_TYPE},
+    Client,
+};
+use serde::{de::DeserializeOwned, Deserialize, Serialize};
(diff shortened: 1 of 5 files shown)


discovery_engine_core/web-api/src/routes.rs
```diff
@@ -64,6 +68,53 @@ fn user_path() -> impl Filter<Extract = (UserId,), Error = Rejection> + Clone {
...
+// PATH /documents/:document_id
+fn document_path() -> impl Filter<Extract = (DocumentId,), Error = Rejection> + Clone {
+    warp::path("documents")
+        .and(warp::path::param::<String>())
+        .and_then(|document_id: String| async move {
+            urlencoding::decode(&document_id)
+                .map_err(Error::DocumentIdUtf8Conversion)
+                .and_then(DocumentId::new)
+                .map_err(warp::reject::custom)
+        })
```
biernacki, line 115: maybe we can add to this filter also "properties" path segment, instead of repeating it in every endopoint, wdyt?
```rust
.and(warp::path("properties"))
```

discovery_engine_core/web-api/src/elastic.rs
```diff
@@ -70,98 +75,165 @@ impl ElasticState {
...
 struct Total {
+    #[allow(dead_code)]
     value: usize,
 }
+
+#[derive(Clone, Debug, Deserialize)]
+struct DocumentPropertiesResponse {
+    #[serde(default)]
+    properties: DocumentProperties,
+}
```
biernacki, line 239: we also have similar struct in our models, do we need both of them?

biernacki (commented):
looks good, just couple of questions

discovery_engine_core/web-api/src/elastic.rs
janpetschexain: i tried to keep them separate to not mix up elasticsearch internals and web-api externals. even though they are quite similar, except for the serde default, i guess it's still better to have a clear separation for type safety.

discovery_engine_core/web-api/src/models.rs
```diff
@@ -50,8 +63,64 @@ pub(crate) enum Error {
...
+        } else if id.contains('\u{0000}') {
+            Err(Error::DocumentIdContainsNul)
+        } else {
+            Ok(Self(id))
+        }
+    }
+}
+
+#[derive(Clone, Debug, Display, Serialize, Deserialize, PartialEq, Eq, Hash)]
+pub struct DocumentPropertyId(String);
```
biernacki, line 97: @acrrd mentioned that we should check this for nul char and empty, same as `UserId` and `DocumentId`

janpetschexain: in this pr we only retrieve and construct `DocumentPropertyId`s from our internal database, we never perform these checks in that case as we assume that we stored them correctly in the first place.

in the case where we have to validate `DocumentPropertyId`s coming from the outside, the code is added in the follow up pr #646, otherwise i had to mark it as dead code here.

biernacki: sure, got it

biernacki (approved):
lgtm 🚀

janpetschexain:
bors merge

PR #646, opened 2022-10-13
ET-3373 single property endpoint

janpetschexain:
**References**

- [ET-3373]
- requires #645

**Summary**

- add GET, PUT and DELETE endpoints for a single document property


discovery_engine_core/web-api/src/models.rs
```diff
@@ -96,6 +105,20 @@ impl DocumentId {
 #[derive(Clone, Debug, Display, Serialize, Deserialize, PartialEq, Eq, Hash)]
 pub struct DocumentPropertyId(String);
 
+impl DocumentPropertyId {
+    pub(crate) fn new(id: impl Into<String>) -> Result<Self, Error> {
```
rustonaut, line 109: :pushpin:  See the other comment I mentioned but we probably want to limit the format of property ids to e.g. `[a-A0-9_]+` or similar.

discovery_engine_core/web-api/src/elastic.rs
```diff
@@ -152,6 +159,64 @@ impl ElasticState {
...
     }
 
+    pub(crate) async fn get_document_property(
+        &self,
+        doc_id: &DocumentId,
+        prop_id: &DocumentPropertyId,
+    ) -> Result<Option<DocumentProperty>, Error> {
+        // https://www.elastic.co/guide/en/elasticsearch/reference/8.4/docs-get.html
+        self.query_elastic_search::<DocumentPropertyResponse>(
+            &format!("_source/{doc_id}?_source_includes=properties.{prop_id}"),
```
rustonaut, line 169: :pushpin: Doesn't this potentially allow injection attacks by choosing problematic document ids and properties.

I would say this is a more generic issue which might be better to fix with a follow up PR but wouldn't it be much better
to use an URL builder or similar to make sure added path segments and properties are appropriately url encoded.

discovery_engine_core/web-api/src/handlers.rs
```diff
@@ -302,3 +305,67 @@ pub(crate) async fn handle_delete_document_properties(
...
+#[instrument(skip(state))]
+pub(crate) async fn handle_get_document_property(
+    doc_id: DocumentId,
+    prop_id: DocumentPropertyId,
+    state: Arc<AppState>,
+) -> Result<Box<dyn Reply>, Infallible> {
+    match state.elastic.get_document_property(&doc_id, &prop_id).await {
+        Ok(Some(property)) => Ok(Box::new(DocumentPropertyResponse::new(property).to_reply()) as _),
+        Ok(None) => Ok(Box::new(StatusCode::NOT_FOUND) as _),
+        Err(Error::Elastic(error)) if matches!(error.status(), Some(StatusCode::NOT_FOUND)) => {
```
rustonaut, line 318: :art: ? I think it might be a bit nicer if we would handle elastic not found errors as part of `get_document_property`.

rustonaut (approved):
lgtm, there are some comments which likely will lead to some changes but we can easily do them in a follow up PR and I probably have to call in sick today so it's probably less friction to do them in a follow up PR.

discovery_engine_core/web-api/src/elastic.rs
janpetschexain: good point, i added a [ticket to the board](https://xainag.atlassian.net/browse/ET-3461)

janpetschexain:
bors merge

PR #648, opened 2022-10-17
Jap demo

acrrd:
Not meant to be merged.


acrrd:
I fixed the stack overflow. That was a weird bug because it was happening on or the Mutex::map or OnceCell::get_or_init calls.
I removed the used of them and now it works.
Other tests are failing because the pretokinizer is adding token to english sentences that we have in the tests.

PR #659, opened 2022-10-24
ET-3498 port config to actix

rustonaut:
- renamed `tracing` to `logging` to avoid naming collisions
   - that we use the tracing library is an implementation detail
   - similary named file for db glue code `db` and not `postrges` and `embedding` instead of `smbert`
- added (sub-)configs for `db`, `elastic` and `embedding` in  modules as well as `net` and `logging` config in existing modules
- removed `Config` trait, instead we can use bounds like `AsRef<server::Config>` or `AsRef<logging::Config>` if we need to be generic
- moved `load_config` into `load_config.rs` to avoid cyclic dependency between modules
- instead of being trait-generic over the config type and app state we now have a fixed config and app state type but with a generic extension field
  - much simpler
  - easier to add async app state creation (for `Pool<Postgres>`)
  - both have anyway mostly the same config
  - having some e.g. smbert ingestion specific options ignored is not an problem 

**References:**

- [ET-3498]

### DIFF
--- a/discovery_engine_core/web-api2/src/db.rs
+++ b/discovery_engine_core/web-api2/src/db.rs
@@ -1,0 +1,105 @@
+// Copyright 2022 Xayn AG
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, version 3.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+use secrecy::{ExposeSecret, Secret};
+use serde::Deserialize;
+use sqlx::{pool::PoolOptions, postgres::PgConnectOptions, Pool, Postgres};
+
+#[derive(Deserialize, Debug)]
+pub struct Config {
+    /// The default base url.
+    #[serde(default = "default_base_url")]
+    base_url: String,
+
+    /// Override port from base url.
+    #[serde(default)]
+    port: Option<u16>,
+
+    /// Override user from base url.
+    #[serde(default)]
+    user: Option<String>,
+
+    /// Override password from base url.
+    #[serde(default)]
+    password: Option<Secret<String>>,
+
+    /// Override db from base url.
+    #[serde(default)]
+    db: Option<String>,
+
+    /// Override default application name from base url.
+    ///
+    /// Defaults to `xayn-web-{CARGO_BIN_NAME}`.
+    #[serde(default = "default_application_name")]
+    application_name: Option<String>,
+}
+
+impl Default for Config {
+    fn default() -> Self {
+        Self {
+            base_url: default_base_url(),
+            user: None,
+            password: None,
+            db: None,
+            port: None,
+            application_name: default_application_name(),
+        }
+    }
+}
+
+fn default_base_url() -> String {
+    "postgres://user:pw@localhost:5432/xayn".into()
+}
+
+fn default_application_name() -> Option<String> {
+    option_env!("CARGO_BIN_NAME").map(|name| format!("xayn-web-{name}"))
+}
+
+impl Config {
+    pub(crate) async fn create_connection_pool(&self) -> Result<Pool<Postgres>, sqlx::Error> {
+        let options = self.create_connection_options()?;
+        PoolOptions::new().connect_with(options).await
+    }
+
+    fn create_connection_options(&self) -> Result<PgConnectOptions, sqlx::Error> {
+        let Self {
+            base_url,
+            port,
+            user,
+            password,
+            db,
+            application_name,
+        } = &self;
+
+        let mut options: PgConnectOptions = base_url.parse()?;
+
+        if let Some(user) = user {
+            options = options.username(user);
+        }
+        if let Some(port) = port {
+            options = options.port(*port);
+        }
+        if let Some(password) = password {
+            options = options.password(password.expose_secret())
+        }
+        if let Some(db) = db {
+            options = options.database(db);
+        }
+        if let Some(application_name) = application_name {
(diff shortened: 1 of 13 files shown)


discovery_engine_core/web-api2/src/db.rs
```diff
@@ -0,0 +1,107 @@
...
+use secrecy::{ExposeSecret, Secret};
+use serde::{Deserialize, Serialize};
+use sqlx::{pool::PoolOptions, postgres::PgConnectOptions, Pool, Postgres};
+
+use crate::utils::serialize_redacted_opt;
+
+#[derive(Debug, Deserialize, Serialize)]
+pub struct Config {
+    /// The default base url.
+    #[serde(default = "default_base_url")]
```
acrrd, line 24: :question: What is the point in having all this default here and in the es part? We can easily have a configuration file that we use for testing.

discovery_engine_core/web-api2/src/elastic.rs
```diff
@@ -0,0 +1,61 @@
...
+
+use secrecy::Secret;
+use serde::{Deserialize, Serialize};
+
+use crate::utils::serialize_redacted;
+
+#[derive(Debug, Deserialize, Serialize)]
+pub struct Config {
+    #[allow(dead_code)]
+    #[serde(default = "default_url")]
```
acrrd, line 23: :question: See comment on the db config.

discovery_engine_core/web-api2/src/logging.rs
```diff
@@ -14,25 +14,31 @@
...
     layer::SubscriberExt,
     util::SubscriberInitExt,
 };
 
+#[derive(Debug, Default, Deserialize, Serialize)]
+pub struct Config {
+    pub(crate) file: Option<PathBuf>,
+}
+
 static INIT_TRACING: Once = Once::new();
```
acrrd, line 32: :question: Do we still need this? with an application should be easy to see that we are calling init_tracing only once in the main.

discovery_engine_core/web-api2/tests/config.rs
```diff
@@ -0,0 +1,74 @@
...
+use std::{
+    collections::HashMap,
+    env,
+    ffi::{OsStr, OsString},
+    os::unix::prelude::OsStrExt,
+    sync::Mutex,
+};
+use trycmd::TestCases;
+
+fn env_guard(
```
acrrd, line 23: :art: 
```suggestion
fn with_env_guard(
```

discovery_engine_core/web-api2/src/logging.rs
rustonaut: Doesn't hurt to keep it, could avoid some wasted time in some integration testing situations.

discovery_engine_core/web-api2/src/db.rs
acrrd: The cases are massively different. The engine is a library where the configuration is composed of parameters for the AI that we want to change mainly for A/B testing. Having a default configuration file would have meant deploying it alongside the library and having the client giving as the path to it making it more difficult for little gain.

Here having a file alongside the binary is pretty easy and there are no sane default values that can be used in production to connect to the database and elasticsearch.

If the automatic system for env vars is not good enough we can set some manually (PREFIX_DB_URL, PREFIX_DB_PASSWORD) if needed for the infrastructure otherwise we can drop that part.
For the priority of the options seems that cli > env > configuration has the most consensus on the internet but we can invert configuration and env. The important part is that we document it.

rustonaut: > no sane default values that can be used in production

We have sane default values we do not change in production for **most** of the config fields, the fields we change to different values depending on production, testing etc. are the minority, by far.

> automatic system for env vars is not good

It is good enough, but if you want to have default values for each config field in the binary then you have to put default values on each config field which is what that code is doing. I don't understand why you have a problem with that.

>  the options seems that cli > env > configuration has the most consensus o

that is not what I mean, what I mean is with local override fiels like `config.test.toml` or similar which should be treated with more priority then e.g. some default `.env` but with less priority then e.g. `.env.local` or manually set env variables. 

The reason why you might want to have a override file is that having a config in tree and changing it for local testing is very error pone. So you want a solution without override files. But juggling multiple config files e.g. one not commited for local testing, one committed for testing without your changes, one with production defaults (or similar) can also get nasty especially the more configs there are in the files. So I would want to have as little configs in eachf file as possible especially given that most rarely change and are most times independent of the context. Which means having defaults in the program, which is seems to also be industry consensus given that this is how most programs behave.

Having default in the program also allows you to e.g. use a different default logging level for debug builds.

>  file alongside the binary is pretty easy 

it's not much easier then having it in a app, in both cases it's pretty trivial

discovery_engine_core/web-api2/tests/cmd/cli_overrides.auto.toml
```diff
@@ -0,0 +1,36 @@
...
+    "max_body_size": 64
+  },
+  "elastic": {
+    "url": "http://localhost:3219",
+    "user": "tic",
+    "password": "[REDACTED]",
+    "index_name": "other_index"
+  },
+  "db": {
+    "base_url": "postgres://user:pw@localhost:5432/xayn",
```
biernacki, line 21: should we maybe hide from stdout the password in the url too?

biernacki (approved):
lgtm, but please maybe wait for @acrrd approval

discovery_engine_core/web-api2/tests/cmd/cli_overrides.auto.toml
rustonaut: I hadn't considered it as I think any password (besides the hard coded one for testing) should never be in the url :thinking: 

I wouldn't want to hide the whole url as it is usefull for debugging but I guess we can hide part of it :thinking:

discovery_engine_core/web-api2/src/db.rs
rustonaut: Two maybe relevant things:

- we should be able to derive `Default` based on `serde`
- there is an open issue/in-progress PR for `default_value` which would make it _way_ more readable/concise

rustonaut: :pushpin: I noted this discussion down in a followup ticket.

rustonaut:
bors merge

PR #660, opened 2022-10-25
[Merged by Bors] - rustfmt groups imports

acrrd:
This enables the `group_imports` option. With this rustfmt automatically groups the imports in `std`, `external`, and `crate.
This is almost what we already wanted but we lose control of some things:
* it is not able to differentiate between workspace and external so all our crates will go in that section,
* it does not understand local import: 
```rust
mod foo {
    struct Foo;
}

use foo::Foo;
use rand::random;
```
instead of having a separate group for `foo`. 
* It does not respect groups that we create manually
```rust
use tokio;

use itertools;
```
the imports will be merged.

Looking at the changeset we are not too bad at maintaining this style. Most of the changes are the position of `crate` wr.t. to `self` or `super`, fusing groups, and moving our workspace crates to the external one.
But this will be our only option to enforce this style with a tool.

https://rust-lang.github.io/rustfmt/?version=v1.5.1&search=#group_imports
https://github.com/rust-lang/rustfmt/issues/4709
https://github.com/rust-lang/rustfmt/issues/4693

### DIFF
--- a/discovery_engine_core/ai/ai/src/coi/context.rs
+++ b/discovery_engine_core/ai/ai/src/coi/context.rs
@@ -211,12 +211,11 @@ mod tests {
     use ndarray::arr1;
     use xayn_discovery_engine_test_utils::assert_approx_eq;
 
+    use super::*;
     use crate::{
         coi::point::tests::{create_neg_cois, create_pos_cois},
         utils::SECONDS_PER_DAY_F32,
     };
-
-    use super::*;
 
     #[test]
     fn test_has_enough_cois() {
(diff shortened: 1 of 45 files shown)


janpetschexain:
looks good imo as it helps with consistency and readability.

> it is not able to differentiate between workspace and external so all our crates will go in that section,

that not perfect, but more like a minor inconvenience.

> it does not understand local import:

it's anyway more clear to use `use self::foo::Foo;` for local imports instead of `use foo::Foo;` to avoid any ambiguity for the reader. so it's good that rustfmt enforces the use of better style.

janpetschexain (approved):
lgtm

acrrd:
bors merge

PR #662, opened 2022-10-25
[Merged by Bors] - ET-3470 restrict format of valid id

krajewskiML:
### References:
- [ET-3470](https://xainag.atlassian.net/browse/ET-3470?atlOrigin=eyJpIjoiNTYxMDk3MjM2YzE1NDEzNmFhYzY4YTMxY2FmOWFmYWQiLCJwIjoiaiJ9)

### Changes
- Each `DocumentId` provided to the ingestion system is checked against regex, if it does not match the document is not added and the id is returned in the `IngestionError`


discovery_engine_core/web-api/src/lib.rs
```diff
@@ -19,6 +19,9 @@ mod routes;
 mod state;
 mod storage;
 
+#[macro_use]
+extern crate lazy_static;
```
janpetschexain, line 23: :wrench: from rust edition 2018 onwards `#[macro_use]` and `extern crate` shouldn't be necessary anymore. you can directly import the macro via `use lazy_static::lazy_static;` in the module where it's needed (not necessarily at `crate` level).

discovery_engine_core/web-api/src/bin/ingestion.rs
```diff
@@ -418,8 +418,16 @@ async fn handle_post_documents(
...
     let start = Instant::now();
 
-    let (documents, failed_documents) = body
+    let (valid_ids_docs, mut invalid_ids) = body
         .documents
+        .into_iter()
+        .map(|document| match document.id.is_id_valid() {
+            true => Ok(document),
+            false => Err(document.id),
+        })
```
janpetschexain, line 427: ❓ instead of handling this error during the ingestion, we could already handle it during the routing. for example in `document_properties_path()` where the id string is extracted and the id type is created via `DocumentId::new()`, we could add the regex check within the `DocumentId::new()` consturctor as well. then this would also reject the invalid request early on before much work is done server-side. wdyt?

janpetschexain (commented):
:wrench: please also add the regex check to the `DocumentPropertyId` and `UserId`.

discovery_engine_core/web-api/Cargo.toml
```diff
@@ -15,7 +15,9 @@ dotenvy = "0.15.5"
 envconfig = "0.10.0"
 futures = { workspace = true }
 itertools = { workspace = true }
+lazy_static = "1.4.0"
```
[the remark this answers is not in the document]
krajewskiML: Alright, I will check crates more thoroughly in the future

discovery_engine_core/web-api/src/lib.rs
krajewskiML: Okay, I went with the tutorial on this one, must have been obsolete. I will change it

discovery_engine_core/web-api/src/bin/ingestion.rs
krajewskiML: I was thinking about this solution, the question is basically whether we want to refuse whole batch of docs because some (maybe all) document IDs are invalid or do we want to process only the ones with the proper ID (if everything else won't fail). I think that in most cases either all IDs will be valid or all IDs will be invalid. Taking that into the account, the approach proposed by you is more ergonomic, but then should we just return `BAD_REQUEST` if we spot single wrong ID and stop the ingestion at this point?

rustonaut: IMHO it's okay the reject the whole batch with a single BAD_REQUEST for the ingestion endpoint if it contains invalid document ids. I also don't think keeping two different error sub-variants for "empty" and "bad chars" makes sense.

There is a clear definition about how the use the API and there is only that much we should do for the case of a obvious wrong API usage. In case of a fully public API like for personalization we probably have to anyway reduce the amount of error details we report but that's a bit off topic.

So preferable all ids would:

- implement `FromStr` (for usage in url path)
- have a check in their deserialization logic for validity 

The later is a bit non obvious to do, you need roughly something like:

```rust
impl<'de> Deserialize<'de> for DocumentId {
    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
    where
        D: Deserializer<'de>,
    {
        let inner = String::deserialize(deserializer)?;
        if is_valid_id(&inner) {
            Ok(Self(inner))
        } else {
            // you probably need  <D::Error as serde::de::Error>::invalid... instead for it to compile
            Err(D::Error::invalid_value(Unexpected::Str(&inner), &"Id must match regext [...]+"))
        }
    }
}
```

With this we:

1. make sure any xxxId is always well formed
2. as long as we pass deserialization errors on (in actix we currently do) the user still gets an err with details for the first malformed document id

discovery_engine_core/web-api/src/lib.rs
rustonaut: :art: Using `once_cell` is a bit nicer (nice enough for it to be in nightly std).

I.e.:

```
use once_cell::sync::Lazy;
static RE: Lazy<Regex> = Lazy::new(||  .... );
```

discovery_engine_core/web-api/Cargo.toml
```diff
@@ -15,7 +15,9 @@ dotenvy = "0.15.5"
...
 ndarray = { workspace = true }
+regex = "1"
```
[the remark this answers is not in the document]
krajewskiML: Ok

discovery_engine_core/web-api/src/bin/ingestion.rs
krajewskiML: I decided to go with the deserialization approach suggested by @rustonaut

discovery_engine_core/web-api/src/models.rs
```diff
@@ -112,7 +112,24 @@ impl DocumentId {
     }
 }
 
-#[derive(Clone, Debug, Display, Serialize, Deserialize, PartialEq, Eq, Hash, AsRef)]
+impl<'de> Deserialize<'de> for DocumentId {
```
janpetschexain, line 115: ❓ this looks quite elegant, but if i'm not mistaken, this is not yet called anywhere in the routes? currently the parameter extraction reads the id as a `String`,  then some url decoding and calls the `new()` constructor, but deserialization into `DocumentId` is not called.

the deserialization is called when we get data from the database though, but it seems unnecessary to check validity again if we assume that we only store valid data in the first place.

it seems that we either have to change the routing to extract the id parameter as a `DocumentId` instead of a `String` (not sure if that's doable) or we need to add the `is_id_valid()` check to the `DocumentId::new()` (and remove it from the serde as this is to a certain extent in vain).

rustonaut: It is called with the ingestion endpoint and any other endpoint taking ids in a json blob.

In actix parsing from path uses `FromStr`.

The postgres database uses `Encode` and sqlx transparent, so it shouldn't affect this.

But when I proposed this I had forgotten that we also get document ids from elastic instead of just giving them to elastic :face_exhaling: 

Maybe we need two `DocumentId` types or something like this. 

I which we could have context dependent serialization etc. similar to how you can use context objects in scala for some things.

rustonaut: > In actix parsing from path uses FromStr.

To avoid confusion with `ToString` we probably should not include the `urldecode` part in `FromStr` but provide some `Urldecode` wrapper which does so, like `Path<(Urldecode<DocumentId>, Urldecode<DocumentPropertyId>)>`. Maybe there is already a helper type like that idk.

discovery_engine_core/web-api/openapi.yaml
```diff
@@ -324,9 +324,11 @@ paths:
 components:
   schemas:
     DocumentId:
-      description: A document id can be any non-empty UTF-8 string that does not contain the null byte.
+      description: |-
+        A document id can be any non-empty UTF-8 string that does not contain the null byte.
+        It may consist of digits, Latin letters, underscores, colons, minus signs, at signs, and dots.
```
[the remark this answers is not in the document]
krajewskiML: Good point

discovery_engine_core/web-api/src/models.rs
```diff
@@ -94,6 +86,14 @@ impl From<DocumentId> for String {
     }
 }
 
+fn is_id_valid(id: &String) -> bool {
```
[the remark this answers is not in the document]
krajewskiML: okay

discovery_engine_core/web-api/openapi.yaml
```diff
@@ -324,11 +324,15 @@ paths:
...
       type: string
-      example: "document id"
+      example: "document_id"
     DocumentPropertyId:
-      description: A document property id can be any non-empty UTF-8 string that does not contain the null byte.
+      description: |-
+        A document property id can be any non-empty UTF-8 string that does not contain the null byte.
+        It may consist of digits, Latin letters, underscores, colons, minus signs, at signs, and dots.
       type: string
       example: "document property id"
```
janpetschexain, line 337: 🔧 space is not valid anymore
```suggestion
      example: "document_property_id"
```

discovery_engine_core/web-api/openapi.yaml
```diff
@@ -368,7 +372,9 @@ components:
         properties:
           $ref: '#/components/schemas/DocumentProperties'
     UserId:
-      description: A user id can be any non-empty UTF-8 string that does not contain the null byte.
+      description: |-
+        A user id can be any non-empty UTF-8 string that does not contain the null byte.
+        It may consist of digits, Latin letters, underscores, colons, minus signs, at signs, and dots.
       type: string
       example: "user id"
```
janpetschexain, line 379: 🔧 space is not valid anymore
```suggestion
      example: "user_id"
```

discovery_engine_core/web-api/src/models.rs
```diff
@@ -94,14 +77,17 @@ impl From<DocumentId> for String {
...
 
-        if id.is_empty() {
-            Err(Error::DocumentIdEmpty)
-        } else if id.contains('\u{0000}') {
-            Err(Error::DocumentIdContainsNul)
+        if !is_valid_id(&id) {
+            Err(Error::InvalidDocumentId { invalid_id: id })
         } else {
             Ok(Self(id))
         }
```
janpetschexain, line 93: :art: since only one if-branch remains, we could simplify the doubled negation. same for the other id constructors.
```suggestion
        if is_valid_id(&id) {
            Ok(Self(id))
        } else {
            Err(Error::InvalidDocumentId { invalid_id: id })
        }
```

janpetschexain (commented):
lgtm, the spec still needs to be fixed though

rustonaut (approved):

discovery_engine_core/web-api/src/models.rs
```diff
@@ -112,17 +98,34 @@ impl DocumentId {
...
 
 impl DocumentPropertyId {
     pub fn new(id: impl Into<String>) -> Result<Self, Error> {
         let id = id.into();
 
-        if id.is_empty() {
-            Err(Error::DocumentPropertyIdEmpty)
-        } else if id.contains('\u{0000}') {
-            Err(Error::DocumentPropertyIdContainsNul)
+        if !is_valid_id(&id) {
```
janpetschexain, line 127: :art: 

> since only one if-branch remains, we could simplify the doubled negation.

same here.

discovery_engine_core/web-api/src/models.rs
```diff
@@ -258,19 +280,36 @@ impl UserInteractionError {
...
-    pub(crate) fn new(id: impl AsRef<str>) -> Result<Self, Error> {
-        let id = id.as_ref();
+    pub(crate) fn new(id: impl Into<String>) -> Result<Self, Error> {
+        let id = id.into();
 
-        if id.is_empty() {
-            Err(Error::UserIdEmpty)
-        } else if id.contains('\u{0000}') {
-            Err(Error::UserIdContainsNul)
+        if !is_valid_id(&id) {
```
janpetschexain, line 290: :art: 

> since only one if-branch remains, we could simplify the doubled negation.

same here.

janpetschexain (approved):

krajewskiML:
bors merge

krajewskiML:
bors merge

PR #664, opened 2022-10-25
[Merged by Bors] - ET-3111 remove storage feature flag

janpetschexain:
**Reference**

- [ET-3111]

**Summary**

each change is handled in a separate commit:
- remove `!cfgFeatureStorage` related code from the document, feed, search & system managers
- cleanup the remaining code of the event handler
- cleanup `cfg(not(feature = "storage"))` related code in the ffi
- move `storage2` code to the `storage` module

[ET-3111]: https://xainag.atlassian.net/browse/ET-3111?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


discovery_engine/lib/src/domain/event_handler.dart
```diff
@@ -213,126 +155,32 @@ class EventHandler {
...
-      sourcePreferenceRepository,
-      availableSources,
-    );
-    _searchManager = SearchManager(
-      engine,
-      activeSearchRepository,
-      documentRepository,
-      activeDataRepository,
-      engineStateRepository,
+    final dartMigrationData = await DartMigrationData.fromRepositories(
```
biernacki, line 162: maybe it makes sense to rename it?

```suggestion
    final dartMigrationData = await DartMigrationData.fromDirectoryPath(
```

discovery_engine/lib/src/domain/engine/engine.dart
```diff
@@ -88,56 +58,31 @@ abstract class Engine {
...
-  /// Gets the next batch of feed documents.
-  Future<List<DocumentWithActiveData>> getFeedDocuments(
-    List<HistoricDocument> history,
-    List<SourceReacted> sources,
-  );
-
-  /// Restores the feed documents, ordered by their global rank (timestamp & local rank).
-  Future<List<DocumentWithActiveData>> restoreFeed();
+  /// Restores the documents which have been fed, i.e. the current feed.
+  Future<List<DocumentWithActiveData>> fed();
```
biernacki, line 76: I think the previous name was more descriptive, `engine.fed()` is a bit vague, why you've changed it?

discovery_engine/lib/src/domain/models/active_data.dart
```diff
@@ -38,12 +41,17 @@ class DocumentWithActiveData with EquatableMixin {
...
   @HiveField(0)
   final Embedding smbertEmbedding;
   @HiveField(1)
   final Map<DocumentViewMode, Duration> viewTime;
 
-  ActiveDocumentData(this.smbertEmbedding) : viewTime = {};
+  ActiveDocumentData()
+      :
+        // ignore: deprecated_member_use_from_same_package
+        smbertEmbedding = Embedding(Float32List(0)),
```
biernacki, line 53: so we don't use the data stored in hive during migration? could you help me understand why we have a dummy value here?

discovery_engine/lib/src/domain/engine/engine.dart
janpetschexain: mainly to match the `searched` and related naming style. if the new names are vague, i'll rename them to `restore_...`.

discovery_engine/lib/src/domain/models/active_data.dart
janpetschexain: i was under the impression that the generated code for hive on `ActiveDocumentData` uses its own constructor, but i just double checked and it uses this one. i'll revert this and construct empty embeddings in the non-migration related code, basically i just wanted to avoid that.

discovery_engine/lib/src/ffi/types/document/document.dart
```diff
@@ -73,25 +64,18 @@ class DocumentFfi with EquatableMixin {
...
-        int batchIndex = 0,
-    bool isSearched = false,
-  }) =>
-      Document(
+  Document toDocument({bool isSearched = false}) => Document(
         documentId: id,
         stackId: stackId,
         resource: resource,
-        batchIndex: batchIndex,
+        batchIndex: 0, // unused
```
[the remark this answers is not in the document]
janpetschexain: i don't know, i thought we have to keep deprecated fields for db read/write consistency? if that's not the case we can remove a bunch of fields in various places.

discovery_engine/lib/src/ffi/types/engine.dart
```diff
@@ -420,24 +316,6 @@ class DiscoveryEngineFfi implements Engine {
     return resultVoidStringFfiAdapter.consumeNative(result);
   }
 
-  @override
-  Future<List<DocumentWithActiveData>> deepSearch(
```
biernacki, line 424: We don't use deep search anymore? It was used for a PoC, on an older app branch, but I would double check with @LeifXayn or @frankpepermans maybe?

discovery_engine/lib/src/ffi/types/trending_topic.dart
```diff
@@ -15,58 +15,40 @@
...
 import 'package:xayn_discovery_engine/src/domain/models/trending_topic.dart';
-import 'package:xayn_discovery_engine/src/domain/models/unique_id.dart';
 import 'package:xayn_discovery_engine/src/ffi/genesis.ffigen.dart';
 import 'package:xayn_discovery_engine/src/ffi/load_lib.dart' show ffi;
-import 'package:xayn_discovery_engine/src/ffi/types/embedding.dart';
 import 'package:xayn_discovery_engine/src/ffi/types/string.dart' show StringFfi;
 import 'package:xayn_discovery_engine/src/ffi/types/uri.dart' show UriFfi;
-import 'package:xayn_discovery_engine/src/ffi/types/uuid.dart';
 
 class TrendingTopicFfi with EquatableMixin {
```
biernacki, line 24: I would also check if we use trending topics anywhere or do we still have plans to use them, maybe we could remove them? cc @LeifXayn, @frankpepermans

discovery_engine/test/repository/hive_active_document_repo_test.dart
```diff
@@ -84,18 +80,6 @@ Future<void> main() async {
         await repo.update(id1, data);
       });
 
-      test('existing smbert embedding', () async {
-        final emb = await repo.smbertEmbeddingById(id1);
-        expect(emb, equals(data.smbertEmbedding));
-      });
```
biernacki, line 90: I would keep the test to read the data from db, because we need that for migration. Other tests we can remove I think.

discovery_engine/lib/src/ffi/types/engine.dart
janpetschexain: the deep search still exists, it was moved to `search_by_id()` for the rust db version, the old name `deep_search()` was kept for the dart db version. also the dart event names and arguments didn't change, so this shouldn't be a breaking change for the app.

discovery_engine/lib/src/ffi/types/trending_topic.dart
janpetschexain: good question. if we want to remove them, then it's probably better to do so in a follow up pr.

discovery_engine_core/bindings/src/lib.rs
```diff
@@ -40,48 +40,32 @@ use std::path::Path;
...
 )]
 impl XaynDiscoveryEngineAsyncFfi {
     /// Initializes the engine.
-    #[allow(clippy::box_collection)]
-    pub async fn initialize(
-        config: Box<InitConfig>,
-        state: Option<Box<Vec<u8>>>,
-        history: Box<Vec<HistoricDocument>>,
-        sources: Box<Vec<WeightedSource>>,
-        dart_migration_data: Option<Box<DartMigrationData>>,
```
biernacki, line 70: do we use migration data somewhere else?

janpetschexain: no, only in the `Engine::from_config()` constructor.

biernacki (commented):
looks good, left some comments, nothing major

discovery_engine_core/bindings/src/lib.rs
biernacki: I saw it was moved into the config 👍

discovery_engine/lib/src/ffi/types/trending_topic.dart
biernacki: yes yes, if we decide to remove it it should be done in a separate PR

discovery_engine/lib/src/ffi/types/document/document.dart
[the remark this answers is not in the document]
biernacki: I would check how type adapters look before and after removing a field, but I think we could remove the unused fields and type adapters will just disregard them, so it should be safe to do

discovery_engine/lib/src/domain/engine/engine.dart
biernacki: yeah, I would add `restore*` in front, it reads better imho

discovery_engine_core/bindings/src/types/document/document.rs
```diff
@@ -18,7 +18,28 @@ use std::ptr::addr_of_mut;
 
 use uuid::Uuid;
 
-use xayn_discovery_engine_core::document::{Document, Embedding, NewsResource, UserReaction};
+use xayn_discovery_engine_core::{
+    document::{Id, NewsResource, UserReaction},
+    stack::Id as StackId,
+};
+
+pub struct Document {
```
finiteprods, line 26: :grey_question: why is it beneficial to introduce this here?

discovery_engine_core/bindings/src/types/init_config.rs
```diff
@@ -167,6 +141,19 @@ pub unsafe extern "C" fn init_config_place_of_de_config(
     unsafe { addr_of_mut!((*place).de_config) }
 }
 
+/// Returns a pointer to the `log_file` field of a configuration.
+///
+/// # Safety
+///
+/// The pointer must point to a valid [`InitConfig`] memory object,
+/// it might be uninitialized.
```
finiteprods, line 149:
```suggestion
/// which might be uninitialized.
```
similarly elsewhere in this and other files... I don't know if you want to change them all here in this pr, up to you

discovery_engine_core/core/src/engine.rs
```diff
@@ -1887,13 +1565,13 @@ pub(crate) mod tests {
...
-                .unwrap()
-                .0;
-        }
+        engine.storage = SqliteStorage::init_storage_system(None, None, &|_| None)
+            .await
+            .unwrap()
+            .0;
+        // This triggers the trusted sources stack to also fetch articles
+        engine.endpoint_config.trusted_sources = Arc::new(RwLock::new(vec!["example.com".into()]));
+        engine.endpoint_config.excluded_sources = Arc::new(RwLock::new(Vec::new()));
```
finiteprods, line 1574: :grey_question: why was this not needed before?

discovery_engine_core/core/src/storage.rs
```diff
@@ -59,6 +60,62 @@ impl From<MalformedBytesEmbedding> for Error {
...
+    /// Embedding from smbert.
+    pub smbert_embedding: Option<Embedding>,
+
+    /// Reaction.
+    pub reaction: UserReaction,
+
+    /// Resource this document refers to.
+    pub resource: NewsResource,
+
+    // If true the document is part of the search OR feed
```
finiteprods, line 102:
```suggestion
    // If true the document is part of the search OR feed.
```

discovery_engine_core/core/src/storage.rs
```diff
@@ -59,6 +60,62 @@ impl From<MalformedBytesEmbedding> for Error {
...
+    pub is_active: bool,
+
+    // If true the document is/was part of the search
```
finiteprods, line 105:
```suggestion
    // If true the document is/was part of the search.
```

discovery_engine_core/core/src/storage.rs
```diff
@@ -59,6 +60,62 @@ impl From<MalformedBytesEmbedding> for Error {
...
+}
+
+/// Represents a result from a query.
+#[derive(Clone, Debug)]
+pub struct MigrationDocument {
+    /// Unique identifier of the document.
+    pub id: document::Id,
+
+    /// Stack from which the document has been taken.
+    /// [`stack::Id::nil()`] is used for documents which are not from a stack
```
finiteprods, line 90:
```suggestion
    /// [`stack::Id::nil()`] is used for documents which are not from a stack.
```

discovery_engine_core/core/src/storage/models.rs
```diff
@@ -0,0 +1,177 @@
...
+            id: view.document_id,
+            stack_id: view.stack_id.unwrap_or_default(),
+            smbert_embedding: view.embedding,
+            reaction: view.user_reaction,
+            resource: (view.news_resource, view.newscatcher_data).into(),
+        }
+    }
+}
+
+/// Represents a news that is delivered by an external content API.
```
finiteprods, line 118:
```suggestion
/// Represents news that is delivered by an external content API.
```

discovery_engine_core/core/src/storage/sqlite/dart_migrations.rs
```diff
@@ -24,15 +24,16 @@ use crate::{
...
+        MigrationDocument,
         Storage,
     },
-    storage2::{DartMigrationData, MigrationDocument},
 };
 
 use super::SqliteStorage;
 
-/// Add the data from the  dart->rust/sqltie migration to the prepared database.
+/// Add the data from the  dart->rust/sqlite migration to the prepared database.
```
finiteprods, line 36:
```suggestion
/// Add the data from the dart -> rust/sqlite migration to the prepared database.
```

finiteprods (commented):
nice to see the code reduction payoff :+1: only some minor comments and a few questions

discovery_engine/lib/src/ffi/types/document/document.dart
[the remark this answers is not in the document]
janpetschexain: removing the field causes the adapter to change the read and write order. if the removed field index leaves a gap in the consecutive hive numbering, then this even causes out of bounds indexing. i also couldn't find helpful information in the hive documenation, i guess we have to carry these fields around until we completely remove the hive db.

.env
```diff
@@ -27,5 +27,5 @@ RUST_NIGHTLY=nightly-2022-09-22
 # cargo install will install CLI tools into this directory
 CARGO_INSTALL_ROOT=cargo-installs
 
-# by default enable the feature
-XAYN_DE_FEATURES=storage
+# by default enable these features
+XAYN_DE_FEATURES=""
```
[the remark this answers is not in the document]
janpetschexain: not really, i'll remove it to simplify the scripts 👍  if we need it again, we can always reintroduce it.

discovery_engine_core/bindings/src/types/document/document.rs
janpetschexain: the `core::document::Document` has the embedding field, which is not used at the dart side anymore. hence i introduced this new type to only transfer the required fields. since this is a ffi detail, i added it to `bindings` instead of `core`.

discovery_engine_core/core/src/engine.rs
janpetschexain: it was done before as well, it just moved out of the `InitConfig`.

finiteprods: ah I see, ok thanks

biernacki (approved):

janpetschexain:
bors merge

PR #665, opened 2022-10-25
ET-3499 Add Actix Web Json parsing configuration

MichaelKMalak:
**Reference**

- [ET-3499]

**Summary**

set up JsonConfig type in actix using the net.max_body_size config then add it to the actix-web App::new()...  function chain in server.rs

[ET-3499]: https://xainag.atlassian.net/browse/ET-3499

### DIFF
--- a/discovery_engine_core/web-api2/src/embedding.rs
+++ b/discovery_engine_core/web-api2/src/embedding.rs
@@ -17,7 +17,7 @@ use std::path::PathBuf;
 use serde::{Deserialize, Serialize};
 
 #[derive(Debug, Deserialize, Serialize)]
-pub struct Embedding {
+pub struct Config {
     #[allow(dead_code)]
     #[serde(default = "default_vocabulary")]
     vocabulary: PathBuf,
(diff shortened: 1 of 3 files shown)


discovery_engine_core/web-api2/src/server.rs
```diff
@@ -77,11 +78,21 @@ where
         let addr = config.bind_address();
         init_tracing(config.log_file());
         let app_state = web::Data::new(A::AppState::try_from(config)?);
+        let json_config = web::JsonConfig::default()
+            .limit(524288)
+            .content_type(|mime| mime == mime::TEXT_PLAIN)
+            .error_handler(|err, _req| {
+                InternalError::from_response(err, HttpResponse::Conflict().finish()).into()
+            });
```
rustonaut, line 86: I'm not sure why you also added this lines from the example.

Given the documentation of actix:

- `content_type` allows you to configure content types we do accept as json bodies, we don't have legacy client sending the `text/plain` as content type for json. So we don't need this.

- `error_handler` allows you do customize the error handling for json parsing failure. We don't need this (we have a middleware doing this uniformly for all "simple" errors)

So basically we just need `JsonConfig::default().limit(....);`.

PR #666, opened 2022-10-25
[Merged by Bors] - ET-3499 Add Actix Web Json parsing configuration 

MichaelKMalak:
**Reference**

- [ET-3499]

**Summary**

set up JsonConfig type in actix using the net.max_body_size config then add it to the actix-web App::new()...  function chain in server.rs

[ET-3499]: https://xainag.atlassian.net/browse/ET-3499

### DIFF
--- a/discovery_engine_core/web-api2/src/server.rs
+++ b/discovery_engine_core/web-api2/src/server.rs
@@ -21,7 +21,7 @@ use serde::{de::DeserializeOwned, Serialize};
 use tracing::error;
 
 use actix_web::{
-    web::{self, ServiceConfig},
+    web::{self, JsonConfig, ServiceConfig},
     App,
     HttpResponse,
     HttpServer,
@@ -68,12 +68,14 @@ where
         let addr = config.net.bind_to;
         init_tracing(config.as_ref());
 
+        let json_config = JsonConfig::default().limit(config.net.max_body_size);
         let app_state = AppState::create(config).await?;
         let app_state = web::Data::new(app_state);
 
         HttpServer::new(move || {
             App::new()
                 .app_data(app_state.clone())
+                .app_data(json_config.clone())
                 .service(web::resource("/health").route(web::get().to(HttpResponse::Ok)))
                 .configure(A::configure)
                 .wrap_fn(wrap_non_json_errors)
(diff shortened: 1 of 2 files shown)


discovery_engine_core/web-api2/src/server.rs
```diff
@@ -67,12 +67,14 @@ where
         let addr = config.net.bind_to;
         init_tracing(config.as_ref());
 
+        let json_config = JsonConfig::default().limit(config.net.max_body_size.try_into().unwrap());
```
rustonaut, line 70: Looks good through given that `max_body_size` only exist to be used in the config it is probably better to change the type of `max_body_size` to match the one `limit` expects as a small improvement.

---

Not specific to this situation but I personally prefer to use `.try_into().unwrap_or(usize::MAX)` i.e. max-clamp the value in situations where:

- you can't change the input (we can change it here)
- it only fails with absurdly to high values
- there is no major problem with using a different value 

through the last point is often not simple to determine and doing it in some situations could not only be confusion for someone later down the line but also might even be a security vulnerability. You could say it's the choice between  "going on with a slightly changed value because crashing is bad" or "aborting because a changed value could have unexpected consequences".

Either way in this case we can just change the input (`max_body_size`) to match the integer type we need and avoid any problematic decisions entirely.

MichaelKMalak: thanks for explaining. here is the new change a104f45d5f9cc58d363b3e88df42103c34c0444f

rustonaut (approved):
lgtm

discovery_engine_core/web-api2/src/server.rs
```diff
@@ -68,12 +68,14 @@ where
...
         init_tracing(config.as_ref());
 
+        let json_config = JsonConfig::default().limit(config.net.max_body_size);
         let app_state = AppState::create(config).await?;
         let app_state = web::Data::new(app_state);
 
         HttpServer::new(move || {
             App::new()
                 .app_data(app_state.clone())
+                .app_data(json_config.clone())
```
frankpepermans, line 78: Not that clone is wrong, but isn't to_owned more idiomatic Rust?

rustonaut: no, `to_owned()` is ideomatic for converting types like `str` to their "owned" representation (`String`). I.e. it's meant for cases where `clone` doesn't work because you are not simply cloning the type but converting it to a different type.

MichaelKMalak:
bors ping

MichaelKMalak:
bors merge

PR #669, opened 2022-10-25
[Merged by Bors] - ET-3516 db migration files in web api2 (4)

rustonaut:
- Copies the `migrations` folder from `web-api` to `web-api2`. 
  - now placed under `web-api2/migrations` which is the default folder location from `sqlx`
- Run db migrations on startup by default.
- Migrations can be skipped by setting `db.skip_migrations` to `true`.
- Also encapsulate the `Pool<Postgres>` type into a `Database` type.

**References:**

- [ET-3516]
- based on 
  - #668 
  - #667
  - #659

[ET-3516]: https://xainag.atlassian.net/browse/ET-3516?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #668, opened 2022-10-25
ET-3515 split into state and routes (3)

rustonaut:
This only moves some code around and doesn't touch anything besides that.

**References:**

- [ET-3515]
- based on 
  - #667
  - #659

PR #667, opened 2022-10-25
[Merged by Bors] - ET-3506 load smbert into app state (2)

rustonaut:
Adds CoiConfig to config and CoiSystem, Embedder to application state.

**References:**

- [ET-3506]
- based on #659 

[ET-3506]: https://xainag.atlassian.net/browse/ET-3506?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #670, opened 2022-10-26
[Merged by Bors] - ET-3518 error handling improvements (5)

rustonaut:
- log errors before rendering them
- update openapi.yaml
  - using camel case for error kinds allows us to use `stringify!($typename)` in a macro to easily implement error simple types
  - all details are now under `details:` this makes implementation a bit easier and avoid name collision problems between error details and additional base error fields
- added implementations for all common errors 
  - `InternalError` uses `anyhow::Error`
- `NotEnoughInteractions` is now `404 Not Found`, it's returned if personalized documents do not yet exist because there where not interactions yet, so 404 is more appropriate then 422 which was designed for post/put/patch requests with a body.
  

**References:**

- [ET-3518]
- based on:
  - #669
  - #668
  - #667
  - #659

[ET-3518]: https://xainag.atlassian.net/browse/ET-3518?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine_core/web-api/openapi.yaml
+++ b/discovery_engine_core/web-api/openapi.yaml
@@ -1,7 +1,7 @@
 openapi: 3.0.3
 info:
   title: Personalization API
-  version: 1.0.0-rc3
+  version: 1.0.0-rc4
   description: |-
     # Overview
     The personalization system is divided into two parts: one to manage documents and one to manage users.
@@ -379,10 +379,17 @@ components:
       example: "user_id"
     BaseError:
       type: object
+      required: [request_id, kind]
       properties:
         request_id:
           description: Request ID optionally generated from the service. It can be communicated to xayn to help debugging.
           type: string
+        kind:
+          description: What kind of error this is.
+          type: String
+        details:
+          description: Additional error details. Might differ depending on debug options.
+          type: object
     PersonalizedDocumentData:
       type: object
       required: [id, score]
@@ -409,7 +416,7 @@ components:
           properties:
             kind:
               type: string
-              enum: [not_enough_interactions]
+              enum: [NotEnoughInteractions]
     UserInteractionType:
       type: string
       enum: [positive]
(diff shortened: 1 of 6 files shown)


discovery_engine_core/web-api/openapi.yaml
```diff
@@ -454,16 +461,20 @@ components:
...
-            documents:
-              type: array
-              items:
-                type: object
-                required: [id]
-                properties:
-                  id:
-                    $ref: '#/components/schemas/DocumentId'
+            details:
+              type: object
```
finiteprods, line 467: :grey_question: why is it necessary to introduce an extra level of indirection in `details`?

discovery_engine_core/web-api2/src/error/common.rs
```diff
@@ -0,0 +1,97 @@
...
+use crate::impl_application_error;
+
+use super::application::ApplicationError;
+
+/// The requested document was not found.
+#[derive(Debug, Error, Display, Serialize)]
+pub struct DocumentNotFound;
+impl_application_error!(DocumentNotFound => NOT_FOUND);
+
+/// The requested property was not found
```
finiteprods, line 30: :art: similarly for other docstrings below
```suggestion
/// The requested property was not found.
```

finiteprods (approved):
lgtm, included a question.

discovery_engine_core/web-api/openapi.yaml
rustonaut: It makes a lot of things simpler:

- keep the fields of the base error and specific errors separate, so accidental field overlaps are impossible
- allows easily to specify (in openapi.yaml) a field for more details as an free-form object (as we don't want to specify additional error details which are not meant to be processed by a client but are purely informational for a user debugging issues; the reason we don't want to specify them is so that they can freely change in both availability and formatting)
- have a easy "cut" point where we can reduce the amount of details an error contains for fully public production deployments
- simpler to implement in code

discovery_engine_core/web-api2/src/error/common.rs
rustonaut: Adding a `.` is in conflict with error formatting using `Display` generated by `display_doc`.

rustonaut:
bors merge

PR #671, opened 2022-10-26
[Merged by Bors] - ET-3519 actix common models (6)

rustonaut:
- copied models from web-api
- implemented `FromStr` for Id types so that they can be used directly in `Path`
- [x] propagate changes from `Id` validation 

**References:**

- [ET-3519]
- based on:
  - #670 
  - #669
  - #668
  - #667
  - #659

[ET-3519]: https://xainag.atlassian.net/browse/ET-3519?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine_core/web-api2/src/models.rs
+++ b/discovery_engine_core/web-api2/src/models.rs
@@ -121,7 +121,7 @@ impl AiDocument for PersonalizedDocumentData {
         &self.id
     }
 
-    fn smbert_embedding(&self) -> &Embedding {
+    fn bert_embedding(&self) -> &Embedding {
         &self.embedding
     }
 }
--- a/discovery_engine_core/web-api2/src/models.rs
+++ b/discovery_engine_core/web-api2/src/models.rs
@@ -19,7 +19,7 @@ use serde::{Deserialize, Serialize};
 
 use xayn_discovery_engine_ai::{Document as AiDocument, Embedding};
 
-use crate::error::common::{InvalidDocumentId, InvalidPropertyId, InvalidUserId};
+use crate::error::common::{InvalidDocumentId, InvalidDocumentPropertyId, InvalidUserId};
 
 /// A unique identifier of a document.
 #[derive(
@@ -73,11 +73,11 @@ pub struct DocumentPropertyId(String);
 
 impl DocumentPropertyId {
     #[allow(dead_code)]
-    pub fn new(id: impl Into<String>) -> Result<Self, InvalidPropertyId> {
+    pub fn new(id: impl Into<String>) -> Result<Self, InvalidDocumentPropertyId> {
         let id = id.into();
 
         if id.is_empty() || id.contains('\u{0000}') {
-            Err(InvalidPropertyId)
+            Err(InvalidDocumentPropertyId)
         } else {
             Ok(Self(id))
         }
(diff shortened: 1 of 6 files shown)


discovery_engine_core/web-api2/src/models.rs
```diff
@@ -0,0 +1,188 @@
...
+#[derive(Clone, Debug, Display, Serialize, Deserialize, PartialEq, Eq, Hash, AsRef)]
+pub struct DocumentPropertyId(String);
+
+impl DocumentPropertyId {
+    #[allow(dead_code)]
+    pub fn new(id: impl Into<String>) -> Result<Self, InvalidPropertyId> {
+        let id = id.into();
+
+        if id.is_empty() || id.contains('\u{0000}') {
+            Err(InvalidPropertyId)
```
biernacki, line 80:
```suggestion
            Err(InvalidDocumentPropertyId)
```

discovery_engine_core/web-api2/src/models.rs
```diff
@@ -0,0 +1,188 @@
...
+        } else {
+            Ok(Self(id))
+        }
+    }
+
+    #[allow(dead_code)]
+    pub fn encode(&self) -> Cow<str> {
```
acrrd, line 87: This implies that there is a canonical encoding for this model, while this is only useful when it is used in paths when querying elasticsearch, the model should not be aware of this and in other contexts the required encoding can be different. 
Encoding the id with urlendoding should be done in the elasticsearch code only.

biernacki (approved):
one suggestion, but otherwise lgtm

discovery_engine_core/web-api2/src/models.rs
rustonaut: I didn't introduce this method here, I moved that code over. 

urlencoding is somewhat of an conancial encoding as it's the only encoding we ever do use and it's not just specific to elastic search but also used when the parameters appear in the path, i.e. it's the encoding _we_ use all over the place, too

I.e. `DocumentId` should implement `FromStr`  (to be usable in `Path` ) later down the line in a way which url decodeds the string. But then we probably also should urlencode it on `Display` as this should match `FromStr`. But then `as_ref()->&str` and `to_string()` would no longer match so we probably should replace `as_ref()` with `as_unencoded_str()` :cry: which doesn't seem nice. Or in other words this is a temporary solution until we decide how to handle the fact that we have two semi-conancial encodings for ids, one is url encoded in path (in our api and elastic search) and the other is not encoded at all in json (in our api and in elastic search).

Butting that aside I really think we should switch to the official elastic search library even if it's in a very early stage, it's still more stable, reliable and well written then any of the ad hoc code we have written for elastic search and using it would safe a bunch of time and we wouldn't have to worry about any encoding of elastic search (we still would have FromStr with url decoding thou, that's in our api).

rustonaut:
@biernacki  I mainly rebased but there are 3 changes:

- included the changes to DocumentId, DocumentPropertyId and UserId from upstream 
- de-duplicate code using a `id_wrapper` macro (the ids all have roughly the same structure/interface)
- use `#[serde(try_from="String")]` instead of a custom deserializer making things simpler

so I need a re-approval

biernacki (approved):

rustonaut:
bors merge

rustonaut:
bors merge

PR #672, opened 2022-10-26
[Merged by Bors] - ET-3517 elastic search client in app state (7)

rustonaut:
Added an elastic search client to `AppState`.

**References:**

- [ET-3517]
- based on:
  - #671   
  - #670 
  - #669
  - #668
  - #667
  - #659

[ET-3517]: https://xainag.atlassian.net/browse/ET-3517?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine_core/web-api2/src/error/application.rs
+++ b/discovery_engine_core/web-api2/src/error/application.rs
@@ -105,11 +105,11 @@ macro_rules! impl_application_error {
                 stringify!($name)
             }
 
-            fn encode_details(&self) -> serde_json::Value {
+            fn encode_details(&self) -> ::serde_json::Value {
                 ::serde_json::to_value(self)
                     .unwrap_or_else(|err| {
-                        error!(%err, "serializing error details failed");
-                        serde_json::Value::Null
+                        ::tracing::error!(%err, "serializing error details failed");
+                        ::serde_json::Value::Null
                     })
             }
         }
--- a/discovery_engine_core/Cargo.lock
+++ b/discovery_engine_core/Cargo.lock
@@ -4606,6 +4606,7 @@ dependencies = [
  "once_cell",
  "pin-project",
  "regex",
+ "reqwest",
  "secrecy",
  "serde",
  "serde_json",
--- a/discovery_engine_core/web-api2/Cargo.toml
+++ b/discovery_engine_core/web-api2/Cargo.toml
@@ -20,6 +20,7 @@ mime = "0.3.16"
 once_cell = { workspace = true }
 pin-project = { workspace = true }
 regex = { workspace = true }
+reqwest = { workspace = true }
 secrecy = { version = "0.8.0", features = ["serde"] }
 serde = { workspace = true }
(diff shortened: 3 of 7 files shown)


discovery_engine_core/web-api2/src/error/application.rs
```diff
@@ -105,11 +105,11 @@ macro_rules! impl_application_error {
                 stringify!($name)
             }
 
-            fn encode_details(&self) -> serde_json::Value {
+            fn encode_details(&self) -> ::serde_json::Value {
                 ::serde_json::to_value(self)
                     .unwrap_or_else(|err| {
-                        error!(%err, "serializing error details failed");
-                        serde_json::Value::Null
+                        ::tracing::error!(%err, "serializing error details failed");
```
janpetschexain, line 111: :wrench: the macro is already imported above, no need for a full path here
```suggestion
                        error!(%err, "serializing error details failed");
```

discovery_engine_core/web-api2/src/error/application.rs
```diff
@@ -105,11 +105,11 @@ macro_rules! impl_application_error {
...
+                        ::serde_json::Value::Null
```
janpetschexain, line 112: ❓ why the additional `::` qualifier? it's quite unlikely that we confuse the external crate with some internal path. even if we decide to take this precautionary measure, it then should be applied everywhere in the code, which should be done in a separate pr.

biernacki:
please keep also in mind that our elastic client in `web-api` had some fixes introduced in #677 and #679

discovery_engine_core/web-api2/src/error/application.rs
rustonaut: The macro is used across the crate, so `error!` might not have been imported.

discovery_engine_core/web-api2/src/error/application.rs
rustonaut: Just general best practice for macros with `#[macro_export]`,  I don't think we should bother applying it everywhere in the code it doesn't add any value until it causes a problem, at which point we can change it. But there is no reason to not use it here.

discovery_engine_core/web-api2/src/error/application.rs
janpetschexain: i missed the `#[macro_export]` which was hidden behind github's folding, my bad.

janpetschexain:
> please keep also in mind that our elastic client in `web-api` had some fixes introduced in #677 and #679

@rustonaut ☝️

rustonaut:
I made a remainder ticket but this PR didn't include any of the affected code I only included the minimal skeleton.

The more specific code is in ET-3501/ET-3502.

janpetschexain (approved):

rustonaut:
bors merge

rustonaut:
bors merge

PR #673, opened 2022-10-26
[Merged by Bors] - ET-3522 impl from specific error for error (8)

rustonaut:
impl From<T> for crate::Error where T is one of

- anyhow::Error
- sqlx::Error
- reqwest::Error
- std::io::Error
- tokio::task::JoinError
- serde_json::Error

all are nearly every time internal errors so this will use `InternalError::from_{std,anyhow}(error).into()`

A wildcard `impl From<E> where E: std::error::Error` or similar is not possible due to limitations of rust.

**References:**

- [ET-3522]
- based on:
  - #672 
  - #671   
  - #670 
  - #669
  - #668
  - #667
  - #659

[ET-3522]: https://xainag.atlassian.net/browse/ET-3522?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #674, opened 2022-10-26
[Merged by Bors] - ET-3369 implement delete documents with actix (9)

rustonaut:
Implement single and batch document deletion on ingestion endpoint.

**References:**

- [ET-3369]
- based on:
  - #673   
  - #672 
  - #671   
  - #670 
  - #669
  - #668
  - #667
  - #659

[ET-3369]: https://xainag.atlassian.net/browse/ET-3369?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine_core/web-api2/src/elastic.rs
+++ b/discovery_engine_core/web-api2/src/elastic.rs
@@ -110,12 +110,13 @@ impl ElasticSearchClient {
             .await?;
 
         if response.errors {
-            let mut errors = HashMap::new();
-            for (idx, mut response) in response.items.into_iter().enumerate() {
+            let mut errors = Vec::new();
+            for mut response in response.items.into_iter() {
                 if let Some(response) = response.remove("delete") {
                     if let Ok(status) = StatusCode::from_u16(response.status) {
                         if status != StatusCode::NOT_FOUND && !status.is_success() {
-                            errors.insert(documents[idx].clone(), response.error);
+                            error!(document_id=%response.id, error=%response.error);
+                            errors.push(response.id);
                         }
                     } else {
                         error!("Non http status code: {}", response.status);
@@ -126,7 +127,10 @@ impl ElasticSearchClient {
             }
 
             if !errors.is_empty() {
-                return Err(FailedToDeleteSomeDocuments { errors }.into());
+                return Err(FailedToDeleteSomeDocuments {
+                    errors: errors.into_iter().map(Into::into).collect(),
+                }
+                .into());
             }
         }
 
@@ -222,6 +226,9 @@ struct BulkResponse {
 
 #[derive(Debug, Deserialize)]
 struct BulkItemResponse {
+    #[serde(rename = "_id")]
+    id: DocumentId,
     status: u16,
+    #[serde(default)]
     error: Value,
 }
--- a/discovery_engine_core/web-api2/src/elastic.rs
+++ b/discovery_engine_core/web-api2/src/elastic.rs
@@ -18,8 +18,9 @@ use reqwest::{
 };
 use secrecy::{ExposeSecret, Secret};
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
+use serde_json::Value;
 
-use crate::{error::common::InternalError, utils::serialize_redacted, Error};
+use crate::{error::common::InternalError, models::DocumentId, utils::serialize_redacted, Error};
 
 #[derive(Debug, Deserialize, Serialize, Clone)]
 pub struct Config {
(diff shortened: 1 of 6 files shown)


rustonaut:
I forgot to uncherry-pick some changes, will be fixed in a moment.

rustonaut:
done, now the changes unrelated to delete documents are moved to the PR they should be in

discovery_engine_core/web-api2/src/elastic.rs
```diff
@@ -65,18 +76,97 @@ impl Default for Config {
...
+                    }
+                } else {
+                    error!("Bulk delete request contains non delete responses: {response:?}");
+                }
+            }
+
+            if !errors.is_empty() {
+                return Err(FailedToDeleteSomeDocuments { errors }.into());
+            }
+        }
```
rustonaut, line 131: :gray_exclamation: I'm really not sure if having that error handling makes sense.

- Non existing documents don't cause `response.errors`  to be `true`.
- The only way we can have some but not all document deletions failing are (AFIK) supper edge case failure cases in which it often 1) isn't possible to fix it by directly retrying 2) it isn't too big of a problem to retry all once the problem was fixed.

discovery_engine_core/web-api2/src/elastic.rs
```diff
@@ -123,3 +213,15 @@ impl ElasticSearchClient {
...
+#[derive(Deserialize)]
+struct BulkResponse {
+    errors: bool,
+    items: Vec<HashMap<String, BulkItemResponse>>,
+}
+
+#[derive(Debug, Deserialize)]
+struct BulkItemResponse {
+    status: u16,
+    error: Value,
```
biernacki, line 226: is error always in the response? I know the ingestion request we used an Option there
```suggestion
    error: Option<Value>,
```

biernacki (commented):
left some minor comments

discovery_engine_core/web-api2/src/elastic.rs
rustonaut: yes, through it's only missing a `#[serde(default)]` but we don't need to wrap it into an option as there is no semantic difference between not given and set to `null`.

discovery_engine_core/web-api2/src/elastic.rs
```diff
@@ -65,18 +76,97 @@ impl Default for Config {
...
+            )
+            .await?;
+
+        if response.errors {
+            let mut errors = HashMap::new();
+            for (idx, mut response) in response.items.into_iter().enumerate() {
+                if let Some(response) = response.remove("delete") {
+                    if let Ok(status) = StatusCode::from_u16(response.status) {
+                        if status != StatusCode::NOT_FOUND && !status.is_success() {
+                            errors.insert(documents[idx].clone(), response.error);
```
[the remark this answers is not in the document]
rustonaut: The order is fixed so both approaches work. I changed it to use the `id` field.

discovery_engine_core/web-api2/src/elastic.rs
```diff
@@ -65,18 +76,101 @@ impl Default for Config {
...
+            for mut response in response.items.into_iter() {
+                if let Some(response) = response.remove("delete") {
+                    if let Ok(status) = StatusCode::from_u16(response.status) {
+                        if status != StatusCode::NOT_FOUND && !status.is_success() {
+                            error!(document_id=%response.id, error=%response.error);
+                            errors.push(response.id);
+                        }
+                    } else {
+                        error!("Non http status code: {}", response.status);
+                    }
```
biernacki, line 123: Not sure if we should add `DocumentId` to errors based on status being a valid http status, but rather based on the presence of `error` field in `BulkItemResponse`, wdyt?

biernacki (approved):

discovery_engine_core/web-api2/src/elastic.rs
biernacki: >the documentation does give guarantees

Let's leave it for now then. We can see later if we can trust their documentation 😉 

>I think we should just use the official rust bindings even if they still are alpha

Let's maybe put a pin in it until Andrea is back. This was an ad hoc decision that was made when the api was a PoC stage, so we can definitely discuss switching to official bindings I think.

rustonaut:
bors merge

rustonaut:
spurious failure in dart :roll_eyes:

rustonaut:
bors merge

rustonaut:
gitthub passing to ci-ok that a task failed while also reporting all other tasks as not having failed :face_exhaling:

rustonaut:
bors merge

rustonaut:
bors merge

PR #681, opened 2022-11-01
[Merged by Bors] - ET-3539 Removed trending topics

biernacki:
**References**:
- [ET-3539]

**Summary**:
- removed all code related to `TrendingTopics`

[ET-3539]: https://xainag.atlassian.net/browse/ET-3539?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


janpetschexain (approved):
lgtm 👍 

❓ do you know if any of the trending topic dart events/request are actually used by the app? if so, we would need to forward this to someone such that the breaking change can be fixed.

biernacki:
afaik the app doesn't use the functionality, but there is a small change needed anyways, because they extend our engine class, and with this PR there is a breaking change which removes `requestTrendingTopics` method. I'll add a task to the board.

biernacki:
bors merge

PR #682, opened 2022-11-01
[Merged by Bors] - ET-3379 change model builder config

janpetschexain:
**References**

- [ET-3379]
- [ET-3380]

**Summary**

- read all configurations for the bert model builder (except for token size) from a toml file & utilize this in the tokenizer and model constructors; apply the new builder whereever bert models are constructed
- split the various bert models which lived in `smbert_v0002` so far into `smbert_v0003` & `smbert_quantized_v0000` & `smbert_mocked_v0000` & `sjbert_v0000`, they are already uploaded to the yellow and staging buckets
- fix/consistify the bert naming: 1) in the rust engine we already used different bert models than smbert (eg sjbert), also with the new config design any bert model should be loadable, hence the engine only refers as bert to the models; 2) the dart app specifically loads a smbert model hence the names are kept

[ET-3379]: https://xainag.atlassian.net/browse/ET-3379?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ET-3380]: https://xainag.atlassian.net/browse/ET-3380?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


discovery_engine/lib/assets/asset_manifest.json
```diff
@@ -1,14 +1,20 @@
...
       "id": "smbertVocab",
-      "url_suffix": "smbert_v0002/vocab.txt",
+      "url_suffix": "smbert_quantized_v0000/vocab.txt",
       "checksum": "ffb2398810fa977ee39e8a0103b07f9c2614f1f9825df53a2c99e71569e3451c",
       "fragments": []
     },
     {
       "id": "smbertModel",
-      "url_suffix": "smbert_v0002/smbert-quantized.onnx",
+      "url_suffix": "smbert_quantized_v0000/model.onnx",
```
biernacki, line 17: why not keep the `v0002` suffix? more then that, I think we embed the manifest into the mobile app, so if we change the paths in S3 that might have some consequences for the mobile users, although probably not too many or maybe we don't care?

discovery_engine_core/ai/bert/examples/bert.rs
```diff
@@ -12,21 +12,18 @@
...
+    let pipeline = Config::new(smbert()?)?
         .with_token_size(64)?
+        .with_pooler::<FirstPooler>()
         .build()?;
-    let embedding = mbert.run("This is a sequence.")?;
-    let size = SMBert::embedding_size();
-
+    let embedding = pipeline.run("This is a sequence.")?;
     println!("{}", *embedding);
-    assert_eq!(embedding.shape(), [size]);
```
biernacki, line 29: why we're removing this assert?

biernacki (approved):
very nice change! just one concern about changing paths in S3, but maybe not a big issue

discovery_engine/lib/assets/asset_manifest.json
janpetschexain: > why not keep the v0002 suffix?

we never had an archive of this name, hence i started from `v0000`. but we can also start from a higher version. since it was split from the `smbert_v0002` and the `config.toml` was added, should we start from `v0003` then?

> more then that, I think we embed the manifest into the mobile app, so if we change the paths in S3 that might have some consequences for the mobile users, although probably not too many or maybe we don't care?

i was always under the impression, that the app doesn't deal with the details of the assets, but just fetches whatever is written in the asset manifest. in the previous asset version changes i also didn't change more than this. am i missing sth?

discovery_engine_core/ai/bert/examples/bert.rs
janpetschexain: seems that this got lost in between, i'll add it again, good catch 👍

discovery_engine/lib/src/domain/assets/asset.dart
```diff
@@ -64,6 +64,7 @@ class Asset {
 
 // Type of an asset.
 enum AssetType {
+  smbertConfig,
   smbertVocab,
   smbertModel,
```
janpetschexain, line 69: regarding the dart side, it would be nice to also just have a single `AssetType.smbert` variant. but that would require refactoring the whole data provider and asset copier code, which would be out of scope of this pr. should we add a ticket to the backlog @biernacki ?

discovery_engine/lib/assets/asset_manifest.json
biernacki: > we never had an archive of this name, hence i started from v0000. but we can also start from a higher version. since it was split from the smbert_v0002 and the config.toml was added, should we start from v0003 then?

If that's a new version of smbert quantized model, then I would start with v0003. If this is a new model altogether then a v0000 should be good.

> i was always under the impression, that the app doesn't deal with the details of the assets, but just fetches whatever is written in the asset manifest. in the previous asset version changes i also didn't change more than this. am i missing sth?

I'm not opposed to changing asset manifest, I'm just worried if we can still support older versions of the manifest that are embedded on user devices (hence we can't change them, the app needs to be updated). So it's more about having references from old manifest still accessible on S3 to download, but it's maybe an edge case for user who installed the app and didn't fetch assets yet.

discovery_engine/lib/src/domain/assets/asset.dart
biernacki: I think it's not so important to change that, so I wouldn't add a ticket for now.

discovery_engine/lib/assets/asset_manifest.json
janpetschexain: > If that's a new version of smbert quantized model, then I would start with v0003. If this is a new model altogether then a v0000 should be good.

i'll change it to `v0003` then 👌 

> I'm not opposed to changing asset manifest, I'm just worried if we can still support older versions of the manifest that are embedded on user devices (hence we can't change them, the app needs to be updated). So it's more about having references from old manifest still accessible on S3 to download, but it's maybe an edge case for user who installed the app and didn't fetch assets yet.

this should still work, the buckets contain all the old model versions. if someone with an old asset manifest starts to fetch them, it will succeed.

biernacki: >this should still work, the buckets contain all the old model versions. if someone with an old asset manifest starts to fetch them, it will succeed.

this is exactly what I wanted to hear 👍

janpetschexain:
bors merge

PR #683, opened 2022-11-01
[Merged by Bors] - ET-3314 Fixes sort for MIND api

biernacki:
**References**:
- [ET-3314]
- https://www.elastic.co/guide/en/elasticsearch/reference/current/paginate-search-results.html#search-after

**Summary**:
- changed type for "sort", as it can be present or not, and it's not a `String`, it might be an array of different types, like:
```json
"sort" : [
  1463538857,
  "654323"
]
```

[ET-3314]: https://xainag.atlassian.net/browse/ET-3314?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/discovery_engine_core/tooling/src/bin/backend/elastic.rs
+++ b/discovery_engine_core/tooling/src/bin/backend/elastic.rs
@@ -13,6 +13,7 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 use serde::Deserialize;
+use serde_json::Value;
 
 #[derive(Clone, Deserialize, Debug)]
 pub struct CountResponse {
@@ -35,7 +36,7 @@ pub struct Hit<T> {
     pub id: String,
     #[serde(rename(deserialize = "_source"))]
     pub source: T,
-    pub sort: String,
+    pub sort: Option<Value>,
 }
 
 /// An article in the MIND dataset.
--- a/discovery_engine_core/tooling/src/bin/backend/elastic.rs
+++ b/discovery_engine_core/tooling/src/bin/backend/elastic.rs
@@ -32,9 +32,9 @@ pub struct Hits<T> {
 
 #[derive(Clone, Deserialize, Debug)]
 pub struct Hit<T> {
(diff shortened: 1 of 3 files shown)


joecks (approved):

discovery_engine_core/tooling/src/bin/backend/routes.rs
```diff
@@ -171,13 +171,13 @@ async fn fetch_popular_results(
     });
 
     // after the first search also include search_after
-    let index = app_state.index.read().await;
     let from_index = app_state.from_index.read().await;
-    if *index > 0 || !from_index.is_empty() {
```
rustonaut, line 176: :question: Shouldn't we still check for `index > 0`?

discovery_engine_core/tooling/src/bin/backend/elastic.rs
```diff
@@ -35,7 +36,7 @@ pub struct Hit<T> {
     pub id: String,
     #[serde(rename(deserialize = "_source"))]
```
rustonaut, line 37: :art:  I have seen this in other places but we should prefer using `#[serde(rename = "...")]` in any case as long as we don't want to explicitly have different serialization then deserialization (which is generally a bad idea). Sure we don't serialize `Hit` but if we ever for whatever reason do so we now would have divergent serialization formats.

discovery_engine_core/tooling/src/bin/backend/routes.rs
biernacki: I've added that back just in case, but `from_index` should be set to `None` for the "first page" call if I understand correctly

discovery_engine_core/tooling/src/bin/backend/elastic.rs
biernacki: changed it, please check again

rustonaut (approved):
lgtm

biernacki:
bors merge

PR #688, opened 2022-11-02
Expose the correct PORT of the backend

joecks:
The port used in the  backend is 8080 
The better change probably would be to pass a param to the backend itself.


biernacki:
If you don't mind (pun intended) we could close this one in favor of changes in #683

PR #704, opened 2022-11-15
Update deps

acrrd:
In particular ` chrono`  made some functions deprecated that needed to be changed.


janpetschexain:
i already handled that in the rustc 1.65 update, but it's not opened yet, sorry for the redundancy :/

PR #706, opened 2022-11-16
[Merged by Bors] - Fix use of cache in CI

acrrd:
* Adds ` zstd` to the docker image
* Invert the setting of the env var `CARGO_BUILD_JOBS`  and the action that handles the cache. rust-cache uses all the env vars that start with `CARGO`  to generate the hash of the cache. Setting the variable after the cache action leads to uploading a cache with a different id than the one we want to restore


janpetschexain (approved):
lgtm

acrrd:
bors merge

PR #708, opened 2022-11-16
[Merged by Bors] - ET-3616 Fix workflows

Robert-Steiner:
fix workflows for ingestion and personalization
https://xainag.atlassian.net/browse/ET-3616

### DIFF
--- a/web-api/Dockerfile
+++ b/web-api/Dockerfile
@@ -2,7 +2,9 @@ FROM debian:bookworm-slim
 ARG api
 LABEL api="$api"
 WORKDIR /app
-#renamed as $api doesn't work in ENTRYPOINT
-COPY ./"$api" ./server.bin
+# copy everything in the cwd (bin and assets)
+COPY ./ ./
+# renamed as $api doesn't work in ENTRYPOINT
+RUN mv ./"$api" ./server.bin
 RUN chmod +x ./server.bin
 ENTRYPOINT ["/app/server.bin"]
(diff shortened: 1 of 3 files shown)


Robert-Steiner:
https://github.com/xaynetwork/xayn_discovery_engine/actions/runs/3481157410
https://github.com/xaynetwork/xayn_discovery_engine/actions/runs/3481204460
i also checked the docker images locally

web-api/Dockerfile
```diff
@@ -2,7 +2,9 @@ FROM debian:bookworm-slim
 ARG api
 LABEL api="$api"
 WORKDIR /app
-#renamed as $api doesn't work in ENTRYPOINT
-COPY ./"$api" ./server.bin
+# copy everything in the cwd (bin and assets)
+COPY ./ ./
+# renamed as $api doesn't work in ENTRYPOINT
+RUN mv ./"$api" ./server.bin
```
acrrd, line 8: :question: 
Is it worth it to have a parameter only to rename a file? When we generate the docker context we can already name the binary as server.

acrrd (approved):

web-api/Dockerfile
Robert-Steiner: i can remove it

Robert-Steiner: done. was a good idea. it makes the pulling faster

Robert-Steiner:
bors merge

PR #710, opened 2022-11-16
[Merged by Bors] - Split ai crate

acrrd:
Split out the coi system. Simplifications or other changes are out of scope.
    
The aim is to separate the code that the web api needs from the one that only the engine needs.
The code related to the engine will be deleted in the future.

Since the code left in `ai/ai`  will be deleted it does not need a thorough review.


ai/ai/src/kps/key_phrase.rs
```diff
@@ -110,7 +107,7 @@ impl KeyPhrases {
     /// Updates the key phrases for the positive coi and market.
     ///
     /// The most relevant key phrases are selected from the existing key phrases and the candidates.
-    fn update(
+    pub(crate) fn update(
```
acrrd, line 110: Now this is called instead of the associated method on `PositiveCoi`.

ai/bert/Cargo.toml
```diff
@@ -22,9 +22,7 @@ tract-onnx = "0.18.3"
 onnxruntime = { version = "0.0.13", optional = true }
 
 [dev-dependencies]
-criterion = { workspace = true }
```
janpetschexain, line 25: ❓:wrench: why is this removed? without it the `bert` bench declared below fails to compile.

ai/bert/Cargo.toml
```diff
@@ -22,9 +22,7 @@ tract-onnx = "0.18.3"
...
 csv = { workspace = true }
-indicatif = { workspace = true }
```
janpetschexain, line 27: ❓:wrench: why is this removed? without it the `validate` example declared below fails to compile.

janpetschexain (commented):
the code split looks good, but there are some compilation failures.

ai/bert/Cargo.toml
acrrd: I trusted `udeps` to tell me which deps where not used. I guess we also need to modify our just file/CI because it is not failing.

acrrd: I guess is my fault, I didn't use `--all-features`

janpetschexain: the ci doesn't check them because they have `required-features` which are not given in the ci action, `cargo check` just skips those. it should for example fail if we add an additional check with `--all-features` (which is also the reason why i see this locally on my machine).

janpetschexain (approved):
👍

PR #711, opened 2022-11-16
[Merged by Bors] - Confine engine related crates in a dedicated directory

acrrd:
On top of #710 

- Move all the crates related to the engine and the app under `app_engine`.
- Move crates out of `ai/<name>` at top of the repository. The crate name in the toml become `xayn-ai-<name>`.
- Rename the `core` crate to `engine` since it is not the core of our project anymore.


janpetschexain (commented):
❓ the `tooling/discovery_engine` and `tooling/backend` are based on the old app engine, but they are needed for the new ai tests. we could either move them here (and later delete them with the app engine), which would require a full rewrite of both for the ai test. but i guess it's better to not move them and to refactor them to use the ai directly instead of the app engine wrapper.

janpetschexain (approved):

acrrd:
We don't have to remove it soon but we should limit the use and change of that code base. We need to prioritise the tests on the web-api because that is the part that will really run.

bors merge

PR #716, opened 2022-11-29
[Merged by Bors] - fix ignore response decoding error

rustonaut:
- also fixes justfile `cargo sort` having a (no longer?) valid option

### DIFF
--- a/justfile
+++ b/justfile
@@ -40,7 +40,7 @@ rust-fmt:
     #!/usr/bin/env bash
     set -eux -o pipefail
     cargo +"$RUST_NIGHTLY" fmt --all -- {{ if env_var_or_default("CI", "false") == "true" { "--check" } else { "" } }};
-    cargo sort --grouped --workspace {{ if env_var_or_default("CI", "false") == "true" { "--check" } else { "" } }}
+    cargo sort --grouped --workspace {{ if env_var_or_default("CI", "false") == "true" { "--check --check-format" } else { "" } }}
 
 # Formats all code (checks only on CI)
 fmt: rust-fmt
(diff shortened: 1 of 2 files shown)


web-api/src/storage/elastic.rs
```diff
@@ -444,7 +444,7 @@ struct DocumentPropertiesResponse {
 }
 
 #[derive(Clone, Debug, Deserialize)]
-struct IgnoredResponse;
+struct IgnoredResponse {}
```
janpetschexain, line 447: ❓ from a rust type perspective this is the same, but if i get it right this is needed because of the json serde, right? then we should add a comment about this implicit detail, such that it doesn't get changed accidentally again.

justfile
```diff
@@ -40,13 +40,13 @@ rust-fmt:
     #!/usr/bin/env bash
     set -eux -o pipefail
     cargo +"$RUST_NIGHTLY" fmt --all -- {{ if env_var_or_default("CI", "false") == "true" { "--check" } else { "" } }};
-    cargo sort --grouped --workspace {{ if env_var_or_default("CI", "false") == "true" { "--check --check-format" } else { "" } }}
+    cargo sort --grouped --workspace {{ if env_var_or_default("CI", "false") == "true" { "--check" } else { "" } }}
```
janpetschexain, line 43: :wrench: this works fine on my machine as well as on the ci. this was introduced with the last toolchain update. just to be sure, have you updated your local `cargo sort` to `1.0.9` to match the ci?
```suggestion
    cargo sort --grouped --workspace {{ if env_var_or_default("CI", "false") == "true" { "--check --check-format" } else { "" } }}
```

rustonaut: :facepalm:

web-api/src/storage/elastic.rs
rustonaut: Unit  structs and empty structs are just mostly the same but differences in their handling is not limit to `serde` but also other derives and some other aspects like type matching/destruction and construction. 

But yes, it's for serde to deserialize it as an empty struct instead of an unit type.

I added documentation.

janpetschexain (approved):
lgtm

rustonaut:
bors merge

PR #717, opened 2022-11-30
ET-3680 memory storage knn search

janpetschexain:
**Reference**

- [ET-3680]
- [ET-3681]

**Summary**

in the memory storage impl:
- split embedding map from document map, they are still under one lock for atomicity though
- add a hnsw index for the embeddings via the `instant-distance` crate (see the ticket for other hnsw crate comparisons), this is self-referential to avoid a lot of allocations but the `ouroboros` crate should safely take care about that
- impl the knn search in `get_by_embedding()`, update the other impls due to the map split

in other modules:
- reduce overhead in `KnnSearchParams` allocation
- move serde of `Embedding<_>` to the type itself to have consistent serde behavior across the code
- reduce overhead for single `cosine_similarity()` computation


web-api/src/storage/memory.rs
```diff
@@ -83,26 +88,69 @@ impl TryFrom<&DocumentId> for models::DocumentId {
...
+
+#[self_referencing]
+struct Embeddings {
+    map: FnvHashMap<DocumentId, Embedding>,
+    #[borrows(map)]
+    #[covariant]
+    index: HnswMap<EmbeddingRef<'this>, &'this DocumentId>,
+}
+
+impl From<FnvHashMap<DocumentId, Embedding>> for Embeddings {
```
rustonaut, line 112: :art: ? Given how computation intensive this can be maybe it's not the best choice to place it in a `From` impl. it kinda makes it easy to overlook that `document.1 = embeddings.into();` isn't just some trivial conversion.

rustonaut (approved):
lgtm, there are some small stylistic thing

web-api/src/storage/memory.rs
```diff
@@ -17,21 +17,27 @@
 //   "U") to allow for faster map access without hashing (ie each u32 is already its own hash)
```
acrrd, line 17: This is not true, an integer is hashed anyway using the internal state of the hasher and some constant. The only difference is in the number of bytes.

Having this wrapper limits the usability of this storage, making it a mind specific storage and not a general memory storage that could be used for other tests.

acrrd (commented):
Unfortunately, non of the proposed optimizations does affect the performance.
In this branch https://github.com/xaynetwork/xayn_discovery_engine/tree/memory_tests I added a benchmark for the insertion, and then I removed (one per commit):
- u32 wrapper around the id
- use of fnv
- self-referential structure
I also removed `map` from `Embeddings` and store them together with the `Document`.
Because generating the index is so demanding, none of the changes affected the performance in any way. The only meaningful impact was given by using mimalloc in the benchmark for around 4% increased performance.

Removing the index generation from the equation makes the optimized code ~200 microseconds faster. I don't think it is a significant improvement, given the amount of code and the additional complexity.

I think we can merge this anyway to unblock some other work based on this. Still, we should subsequently simplify the code to have less code, complexity and dependencies with the added benefit that we can use the in-memory storage for other tests.

In general, we should apply this kind of optimization when needed and with benchmarks to have data to support our decisions.

bert/src/pooler.rs
```diff
@@ -86,6 +108,24 @@ impl TryFrom<Vec<u8>> for Embedding<Ix1> {
...
+}
+
+impl<'de> Deserialize<'de> for Embedding<Ix2> {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        Array2::<f32>::deserialize(deserializer).map(Self)
+    }
+}
```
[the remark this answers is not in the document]
janpetschexain: it was auto-derived before, but i guess we don't need it and it doesn't break anything if we remove it 👌

web-api/src/storage/memory.rs
janpetschexain: > This is not true, an integer is hashed anyway using the internal state of the hasher and some constant. The only difference is in the number of bytes.

i forgot to update the comment, this refers to the identity hasher (which just passed through the `u32`) before `fnv` was added to replace it. i'll update the comments after the code cleanup.

janpetschexain:
> Unfortunately, none of the proposed optimizations does affect the performance.  [...]

that's somewhat suprising tbh, i'll simplify the code to also make it more generally usable for other tests than mind.

> In general, we should apply this kind of optimization when needed and with benchmarks to have data to support our decisions.

the decision for the hasher optimizations are based on blogposts and crate docs regarding the performance benefits, and using either one or the other is merely a matter of crate import, so it seemed like an obvious choice without much additional work.

regarding the self referential map, did you also bench the memory footprint and the corresponding cache performance for very large test data? it should make a substantial difference for example for one million documents if we have to store the embeddings once plus a reference or twice in memory (520MB vs 1024MB)?

janpetschexain:
bors merge

PR #720, opened 2022-12-01
[Merged by Bors] - ET 3701 Add permissive cors for the api

acrrd:
We have a puplic api that must support every domain

### DIFF
--- a/web-api/src/server.rs
+++ b/web-api/src/server.rs
@@ -16,6 +16,7 @@ mod app_state;
 mod cli;
 mod config;
 
+use actix_cors::Cors;
 use actix_web::{
     middleware,
     web::{self, JsonConfig, ServiceConfig},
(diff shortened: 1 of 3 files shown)


web-api/src/server.rs
```diff
@@ -91,6 +92,7 @@ where
                 .wrap_fn(wrap_non_json_errors)
                 .wrap_fn(tracing_log_request)
                 .wrap(middleware::Compress::default())
+                .wrap(Cors::permissive())
```
rustonaut, line 95: :pin: for now good enough, but I guess in the future we want to only support the domains of the specific customer probably by having a config entry with a list of allowed domains.

rustonaut (approved):
lgtm (see comment for future improvements)

acrrd:
bors merge

PR #721, opened 2022-12-06
[Merged by Bors] - ET-3722 Fix OpenApi spec issues

frankpepermans:
This PR addresses a few minor issues found in the spec:

- rename `String` to `string`, as lowercase is the correct Type name
- add operationId entries on all API calls, this is used by codegen to create the method name for that API call, if omitted, codegen will instead generate a very ugly name, eg `useruseridPersonalizedDocuments`
- add securityScheme for our token, so that codegen adds it to the Header when performing calls


acrrd (approved):

frankpepermans:
bors merge

frankpepermans:
bors merge

PR #722, opened 2022-12-06
[Merged by Bors] - Initial validation and linting of the openapi specification

acrrd:
Adds an initial linter for the openapi specification.
The linter has been added to the docker image.
A just file command has also been added. Currently, that command is not called during `just pp` since it requires to have `node.js` installed. We can talk about if we want this also there or not.

Lots of lints are currently disabled. Before fixing them, we need to refactor the specification to remove the duplication that we have. Once that is done, we can enable more lints. In the meanwhile, we at least have some sanity checks for the specification.


.spectral.js
```diff
@@ -0,0 +1,55 @@
...
+          type: 'pascal'
+        }
+      }
+    },
+    'major-version-in-path': 'off',
+    'string-boundary': 'off',
+    'array-boundary': 'off',
+    'schema-description': 'off',
+    'inline-property-schema': 'off',
+    'duplicate-path-parameter': 'off',
```
rustonaut, line 39: :question:  Why it this off?

acrrd: It's not about a duplicate parameter in one path but is about having multiple paths with the same parameter, and it suggests to create a path section to define it only once. I'm going to do that in the next PR.

rustonaut (approved):

acrrd:
bors merge

PR #725, opened 2022-12-06
[Merged by Bors] - ET-3587 fetching personalized news independent of date

krajewskiML:
**Reference**

[ET-3587](https://xainag.atlassian.net/browse/ET-3587)

**Summary**

Adds a structure that is responsible for reading in the articles, sampling from them and getting articles by specified id.


web-api/src/mind.rs
```diff
@@ -41,6 +45,39 @@ struct Article {
...
+        self.articles
+            .values()
+            .cloned()
+            .collect_vec()
+            .iter()
+            .choose_multiple(&mut rng, n)
+            .into_iter()
+            .cloned()
+            .collect()
+    }
```
PanicButtonPressed, line 74: Looks like it could be optimized but didn't find valuable solutions for this, maybe @janpetschexain

web-api/src/mind.rs
```diff
@@ -41,6 +45,39 @@ struct Article {
...
+    fn new(path: &str) -> Result<Self, anyhow::Error> {
+        let mut articles = HashMap::new();
+        for article in read::<Article>(path)? {
+            let article = article?;
+            articles.insert(article.id.clone(), article);
+        }
+        Ok(Self { articles })
+    }
+
+    fn sample(&self, n: usize) -> Vec<Article> {
```
PanicButtonPressed, line 63: We could think about directly sorting out the retrieved elements from the dataset. However, we could also wait for further logic to know what exactly we want to achieve and how we will reset the dataset for every user

PanicButtonPressed:
Also lets rebase

web-api/src/mind.rs
janpetschexain: do we actually need to return owned articles (i guess not, even if we need to own some of the articles later on we could still clone them when needed)? if not we could simplify and save quite a lot by
```rust
    fn sample(&self, n: usize) -> Vec<&Article> {
        self.articles.values().choose_multiple(&mut thread_rng(), n)
    }
```

if we have to return owned articles, then we need to clone the chosen articles, but we can save the first cloning & collecting (this works because `values()` is already an iterator where we can directly apply the iterator extension `choose_multiple()` onto) by
```rust
    fn sample(&self, n: usize) -> Vec<Article> {
        self.articles
            .values()
            .choose_multiple(&mut thread_rng(), n)
            .into_iter()
            .cloned()
            .collect()
    }
```

in any case the handle to the rng could be directly supplied to the iterator instead of first binding it to a variable.

web-api/src/mind.rs
```diff
@@ -41,6 +45,39 @@ struct Article {
...
+            .cloned()
+            .collect()
+    }
+
+    fn get(&self, id: &str) -> Option<Article> {
+        self.articles.get(id).cloned()
+    }
+}
+
 struct SnippetLabelPair(String, String);
```
janpetschexain, line 81: :art: somewhat related to the `get()` method, we might not even need owned snippets and labels in many contexts. so for example we could operate on references like
```suggestion
struct SnippetLabelPair<'a, 'b>(&'a str, &'b str);
```
this might not fit all our purposes though. if we need to be flexible with owned and referenced types and still want to avoid unnecessary cloning, then feel free to have a look at the `std::borrow::Cow` type:
```suggestion
struct SnippetLabelPair<'a, 'b>(Cow<'a, str>, Cow<'b, str>);
```

no need to change it right away, but it might be a nice exercise for a follow up pr.

janpetschexain (commented):
i left some additional suggestions

web-api/src/mind.rs
```diff
@@ -17,31 +17,116 @@
...
+    D: Deserializer<'de>,
+{
+    let s = String::deserialize(deserializer)?;
+    let viewed_documents = s.split(' ');
+    viewed_documents
+        .map(|document| -> Result<_, D::Error> {
+            let document_id = DocumentId::new(document).map_err(de::Error::custom)?;
+            Ok(document_id)
+        })
+        .collect()
```
janpetschexain, line 83: :art: we can write this more concise for better readability. the deserialization, iteration and mapping can be chained without the need to assign to intermediate variables. the compiler is also ok to have the closure without an additional type hint. and the result inside the closure can be directly used as output, instead of unwrapping and wrapping it again.
```suggestion
    String::deserialize(deserializer)?
        .split(' ')
        .map(|document| DocumentId::new(document).map_err(de::Error::custom))
        .collect()
```

web-api/src/mind.rs
```diff
@@ -17,31 +17,116 @@
...
+}
+
+fn deserialize_viewed_documents<'de, D>(deserializer: D) -> Result<Vec<ViewedDocument>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let s = String::deserialize(deserializer)?;
+    let viewed_documents = s.split(' ');
+    viewed_documents
+        .map(|viewed_document| -> Result<_, D::Error> {
```
janpetschexain, line 53: :art: we can write this more concise for better readability. the deserialization, iteration and mapping can be chained without the need to assign to intermediate variables. the compiler is also ok to have the closure without an additional type hint.
```suggestion
    String::deserialize(deserializer)?
        .split(' ')
        .map(|viewed_document| {
```

web-api/src/mind.rs
```diff
@@ -17,31 +17,116 @@
...
+    articles: HashMap<DocumentId, Document>,
+}
+
+impl DocumentProvider {
+    fn new(path: &str) -> Result<Self, anyhow::Error> {
+        let articles = read::<Document>(path)?
+            .map(|article| article.map(|article| (article.id.clone(), article)))
+            .try_collect()?;
+        Ok(Self { articles })
+    }
```
janpetschexain, line 118: 🔧 renaming should be done consistently (ff)
```suggestion
struct DocumentProvider {
    documents: HashMap<DocumentId, Document>,
}

impl DocumentProvider {
    fn new(path: &str) -> Result<Self, anyhow::Error> {
        let documents = read::<Document>(path)?
            .map(|document| document.map(|document| (document.id.clone(), document)))
            .try_collect()?;
        Ok(Self { documents })
    }
```

web-api/src/mind.rs
```diff
@@ -63,9 +148,7 @@ where
 
 /// Runs the user-based mind benchmark
 fn run_benchmark() -> Result<(), anyhow::Error> {
-    let articles = read("news.tsv")?
-        .map(|result| result.map(|article: Article| (article.id.clone(), article)))
-        .collect::<Result<HashMap<_, _>, _>>()?;
+    let article_provider = DocumentProvider::new("news.tsv")?;
```
janpetschexain, line 151: :wrench:
```suggestion
    let document_provider = DocumentProvider::new("news.tsv")?;
```

janpetschexain (commented):
mainly some naming issues left. i also suggested some additional improvements.

web-api/src/mind.rs
```diff
@@ -17,31 +17,116 @@
...
+}
+
+impl ViewedDocument {
+    fn new(document_id: DocumentId, was_clicked: bool) -> Self {
+        Self {
+            document_id,
+            was_clicked,
+        }
+    }
+}
```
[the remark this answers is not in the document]
krajewskiML: That makes sense

janpetschexain (approved):

krajewskiML:
bors merge

PR #728, opened 2022-12-08
[Merged by Bors] - Remove required-features from benchmark

acrrd:
Some bench was enabled only with a feature enable but that feature was removed.
cargo prints a warning for that that was missed, I could not find a way to turn that warning in an error :/


acrrd:
bors merge

PR #730, opened 2022-12-08
[Merged by Bors] - ET-3731 user & persona test workflow handlers

janpetschexain:
**Reference**

- [ET-3731]
- [ET-3732]

**Summary**

- split the personalization handlers from the routes such that the handlers don't directly depend on the app state anymore
- move the knn search block from the personalized documents handler into `search_knn_documents()`
- add a `State` for mind tests which holds configs, storage & coi system, this provides utility to insert & interact & personalize

**Example**

the `State` can be used for example in the following way to set up tests for mind, the default storage could later be exchanged to load from a serialized blob to avoid to recompute the embeddings everytime:
```rust
#[tokio::test]
async fn test_user_workflow() {
    let state = State::new(Storage::default()).unwrap();
    state.insert(vec![/* documents */]).await.unwrap();

    let user = UserId::new("U123456").unwrap();
    let documents = state
        .personalize(&user, PersonalizeBy::Documents(&[/* document ids */]))
        .await
        .unwrap();
    println!("{documents:?}");
}

#[tokio::test]
async fn test_persona_workflow() {
    let state = State::new(Storage::default()).unwrap();
    state.insert(vec![/* documents */]).await.unwrap();

    let user = UserId::new("U123456").unwrap();
    state.interact(&user, &[/* document ids */]).await.unwrap();
    let documents = state
        .personalize(&user, PersonalizeBy::KnnSearch(10))
        .await
        .unwrap();
    println!("{documents:?}");
}
```

[ET-3731]: https://xainag.atlassian.net/browse/ET-3731?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ET-3732]: https://xainag.atlassian.net/browse/ET-3732?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/web-api/Cargo.toml
+++ b/web-api/Cargo.toml
@@ -43,7 +43,7 @@ xayn-ai-coi = { path = "../coi" }
 [dev-dependencies]
 bincode = { workspace = true }
 chrono = { workspace = true, features = ["clock", "serde"] }
-instant-distance = { version = "0.6.0" }
-ouroboros = { version = "0.15.5" }
+instant-distance = "0.6.0"
+ouroboros = "0.15.5"
 tokio = { workspace = true, features = ["sync"] }
 trycmd = "0.14.4"
--- a/web-api/Cargo.toml
+++ b/web-api/Cargo.toml
@@ -11,6 +11,8 @@ actix-cors = "0.6.4"
 actix-web = { workspace = true }
 anyhow = { workspace = true }
 async-trait = { workspace = true }
+bincode = { workspace = true, optional = true }
+chrono = { workspace = true, features = ["clock", "serde"], optional = true }
 clap = { workspace = true }
 csv = { workspace = true }
 derive_more = { workspace = true }
@@ -18,10 +20,12 @@ displaydoc = { workspace = true }
 dotenvy = "0.15.6"
 figment = { workspace = true, features = ["env", "toml"] }
 futures-util = "0.3.25"
+instant-distance = { version = "0.6.0", optional = true }
 itertools = { workspace = true }
 mime = "0.3.16"
 ndarray = { workspace = true }
 once_cell = { workspace = true }
+ouroboros = { version = "0.15.5", optional = true }
 pin-project = "1.0.12"
 rand = { workspace = true }
 regex = { workspace = true }
@@ -41,9 +45,7 @@ xayn-ai-bert = { path = "../bert" }
 xayn-ai-coi = { path = "../coi" }
 
 [dev-dependencies]
-bincode = { workspace = true }
-chrono = { workspace = true, features = ["clock", "serde"] }
-instant-distance = "0.6.0"
-ouroboros = "0.15.5"
-tokio = { workspace = true, features = ["sync"] }
 trycmd = "0.14.4"
+
+[features]
+mind = ["dep:bincode", "dep:chrono", "dep:instant-distance", "dep:ouroboros", "tokio/sync"]
--- a/web-api/Cargo.toml
+++ b/web-api/Cargo.toml
@@ -11,8 +11,6 @@ actix-cors = "0.6.4"
 actix-web = { workspace = true }
 anyhow = { workspace = true }
 async-trait = { workspace = true }
-bincode = { workspace = true, optional = true }
-chrono = { workspace = true, features = ["clock", "serde"], optional = true }
 clap = { workspace = true }
 csv = { workspace = true }
 derive_more = { workspace = true }
@@ -20,12 +18,10 @@ displaydoc = { workspace = true }
 dotenvy = "0.15.6"
 figment = { workspace = true, features = ["env", "toml"] }
 futures-util = "0.3.25"
-instant-distance = { version = "0.6.0", optional = true }
 itertools = { workspace = true }
 mime = "0.3.16"
 ndarray = { workspace = true }
 once_cell = { workspace = true }
-ouroboros = { version = "0.15.5", optional = true }
 pin-project = "1.0.12"
 rand = { workspace = true }
 regex = { workspace = true }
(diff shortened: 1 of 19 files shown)


web-api/Cargo.toml
```diff
@@ -41,9 +45,7 @@ xayn-ai-bert = { path = "../bert" }
...
 [dev-dependencies]
-bincode = { workspace = true }
-chrono = { workspace = true, features = ["clock", "serde"] }
-instant-distance = "0.6.0"
-ouroboros = "0.15.5"
-tokio = { workspace = true, features = ["sync"] }
 trycmd = "0.14.4"
+
+[features]
+mind = ["dep:bincode", "dep:chrono", "dep:instant-distance", "dep:ouroboros", "tokio/sync"]
```
janpetschexain, line 51: this is probably not the nicest solution, but the main issue comes from the fact that we can't simply make the routes generic over any `state: Data<AppState>` via traits because some of traits already implemented on the `AppState`s are not object safe.

web-api/src/mind.rs
```diff
@@ -11,16 +11,253 @@
...
+            .map(|documents| documents.iter().map(TryInto::try_into).try_collect())
+            .transpose()?;
+
+        Ok(Query(Self { count, documents }))
+    }
+}
+
+trait PersonalizedDocumentsResponder
+where
+    Self: Responder<Body = EitherBody<String>> + Sized,
```
janpetschexain, line 196: i tried to minimize the changes on the routes, which requires to have some cumbersome unpacking of the type-erased responses. the other option would be to split the handler logic from the routing code and run the tests directly on the handlers instead of the routes.

web-api/src/personalization/routes.rs
```diff
@@ -150,84 +158,16 @@ async fn personalized_documents(
         return Err(NotEnoughInteractions.into());
     }
 
-    let cois = &user_interests.positive;
```
janpetschexain, line 153: all the code just moved below into the `search_knn_documents()` function

web-api/src/personalization.rs
```diff
@@ -56,11 +57,11 @@ pub struct ConfigExtension {
 
     #[as_ref]
     #[serde(default)]
-    pub(crate) personalization: PersonalizationConfig,
+    pub(crate) personalization: Config,
 }
 
 #[derive(Debug, Deserialize, Serialize)]
-pub(crate) struct PersonalizationConfig {
```
acrrd, line 63: Don't rename if not needed to reduce the changes to only the ones that are needed for this PR.

web-api/src/personalization/routes.rs
```diff
@@ -116,12 +120,17 @@ async fn update_interactions(
...
 #[derive(Debug, Clone, Deserialize)]
-struct PersonalizedDocumentsQuery {
-    count: Option<usize>,
+pub(crate) struct PersonalizedDocumentsQuery {
+    pub(crate) count: Option<usize>,
+    /// Personalize only the specified documents, otherwise fetch documents to be personalized.
+    // Note: documents can't be provided via the public server api, since Query<Self> always fails
+    // to deserialize if documents is present in a query string (because it's a sequence) and the
+    // route returns a bad request
+    pub(crate) documents: Option<Vec<DocumentId>>,
```
acrrd, line 129: :wrench: We should touch the production endpoint only to refactor the code and nothing else. The property for which this is not exposed to the public API is not trivial and we should keep it simple.

The functionalities for the tests can be implemented in dedicated handlers that have code in common with this one.

web-api/src/server/app_state.rs
```diff
@@ -20,18 +20,18 @@ use crate::{
 };
 
 #[derive(Deref, AsRef)]
-pub(crate) struct AppState<CE, AE, S> {
+pub(crate) struct AppState<C, E, S> {
```
acrrd, line 23: :wrench: Not needed renames

acrrd (commented):
I don´t think we should touch the current production code, if not only to factor out common parts used by the mind tests.
We can avoid touching the ingestion service since we can add the documents directly to the storage. We should precompute the embeddings to avoid doing it every time since is very time-consuming.
We should be able to add only a custom `AppState` for testing that uses the same storage of the personalization endpoint that offers a handler that, given a set of documents, ranks them. For that, we don't need to pass through the knn.

acrrd:
> I don´t think we should touch the current production code, if not only to factor out common parts used by the mind tests. We can avoid touching the ingestion service since we can add the documents directly to the storage. We should precompute the embeddings to avoid doing it every time since is very time-consuming. We should be able to add only a custom `AppState` for testing that uses the same storage of the personalization endpoint that offers a handler that, given a set of documents, ranks them. For that, we don't need to pass through the knn (unless this will be needed by the persona tests).

acrrd:
Ok, I missed that `personalized_document` require a particular `AppState`, that is wrong, it should depend on a generic `Storage` instead.

janpetschexain:
> Ok, I missed that `personalized_document` require a particular `AppState`, that is wrong, it should depend on a generic `Storage` instead.

also regarding https://github.com/xaynetwork/xayn_discovery_engine/pull/730#discussion_r1043374250 & https://github.com/xaynetwork/xayn_discovery_engine/pull/730#pullrequestreview-1213128067, i applied the discussed changes, please have a 2nd look @acrrd

acrrd (approved):

web-api/src/mind.rs
```diff
@@ -24,9 +27,107 @@ use rand::{
...
-use crate::models::DocumentId;
+    async fn insert(&self, documents: Vec<Document>) -> Result<(), Error> {
+        let documents = documents
+            .into_iter()
+            .map(|document| {
+                let snippet = if document.snippet.is_empty() {
+                    document.title.clone()
+                } else {
+                    document.snippet.clone()
+                };
```
acrrd, line 79: :question: In the ingestion, we use only this snippet. Any reason to use the title instead here?

PanicButtonPressed (approved):
Thanks for the code walkthrough

janpetschexain:
bors merge

PR #733, opened 2022-12-12
[Merged by Bors] - ET-3715 document race condition: check existence

janpetschexain:
**Reference**

- [ET-3715]
- requires #732

**Summary**

- check existence of documents in postgres before calling anything in elastic

[ET-3715]: https://xainag.atlassian.net/browse/ET-3715?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/web-api/src/storage/elastic.rs
+++ b/web-api/src/storage/elastic.rs
@@ -284,7 +284,7 @@ impl storage::Document for Storage {
         let body = Some(json!({
             "query": {
                 "ids" : {
-                    "values" : ids
+                    "values" : self.postgres.documents_exist(ids).await?
                 }
             }
         }));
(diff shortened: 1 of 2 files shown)


web-api/src/storage/elastic.rs
```diff
@@ -284,7 +284,7 @@ impl storage::Document for Storage {
         let body = Some(json!({
             "query": {
                 "ids" : {
-                    "values" : ids
+                    "values" : self.postgres.documents_exist(ids).await?
```
rustonaut, line 287: :question:  I just realized that now but there might be one problem: For elastic search `filter.id.values` can easily  become way to big to be included into every request for personalized documents :sweat: 

Maybe we need to filter it in postgres after we got the ids, at the risk of returning less then expected documents (Which might be fine given that this should be an rare situation).

janpetschexain: i would also expect it to be too much work for elastic for larger numbers of ids. i implemented it in this way to uphold the current requirement that we return exactly k documents (if enough exist). can we relax this requirement to ease the load on elastic, @acrrd?

rustonaut (approved):
LGTM

PR #734, opened 2022-12-13
[Merged by Bors] - ET 3730 Don't log ingestion requests

acrrd:
They takes lots of space in the log with not much additional value

### DIFF
--- a/web-api/src/ingestion/routes.rs
+++ b/web-api/src/ingestion/routes.rs
@@ -81,7 +81,7 @@ struct IngestionRequestBody {
     documents: Vec<IngestedDocument>,
 }
 
-#[instrument(skip(state))]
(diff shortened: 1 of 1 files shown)


web-api/src/ingestion/routes.rs
```diff
@@ -81,7 +81,7 @@ struct IngestionRequestBody {
     documents: Vec<IngestedDocument>,
 }
 
-#[instrument(skip(state))]
+#[instrument(skip(state, body))]
```
janpetschexain, line 84: :art: since we now skip all arguments, we can instrument it like
```suggestion
#[instrument(skip_all)]
```

janpetschexain (approved):

acrrd:
bors merge

PR #737, opened 2022-12-15
ET-3590 setup persona related test workflow

krajewskiML:
**References**

[ET-3590](https://xainag.atlassian.net/browse/ET-3590?atlOrigin=eyJpIjoiNDExOGEyMjM0MzU0NDI0ZmEwYWQzY2E1YWI2OGEzZjciLCJwIjoiaiJ9)

**Summary**
This PR introduces persona-related workflow into `mind.rs` script.


web-api/src/mind.rs
```diff
@@ -231,6 +287,121 @@ where
...
+    let amount_of_doc_used_to_rerank = 30;
+
+    //probability of a document being clicked if it is interesting to the user
+    let click_probability = 0.5;
+
+    //define how many iterations to perform
+    let iterations = 10;
+
+    // define how many documents to use to calculate NDCG
+    let nranks = vec![3, 5];
```
janpetschexain, line 343: :art: looks like all those values could go into a config struct, maybe not now, but in a later pr to make it easer to setup different test cases.

web-api/src/mind.rs
```diff
@@ -209,6 +221,50 @@ impl DocumentProvider {
...
+    // get all documents that matches user's interest
+    fn get_all_interest(&self, interests: &UserInterests) -> Vec<&Document> {
+        self.documents
+            .values()
+            .filter(|doc| doc.is_interesting(interests))
+            .collect()
+    }
+}
+
+// struct storing users interests. It's a list of user ids and a list of their interests
```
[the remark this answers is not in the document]
krajewskiML: Yes,  you are right, I have probably been trying to find out how to split the structures in the best way and while I was doing that I forgot about the comments

web-api/src/mind.rs
```diff
@@ -209,6 +221,50 @@ impl DocumentProvider {
...
+#[derive(Debug, Deserialize)]
+struct UserInterests {
+    user_id: String,
+    interests: Vec<String>,
+}
+
+#[derive(Debug, Deserialize)]
+struct UsersInterests {
+    user_interests: Vec<UserInterests>,
+}
```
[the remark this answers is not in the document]
krajewskiML: Good suggestions, I will stick to it

web-api/src/mind.rs
```diff
@@ -209,6 +221,50 @@ impl DocumentProvider {
...
+            .values()
+            .filter(|doc| doc.is_interesting(interests))
+            .collect()
+    }
+}
+
+// struct storing users interests. It's a list of user ids and a list of their interests
+#[derive(Debug, Deserialize)]
+struct UserInterests {
+    user_id: String,
```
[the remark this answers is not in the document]
krajewskiML: Of course, I will replace it

web-api/src/mind.rs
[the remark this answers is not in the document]
krajewskiML: Good point, thanks for the explanation!

web-api/src/mind.rs
```diff
@@ -209,6 +221,50 @@ impl DocumentProvider {
...
+                    user_id: user_id.to_string(),
+                    interests: interests
+                        .as_array()
+                        .unwrap()
+                        .iter()
+                        .map(|interest| interest.as_str().unwrap().to_string())
+                        .collect(),
+                })
+                .collect(),
+        })
```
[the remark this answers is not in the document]
krajewskiML: Good idea, I will apply it

web-api/src/mind.rs
krajewskiML: Good idea, I will create a such struct

web-api/src/mind.rs
```diff
@@ -231,6 +287,121 @@ where
...
+                if document.is_interesting(&user_interest) {
+                    // user clicked on the document if the probability is greater than click_probability
+                    if rng.gen::<f32>() < click_probability {
+                        println!("The document {:?} was clicked.", document);
+                    }
+                }
+            }
+        }
+        // add ndcgs for the user to the results
+        results.push(Axis(2), ArrayView::from(&ndcgs)).unwrap();
```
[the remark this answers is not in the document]
krajewskiML: I have tried to implement this, but I will leave it unresolved as I am not sure if I have done this correctly according to the instructions

web-api/src/mind.rs
```diff
@@ -166,6 +168,44 @@ where
...
+        Self {
+            click_probability,
+            n_documents,
+            iterations,
+            amount_of_doc_used_to_prepare,
+            nranks,
+        }
+    }
+
+    fn default() -> Self {
```
janpetschexain, line 198: 🔧 for default impls the `Default` trait should be used to make it usable in other default contexts
```suggestion
    }
}

impl Default for PersonaBasedConfig {
    fn default() -> Self {
```

web-api/src/mind.rs
```diff
@@ -231,6 +315,124 @@ where
         .map_err(Into::into)
 }
 
+// function that assigns a score to a vector of documents based on the user's interests
+// score is equal to 1 if the document is interesting to the user, 0 otherwise
```
janpetschexain, line 319: 🔧 the docs doesn't match the function logic anymore.

web-api/src/mind.rs
```diff
@@ -231,6 +315,124 @@ where
...
+        .iter()
+        .map(|document| {
+            if document.is_interesting(user_interests) {
+                2.0
+            } else {
+                0.0
+            }
+        })
+        .collect::<Vec<f32>>();
+    Array1::from(scores)
```
janpetschexain, line 331: :art: everything which has a `From` impl automatically derives a `Into` impl as well, which can make some calls more ergonomically. from time to time it can have some type hint issues though, but here the functions signature should be enough for the compiler to infer correctly.
```suggestion
    documents
        .iter()
        .map(|document| {
            if document.is_interesting(user_interests) {
                2.0
            } else {
                0.0
            }
        })
        .collect::<Vec<f32>>()
        .into()
```

web-api/src/mind.rs
```diff
@@ -231,6 +315,124 @@ where
...
+        )
+        .await
+        .unwrap();
+    let benchmark_config = PersonaBasedConfig::default();
+    // create 3d array of zeros with shape (nranks, iterations, n_documents)
+    let mut results = Array3::<f32>::zeros(Dim([
+        benchmark_config.nranks.len(),
+        benchmark_config.iterations,
+        benchmark_config.n_documents,
+    ]));
```
janpetschexain, line 373: :art: the array constructors accepts any kind of shape-like object, we could just pass an array and avoid the extra `Dim`
```suggestion
    let mut results = Array3::<f32>::zeros([
        benchmark_config.nranks.len(),
        benchmark_config.iterations,
        benchmark_config.n_documents,
    ]);
```

web-api/src/mind.rs
```diff
@@ -209,6 +259,40 @@ impl DocumentProvider {
...
+    fn get_all_interest(&self, interests: &[String]) -> Vec<&Document> {
+        self.documents
+            .values()
+            .filter(|doc| doc.is_interesting(interests))
+            .collect()
+    }
+}
+
+#[derive(Debug, Deserialize)]
+struct Users(HashMap<UserId, Vec<String>>);
```
janpetschexain, line 273: :art: if we also derive `Deref` for `Users`, then we can forward method calls directly to the inner type, for example instead of `users.0.iter()` we can just call `user.iter()`. this is not available in the `std` lib though, but the `derive_more` crate got this covered (it's already a dependency of our crate).
```suggestion
#[derive(Debug, derive_more::Deref, Deserialize)]
struct Users(HashMap<UserId, Vec<String>>);
```

web-api/src/mind.rs
```diff
@@ -231,6 +315,124 @@ where
...
+                    user_id,
+                    PersonalizeBy::KnnSearch(benchmark_config.n_documents),
+                )
+                .await
+                .unwrap();
+            // get documents based on the personalised documents ids
+            let documents = personalised_documents
+                .iter()
+                .map(|id| document_provider.get(id).unwrap())
+                .collect::<Vec<_>>();
```
janpetschexain, line 403: :art: the `Itertools` trait provides a nice shorthand for this (it doesn't allow for type hints for the elements though, if that's an issue)
```suggestion
                .collect_vec();
```

web-api/src/mind.rs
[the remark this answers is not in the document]
janpetschexain: it's fine the way you implemented it now 👍  the other suggestions are more into the direction of avoiding to allocate the `results` array at all. this optimization is not necessary for our test cases, it was more a suggestion in case you want to look further into `Write`rs and the corresponding traits.

web-api/src/mind.rs
krajewskiML: Okay, I did not know about this feature, thanks for the suggestion!

web-api/src/mind.rs
krajewskiML: I don't know why it was not working previously, but now when I tried this it worked

web-api/src/mind.rs
krajewskiML: That's a nice feature, I will apply it

web-api/src/mind.rs
```diff
@@ -231,6 +299,122 @@ where
...
+            // interact with some of the retrieved documents based on whether the document is interesting to the user and probability of clicking
+            // interact with documents
+            state
+                .interact(
+                    user_id,
+                    &personalised_documents
+                        .iter()
+                        .zip(scores.iter())
+                        .filter(|(_, &score)| {
+                            (score - 2.0).abs() < 0.001
```
PanicButtonPressed, line 402: Could we not just check if the score is 2?

PanicButtonPressed (approved):
Just minor comments left, let's wait for @janpetschexain final approval

web-api/src/mind.rs
```diff
@@ -231,6 +299,122 @@ where
...
+        .begin_nd()?;
+    writer.extend(c_order_items)?;
+    writer.finish()
+}
+#[doc = "
+
+# Panics
+
+The function panics if the provided filenames are not correct.
+"]
```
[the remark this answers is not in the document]
janpetschexain: it's technically correct, but the more readable format is to use `///` instead of `#[doc = ""]`
```suggestion
/// # Panics
///
/// The function panics if the provided filenames are not correct.
```

web-api/src/mind.rs
```diff
@@ -231,6 +299,122 @@ where
...
+        let ids_of_documents_to_prepare = interesting_documents
+            .choose_multiple(&mut rng, benchmark_config.amount_of_doc_used_to_prepare)
+            .map(|doc| doc.id.clone())
+            .collect_vec();
+        // prepare reranker by interacting with documents to prepare
+        state
+            .interact(user_id, &ids_of_documents_to_prepare)
+            .await
+            .unwrap();
+        // perform iterations
```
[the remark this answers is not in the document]
janpetschexain: as a rule of thumb, if the information is already expressed via the variable name or type, then a comment which just repeats that is somewhat redundant. if there is anything non-obvious, then it can make sense to add addtional comments.

for example here the comment is directly followed by an iterator in a loop, and the comment doesn't give any other info than that an iteration is performed. overall this makes the code a bit more noisy for a reviewer.

web-api/src/mind.rs
janpetschexain: exact float comparrison is usually error prone because of float rounding precision errors

janpetschexain (approved):

web-api/src/mind.rs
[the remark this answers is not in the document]
krajewskiML: Alright I will get rid of these redundant comments

krajewskiML:
bors merge

PR #739, opened 2022-12-15
[Merged by Bors] - ET-3714 rename category to tags

janpetschexain:
**Reference**

- [ET-3714]

**Summary**

- rename `category` to `tags` (currently it's still a single `Option<String>`, which will be changed in the following pr)
- add postgres and elastic migration

[ET-3714]: https://xainag.atlassian.net/browse/ET-3714?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/web-api/src/models.rs
+++ b/web-api/src/models.rs
@@ -112,7 +112,7 @@ pub(crate) struct PersonalizedDocument {
     /// Contents of the document properties.
     pub(crate) properties: DocumentProperties,
 
-    /// The metadata of the document.
+    /// The tags associated to the document.
     pub(crate) tags: Option<String>,
 }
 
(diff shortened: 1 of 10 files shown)


web-api/src/models.rs
```diff
@@ -159,9 +159,9 @@ pub(crate) struct IngestedDocument {
     /// Contents of the document properties.
     pub(crate) properties: DocumentProperties,
 
-    /// The high-level category the document belongs to.
+    /// The metadata of the document.
```
acrrd, line 162: See above

web-api/src/storage/elastic.rs
```diff
@@ -247,7 +247,8 @@ pub struct Document {
     pub snippet: String,
     pub properties: DocumentProperties,
     pub embedding: Embedding,
-    pub category: Option<String>,
+    #[serde(alias = "category")] // deprecated
```
acrrd, line 250: It is not in the spec so I think is ok to not support it. Is there some case I'm not aware of?

janpetschexain: i guess it only matters if we actually had data stored in elastic. i'll remove it.

PR #741, opened 2022-12-19
[Merged by Bors] - Openpapi linters

acrrd:
Enable some of linters, these are the ones that do not break the API (too much).

### DIFF
--- a/web-api/openapi/front_office.yaml
+++ b/web-api/openapi/front_office.yaml
@@ -117,7 +117,7 @@ components:
               enum: [NotEnoughInteractions]
     UserInteractionType:
       type: string
-      enum: [positive]
+      enum: [Positive]
     UserInteractionData:
       type: object
       properties:
--- a/web-api/openapi/front_office.yaml
+++ b/web-api/openapi/front_office.yaml
@@ -1,7 +1,7 @@
 openapi: 3.0.3
 info:
   title: Front Office API
-  version: 1.0.0
+  version: 1.0.0-rc6
   description: |-
     # Front Office
     The system identifies a user with only an id.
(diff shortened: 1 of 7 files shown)


web-api/openapi/document_management.yaml
```diff
@@ -31,7 +31,7 @@ paths:
       description: |-
         Add documents to the system. The system will create a representation of the document
         that will be used to match it against the preferences of a user.
-      operationId: ingestDocuments
+      operationId: createDocuments
```
acrrd, line 34: One linter requires `operationId` to start with a verb that identifies the type of the operation depending on which HTTP method we use. I think it is clearer this way and less likely to conflict with future API updates where we add new methods. But I don't have a strong opinion on this, so we can revert this commit.

frankpepermans: The only problem with `createDocuments` right now, is that you can just update existing documents via `createDocuments` as well.

Maybe `sendDocuments` or just `documents` as the POST method already implies a push?

acrrd: This linter only allows for `create` if you think it is a problem, we can disable it.
In this specific case, I think that, since we don't like too much that a document can be updated without a deletion, giving an API that doesn't encourage you to do it is good. Also, from our perspective, it is like you are putting in a new document since we are completely replacing the old one.

web-api/openapi/front_office.yaml
```diff
@@ -65,7 +65,7 @@ paths:
       summary: Add interaction between a user and a document
       description: |-
         The interaction is used to provide personalized documents to the user.
-      operationId: documentInteraction
+      operationId: updateUserInteractions
```
frankpepermans, line 68: Should this not be createUserInteraction?
As if you do 2 positive interactions, it just runs twice on the CoI level, right?

frankpepermans (approved):

acrrd:
bors merge

PR #743, opened 2022-12-19
[Merged by Bors] - ET-3674 multiple tags per document

janpetschexain:
**Reference**

- [ET-3674]

**Summary**

- change document tags from `Option<String>` to `Vec<String>`
- sort documents by tags via accumulated tag weight

[ET-3674]: https://xainag.atlassian.net/browse/ET-3674?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/web-api/src/models.rs
+++ b/web-api/src/models.rs
@@ -113,7 +113,7 @@ pub(crate) struct PersonalizedDocument {
     pub(crate) properties: DocumentProperties,
 
     /// The tags associated to the document.
-    pub(crate) tags: Option<String>,
+    pub(crate) tags: Vec<String>,
 }
 
 impl AiDocument for PersonalizedDocument {
@@ -160,8 +160,7 @@ pub(crate) struct IngestedDocument {
     pub(crate) properties: DocumentProperties,
 
     /// The tags associated to the document.
-    #[serde(default, deserialize_with = "deserialize_empty_option_string_as_none")]
-    pub(crate) tags: Option<String>,
+    pub(crate) tags: Vec<String>,
 }
 
 fn deserialize_string_not_empty_or_zero_byte<'de, D>(deserializer: D) -> Result<String, D::Error>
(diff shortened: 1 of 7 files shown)


web-api/src/models.rs
```diff
@@ -160,8 +160,7 @@ pub(crate) struct IngestedDocument {
     pub(crate) properties: DocumentProperties,
 
     /// The tags associated to the document.
-    #[serde(default, deserialize_with = "deserialize_empty_option_string_as_none")]
-    pub(crate) tags: Option<String>,
+    pub(crate) tags: Vec<String>,
```
acrrd, line 163: :pushpin: We could reject 'tags` that are empty strings or invalid id. This is not exposed in the specification so we have a bit of time to finalize this.

web-api/src/personalization/routes.rs
```diff
@@ -333,11 +331,9 @@ pub(crate) async fn personalize_documents_by(
...
         .iter()
         .map(|document| {
-            let weight = document
-                .tags
-                .as_deref()
-                .and_then(|tag| tags.get(tag).copied())
-                .unwrap_or_default();
+            let weight = document.tags.iter().fold(0, |weights, tag| {
+                tags.get(tag).map_or(weights, |weight| weights + *weight)
+            });
```
acrrd, line 336: :art:
I think this could be also written as something like:
```suggestion
            let weight = document.tags.iter().map(|tag| tags.get(tag).unwrap_or(0)).sum();
```

acrrd (approved):

web-api/src/models.rs
janpetschexain: i'll add a follow up ticket to the backlog.

janpetschexain:
bors merge

PR #744, opened 2022-12-19
[Merged by Bors] - More api changes

acrrd:
Depends on #741 
* Document ingestion now returns 201 instead of 204; error code 422 has been replaced with 409.
* Request bodies have been marked as required.
* A response always returns a body, even when empty.
* 404 is returned only if the request does not enter the handler. If it does, then a 400 will be returned instead.
* Now every error is a `GenericError`
* Specific error are now only one possible error that the endpoint can return, a generic error is always possible. In this way we can extend the kind of error we can return form an endpoint.

### DIFF
--- a/.spectral.js
+++ b/.spectral.js
@@ -46,7 +46,6 @@ module.exports = {
     'response-example-provided': 'off',
     'collection-array-property': 'off',
     'response-error-response-schema': 'off',
-    'enum-case-convention': 'off',
     'operation-id-case-convention': 'off',
     'patch-request-content-type': 'off',
   }
--- a/.spectral.js
+++ b/.spectral.js
@@ -34,7 +34,6 @@ module.exports = {
     'major-version-in-path': 'off',
     'schema-description': 'off',
     'inline-property-schema': 'off',
-    'content-entry-provided': 'off',
     'property-description': 'off',
     'oas3-api-servers': 'off',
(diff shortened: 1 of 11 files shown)


web-api/openapi/user_management.yaml
```diff
@@ -138,11 +145,13 @@ components:
...
-        - $ref: './schemas/error.yml#/BaseError'
-        - type: object
-          required: [kind]
-          properties:
-            kind:
-              type: string
-              enum: [InvalidUserId, InvalidDocumentId]
+      oneOf:
+        - $ref: './schemas/error.yml#/GenericError'
+        - allOf:
```
rustonaut, line 150: Depending on the code generation we might need to switch the order of the `oneOf` around to go from most-precise to least-precise.

rustonaut (approved):
lgtm, through we might need one follow up depending on code gen (it might also be right just the way it is now, or it might not matter at all)

web-api/openapi/user_management.yaml
acrrd: @frankpepermans

acrrd: I removed this commit because it breaks the generation in all targets :(

acrrd:
I reverted the change on the errors because it broke code generation.

bors merge

PR #748, opened 2022-12-20
[Merged by Bors] - [ET-3593] Use data handler to personalize documents

PanicButtonPressed:
**Reference**

[ET-3593](https://xainag.atlassian.net/browse/ET-3593)

**Summary**

Use the state handler to interact and personalize the documents according to `behavior.csv`.

### DIFF
--- a/web-api/src/mind.rs
+++ b/web-api/src/mind.rs
@@ -25,7 +25,8 @@ use ndarray::{Array, Array3, ArrayView};
 use npyz::WriterBuilder;
 use rand::{
     seq::{IteratorRandom, SliceRandom},
-    thread_rng, Rng,
+    thread_rng,
+    Rng,
 };
 use serde::{de, Deserialize, Deserializer};
 use xayn_ai_coi::{nan_safe_f32_cmp_desc, CoiConfig, CoiSystem};
@@ -35,7 +36,10 @@ use crate::{
     models::{DocumentId, DocumentProperties, IngestedDocument, UserId, UserInteractionType},
     personalization::{
         routes::{
-            personalize_documents_by, update_interactions, PersonalizeBy, UserInteractionData,
+            personalize_documents_by,
+            update_interactions,
+            PersonalizeBy,
+            UserInteractionData,
         },
         PersonalizationConfig,
     },
@@ -231,7 +235,7 @@ impl DocumentProvider {
         self.documents.get(id)
     }
 
-    fn get_documents(&self) -> Vec<Document> {
+    fn to_documents(&self) -> Vec<Document> {
         self.documents.values().cloned().collect()
     }
 
@@ -400,15 +404,14 @@ async fn run_persona_benchmark() -> Result<(), Error> {
 }
 
 /// Runs the user-based mind benchmark
-#[test]
+#[tokio::test]
 #[ignore]
-fn async run_user_benchmark() -> Result<(), Error> {
+async fn run_user_benchmark() -> Result<(), Error> {
     let document_provider = DocumentProvider::new("news.tsv")?;
-    let impressions: DeserializeRecordsIntoIter<File, _> = read("behaviors.tsv")?;
 
     let state = State::new(Storage::default()).unwrap();
     state
-        .insert(document_provider.get_documents())
+        .insert(document_provider.to_documents())
         .await
         .unwrap();
 
@@ -417,7 +420,7 @@ fn async run_user_benchmark() -> Result<(), Error> {
 
     // Loop over all impressions, prepare reranker with news in click history
     // and rerank the news in an impression
-    for impression in impressions {
+    for impression in read("behaviors.tsv")? {
         let impression: Impression = impression?;
         let user = UserId::new(impression.user_id).unwrap();
 
@@ -435,11 +438,14 @@ fn async run_user_benchmark() -> Result<(), Error> {
             .unwrap()
             .iter()
             .map(|reranked_id| {
-                impression.news[document_ids
-                    .iter()
-                    .position(|&actual_id| actual_id == reranked_id)
-                    .unwrap()]
-                .was_clicked as i32 as f32
+                u8::from(
+                    impression.news[document_ids
+                        .iter()
+                        .position(|&actual_id| actual_id == reranked_id)
+                        .unwrap()]
+                    .was_clicked,
+                )
+                .into()
             })
             .collect_vec();
 
--- a/web-api/src/mind.rs
+++ b/web-api/src/mind.rs
@@ -36,10 +36,7 @@ use crate::{
     models::{DocumentId, DocumentProperties, IngestedDocument, UserId, UserInteractionType},
     personalization::{
         routes::{
-            personalize_documents_by,
-            update_interactions,
-            PersonalizeBy,
-            UserInteractionData,
+            personalize_documents_by, update_interactions, PersonalizeBy, UserInteractionData,
         },
         PersonalizationConfig,
     },
@@ -272,8 +269,6 @@ impl Users {
                 .collect(),
         ))
 }
-
-struct SnippetLabelPair(String, bool);
 
 fn read<T>(path: &str) -> Result<DeserializeRecordsIntoIter<File, T>, Error>
 where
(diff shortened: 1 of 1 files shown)


web-api/src/mind.rs
```diff
@@ -235,32 +242,35 @@ fn run_benchmark() -> Result<(), Error> {
...
-                document_provider
-                    .get(&viewed_document.document_id)
-                    .map(|document| {
-                        SnippetLabelPair(document.snippet.clone(), viewed_document.was_clicked)
-                    })
+            .map(|document| document.document_id)
+            .collect::<Vec<_>>();
+
+        let documents = state
+            .personalize(&user, PersonalizeBy::Documents(document_ids.as_slice()))
```
PanicButtonPressed, line 256: @janpetschexain any idea how I can resolve that `&[&models::DocumentId]` is expected instead of `&[models::DocumentId]`?

janpetschexain: in general, if you have a `Vec<T>` and want to get a `Vec<&T>` you can iterate and collect them like `let referenced_vec = owned_vec.iter().collect();`. but in this case the actual issue lies in collecting the document ids in l.252, which tries to move an owned value out of a reference (this is only possible in a sense for `Copy` values). if you take a reference instead, then that will also solve the issue in l.256: `.map(|document| &document.document_id)`.

PanicButtonPressed: Yea makes sense, did not notice the owning in that statement

web-api/src/mind.rs
```diff
@@ -300,8 +302,8 @@ fn ndcg(relevance: &[f32], k: &[usize]) -> Vec<f32> {
         .collect()
 }
 
-fn main() {
-    if let Err(e) = run_benchmark() {
+async fn main() {
+    if let Err(e) = run_benchmark().await {
```
janpetschexain, line 306: :wrench: the compilers is unhappy because it can't find the `run_bechmark()` anymore, which was renamed. you could also remove the `main()` function here, as this is going to be resolved with #737 anyways.

web-api/src/mind.rs
```diff
@@ -200,6 +197,10 @@ impl DocumentProvider {
     fn get(&self, id: &DocumentId) -> Option<&Document> {
         self.documents.get(id)
     }
+
+    fn get_documents(&self) -> Vec<Document> {
```
janpetschexain, line 201: :art: getters are usually implemented via (mutable) references in rust, whenever heavy work like cloning is involved, the `into` (from owned)/`to` (from referenced) prefixes are more common.
```suggestion
    fn to_documents(&self) -> Vec<Document> {
```

web-api/src/mind.rs
```diff
@@ -223,10 +224,15 @@ where
 }
 
 /// Runs the user-based mind benchmark
-fn run_benchmark() -> Result<(), Error> {
+async fn run_user_benchmark() -> Result<(), Error> {
     let document_provider = DocumentProvider::new("news.tsv")?;
+    let impressions: DeserializeRecordsIntoIter<File, _> = read("behaviors.tsv")?;
```
janpetschexain, line 229: :art: the type hint is not required
```suggestion
    let impressions = read("behaviors.tsv")?;
```

also, we could just call the iterator construction in the loop head where it's actually used, in that way the iterator construction is not visually detached from the iterator usage (in l.242):
```rust
    for impression in read("behaviors.tsv")? {
```

web-api/src/mind.rs
```diff
@@ -235,35 +241,31 @@ fn run_benchmark() -> Result<(), Error> {
...
+            .await
+            .unwrap()
             .iter()
-            .map(|snippet_label| if snippet_label.1 { 1.0 } else { 0.0 })
+            .map(|reranked_id| {
+                impression.news[document_ids
+                    .iter()
+                    .position(|&actual_id| actual_id == reranked_id)
+                    .unwrap()]
+                .was_clicked as i32 as f32
```
janpetschexain, line 264: :wrench: once the other compilation issue is fixed, the compiler will complain about the `bool as i32 as f32` type cast (because of potential precision loss). we can handle this with a more restrictive type cast `bool as u8 as f32`:
```suggestion
                u8::from(
                    impression.news[document_ids
                        .iter()
                        .position(|&actual_id| actual_id == reranked_id)
                        .unwrap()]
                    .was_clicked,
                )
                .into()
```

janpetschexain (commented):
looks good, there are some compiler issues though

web-api/src/mind.rs
PanicButtonPressed: removed the `main`

web-api/src/mind.rs
PanicButtonPressed: changed

web-api/src/mind.rs
PanicButtonPressed: makes sense

web-api/src/mind.rs
PanicButtonPressed: Great hint

web-api/src/mind.rs
```diff
@@ -402,49 +404,52 @@ async fn run_persona_benchmark() -> Result<(), Error> {
...
+        let user = UserId::new(impression.user_id).unwrap();
 
-        // Placeholder for interacting with the entire click history
-        for click in impression.clicks {
-            match document_provider.get(&click) {
-                Some(document) => println!("The document {:?} was interacted.", document),
-                None => println!("Document id {} not found.", click),
-            }
-        }
+        state.interact(&user, &impression.clicks).await.unwrap();
```
krajewskiML, line 427: The history of clicks of a single user does not change between his/her different impressions, which means we could interact with clicked news only once e.g when we encounter a new user id. We could use a set of user ids for this

web-api/src/mind.rs
```diff
@@ -402,49 +404,52 @@ async fn run_persona_benchmark() -> Result<(), Error> {
...
-            .map(|snippet_label| if snippet_label.1 { 1.0 } else { 0.0 })
+            .map(|reranked_id| {
+                u8::from(
+                    impression.news[document_ids
+                        .iter()
+                        .position(|&actual_id| actual_id == reranked_id)
+                        .unwrap()]
+                    .was_clicked,
+                )
+                .into()
```
krajewskiML, line 448: I have been using `2.0` as a value for "clicked" news, if I understand correctly this code will return `1.0` for clicked articles. For the representative results, we should agree on a single value in both tests. I advocate for 2 because we have been using it in previous tests performed in python and we could maybe compare some results. WDYT?

web-api/src/mind.rs
PanicButtonPressed: True, we could do this. However, we also need to maintain the state of the user id and keep track of whether the user id was already interacted with. Would mark this as further optimization as long as we don't have any need in increasing the computational speed. In the worst case, it takes around 1.5 times longer than before.

web-api/src/mind.rs
rustonaut: generally `as` casting is discouraged in most situations by now with `.into()` and `.try_into()` being recommended instead.

Some exceptions are coercion to a trait object, some micro optimizations and some same size bit-casting you normally don't want to do in most situations.

rustonaut:
Nothing to add on my side wrt. the rust parts so once the open logic related questions are resolved it can be merged I think.

rustonaut (approved):
Nothing to add on my side wrt. the rust parts so once the open logic related questions are resolved it can be merged I think.

web-api/src/mind.rs
PanicButtonPressed: Although always tending towards consistency, I am against it in this case as the results of the user benchmark are not comparable to other externals anymore. Doesn't matter that much for the persona-related one, do we have any representative results in python documented somewhere that we could use reference?

PanicButtonPressed:
> Nothing to add on my side wrt. the rust parts **so once the open logic related questions are resolved** it can be merged I think.
> 
> Sadly there is no "partial approval" in github.

Agree, would be a valuable feature. Will clarify the other things first

web-api/src/mind.rs
krajewskiML: I think it is not only a speed issue, but also interacting with the same articles in every impression may change the results as the weights of topics are incremented multiple times and also CoIs are changed

krajewskiML (approved):
Good job!

PanicButtonPressed:
bors merge

PR #750, opened 2022-12-20
[Merged by Bors] - ET-3766 exclude unnecessary fields from elastic queries

janpetschexain:
**Reference**

- [ET-3766]

**Summary**

- return the smaller `PersonalizedDocumentData` instead of the `PersonalizedDocument` from the `personalized_documents` route
- restrict elastic to only return the necessary fields in the storage calls

[ET-3766]: https://xainag.atlassian.net/browse/ET-3766?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/web-api/src/models.rs
+++ b/web-api/src/models.rs
@@ -96,16 +96,21 @@ pub(crate) struct DocumentProperty(serde_json::Value);
 /// Arbitrary properties that can be attached to a document.
 pub(crate) type DocumentProperties = HashMap<DocumentPropertyId, DocumentProperty>;
 
-#[derive(Debug, Deserialize)]
+/// Represents a result from an interaction query.
+#[derive(Debug)]
 pub(crate) struct InteractedDocument {
+    /// Unique identifier of the document.
     pub(crate) id: DocumentId,
+
+    /// Embedding from smbert.
     pub(crate) embedding: Embedding,
-    #[serde(default)]
+
+    /// The tags associated to the document.
     pub(crate) tags: Vec<String>,
 }
 
 /// Represents a result from a personalization query.
-#[derive(Debug, Deserialize)]
+#[derive(Debug)]
 pub(crate) struct PersonalizedDocument {
     /// Unique identifier of the document.
     pub(crate) id: DocumentId,
@@ -117,11 +122,9 @@ pub(crate) struct PersonalizedDocument {
     pub(crate) embedding: Embedding,
 
     /// Contents of the document properties.
-    #[serde(default)]
     pub(crate) properties: DocumentProperties,
 
     /// The tags associated to the document.
-    #[serde(default)]
     pub(crate) tags: Vec<String>,
 }
 
--- a/web-api/src/models.rs
+++ b/web-api/src/models.rs
@@ -91,13 +91,13 @@ id_wrapper!(DocumentPropertyId, is_valid_id, InvalidDocumentPropertyId);
 id_wrapper!(UserId, is_valid_id, InvalidUserId);
 
 #[derive(Clone, Debug, Deserialize, Serialize)]
-pub struct DocumentProperty(serde_json::Value);
+pub(crate) struct DocumentProperty(serde_json::Value);
 
 /// Arbitrary properties that can be attached to a document.
 pub(crate) type DocumentProperties = HashMap<DocumentPropertyId, DocumentProperty>;
 
 /// Represents a result from a query.
-#[derive(Debug, Clone, Serialize)]
+#[derive(Debug, Deserialize)]
 pub(crate) struct PersonalizedDocument {
     /// Unique identifier of the document.
     pub(crate) id: DocumentId,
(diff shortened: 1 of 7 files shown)


acrrd:
I think there has been some misunderstanding. `personalized_documents` must return the properties to avoid other calls. Also at the moment, those endpoints are not exposed to the user but only to the customer.
What we can do to reduce the traffic between elastic and our service is:
* we never read the snippet
* in the interaction endpoint we get only the embedding from elastic and not the whole document

janpetschexain:
> I think there has been some miscommunication. `personalized_documents` must return the properties to avoid other calls. Also, at the moment, those endpoints are not exposed to the user but only to the customer. What we can do to reduce the traffic between elastic and our service is:
> 
>     * we never read the snippet
> 
>     * in the interaction endpoint we get only the embedding from elastic and not the whole document

i'll update it accordingly

web-api/src/models.rs
```diff
@@ -91,13 +91,26 @@ id_wrapper!(DocumentPropertyId, is_valid_id, InvalidDocumentPropertyId);
...
+#[derive(Debug)]
+pub(crate) struct InteractedDocument {
+    /// Unique identifier of the document.
+    pub(crate) id: DocumentId,
+
+    /// Embedding from smbert.
+    pub(crate) embedding: Embedding,
+
+    /// The tags associated to the document.
+    pub(crate) tags: Vec<String>,
```
rustonaut, line 109: Maybe still add `#[serde(default)]` to make it a bit more robust :thinking:

web-api/src/models.rs
```diff
@@ -106,15 +119,12 @@ pub(crate) struct PersonalizedDocument {
...
     /// Embedding from smbert.
-    #[serde(skip_serializing)]
     pub(crate) embedding: Embedding,
 
     /// Contents of the document properties.
-    #[serde(default)]
     pub(crate) properties: DocumentProperties,
 
     /// The tags associated to the document.
-    #[serde(default)]
```
rustonaut, line 117: I'm not sure if removing the default is the best idea here having it a bit more robust by default for a non-fixed schema backend like elastic search is a good idea even if we don't  take advantage of it.

web-api/src/storage.rs
```diff
@@ -66,7 +67,12 @@ pub(crate) enum DeletionError {
 
 #[async_trait]
 pub(crate) trait Document {
-    async fn get_by_ids(&self, ids: &[&DocumentId]) -> Result<Vec<PersonalizedDocument>, Error>;
+    async fn get_interacted(&self, ids: &[&DocumentId]) -> Result<Vec<InteractedDocument>, Error>;
+
+    async fn get_personalized(
```
rustonaut, line 72: :art: We can keep the naming, but I'm not sure if it's most appropriate as they do get used in other contexts.

web-api/src/storage/elastic.rs
```diff
@@ -231,32 +224,54 @@ struct Hit<T> {
...
+    }
+}
+
+#[derive(Debug, Deserialize)]
+struct PersonalizedDocument {
+    #[serde(default)]
+    properties: DocumentProperties,
+    embedding: Embedding,
+    #[serde(default)]
+    tags: Vec<String>,
```
rustonaut, line 265: Couldn't we just use `[models](models::PersonalizedDocument)` like before, this would keep things a bit more simple, less indirection and confusions about what  does serialize in which way.

rustonaut: Same for IngestedDocument.

web-api/src/storage.rs
janpetschexain: i agree that the names aren't perfect, i just could come up with any better, i'm happy to change them for new suggestions.

web-api/src/storage/elastic.rs
janpetschexain: i'd say no for a few reasons:
- it is the specific purpose of this task to minimize traffic in elastic queries, so all unnecessary fields are removed in `elastic::PersonalizedDocument` compared to the `models::PersonalizedDocument` which has additional fields
- the `models::PersonalizedDocument` is the general kind of a personalized document used in storage contexts, whereas the `elastic::PersonalizedDocument` is the specific kind for the elastic storage, depending on the concrete storage impl we might also require further personalized document kinds in the future

rustonaut: I still believe it would be better to only have `models::PersonalizedDocument` and `models::InteractedDocument`  at least as long as the storage representation does (mostly) match the runtime representation, which I think we always should keep matching.

I also still don't think having more then one storage engine was a good idea, or that we ever should add another one.

rustonaut (approved):

web-api/src/storage/elastic.rs
janpetschexain: > I still believe it would be better to only have models::PersonalizedDocument and models::InteractedDocument at least as long as the storage representation does (mostly) match the runtime representation, which I think we always should keep matching.

it would be nicer, but it's not possible. even if we only had elastic as storage and no storage abstractions, we would still need different types, for example the id field is retrieved in a different structure layout than we require according to the openapi spec. also it wouldn't be possible to save traffic, which is the whole purpose of this task.

janpetschexain:
bors merge

janpetschexain:
bors merge

PR #751, opened 2022-12-20
[Merged by Bors] - Update openapi docs

frankpepermans:
Updates the docs

### DIFF
--- a/web-api/openapi/back_office.yaml
+++ b/web-api/openapi/back_office.yaml
@@ -4,20 +4,39 @@ info:
   version: 1.0.0-rc6
   description: |-
     # Back Office
-    For this system, a document is anything that has an id, a snippet, and an arbitrary set of properties.
-    The id uniquely identifies the document in the system.
-    The system uses the snippet to create a document representation that can be matched against user interests.
+    The back office is typically used within server-side apps.
+    This SDK acts as a create/read/update/delete interface for anything related to documents.
 
-    It is possible to attach any properties to a document; the system will not use them but will be returned together with the document when asking for personalized documents for a user. They could help store information like a title and the URL where it is possible to find the document. Each document can have its own set of properties.
+    ## Document
+    By document, we refer to a cohesive text, for example a complete news article.
+    However, we only require a simple representation for our system to work. Consisting just of a unique ID, a text snippet and optional properties.
+    The text snippet is ideally a short, meaningful representation of the larger document, reduced to just one or two paragraphs.
 
-    The API offers an endpoint that allows uploading the documents to the system. After a successful upload, it will be possible to use the document in other endpoints.
+    ### Id
+    The document_id is a unique identifier for a single document.
+    We don't require a specific format, but would suggest using UUID for example.
 
-    # Auth
-    To authenticate with the API, a token will be provided.
+    ## Snippet
+    A snippet is a reduced representation of a larger text, for example if the document would be a news article, then it's text would be the news article in a pure textual format.
+    In order for our system to work correctly, it is important that the snippet is just the size of one or two paragraphs, and that the snippet's text clearly summarises the larger text.
+    For example, let's take a news article which talks about the effects of inflation.
+    Our snippet could then be: "Inflation worries as prices keep rising. People having budgetary difficulties as a result. Government pressured to take action.".
+    A bad snippet would for example be just taking the very first paragraph of a document, which could sometimes work, but could also lead to:
+    "December 20th 2020. Article written by Jane Doe. On this bright and sunny day, people might forget about their worries sometimes."
+
+    ## Properties
+    Properties are optional entries for documents, which are usually needed to properly show the document back to the user, when it returns as a personalized document.
+    If for example, you'd wish to integrate a type of carousel view, listing a total of 10 personalized documents in a "for you"-section, then you might choose to display each document as an image and title, with a url to link the user to when pressed.
+    For this, we'd need three document properties: `image`, `link` and `title`.
+    You are free to add as many properties as needed.
+
+    ## Auth
+    The API requires an authenticationToken when used.
+    Please see further down in this README.md file, or refer to the test included in this SDK's codebase, on how to set this token correctly.
 
 tags:
   - name: back office
-    description: Management of the documents in the system.
+    description: Manage documents in a typical create/read/update/delete manner.
 
 security:
   - ApiKeyAuth: []
--- a/web-api/openapi/back_office.yaml
+++ b/web-api/openapi/back_office.yaml
@@ -17,7 +17,7 @@ info:
     We don't require a specific format, but would suggest using UUID for example.
 
     ## Snippet
-    A snippet is a reduced representation of a larger text, for example if the document would be a news article, then it's text would be the news article in a pure textual format.
+    A snippet is a reduced representation of a larger text, for example if the document would be a news article, then its text would be the news article in a pure textual format.
     In order for our system to work correctly, it is important that the snippet is just the size of one or two paragraphs, and that the snippet's text clearly summarises the larger text.
     For example, let's take a news article which talks about the effects of inflation.
     Our snippet could then be: "Inflation worries as prices keep rising. People having budgetary difficulties as a result. Government pressured to take action.".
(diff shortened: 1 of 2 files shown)


web-api/openapi/back_office.yaml
```diff
@@ -4,20 +4,39 @@ info:
...
-    It is possible to attach any properties to a document; the system will not use them but will be returned together with the document when asking for personalized documents for a user. They could help store information like a title and the URL where it is possible to find the document. Each document can have its own set of properties.
+    ## Document
+    By document, we refer to a cohesive text, for example a complete news article.
+    However, we only require a simple representation for our system to work. Consisting just of a unique ID, a text snippet and optional properties.
+    The text snippet is ideally a short, meaningful representation of the larger document, reduced to just one or two paragraphs.
 
-    The API offers an endpoint that allows uploading the documents to the system. After a successful upload, it will be possible to use the document in other endpoints.
+    ### Id
+    The document_id is a unique identifier for a single document.
+    We don't require a specific format, but would suggest using UUID for example.
```
acrrd, line 17: We don't require a specific format; it must comply with the following regexp `^[a-zA-Z0-9_:@.-]+$`.

web-api/openapi/back_office.yaml
```diff
@@ -4,20 +4,39 @@ info:
...
-    The id uniquely identifies the document in the system.
-    The system uses the snippet to create a document representation that can be matched against user interests.
+    The back office is typically used within server-side apps.
+    This SDK acts as a create/read/update/delete interface for anything related to documents.
 
-    It is possible to attach any properties to a document; the system will not use them but will be returned together with the document when asking for personalized documents for a user. They could help store information like a title and the URL where it is possible to find the document. Each document can have its own set of properties.
+    ## Document
+    By document, we refer to a cohesive text, for example a complete news article.
+    However, we only require a simple representation for our system to work. Consisting just of a unique ID, a text snippet and optional properties.
+    The text snippet is ideally a short, meaningful representation of the larger document, reduced to just one or two paragraphs.
```
acrrd, line 13: Let's go for only one paragraph.

web-api/openapi/back_office.yaml
```diff
@@ -4,20 +4,39 @@ info:
...
+    For example, let's take a news article which talks about the effects of inflation.
+    Our snippet could then be: "Inflation worries as prices keep rising. People having budgetary difficulties as a result. Government pressured to take action.".
+    A bad snippet would for example be just taking the very first paragraph of a document, which could sometimes work, but could also lead to:
+    "December 20th 2020. Article written by Jane Doe. On this bright and sunny day, people might forget about their worries sometimes."
+
+    ## Properties
+    Properties are optional entries for documents, which are usually needed to properly show the document back to the user, when it returns as a personalized document.
+    If for example, you'd wish to integrate a type of carousel view, listing a total of 10 personalized documents in a "for you"-section, then you might choose to display each document as an image and title, with a url to link the user to when pressed.
+    For this, we'd need three document properties: `image`, `link` and `title`.
```
acrrd, line 30: We don't need anything; they do.
```suggestion
    For this, you would need three document properties: `image`, `link` and `title`.
```

web-api/openapi/back_office.yaml
```diff
@@ -4,20 +4,39 @@ info:
...
+    You are free to add as many properties as needed.
+
+    ## Auth
+    The API requires an authenticationToken when used.
```
acrrd, line 34:
```suggestion
    The API requires an authentication token when used.
```

web-api/openapi/back_office.yaml
```diff
@@ -4,20 +4,39 @@ info:
...
+    Please see further down in this README.md file, or refer to the test included in this SDK's codebase, on how to set this token correctly.
 
 tags:
   - name: back office
-    description: Management of the documents in the system.
+    description: Manage documents in a typical create/read/update/delete manner.
```
acrrd, line 39: "All you to manage the system, including the document and their properties" or something similar?
We changed the name to back office to avoid relating too strongly with only documents. Also, we might add some reading to the front-office part.

web-api/openapi/front_office.yaml
```diff
@@ -4,29 +4,41 @@ info:
...
-    The system identifies a user with only an id.
-    There are two main endpoints: one that allows registering an interaction between a user and a document (a user clicked or liked a document), and one to retrieve a personalized set of documents for the user. The system needs a few user interactions with documents before it can generate a personalized list.
+    The front office is typically used within front-end apps, for example a website or a mobile application.
+    With this SDK, you can handle interactions with documents and request a list of personalized documents.
 
-    # Auth
-    To authenticate with the API, a token will be provided.
+    ## User
+    Each method requires a `user_id`.
+    From Xayn's perspective, a `user_id` is simply required to group interactions together. We don't need to know who that user is, so it is preferred to create this `user_id` in a privacy-preserving way.
```
acrrd, line 12:
```suggestion
    From our perspective, a `user_id` is simply required to group interactions together. We don't need to know who that user is, so it is preferred to create this `user_id` in a privacy-preserving way.
```
no strong opinion

web-api/openapi/front_office.yaml
```diff
@@ -4,29 +4,41 @@ info:
...
+    For example, create a hash method which transforms your user into an ID hash.
+
+    ## Document
+    By document, we refer to a cohesive text, for example a complete news article.
+    However, we only require a simple representation for our system to work. Consisting just of a unique ID, a text snippet and optional properties.\
+    The text snippet is ideally a short, meaningful representation of the larger document, reduced to just one or two paragraphs.
```
acrrd, line 18: This api does not know about the snippet. A document here has an id, some properties, and a score associated to it.

web-api/openapi/front_office.yaml
```diff
@@ -4,29 +4,41 @@ info:
...
+
+    ## Auth
+    The API requires an authenticationToken when used.
+    Please see further down in this README.md file, or refer to the test included in this SDK's codebase, on how to set this token correctly.
 
 tags:
   - name: front office
-    description: Interaction of the user and personalized document.
+    description: Manage user interactions and fetch personalized documents.
```
acrrd, line 26: Here I would put something more like the intro of this document about this being the API the user interacts with, and there is no managing involved.

web-api/openapi/front_office.yaml
```diff
@@ -4,29 +4,41 @@ info:
...
-  - ApiKeyAuth: []    
+  - ApiKeyAuth: []
 
 paths:
   /users/{user_id}/personalized_documents:
     get:
       tags:
         - front office
-      summary: Get personalized documents for the user
+      summary: Gets personalized documents for a user.
```
acrrd, line 36: I would keep "the" because it refers to the user specified by `user_id`.

web-api/openapi/front_office.yaml
```diff
@@ -60,9 +72,12 @@ paths:
     patch:
       tags:
         - front office
-      summary: Add interaction between a user and a document
+      summary: Adds an interaction between a user and a document.
       description: |-
-        The interaction is used to provide personalized documents to the user.
+        Use this method to register an interaction between a user and a document.
+        Currently, we only support a `Positive` interaction, which is equivalent to a user who "likes" a certain document.
+        For web sites, consider triggering this method whenever a certain document url loads.
```
acrrd, line 79: We don't want to get when the URL load, but if the user stays "long enough" or scroll a certain amount of the article. Something that could point out that the user might be interested in this and is not a misclick.

acrrd (approved):

web-api/openapi/back_office.yaml
acrrd: Also, it is max 256 long and cannot be empty.
This is the specification https://github.com/xaynetwork/xayn_discovery_engine/blob/main/web-api/openapi/schemas/id.yml

acrrd: The limit is not technical so if needed can be increased. Is to have some bound and avoid ids that are too long because it can hurt performance.

web-api/openapi/front_office.yaml
frankpepermans: They need to already show suggested articles when a user opens a web page,
so the flow is await updateInteractions (current document viewed)
then, get personalized content.

This is also why we went from 2 to 1 interaction, so we could do this immediately.

In the end, it's up to them to implement it of course

web-api/openapi/back_office.yaml
frankpepermans: Added the regexp before I saw the other comments, I guess it's fine, can also mention the limit in the email

frankpepermans:
Bors merge

PR #755, opened 2022-12-22
[Merged by Bors] - ET-3781 update rust toolchain to 1.66.0

janpetschexain:
**Reference**

- [ET-3781]
- followed by #756

**Summary**

- update toolchain to `rustc 1.66.0`, `rustc nightly-2022-12-15` and `just 1.9.0`
- update rust deps & transitive deps
- cleanup rust workspace deps, after the removal of the app engine some are unused and some are only crate deps
- fix new rust lints
- update ci workflow deps
- release ci docker image `v8`

[ET-3781]: https://xainag.atlassian.net/browse/ET-3781?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


web-api/src/storage/elastic.rs
```diff
@@ -562,7 +562,7 @@ impl storage::DocumentProperty for Storage {
             ["_source", document_id.as_ref()],
             [(
                 "_source_includes",
-                Some(&*format!("properties.{}", property_id)),
+                Some(&*format!("properties.{property_id}")),
```
frankpepermans, line 565: Finally!
Hated the old way :)

janpetschexain: this is actually possible since a few versions :) now there is a lint to enforce it. the main caveat is that this currently only works with variable names, if you have a functions call or a a object field reference etc the old interpolation syntax is still required :/

frankpepermans: Probably because it's still a macro of course, would be quite complex to implement e.g. a function call :/

acrrd (approved):

web-api/src/mind.rs
```diff
@@ -420,8 +420,8 @@ fn run_user_benchmark() -> Result<(), Error> {
         // Placeholder for interacting with the entire click history
         for click in impression.clicks {
             match document_provider.get(&click) {
-                Some(document) => println!("The document {:?} was interacted.", document),
-                None => println!("Document id {} not found.", click),
+                Some(document) => println!("The document {document:?} was interacted."),
```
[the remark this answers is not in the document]
acrrd: I tested it, and it works correctly. The lint is only in `clippy:pedantic` so there isn't much consensus on it and we could also have disabled it.

janpetschexain:
bors merge

PR #758, opened 2023-01-03
[Merged by Bors] - ET-3565 update spec

rustonaut:
-  `publication_date` is now in `properties` 
    -  it's updatable, deletable but limited to a RFC3339 timestamp
 -  conceptually documents with a publication_date in the future, or without one are no treated as unpublished. This means they will be filtered out from personalized documents. But they will not be directly discarded. They still will be ingested and the date can be set later allowing for scheduled or delayed publishing. Making it easier to integrate our service into a typical news paper workflow (which often involves scheduled publishing, canceled scheduled publishing, delayed publishing, ubpublishing etc. all now possible)

### DIFF
--- a/.spectral.js
+++ b/.spectral.js
@@ -39,7 +39,7 @@ module.exports = {
     'delete-body': 'off',
     'prohibit-summary-sentence-style': 'off',
     'collection-array-property': 'off',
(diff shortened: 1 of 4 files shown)


frankpepermans (approved):

rustonaut:
bors merge

PR #759, opened 2023-01-03
ET-3564 date filter elastic search

rustonaut:

PR #761, opened 2023-01-05
[Merged by Bors] - ET-3845 cherrypick elastic search mapping refactoring

rustonaut:
- Introduces `web-api/elastic-search/mapping.json` as source of truth for what the ES mapping should be.
- Split out helper script to set that mapping 
- Refactored `compose.db.yml` to use that mapping and helper script.

 This is needed for following reasons:

- a source of truth so that we can properly update the mapping
- using a single source of truth in different contexts (compose.db.yml and integration tests for now) 
- set the foundation for features we likely will need in the future:
  - we can add CI functionality to use the file automatically for new deployments
  - we can extend it to support mapping "migrations"  in some way to allow updating
    existing deployments in a semi-automatic way 

References:

- ET-3845
- ET-3564
- #759


Robert-Steiner (approved):

rustonaut:
bors merge

PR #765, opened 2023-01-10
[Merged by Bors] - Fix on conflict do update failure with two interactions.

rustonaut:
**Problem:**

In some situations we get `ON CONFLICT DO UPDATE command cannot affect row a second time `.

This happens if in a `INSERT` statments values list there are multiple tuples with the same key and the key is already in the database triggering multiple ` ON CONFLICT DO UPDATE` for the same key in the same insert statement.

**Changes:**

- changed various `upsert_*` to accept maps instead of iterators to avoid the problem by enforcing unique keys, most functions created the iterators from maps anyway

- removed some unnecessary "if empty return Ok" micro-optimizations

### DIFF
--- a/web-api/src/storage/postgres.rs
+++ b/web-api/src/storage/postgres.rs
@@ -35,7 +35,7 @@ use xayn_ai_coi::{CoiStats, Embedding, NegativeCoi, PositiveCoi, UserInterests};
 
 use super::InteractionUpdateContext;
 use crate::{
-    models::{DocumentId, UserId, UserInteractionType},
+    models::{DocumentId, InteractedDocument, UserId, UserInteractionType},
     storage::{self, utils::SqlxPushTupleExt, Storage},
     utils::serialize_redacted,
     Error,
@@ -353,18 +353,13 @@ impl Database {
         Ok(())
     }
 
-    async fn upsert_interactions<'d, I>(
+    async fn upsert_interactions(
         tx: &mut Transaction<'_, Postgres>,
         user_id: &UserId,
         now: DateTime<Utc>,
-        interactions: I,
-    ) -> Result<(), Error>
-    where
-        I: IntoIterator<Item = (&'d DocumentId, UserInteractionType)>,
-        I::IntoIter: ExactSizeIterator,
-    {
-        let interactions = interactions.into_iter();
-        if interactions.len() == 0 {
+        interactions: &HashMap<&DocumentId, (&InteractedDocument, UserInteractionType)>,
+    ) -> Result<(), Error> {
+        if interactions.is_empty() {
             return Ok(());
         }
         //FIXME micro benchmark and chunking+persist abstraction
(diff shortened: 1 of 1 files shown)


web-api/src/storage/postgres.rs
```diff
@@ -322,7 +318,9 @@ impl Database {
             last_view
         ) ",
         );
-        for chunk in cois.chunks(Database::BIND_LIMIT / 7) {
```
rustonaut, line 325: :exclamation: : Itertools `chunks` extension is not compatible with `await` (it's not `Send`).

web-api/src/storage/postgres.rs
```diff
@@ -530,36 +518,28 @@ impl storage::Interaction for Storage {
...
+            if let Some((document, _)) = document_map.get_mut(id) {
+                let updated_coi = update_logic(InteractionUpdateContext {
                     document,
                     tag_weight_diff: &mut tag_weight_diff,
                     positive_cois: &mut interests.positive,
-                }));
+                });
+                // We might update the same coi min `interests` multiple times,
+                // if we do we only want to keep the latest update.
+                updates.insert(updated_coi.id, updated_coi);
```
rustonaut, line 532: :exclamation: This was one of the places which could trigger the bug.

web-api/src/storage/postgres.rs
```diff
@@ -353,38 +351,33 @@ impl Database {
...
-    where
-        I: IntoIterator<Item = (&'d DocumentId, UserInteractionType)>,
-        I::IntoIter: ExactSizeIterator,
-    {
-        let interactions = interactions.into_iter();
-        if interactions.len() == 0 {
-            return Ok(());
-        }
+        interactions: &HashMap<&DocumentId, (&InteractedDocument, UserInteractionType)>,
+    ) -> Result<(), Error> {
```
rustonaut, line 359: As this was created from a map before this couldn't trigger the bug before, I still changed it to avoid running into the bug again in the future.

web-api/src/storage/postgres.rs
```diff
@@ -398,18 +391,13 @@ impl Database {
...
         user_id: &UserId,
-        updates: I,
-    ) -> Result<(), Error>
-    where
-        I: IntoIterator<Item = (&'c str, i32)>,
-    {
-        let updates = updates.into_iter();
-
+        updates: &HashMap<&str, i32>,
+    ) -> Result<(), Error> {
```
rustonaut, line 398: As this was created from a map before this couldn't trigger the bug before, I still changed it to avoid running into the bug again in the future.

web-api/src/storage/postgres.rs
rustonaut: not sure what you mean?

Update coi in-place updated the `Vec` of cois so it always has the updated cois.

rustonaut: This is not part of my design but a left over from CoI handling in the app.

janpetschexain: > There is an additional problem here, the updated_coi must be fed to the next interaction.

not sure if i understand this? all cois are modified via the context as a ref mut, and the updated coi is stored in the map, so shouldn't all information be available to the next interaction without a need to pass the value?

janpetschexain (approved):

web-api/src/storage/postgres.rs
acrrd: than everything is fine. I didn't look at what the update function was doing

rustonaut:
bors merge

PR #766, opened 2023-01-12
NO_ISSUE: remove cartesian_product from pairwise_cosine_similarity

frankpepermans:
Leftover from an old PR which improves performance a bit.

Can be ignored/closed if no longer relevant.

- avoids unnecessary iterations via cartesian_product
- minor cleanup to avoid copy


janpetschexain:
since we remove that method all together in #768, we can probably close this pr?

PR #767, opened 2023-01-12
[Merged by Bors] - NO_ISSUE: Remove dependency on unix

frankpepermans:
I'm running Windows (yes.. I know...)
While I can also run a Linux subsystem, it's finically to work with.

and os::unix is unavailable, this small change should get rid of that dependency


rustonaut (approved):
lgtm, the dangers of auto imports :smile:

frankpepermans:
bors merge

PR #768, opened 2023-01-12
[Merged by Bors] - ET-3891: Remove obsolete code

frankpepermans:


coi/src/embedding.rs
```diff
@@ -37,62 +36,18 @@ where
...
-        similarities[[i, j]] = (a.dot(&b) / norms[i] / norms[j]).clamp(-1., 1.);
-        similarities[[j, i]] = similarities[[i, j]];
-    }
-
-    similarities
-}
-
 /// Computes the cosine similarity of two vectors.
-///
-/// See [`pairwise_cosine_similarity`] for details.
```
janpetschexain, line 95: 🔧 we need to replace this reference with the deleted paragraph from the other docs, otherwise the references at the constants above point into void.
```suggestion
/// Each value is bounded in `[-1, 1]`. The zero vector is always "similar" to all other vectors,
/// thus will yield a similarity of 1.
```

janpetschexain (approved):
just an issue with the docs, i'll already approve it

coi/src/embedding.rs
frankpepermans: Yes will change

janpetschexain (approved):
lgtm

frankpepermans:
bors merge

PR #770, opened 2023-01-15
[Merged by Bors] - ET-3862 Embedding Normalization

frankpepermans:
This PR moves the normalization of embeddings to the ingestion, then adapts code throughout so that we expect normalized embeddings everywhere.

Caveat: many tests are written with regular embeddings, a helper function was therefore added, so that we can just convert value Arrays into normalized value Arrays.

Perks we now get:
- Invalid embedding vectors are now detected at ingestion time, and can be rejected early
- Cosine similarity is now just one matmul, instead of 3 before

### DIFF
--- a/bert/src/pooler.rs
+++ b/bert/src/pooler.rs
@@ -114,23 +114,6 @@ impl TryFrom<Vec<u8>> for Embedding<Ix1> {
     }
 }
 
-/// Computes the l2 norm (euclidean metric).
-///
-/// *NOTE* This used to `panic` before, but as we now calculate l2 norm upon ingestion,
-/// this behavior changed to become `None`, so that underlying code could still follow up.
-fn l2_norm<S>(
-    a: ArrayBase<S, Ix1>,
-    b: ArrayBase<S, Ix1>,
-) -> Result<f32, InvalidVectorEncounteredError>
-where
-    S: Data<Elem = f32>,
-{
-    match a.view().dot(&b.view()).sqrt() {
-        n if n.is_finite() => Ok(n),
-        _ => Err(InvalidVectorEncounteredError),
-    }
-}
-
 /// A 2-dimensional sequence embedding.
 ///
 /// The embedding is of shape `(token_size, embedding_size)`.
@@ -308,37 +291,5 @@ mod tests {
         let mask = arr2(&[[1, 1]]).into();
         let embedding = AveragePooler::pool(&prediction.into(), &mask).unwrap();
         assert_eq!(embedding, arr1(&[2.5, 3.5, 4.5]));
-    }
-
-    #[test]
-    fn test_l2_norm() {
-        xayn_ai_test_utils::assert_approx_eq!(
-            f32,
-            l2_norm(arr1(&[1., 2., 3.]), arr1(&[1., 2., 3.])).unwrap(),
-            3.741_657_5
-        );
-    }
-
-    #[test]
-    fn test_l2_norm_nan() {
-        assert!(l2_norm(arr1(&[1., f32::NAN, 3.]), arr1(&[1., f32::NAN, 3.])).is_err());
-    }
-
-    #[test]
-    fn test_l2_norm_inf() {
-        assert!(l2_norm(
-            arr1(&[1., f32::INFINITY, 3.]),
-            arr1(&[1., f32::INFINITY, 3.])
-        )
-        .is_err());
-    }
-
-    #[test]
-    fn test_l2_norm_neginf() {
-        assert!(l2_norm(
-            arr1(&[1., f32::NEG_INFINITY, 3.]),
-            arr1(&[1., f32::NEG_INFINITY, 3.])
-        )
-        .is_err());
     }
 }
--- a/bert/src/pooler.rs
+++ b/bert/src/pooler.rs
@@ -46,7 +46,7 @@ impl Embedding<Ix1> {
 
     pub fn normalized(&self) -> Result<Self, InvalidVectorEncounteredError> {
         match self.view().l2_norm()? {
-            norm if norm < 0. => Ok(self.clone()),
+            norm if norm <= 0. => Ok(self.mapv(|_| f32::MAX).into()), // very far away
             norm => Ok(self.mapv(|x| x / norm).into()),
         }
     }
--- a/bert/src/pooler.rs
+++ b/bert/src/pooler.rs
@@ -44,8 +44,9 @@ impl Embedding<Ix1> {
         self.iter().flat_map(|value| value.to_le_bytes()).collect()
     }
 
-    pub fn normalized(&self) -> Result<Self, InvalidVectorEncounteredError> {
-        match self.view().l2_norm()? {
+    pub fn normalize(&self) -> Result<Self, InvalidVectorEncounteredError> {
+        let view = self.view();
+        match l2_norm(view, view)? {
             norm if norm <= 0. => Ok(self.mapv(|_| f32::MAX).into()), // very far away
             norm => Ok(self.mapv(|x| x / norm).into()),
         }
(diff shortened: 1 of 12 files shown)


bert/src/pooler.rs
```diff
@@ -43,6 +43,13 @@ impl Embedding<Ix1> {
     pub fn to_bytes(&self) -> Vec<u8> {
         self.iter().flat_map(|value| value.to_le_bytes()).collect()
     }
+
+    pub fn normalized(&self) -> Result<Self, InvalidVectorEncounteredError> {
+        match self.view().l2_norm()? {
+            norm if norm <= 0. => Ok(self.mapv(|_| f32::MAX).into()), // very far away
+            norm => Ok(self.mapv(|x| x / norm).into()),
+        }
```
janpetschexain, line 51: :wrench: this mapping (as you mentioned on slack) seems to be related to some previous handling in the cosine similarity computation. since this function is about normalization, and the metric is handled in another processing step, we should map the edge case to a zero vector instead of a f32::MAX vector, otherwise that would be quite unexpected for normalization. then the cosine similarity/dot product can still act on it later on.

in combination with moving in the parts from the trait impl, i'd suggest sth like:
```suggestion
        let view = self.view();
        let norm = view.dot(&view).sqrt();
        if norm.is_finite() {
            let normalized = if norm <= 0. {
                Array1::zeros(self.len())
            } else {
                &self.0 / norm
            };
            Ok(normalized.into())
        } else {
            Err(InvalidVectorEncounteredError)
        }
```

we can use broadcasting to perform the division. also as a side note: the norm can never be strictly less than zero, neither numerically nor mathematically, we only had `<= 0` instead of `== 0` to avoid float comparison lints.

coi/src/embedding.rs
```diff
@@ -94,82 +69,25 @@ where
 ///
 /// See [`pairwise_cosine_similarity`] for details.
 pub fn cosine_similarity(a: ArrayView1<'_, f32>, b: ArrayView1<'_, f32>) -> f32 {
```
janpetschexain, line 71: 🔧 we need to rename the function to reflect on the new body, maybe sth like
```suggestion
pub fn normalized_dot_product(a: ArrayView1<'_, f32>, b: ArrayView1<'_, f32>) -> f32 {
```
it's probably also better to merge #768 first to avoid conflicts with `pairwise_cosine_similarity()`.

coi/src/embedding.rs
```diff
@@ -94,82 +69,25 @@ where
...
-        return 1.;
-    }
-
-    let norm_b = l2_norm(b.view());
-    if norm_b <= 0. {
-        return 1.;
-    }
-
-    (a.dot(&b) / norm_a / norm_b).clamp(-1., 1.)
+    a.dot(&b).clamp(-1., 1.)
```
janpetschexain, line 72: :art:/:wrench: with the proposed changes to `normalized()`, we would need to map the edge case here, such that we still have the desired property such that the zero vector is similar to all other vectors (in reality this should never occur anyways), we have to be careful though that we don't confuse orthogonal vectors with the all-zero vector
```suggestion
    if a.iter().any(|&v| v != 0.) || b.iter().any(|&v| v != 0.) {
        a.dot(&b).clamp(-1., 1.)
    } else {
        1.
    }
```
since the check is short-circuiting, it should be very cheap in almost all cases.

coi/src/utils.rs
```diff
@@ -111,6 +114,23 @@ pub(crate) mod serde_duration_as_days {
...
+/// Will panic if the array cannot be normalized.
+pub fn normalize_array<const N: usize>(array: [f32; N]) -> [f32; N] {
+    let embedding: Embedding<Ix1> = Array::from_vec(array.into()).into();
+    embedding
+        .normalized()
+        .unwrap()
+        .as_slice()
+        .unwrap()
+        .try_into()
+        .unwrap()
```
janpetschexain, line 131: :art: we can create the embedding directly from the array, also using `to_vec` instead of `as_slice` makes this more robust in case of non-continguous memory
```suggestion
    Embedding::from(array)
        .normalized()
        .unwrap()
        .to_vec()
        .try_into()
        .unwrap()
```

web-api/src/ingestion/routes.rs
```diff
@@ -99,14 +100,29 @@ async fn new_documents(
...
-                error!(
-                    "Document with id '{}' caused a PipelineError: {:#?}",
-                    document.id, err,
-                );
-                Err(document.id.into())
+        .map(|document| {
+            match state
+                .embedder
+                .run(&document.snippet)
+                .and_then(|res| res.normalized().map_err(InternalError::from_std))
```
janpetschexain, line 107: ❓ wdyt about moving the call to `normalized()` into the `embedder.run()`? then we could guarantee that we never forget to normalize embeddings, independent of the call site.

bert/src/pooler.rs
```diff
@@ -228,6 +235,35 @@ impl AveragePooler {
...
+#[derive(Clone, Debug, Display, Error)]
+/// Bytes do not represent a valid embedding.
+pub struct InvalidVectorEncounteredError;
+
+/// Computes the l2 norm (euclidean metric).
+///
+/// When l2 norm could not be calculated, `None` is returned.
+///
+/// *NOTE* This used to `panic` before, but as we now calculate l2 norm upon ingestion,
+/// this behavior changed to become `None`, so that underlying code could still follow up.
```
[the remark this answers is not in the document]
frankpepermans: Yes, Had an Option before, but a Result seemed more correct, as an Err > None
Will adapt

bert/src/pooler.rs
```diff
@@ -228,6 +235,35 @@ impl AveragePooler {
...
+    S: Data<Elem = f32>,
+{
+    fn l2_norm(&self) -> Result<f32, InvalidVectorEncounteredError> {
+        let view = self.view();
+        match view.dot(&view).sqrt() {
+            n if n.is_finite() => Ok(n),
+            _ => Err(InvalidVectorEncounteredError),
+        }
+    }
+}
```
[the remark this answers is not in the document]
frankpepermans: Yup, can do

bert/src/pooler.rs
```diff
@@ -43,6 +43,13 @@ impl Embedding<Ix1> {
     pub fn to_bytes(&self) -> Vec<u8> {
         self.iter().flat_map(|value| value.to_le_bytes()).collect()
     }
+
+    pub fn normalized(&self) -> Result<Self, InvalidVectorEncounteredError> {
```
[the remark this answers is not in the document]
frankpepermans: can do, I was assuming "normalize" could wrongfully be interpreted as changing the inner values, so treating the Embedding as mut. 
But fine to change it oc

web-api/src/ingestion/routes.rs
frankpepermans: okido

coi/src/utils.rs
frankpepermans: yup!

coi/src/embedding.rs
frankpepermans: yup, wasn't really sure in these cases, will take your snippet

bert/src/pooler.rs
frankpepermans: ok will change

coi/src/utils.rs
```diff
@@ -111,6 +114,17 @@ pub(crate) mod serde_duration_as_days {
...
 
+/// # Panics
+///
+/// This method is primarily used for testing.
+/// It simply converts an Array into a normalized Array.
+///
+/// Will panic if the array cannot be normalized.
+pub fn normalize_array<const N: usize>(array: [f32; N]) -> [f32; N] {
+    let embedding: Embedding<Ix1> = Array::from_vec(array.into()).into();
+    embedding.normalize().unwrap().to_vec().try_into().unwrap()
```
janpetschexain, line 125: :art: we could still simplify this by (importing `use xayn_ai_bert::Embedding1;` instead)
```suggestion
    Embedding1::from(array)
        .normalize()
        .unwrap()
        .to_vec()
        .try_into()
        .unwrap()
```

web-api/src/embedding.rs
```diff
@@ -49,7 +49,11 @@ pub(crate) struct Embedder {
 
 impl Embedder {
     pub(crate) fn run(&self, s: &str) -> Result<Embedding, InternalError> {
-        self.bert.run(s).map_err(InternalError::from_std)
+        let embedding = self.bert.run(s);
+        match embedding {
```
janpetschexain, line 53: :art: we could match the result right away without an intermediate assignment
```suggestion
        match self.bert.run(s) {
```

janpetschexain (approved):
lgtm, there is a comment left though, i'll already approve it

PR #775, opened 2023-01-17
[Merged by Bors] - Don't consider partial files as downloaded

acrrd:
The current download script download an asset directly in the name we give. If the connections drops that we think that the file has been downloaded. This PR fix this by downloading to a temporary file, this also allows curl to resume the download from the previous state instead of starting from scratch.


acrrd:
bors merge

PR #780, opened 2023-01-23
[Merged by Bors] - ET-3893 remove not enough interactions error

janpetschexain:
**Reference**

- [ET-3893]

**Summary**

- move `UserInterests` from the `point` module to the `context` module and simplify the types for closest coi search
- remove the `context::Error`s and use `Option`s instead, all error occurences could only ever happen if the cois were empty
- return `Ok(NotEnoughInteractions)` instead of `Err(NotEnoughInteractions)` in the `personalized_documents` route

[ET-3893]: https://xainag.atlassian.net/browse/ET-3893?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/coi/benches/benches.rs
+++ b/coi/benches/benches.rs
@@ -18,13 +18,7 @@ use criterion::{black_box, criterion_group, BatchSize, Criterion};
 use itertools::Itertools;
 use rand::Rng;
 use rand_distr::Uniform;
-use xayn_ai_coi::{
-    compute_coi_relevances,
-    stats::compute_coi_decay_factor,
-    CoiId,
-    CoiPoint,
-    PositiveCoi,
-};
+use xayn_ai_coi::{compute_coi_decay_factor, compute_coi_relevances, CoiId, CoiPoint, PositiveCoi};
 
 fn create_positive_coi(n: usize, embedding_size: usize) -> Vec<PositiveCoi> {
     let range = Uniform::new(-1., 1.);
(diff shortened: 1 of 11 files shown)


acrrd:
Since we don't want to remove the error, should we close this?

janpetschexain:
> Since we don't want to remove the error, should we close this?

i reverted the changes in the `web-api` crate, but i would still like to merge the simplifications in the `coi` crate.

regarding the the error filtering via tracing, i looked further into that. with instrumentation there seems to be no way to exclude a specific error. what we could do, is adding a layer to the subscriber with a dynamic filter, where we might be able to access the values of the span fields in the `Layer::on_new_span()` or `Layer::on_event()` and filter for the `personalized_documents` path (sth along the lines of https://stackoverflow.com/questions/66575682/how-to-access-the-value-of-a-field-in-a-tracing-span). we would still need a way to filter for the `NotEnoughInteractions` error type, depends a bit how the type is recorded in the event, but this usually happens via a `dyn Debug/Display`, hence we would need to filter via strings anyways. overall this is quite complex for such a seemingly simple problem, the tracing is more designed for opt-in instead of opt-out. also dynamic filtering is of course more expensive than static filtering.

since you want to keep the distinction of no cois vs empty results, wdyt about adding a different return code for that? instead of a 409 (conflict) we could return a 204 (no content), then no error would be logged, but the user still has the possibility for distinction. this would also be more in line with the understanding that this is no actual error, but just an empty result for a specific reason.

acrrd (approved):

janpetschexain:
bors merge

PR #781, opened 2023-01-23
[Merged by Bors] - improve asset dir handling

rustonaut:
This fixes two problems:

1. Fix edge cases where (re-)serializing `RelativePathBuf` does not work.
2. Better error handling if the asset dir or config file in the asset dir is missing.

### DIFF
--- a/web-api/src/embedding.rs
+++ b/web-api/src/embedding.rs
@@ -12,6 +12,7 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
+use anyhow::bail;
 use serde::{Deserialize, Serialize};
 use xayn_ai_bert::{
     tokenizer::bert,
@@ -62,7 +63,21 @@ impl Embedder {
     }
 
     pub(crate) fn load(config: &Config) -> Result<Self, SetupError> {
-        let bert = BertConfig::new(config.directory.relative())?
+        let path = config.directory.relative();
+        if !path.exists() {
+            bail!(
+                "Fail to load Embedder: asset dir missing: {}",
+                path.display()
+            );
+        }
+        let config_file = path.join("config.toml");
+        if !config_file.exists() {
+            bail!(
+                "Fail to load Embedder: <assets>/config.toml doesn't exist: {}",
+                config_file.display()
+            );
+        }
+        let bert = BertConfig::new(path)?
             .with_tokenizer::<bert::Tokenizer>()
             .with_pooler::<AveragePooler>()
             .with_token_size(config.token_size)?


web-api/src/utils.rs
```diff
@@ -33,6 +31,31 @@ impl From<&str> for RelativePathBuf {
...
+        }
+    }
+}
+
+impl<'de> Deserialize<'de> for RelativePathBuf {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let value = String::deserialize(deserializer)?;
```
rustonaut, line 54: We can't use `<&str>::deserialize` due to how `serde(flatten)` works, the bug btw. was also caused by the subtle interactions of default values and `serde(flatten)`.

rustonaut:
This turned out to be much much more troublesome to fix.

Basically `#[serde(flatten)]` and `RelativePathBuf` are fundamentally (but subtle) incompatible  with each other
(there is no fix).

So I will limit this PR only to add the better error messages.

---

**To be precise:**

- `Figment` provides a special `Deserializer` which is `deserialize_struct` is called with one of the "magic" struct names (e.g. `___figment_relative_path_buf`)  it will call a special function not part of the serde framework (e.g. `<RelativePathBuf as Magic>::deserialize_from`) which will inspect the data and ad-hoc modify it. In case of `RelativePathBuf` this  means creating a `{ ___figment_relative_metadata_path: <meta>,  ___figment_relative_path: <value> }`  map from whatever there had been before (with some cleverness ) and then de-serializing  with that as deserializer/data.

- `serde(flatten)` runs a 2 pass deserialization:
   1. de-serialize the struct normally and collect all remaining/unknown fields into a `Vec<Option<(Content, Content)>>` (basically a optimized/flat arbitrary key value map
   2. use the collected remaining/unknown fields as deserializer/data to de-serialize the flattened struct
   
Depending on how you (de-)serialize `RelativePathBuf` you therefore run into one of two errors:

-  `RelativePathBuf` fails to deserialize because it expects a  `{ ___figment_relative_metadata_path: <meta>,  ___figment_relative_path: <value> }`  but gets a `String`, `&'de str` or similar.

- `RelativePathBuf` does deserialize but is not relative to the config file, i.e. it acts exactly like `PathBuf`

It's unfixable because to fix it you would need to interleave functionality which fundamentally can't know statically about each others existence in a "proper"/generic way  (so you would need to make serde non static with boxed callback maps and a lot of `dyn` and additional value allocations etc.).  I mean you could just hard code the transformation of any `embedding.directory` keys value to `{ ___figment_relative_metadata_path: <meta>,  ___figment_relative_path: <value> }` but that's prone to change errors as it needs to be done for any `RelativePath` in a flattened field.

**How to proceed:**

In deployment we only use the default value, same for most other places except some local testing use cases: So we are very free to change it in any way.

The only reasonable option would be:

1. to remove `RelativePath` from `Config` everywhere at the cost of paths in a config no longer being relative to a config
2. removing all usages of `serde(flatten)` 

The first option is the simplest, and we need it configurable for local testing.

But I have been unhappy with the config/app state extension setup in the sense that I think it would be better to make the shared parts it's own library (including most parts in server.rs)  and then making a crate for `ingestion` and a separte crate for personalization.  Backporting some of the refactoring in the private nc repo wrt. code stemming from server.rs would allow having the code of server.rs a nice library without any "extension" generic parts (you would have separate `Config` struct for both instead of generic + serde(flatten) and reuse the sub-configs). I think this is a cleanup we should do anyway but maybe not now :thinking:

janpetschexain (approved):

rustonaut:
bors merge

PR #782, opened 2023-01-24
[Merged by Bors] - ET-3940 validate route parameters

janpetschexain:
**Reference**

- [ET-3940]

**Summary**

- validate parameters in the routes: instead of deserializing directly into the type which skips validation, we now take parameters of basic type like `String` and then validate them to be a well formed type like `DocumentId` before calling further logic. we don't want to change the serde of types like `DocumentId` to be able to still do unvalidated deserialization from the databases.

[ET-3940]: https://xainag.atlassian.net/browse/ET-3940?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/web-api/src/error/common.rs
+++ b/web-api/src/error/common.rs
@@ -115,8 +115,9 @@ impl From<String> for BadRequest {
 }
 
 #[derive(Serialize, Debug, From)]
+#[from(types(DocumentId))]
 pub(crate) struct DocumentIdAsObject {
-    pub(crate) id: DocumentId,
+    pub(crate) id: String,
 }
 
 /// Internal Error: {0}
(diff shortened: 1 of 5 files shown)


rustonaut (approved):
I don't really like this, it IMHO creates a bunch of uncessary verbosity and messiness. 

But it does work and solve our problem for now, so I will approve it.

Generally I would argue that when we get data from PG/ES or similar we have to live with what we get and proceed as far as possible.  So I think it would be a better solution to only soft-guarantee correctness with the string new types like `DocumentId` and have some generic wrapper type which enforces this in interface types, e.g. `Checked<DocumentId>` adding a de-serialization impl. which enforces validity checks. And then we can use this in all the `Path` and `xxxRequest` types and still have type safety from the get to go. (probably also not impl `Deserialize` on `DocumentId` but have `Lenient<DocumentId>` or similar).

But I don't think we have currently time for refactoring like that.

janpetschexain:
bors merge

PR #783, opened 2023-01-24
[Merged by Bors] - ET-3895 define & impl tags constraints

janpetschexain:
**Reference**

- [ET-3894]
- [ET-3895]
- requires #782

**Summary**

- define tags in the spec
- impl tag validation, validate tags from incomming web requests & use validated tags internally

[ET-3894]: https://xainag.atlassian.net/browse/ET-3894?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ET-3895]: https://xainag.atlassian.net/browse/ET-3895?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/web-api/openapi/schemas/document.yml
+++ b/web-api/openapi/schemas/document.yml
@@ -28,7 +28,7 @@ DocumentPropertyId:
 DocumentTag:
   description: A tag of a document can be any non-empty, UTF-8-encoded string which doesn't contain a zero byte.
   type: string
-  pattern: '[^\x00]+$'
+  pattern: '^[^\x00]+$'
   minLength: 1
   maxLength: 256
   example: "tag"
--- a/web-api/openapi/schemas/document.yml
+++ b/web-api/openapi/schemas/document.yml
@@ -26,9 +26,9 @@ DocumentPropertyId:
   example: "title"
 
 DocumentTag:
-  description: A tag of a document can be any non-empty string that consists of alphanumerical characters, common punctuation and whitespace.
+  description: A tag of a document can be any non-empty, UTF-8-encoded string which doesn't contain a zero byte.
   type: string
-  pattern: '^[a-zA-Z0-9.:,;\-_+*!?%&/\\() ]{1,100}$'
+  pattern: '[^\x00]+$'
   minLength: 1
-  maxLength: 100
+  maxLength: 256
   example: "tag"
--- a/web-api/openapi/schemas/document.yml
+++ b/web-api/openapi/schemas/document.yml
@@ -24,3 +24,11 @@ DocumentPropertyId:
   description: Id of a property of a document.
   $ref: './id.yml#/Id'
   example: "title"
+
+DocumentTag:
+  description: A tag of a document can be any non-empty string that consists of alphanumerical characters, common punctuation and whitespace.
+  type: string
+  pattern: '^[a-zA-Z0-9.:,;\-_+*!?%&/\\() ]{1,100}$'
+  minLength: 1
+  maxLength: 100
+  example: "tag"
--- a/web-api/openapi/schemas/id.yml
+++ b/web-api/openapi/schemas/id.yml
@@ -2,7 +2,7 @@ Id:
   description: |-
     An id can be any non-empty string that consist of digits, latin letters, underscores, colons, minus signs, at signs, and dots.
   type: string
-  pattern: '^[a-zA-Z0-9_:@.-]+$'
+  pattern: '^[a-zA-Z0-9_:@.\-]{1,256}$'
   minLength: 1
   maxLength: 256
   example: "valid_id1"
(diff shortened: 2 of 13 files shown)


web-api/openapi/schemas/document.yml
```diff
@@ -24,3 +24,11 @@ DocumentPropertyId:
   description: Id of a property of a document.
   $ref: './id.yml#/Id'
   example: "title"
+
+DocumentTag:
+  description: A tag of a document can be any non-empty string that consists of alphanumerical characters, common punctuation and whitespace.
+  type: string
+  pattern: '^[a-zA-Z0-9.:,;\-_+*!?%&/\\() ]{1,100}$'
```
rustonaut, line 31: I don't think we should set a pattern for this.

If we do we must allow unicode letters (e.g. äöüß, Frensh, and tags might even be in japanese even if the articles are not).

This is especially important given that some of our (potential) customers use author names as tags.

So at least all character classes and related non-visible unicode needs to be supported (combinators, zero-sized whitespace, text order changers etc.). For especially rare cases symbols need to be supported, too. And naturally non ascii numeric, too.

I think the easiest would be to allow everything except control characters (assuming bidi control doesn't count as control) as well as tab and newline, but then there is no real reason not to allow them as we still have a max bytes limit anyway.

The regex would be something like `^\P{Control}{1,100}$` (`\P{}` is inverse of `\p` and it's here enough as `\t`, `\r` and `\n` are all control characters. The bidi marks are not covered by `\p{Control}` so this aspect works, too.

Note that  this is 100 characters, while the current patters is 100 bytes (as bytes happen to be characters).

rustonaut: Also one problem with that approach is that it might not work for some of the generated sdks.

For example in the browser `/^\P{Control}{1,100}$/` does compile but not behave at all like expected, instead you need to use `/^\P{Control}{1,100}$/u` and no idea if the code gen does so.

So probably better to not set a patter and only set a byte limit :thinking: 

Or only reject `\0`.

web-api/openapi/schemas/id.yml
```diff
@@ -2,7 +2,7 @@ Id:
   description: |-
     An id can be any non-empty string that consist of digits, latin letters, underscores, colons, minus signs, at signs, and dots.
   type: string
-  pattern: '^[a-zA-Z0-9_:@.-]+$'
+  pattern: '^[a-zA-Z0-9_:@.\-]{1,256}$'
   minLength: 1
   maxLength: 256
```
rustonaut, line 7: This now duplicates the length constraints making them more prone to change induced bugs/desync.

web-api/openapi/schemas/document.yml
janpetschexain: i guess only rejecting `\0` and having a byte limit is the most flexible, i'll change it.

web-api/openapi/schemas/id.yml
janpetschexain: but with just the `+` repetition, the pattern contradicts the limits. so how should we formulate this instead? i can also revert this and we can find a solution in a follow up, but this conflict should be resolved.

rustonaut: I would see it as the length further restricting the generic pattern in a similar way you would have "english grammar" and text size limitations of a book separate.

web-api/openapi/schemas/document.yml
```diff
@@ -24,3 +24,11 @@ DocumentPropertyId:
   description: Id of a property of a document.
   $ref: './id.yml#/Id'
   example: "title"
+
+DocumentTag:
+  description: A tag of a document can be any non-empty, UTF-8-encoded string which doesn't contain a zero byte.
+  type: string
+  pattern: '[^\x00]+$'
```
rustonaut, line 31:
```suggestion
  pattern: '^[^\x00]+$'
```

rustonaut (approved):
lgtm, there is one small typo issue which needs to be fixed before merging

web-api/openapi/schemas/id.yml
janpetschexain: mhh ok, if the constraints are intersected, than it's sufficient 👌

web-api/openapi/schemas/document.yml
janpetschexain: fixed. i also updated the regex in the rust code, for some reason i introduced another bug additionally to this one as well. now they are on par.

janpetschexain:
bors merge

PR #788, opened 2023-01-30
[Merged by Bors] - ET-3945 replace flattened config extension

janpetschexain:
**Reference**

- [ET-3945]

**Summary**

- replace the config extension of each application with distinct configs

[ET-3945]: https://xainag.atlassian.net/browse/ET-3945?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/web-api/src/ingestion.rs
+++ b/web-api/src/ingestion.rs
@@ -20,44 +20,60 @@ use serde::{Deserialize, Serialize};
 
 use crate::{
     embedding::{self, Embedder},
-    server::{self, Application},
-    storage::Storage,
+    logging,
+    server::{self, Application, NetConfig},
+    storage::{self, Storage},
 };
 
 pub struct Ingestion;
 
 impl Application for Ingestion {
+    type Config = Config;
     type AppStateExtension = AppStateExtension;
-    type ConfigExtension = ConfigExtension;
 
     fn configure_service(config: &mut ServiceConfig) {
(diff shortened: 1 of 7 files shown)


rustonaut (approved):
At some point there was the idea of having `AsRef<>` for all sub-configs so you would just derive the `AsRef` and then use `AsRef<logging::Config>`, `AsRef<net::Config>` etc.

But this doesn't work that grate with type inference and doing a `let net = config.as_ref()` wold conflict with the move below needing some manual drop or similar.

So I guess using a trait here is fine, too.

janpetschexain:
> At some point there was the idea of having `AsRef<>` for all sub-configs so you would just derive the `AsRef` and then use `AsRef<logging::Config>`, `AsRef<net::Config>` etc.
> 
> But this doesn't work that grate with type inference and doing a `let net = config.as_ref()` would conflict with the move below needing some manual drop or similar.
> 
> So I guess using a trait here is fine, too.

the `as_ref` approach also only works if the fields are known to the type, that means we can use it for a `Config<CE>` with struct `Config` and a generic extension `CE` but not for a generic config `C`, which is the reason why i introduced the `Config` trait.

janpetschexain:
bors merge

PR #791, opened 2023-01-30
[Merged by Bors] - replace trait config with asref config

janpetschexain:
**Reference**

- follow up from the comment https://github.com/xaynetwork/xayn_discovery_engine/pull/788#pullrequestreview-1275230959

**Summary**

- replace the `Config` trait bound with `AsRef<logging::Config>` trait bound etc
- use the available associated types of the `Application` trait to simplify the generics on `AppData`

### DIFF
--- a/web-api/src/server.rs
+++ b/web-api/src/server.rs
@@ -24,26 +24,34 @@ use actix_web::{
     HttpResponse,
     HttpServer,
 };
+use async_trait::async_trait;
 use clap::Parser;
 use serde::{de::DeserializeOwned, Serialize};
 use tracing::error;
 
-pub use self::config::Config;
-pub(crate) use self::{
-    app_state::AppState,
-    config::{impl_config, NetConfig},
-};
+pub(crate) use self::{app_state::AppState, config::Config};
 use crate::{
     load_config::load_config,
+    logging,
     logging::init_tracing,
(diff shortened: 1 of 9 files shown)


web-api/src/server.rs
```diff
@@ -53,21 +61,21 @@ pub trait Application {
...
-    fn create_app_state_extension(
-        config: &Self::Config,
-    ) -> Result<Self::AppStateExtension, SetupError>;
+    fn create_extension(config: &Self::Config) -> Result<Self::Extension, ApplicationError>;
+
+    async fn setup_storage(config: &storage::Config) -> Result<Self::Storage, ApplicationError>;
 }
 
-pub type SetupError = anyhow::Error;
+pub(crate) type ApplicationError = anyhow::Error;
```
rustonaut, line 71: I'm not sure why you changed the name.

It's not an application error (as in it happens during the runtime of the application), it's an error which basically only can happen during startup (and in very very unlikely cases shutdown).

janpetschexain: i renamed it because it's mainly the error which is returned in the `Application` trait methods, which seems more fitting. i'll revert it though.

rustonaut (approved):
lgtm

janpetschexain:
bors merge

PR #792, opened 2023-01-30
[Merged by Bors] - ET-3882 impl semantic search

rustonaut:
Implement `/semantic_search/{document_id}` endpoint as specified in `front_office.yaml`.

**References:**

- ET-3882

### DIFF
--- a/web-api/src/personalization/routes.rs
+++ b/web-api/src/personalization/routes.rs
@@ -34,11 +34,11 @@ use xayn_ai_coi::{
     PositiveCoi,
 };
 
-use super::{AppState, PersonalizationConfig};
+use super::{AppState, PersonalizationConfig, SemanticSearchConfig};
 use crate::{
     error::{
         application::WithRequestIdExt,
-        common::{BadRequest, InternalError},
+        common::{BadRequest, DocumentNotFound, InternalError},
     },
     models::{DocumentId, DocumentProperties, PersonalizedDocument, UserId, UserInteractionType},
     storage::{self, KnnSearchParams},
@@ -46,7 +46,7 @@ use crate::{
 };
 
 pub(super) fn configure_service(config: &mut ServiceConfig) {
-    let scope = web::scope("/users/{user_id}")
+    let users = web::scope("/users/{user_id}")
         .service(
             web::resource("interactions")
                 .route(web::patch().to(interactions.error_with_request_id())),
@@ -56,7 +56,10 @@ pub(super) fn configure_service(config: &mut ServiceConfig) {
                 .route(web::get().to(personalized_documents.error_with_request_id())),
         );
 
-    config.service(scope);
+    let semantic_search = web::resource("/semantic_search/{document_id}")
+        .route(web::get().to(semantic_search.error_with_request_id()));
+
+    config.service(users).service(semantic_search);
 }
 
 /// Represents user interaction request body.
@@ -186,14 +189,7 @@ async fn personalized_documents(
     .map(|documents| {
         if let Some(documents) = documents {
             Either::Left(Json(PersonalizedDocumentsResponse {
-                documents: documents
-                    .into_iter()
-                    .map(|document| PersonalizedDocumentData {
-                        id: document.id,
-                        score: document.score,
-                        properties: document.properties,
-                    })
-                    .collect(),
+                documents: documents.into_iter().map(Into::into).collect(),
             }))
         } else {
             Either::Right((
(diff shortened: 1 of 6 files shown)


web-api/src/storage/memory.rs
```diff
@@ -283,6 +283,11 @@ impl storage::Document for Storage {
         Ok(documents)
     }
 
+    async fn get_embedding(&self, id: &DocumentId) -> Result<Option<NormalizedEmbedding>, Error> {
+        let documents = self.documents.read().await;
+        Ok(documents.1.borrow_map().get(id).cloned())
```
janpetschexain, line 288: :art: there is no need for an intermediate assignment
```suggestion
        Ok(self.documents.read().await.1.borrow_map().get(id).cloned())
```

web-api/src/personalization/routes.rs
```diff
@@ -195,6 +191,16 @@ struct PersonalizedDocumentData {
     properties: DocumentProperties,
 }
 
+impl From<PersonalizedDocument> for PersonalizedDocumentData {
+    fn from(value: PersonalizedDocument) -> Self {
+        Self {
+            id: value.id,
+            score: value.score,
+            properties: value.properties,
```
janpetschexain, line 199: :art: the argument name could be a bit less generic
```suggestion
    fn from(document: PersonalizedDocument) -> Self {
        Self {
            id: document.id,
            score: document.score,
            properties: document.properties,
```

web-api/src/personalization/routes.rs
```diff
@@ -169,14 +172,7 @@ async fn personalized_documents(
...
             Either::Left(Json(PersonalizedDocumentsResponse {
-                documents: documents
-                    .into_iter()
-                    .map(|document| PersonalizedDocumentData {
-                        id: document.id,
-                        score: document.score,
-                        properties: document.properties,
-                    })
-                    .collect(),
+                documents: documents.into_iter().map(Into::into).collect(),
```
janpetschexain, line 175: :art: `Itertools` provides a shorthand for that
```suggestion
                documents: documents.into_iter().map_into().collect(),
```

web-api/src/personalization/routes.rs
```diff
@@ -414,6 +420,59 @@ pub(crate) async fn personalize_documents_by(
...
+}
+
+impl SemanticSearchQuery {
+    fn document_count(&self, config: &SemanticSearchConfig) -> Result<usize, Error> {
+        let count = self.count.map_or(config.default_number_documents, |count| {
+            count.min(config.max_number_documents)
+        });
+
+        if count > 0 {
+            Ok(count)
```
janpetschexain, line 435: ❓ partly related to this since we use the same logic for the personalized documents `count`: why do we only error for `count < 1` but not for `count > max_number_documents`? shouldn't we either error on both or none of them? i'd prefer to not error at all and just clamp the `count`, clamping should be expectable regarding the spec.

as a side note, the spec states a fixed `100` as `max_number_documents`, but it is actually dependend on the config, so i guess we should also update the spec to reflect that. this also affects the max length in the `PersonalizedDocumentsResponse` & `SemanticSearchResponse`.

web-api/src/personalization/routes.rs
```diff
@@ -414,6 +420,59 @@ pub(crate) async fn personalize_documents_by(
...
+            embedding: &embedding,
+            k_neighbors: count,
+            num_candidates: count,
+            published_after: None,
+        },
+    )
+    .await?;
+
+    Ok(Json(SemanticSearchResponse {
+        documents: documents.into_iter().map(Into::into).collect(),
```
janpetschexain, line 472: 🎨 `Itertools` provides a shorthand for that
```suggestion
        documents: documents.into_iter().map_into().collect(),
```

web-api/src/storage/memory.rs
rustonaut: readability, but I can change it.

web-api/src/personalization/routes.rs
rustonaut: I'm not sure why we choose to go with that :thinking: 

I think the idea was something on the line of "we might always return less documents then requested and with config changes requesting more then max can always happen so no reason to error there but requesting 0 is always a mistake so lets error".

I would argue we should just return 0 documents if the user explicitly requests 0 documents :thinking:

rustonaut: Also I would make this a  :pushpin:  fixing it for both cases in a follow up PR.

web-api/src/personalization/routes.rs
rustonaut: I will change it but I'm not the biggest fan of it because `map(Into::into)` works on all monade like types but `map_into` only works on iterators.

rustonaut:
bors merge

PR #793, opened 2023-01-31
simplify container-wide derived impls

janpetschexain:
**Summary**

simplify some container-wide derived impls for code size reduction and readability:
- if a struct derives `Default` and `Deserialize` and annotates all fields with `#[serde(default)]`, then the whole struct can instead be annotated. this also removes the need for the single `default_NAME()` functions as the values can be declared directly in the `Default` impl
- if a struct derives `AsRef` and annotates all fields with `#[as_ref]`, then the annotation on the fields can just be skipped as the derived `AsRef` applies to all of them anyways (if the derived impls are not conflicting, which fails to compile with the attributes as well)


rustonaut (approved):
lgtm

janpetschexain:
bors merge

PR #795, opened 2023-01-31
[Merged by Bors] - ET-3957 optionally store user history

janpetschexain:
**Reference**

- [ET-3957]

**Summary**

- add `store_user_history` flag to the `PersonalizationConfig`, defaults to `true`
- `interactions` endpoint: only store the user history in the `interaction` table if the flag is set
- `personalized_documents` endpoint: only fetch the user history from the `interaction` table (for the excluded docs list) if the flag is set (otherwise no docs are excluded)

[ET-3957]: https://xainag.atlassian.net/browse/ET-3957?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/web-api/src/personalization/routes.rs
+++ b/web-api/src/personalization/routes.rs
@@ -123,6 +123,7 @@ pub(crate) async fn update_interactions(
         storage,
         user_id,
         &document_ids,
+        store_user_history,
         |context| {
             match document_id_to_interaction_type[&context.document.id] {
                 UserInteractionType::Positive => {
@@ -139,7 +140,6 @@ pub(crate) async fn update_interactions(
                 }
             }
         },
-        store_user_history,
     )
     .await?;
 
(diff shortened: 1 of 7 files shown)


web-api/src/personalization/routes.rs
```diff
@@ -111,19 +119,28 @@ pub(crate) async fn update_interactions(
...
+                        context.positive_cois,
+                        &context.document.embedding,
+                    )
                     .clone()
+                }
             }
-        }
-    })
+        },
+        store_user_history,
```
rustonaut, line 142: :art: Maybe we should move it before the closure, I'm not sure if it's just because of the diff, but it's quite easy to miss.

rustonaut (approved):
lgtm, maybe move the position of the `store_user_history` in the parameter list

janpetschexain:
bors merge

PR #796, opened 2023-01-31
[Merged by Bors] - Specify better what documents are returned

acrrd:


web-api/openapi/front_office.yaml
```diff
@@ -39,7 +39,8 @@ paths:
         Returns a list of documents personalized for the given `user_id`.
         Each document contains the id, the score and the properties that are attached to the document.
         The score is a value between 0 and 1 where a higher value means that the document matches the preferences of the user better.
-        Note that you can request personalized documents for a specific `user_id`, only after that same `user_id` has made at least one interaction via our system.
+        Documents that have been interacted with by the user are filtered out from the result.        
+        Note that you can request personalized documents for a specific `user_id`, only after that same `user_id` has made enough interaction via our system.
```
acrrd, line 43: Here I changed "one" with "enough". The number depends on the configuration of the system. Our default is one, but it is not necessarily one. For some customers, the value will be different.

janpetschexain, line 43: :art: typo
```suggestion
        Note that you can request personalized documents for a specific `user_id`, only after that same `user_id` has made enough interactions via our system.
```

janpetschexain (commented):
❓ don't we have to bump the version to `rc9` in both yaml files?

acrrd:
I would avoid that for these changes. We didn't change how the API behaves, but we write it down in the spec so everyone is aware.

web-api/openapi/front_office.yaml
janpetschexain: not sure if there is a push missing, but the typo is still there even after this was marked as resolved.

acrrd: I told Github to commit this from the interface. Apparently that didn't work

janpetschexain (approved):

acrrd:
bors merge

PR #797, opened 2023-01-31
[Merged by Bors] - backport refactorings and improvements from nc repo

rustonaut:
This PR back-ports various refactoring/improvements from the nc repo.

Here a list of changes in order they where committed to the branch:

1.  separated  config loading from the rest of the application
    - for now this is just about separation of concerns
    - long term it allows us to reuse it as library/crate, which is needed if we either have more projects or need to change/
    extend the nc project more then we expect
2. separated net/actix setup code from the rest of the application
    - same reason as config separation
    - not back-ported, but  due to differences it is really useful for later changes (`AppHandle`)
3. inline config
    - allow passing the config as cmdline arg by prefixing it with `inline:`
    - nice for testing with test specific confics without spreading test config/code across multiple files
5. spawn server and return `AppHandle`
    - spawn server so that we can continue with running tests
    - `AppHandle` allows stopping the server
    - `AppHandle` allows waiting for the server to stop and get the execution result
6. use `env!("CARGO_BIN_NAME")` for env prefix

A good part of the changes are code which was moved in this repo (like a 1/3 or more), or moved from the nc repo.


web-api/src/net.rs
```diff
@@ -0,0 +1,183 @@
+// Copyright 2022 Xayn AG
```
rustonaut, line 1: This file is basically [web-api/src/server/config.rs](https://github.com/xaynetwork/xayn_discovery_engine/pull/797/files#diff-47f02916c7b3ffff14ac94561d1f7071709ddcc9ca1c8a7be2e137a3e7f9b61b) with an additional `start_actix_server` function, separating most "net" related server setup (bind and tcp option setup, start/stopping listening, transport compression, etc.)

web-api/src/net.rs
```diff
@@ -0,0 +1,183 @@
...
+    let app_state = web::Data::new(app_state);
+
+    let server = HttpServer::new(move || {
+        App::new()
+            .app_data(app_state.clone())
+            .app_data(json_config.clone())
+            .service(web::resource("/health").route(web::get().to(HttpResponse::Ok)))
+            .configure(&configure_services)
+            .wrap_fn(wrap_non_json_errors)
+            .wrap_fn(tracing_log_request)
```
rustonaut, line 110: perfectly would be to not have this here because while most other things can still count as net related (with a bit of hand waving)  this clearly is not. But due to how the `App` type changes with wrappers it's just much easier to keep it here. At least as long as we don't want to make this a lib we re-use from other services.

web-api/src/config.rs
```diff
@@ -0,0 +1,80 @@
...
+use std::{ffi::OsString, fmt::Display, process::exit};
+
+use clap::{CommandFactory, Parser};
+use serde::{de::DeserializeOwned, Serialize};
+
+use self::{cli::Args, load_config::load_config};
+
+/// Loads the config with custom CLI args.
+///
+/// See [`Config.load()`].
```
janpetschexain, line 27: :wrench: this points to a non-existent item, i'm not sure why the ci doesn't complain about this 🤔 
```suggestion
/// See [`load()`].
```

web-api/src/net.rs
```diff
@@ -0,0 +1,183 @@
...
+        client_request_timeout,
+    } = app_state.as_ref();
+
+    let json_config = JsonConfig::default().limit(max_body_size);
+    let app_state = web::Data::new(app_state);
+
+    let server = HttpServer::new(move || {
+        App::new()
+            .app_data(app_state.clone())
+            .app_data(json_config.clone())
```
janpetschexain, line 106: :art: we could avoid to clone the whole `JsonConfig`
```suggestion
    let app_state = web::Data::new(app_state);
    let server = HttpServer::new(move || {
        App::new()
            .app_data(app_state.clone())
            .app_data(JsonConfig::default().limit(max_body_size))
```

web-api/src/config/load_config.rs
```diff
@@ -12,7 +12,7 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
```
janpetschexain, line 14: :art: since this module is now a submodule of `config` we could simplify the name to `load` such that we import from `crate::config::load::*` instead of `crate::config::load_config::*`.

rustonaut: I actually wanted to fold it into `config.rs` as the methods in there are just thin wrappers around this, but didn't do that so that github shows it as moved instead of removed+added. I will do so now.

rustonaut: maybe now rename the parent into `load_config.rs`?? Idk.

web-api/src/net.rs
rustonaut: this is now re-creating it every time instead of cloning it, which is "in generally" something you should avoid with things passed to `app_data` as such things (if correctly done) are made to be cheap to copy by using `Arc` internally.

In the case of  `JsonConfig` it happens to not really make a difference in the way we use it as all `Arc`s inside of `JsonConfig`  are `Option<Arc<>>` and `None` with the settings we have, still I would prefer using the "recommended" stile as it the chance someone who add a setting to `JsonConfig` which involves an `Arc` would be prone to not reverting the proposed change and use `clone()` (at least if it's not one of us).

janpetschexain (commented):
it looks like you forgot to push some of the changes you commented on

janpetschexain (approved):
lgtm

rustonaut:
bors merge

PR #799, opened 2023-02-01
[Merged by Bors] - ET-3883 semantic search threshold

rustonaut:
Only consider candidates with a given minimal similarity for semantic search.

**References:**

- bases in #792 
- issue: ET-3883
- story: ET-3877

### DIFF
--- a/web-api/openapi/front_office.yaml
+++ b/web-api/openapi/front_office.yaml
@@ -1,7 +1,7 @@
 openapi: 3.0.3
 info:
   title: Front Office API
-  version: 1.0.0-rc8
+  version: 1.0.0-rc9
   description: |-
     # Front Office
     The front office is typically used within front-end apps, for example a website or a mobile application.
--- a/web-api/openapi/front_office.yaml
+++ b/web-api/openapi/front_office.yaml
@@ -129,6 +129,16 @@ paths:
             minimum: 1
             maximum: 100
             default: 10
+        - name: min_similarity
+          in: query
+          description: Minimum similarity a document has to have to be included.
+          required: false
+          schema:
+            type: number
+            format: float
+            minimum: 0
+            maximum: 1
+            default: 0
       responses:
         '200':
           description: successful operation
--- a/web-api/src/personalization/routes.rs
+++ b/web-api/src/personalization/routes.rs
@@ -307,6 +307,7 @@ async fn search_knn_documents(
                     k_neighbors,
                     num_candidates: count,
                     published_after,
+                    min_similarity: None,
                 },
             )
             .await
@@ -447,6 +448,7 @@ pub(crate) async fn personalize_documents_by(
 #[derive(Deserialize)]
 struct SemanticSearchQuery {
     count: Option<usize>,
+    min_similarity: Option<f32>,
 }
 
 impl SemanticSearchQuery {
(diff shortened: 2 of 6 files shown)


web-api/openapi/front_office.yaml
```diff
@@ -129,6 +129,16 @@ paths:
...
             maximum: 100
             default: 10
+        - name: min_similarity
+          in: query
+          description: Minimum similarity a document has to have to be included.
+          required: false
+          schema:
+            type: number
+            format: float
+            minimum: 0
```
janpetschexain, line 139: :wrench: we should set the minimum to `-1`. if someone provides this parameter but still wants to include all documents, then that wouldn't be possible with minimum `0`. it's likely not a common use case, but there is no reason to restrict this artificially.

web-api/src/personalization/routes.rs
```diff
@@ -461,6 +463,10 @@ impl SemanticSearchQuery {
             Err(BadRequest::from("count has to be at least 1").into())
         }
     }
+
+    fn min_similarity(&self) -> Option<f32> {
+        self.min_similarity.map(|value| value.clamp(0., 1.))
```
janpetschexain, line 468: :wrench: see the other comment
```suggestion
        self.min_similarity.map(|value| value.clamp(-1., 1.))
```

janpetschexain:
also, since this changes the api, the version in both yaml files need to be bumped.

web-api/openapi/front_office.yaml
rustonaut: AFIK scores of ES KNN search are between 0 and 1 so 0 should always include all values :thinking:

rustonaut:
:pushpin:  

> the version in both yaml files need to be bumped.

If we always want to keep the API version in sync could we `$ref` that to a `version.yaml` file? Or will tooling brake?

web-api/openapi/front_office.yaml
janpetschexain: ah i see, then there is a misunderstanding, the elastic score is indeed in `[0, 1]`. but since you named the parameter `min_similarity`, i automatically assumed that you are refererring to cosine similarity (i could have checked better the code using the param). to avoid such confusion also for others, i guess it's best to choose a different name for the param to better reflect its purpose. `min_score` would be an option, maybe also sth else.

janpetschexain:
> 📌
> 
> > the version in both yaml files need to be bumped.
> 
> If we always want to keep the API version in sync could we `$ref` that to a `version.yaml` file? Or will tooling brake?

i don't know if this breaks tooling, but it would indeed be better to specifiy the version only once.

web-api/openapi/front_office.yaml
rustonaut: The problem with `min_score` is that it's not the min score of `PersonalizedDocument.score` but a threshold applied to the similarity search only.  And as far as I can tell the score of ES is a similarity of some form, probably a `[-1;1]` similarity normalized to `[0;1]`. 

Also the mathematical properties of cosin similarity and similar should not be aspects our customers should worry about. The customer just knows there is some similarity search followed by some personalization and the threshold is on the similarity search.

rustonaut: Even for the scoring of the reranker we do there are reasons why we might want to transform it from `[-1;1]` to `[0;1]` depending on how we might want to handle certain things like negative reactions and differentiating between very dissimilar to a CoI (doesn't mean the user dislikes it, could be returned if we have nothing better) and disliked by a user (should mostly not be returned).

janpetschexain: normalization of scores is indeed a good point. so i guess we should leave this for now and handle it in a follow up 👌

rustonaut:
bors merge

PR #800, opened 2023-02-01
[Merged by Bors] - ET-3884 personalization of semantic search

rustonaut:
After finding similar documents (during which `min_similarity` is applied) does an additional personalization step.

Following should be done:

- [x] run similarity search normally (use `min_similarity` there) to get documents and get rank by similarity
- [x] code for resolving/parsing new query parameters
- [x]  get tag weights for documents and rank by tag weight 
- [x]  get cois and rank by interest
- [x] combine all ranks with a weighting
      - the ratio between similarity and interest+tag is `personalize_ratio`
      - the ratio between interest and tag is configured with the same config as it is for normal document personalization
- [x] set a new score based on the combined ranks 
- [ ]  ???~only apply the personlization on the top X found similar documents~???
- [x] update the openapi spec
     - [x] `personalize_for`
     - [x] `personalize_ratio`
     - [ ] ???~`personalization_count`~???
- [ ] decide what to do in case of not enough user interaction (fail or just as if `ratio == 0`?)
- [x] potentially update to combine scores instead of ranks

---

**References:**

- based on: #799 , #792 , #810
- issue: ET-3884
- story: ET-3877

### DIFF
--- a/web-api/src/personalization/rerank.rs
+++ b/web-api/src/personalization/rerank.rs
@@ -15,10 +15,28 @@
 use std::{cmp::Ordering, collections::HashMap, hash::Hash};
 
 use chrono::{DateTime, Utc};
-use itertools::{izip, Itertools};
+use itertools::Itertools;
 use xayn_ai_coi::{nan_safe_f32_cmp, nan_safe_f32_cmp_desc, CoiSystem, UserInterests};
 
 use crate::models::{DocumentId, DocumentTag, PersonalizedDocument};
+
+fn rank_keys_by_score<K, S>(
+    keys_with_score: impl IntoIterator<Item = (K, S)>,
+    mut sort_by: impl FnMut(&S, &S) -> Ordering,
+) -> HashMap<K, f32>
+where
+    K: Eq + Hash,
+{
+    keys_with_score
+        .into_iter()
+        .sorted_unstable_by(|(_, s1), (_, s2)| sort_by(s1, s2))
+        .enumerate()
+        .map(
+            #[allow(clippy::cast_precision_loss)] // index is small enough
+            |(index, (key, _))| (key, 1. / (1 + index) as f32),
+        )
+        .collect()
+}
 
 pub(super) fn rerank_by_interest(
     coi_system: &CoiSystem,
@@ -28,7 +46,7 @@ pub(super) fn rerank_by_interest(
 ) -> HashMap<DocumentId, f32> {
     let scores = coi_system.score(documents, interests, time);
     rank_keys_by_score(
-        izip!(documents.iter().map(|doc| doc.id.clone()), scores),
+        documents.iter().map(|doc| doc.id.clone()).zip(scores),
         nan_safe_f32_cmp_desc,
     )
 }
@@ -55,7 +73,7 @@ pub(super) fn rerank_by_tag_weight(
 /// The `score_weights` determine the ratios of the scores, it is ordered as
 /// `[interest_weight, tag_weight, elasticsearch_weight]`. The final score/ranking per document is
 /// calculated as the weighted sum of the scores.
-pub(super) fn rerank_by_score_interest_and_tag_weight(
+pub(super) fn rerank_by_scores(
     coi_system: &CoiSystem,
     documents: &mut [PersonalizedDocument],
     interests: &UserInterests,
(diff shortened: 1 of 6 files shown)


web-api/src/personalization/routes.rs
```diff
@@ -532,7 +555,29 @@ async fn semantic_search(
...
+    if let Some(personalize) = personalize {
+        let interest = storage::Interest::get(&state.storage, &personalize.for_user).await?;
+        let tag_weights = storage::Tag::get(&state.storage, &personalize.for_user).await?;
+
+        if !interest.has_enough(state.config.as_ref()) {
+            return Ok(Either::Right((
+                Json(PersonalizedDocumentsError::NotEnoughInteractions),
+                StatusCode::CONFLICT,
+            )));
+        }
```
rustonaut, line 567: :question: : It is not clear if this is the behavior we want or if we simple want to not personalize if there are not enough interactions.

web-api/src/personalization/rerank.rs
```diff
@@ -81,8 +95,11 @@ pub(super) fn rerank_by_interest_and_tag_weight(
...
     for document in documents.iter_mut() {
-        document.score = interest_ranks[&document.id]
+        let personalization_score = interest_ranks[&document.id]
             .merge_as_score(tag_weight_ranks[&document.id], interest_tag_bias);
+
+        document.score = personalization_ratio * personalization_score
+            + (1. - personalization_ratio) * document.score;
     }
 
     let secondary_sorting_factor = if interest_tag_bias >= 0.5 {
```
rustonaut, line 105: :question:  I just keep it the way it is, but it's not clear if we might want to change this depending on `persoanlization_bias`.

E.g. if personalization_bias < 0.5 sort by `(score, original_score)` and if it's `>= 0.5` sort by `(score, secondary_sorting_factor)` or similar.

web-api/src/personalization/rerank.rs
```diff
@@ -69,9 +69,23 @@ pub(super) fn rerank_by_tag_weights(
...
+///
+/// The `interest_tag_bias` determines the ratio of interest scoring to tag weight scoring.
+///
+/// The `personalization_ratio` determines the ratio of the current score to the combined score of
+/// interest and tag weight.
+///
+/// So the final score/ranking per document is calculated as:
+///
+/// 1. `personalization_score = interest_tag_bias * interest_score + (1-interest_tag_bias) * tag_score`
+/// 2. `personalization_ratio * personalization_score + (1-personalization_ratio) *  current_score`
```
janpetschexain, line 84: 🔧 an array of weights is more readable and it also doesn't hide quadratic influences like the current weight formula has, ie we would have a weight array `[interest_weight, tag_weight, elasticsearch_weight]` and the final score would be computed as a weighted sum `interest_weight * interest_score + tag_weight * tag_score + elasticsearch_weight * elasticsearch_score` (renamed `current_` to `elasticsearch_`).

i just discussed with @acrrd regarding the default config and we concluded that we need two different weight arrays for the `semantic_search` and the `personalized_documents` endpoints, we decided for `[0.4, 0.4, 0.2]` and `[0.5, 0.5, 0.0]` as default weights.

web-api/src/personalization/rerank.rs
janpetschexain: with the array of weights in mind, we can break ties by selecting the scores map related to the highest weight in the array.

web-api/src/personalization/routes.rs
janpetschexain: imo it's ok to fail in that case, i will ask in the daily about opinions on this.

janpetschexain (commented):
:wrench: the update for the spec is missing

web-api/src/personalization/rerank.rs
```diff
@@ -15,47 +15,46 @@
...
-use itertools::{izip, Itertools};
-use xayn_ai_coi::{nan_safe_f32_cmp_desc, CoiSystem, UserInterests};
+use itertools::Itertools;
+use xayn_ai_coi::{nan_safe_f32_cmp, nan_safe_f32_cmp_desc, CoiSystem, UserInterests};
 
 use crate::models::{DocumentId, DocumentTag, PersonalizedDocument};
 
-#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
-#[repr(transparent)]
-pub(super) struct Rank(usize);
```
janpetschexain, line 25: this wrapper doesn't work well with the elasticsearch scores, since we would have to resort those to get an index just to get the score again right afterwards, and the elasticsearch score is also computed differently than the interest and tag scores. hence i just moved the `to_score()` logic into `rank_keys_by_score()`. the `merge_as_score()` logic is extended to more than two inputs and moved into `rerank_by_scores()`.

acrrd (approved):

janpetschexain:
bors merge

PR #802, opened 2023-02-01
[Merged by Bors] - Don't special case an interest_tag_bias of exactly 0.5.

rustonaut:


janpetschexain (changes requested):
the arbitrary ordering in case that of bias 0.5 and equal weights was a requirement by pink, please check with @PanicButtonPressed first if it's ok to change that.

rustonaut:
With 0.5 The ordering of "equal" scored documents depends on a combination of the previous order, `sort_unstable` internals and the position of other non-equal documents before sorting, which means it currently implicitly already depends on the interest scoring just in a messy unstable arbitrary way, but this might also arbitrary change with valid reasonable code changes. So I really don't think this is any intended algorithmic part.

PanicButtonPressed:
Wasn't any strong requirement anyway. Fine with changing that. The case should not happen frequently

PanicButtonPressed (approved):

rustonaut:
bors merge

janpetschexain (approved):

rustonaut:
bors merge

PR #803, opened 2023-02-06
[Merged by Bors] - ET-3960 integration tests for store_user_history

janpetschexain:
**Reference**

- [ET-3960]

**Summary**

- add integration tests with and without `store_user_history`
- fix test crates naming: renamed `integration-tests` to `xayn-integration-tests` and `xayn-ai-test-utils` to `xayn-test-utils`
- simplify `set_config_option!` usage: there is no need to maintain this additional code, in cases where we create and edit a new config like `let mut config = Table::default(); set_config_option! { for config => ...}` we can simplify this to `let mut config = toml! { ... };`, and in cases where we modify an existing config like `|config| set_config_option! { for config => ... }` we can use `|config| config.extend(toml! { ... })`, otherwise we also need to take care of type export/import which are used in the macro but not available when the macro is called in another crate, additionally we can know use valid toml syntax ie no trailing semicolons

[ET-3960]: https://xainag.atlassian.net/browse/ET-3960?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/web-api/tests/store_user_history.rs
+++ b/web-api/tests/store_user_history.rs
@@ -18,7 +18,7 @@ use reqwest::StatusCode;
 use serde::Deserialize;
 use serde_json::json;
 use toml::toml;
-use xayn_integration_tests::{send_assert, send_assert_json, test_two_apps};
+use xayn_integration_tests::{extend_config, send_assert, send_assert_json, test_two_apps};
 use xayn_web_api::{Ingestion, Personalization};
 
 #[derive(Deserialize)]
@@ -35,10 +35,13 @@ async fn store_user_history(enabled: bool) {
     test_two_apps::<Ingestion, Personalization, _>(
         |_| {},
         |config| {
-            config.extend(toml! {
-                [personalization]
-                store_user_history = enabled
-            });
+            extend_config(
+                config,
+                toml! {
+                    [personalization]
+                    store_user_history = enabled
+                },
+            );
         },
         |client, ingestion, personalization, _| async move {
             send_assert(
--- a/web-api/tests/store_user_history.rs
+++ b/web-api/tests/store_user_history.rs
@@ -1,0 +1,109 @@
+// Copyright 2023 Xayn AG
+//
+// This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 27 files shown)


rustonaut (approved):
lgtm, 

there is one major difference between `test_app` here and in #804: In #804 the function makes sure any executions of `test_app` are mostly isolated from each other (except for globals and the bit of env which is not cleared). So things would need to be structured slightly different (having the test app wrapping `ingest().await; personalize(bool).await;` instead of each call using `test_app`).

janpetschexain:
> lgtm,
> 
> there is one major difference between `test_app` here and in #804: In #804 the function makes sure any executions of `test_app` are mostly isolated from each other (except for globals and the bit of env which is not cleared). So things would need to be structured slightly different (having the test app wrapping `ingest().await; personalize(bool).await;` instead of each call using `test_app`).

i'll rebase this pr after #804 is merged :ok_hand:

rustonaut (approved):
lgtm

web-api/tests/store_user_history.rs
```diff
@@ -0,0 +1,110 @@
...
+}
+
+async fn store_user_history(enabled: bool) {
+    test_two_apps::<Ingestion, Personalization, _>(
+        |_| {},
+        |config| {
+            config.extend(toml! {
+                [personalization]
+                store_user_history = enabled
+            });
```
rustonaut, line 41: While this works in this case it is bug prone compared to the `set_config_option` macro.

:wrench:  The difference is that `extend` is based on the `Map` type while the `set_config_option` macro is based on the full path per option.

So if you e.g. `extend`  with a hypothetical  `[storage.postgres] max_connections=2` then it will replace the _all_ storage entries with just the `max_connections` setting including the url.

Through this is unrelated to the CI failure.

janpetschexain: i didn't expect that, i'll fix it before merging.

janpetschexain:
bors merge

PR #806, opened 2023-02-07
[Merged by Bors] - ET-4012 stateless personalization openapi

rustonaut:
- [x] new endpoint for "stateless" personalization
- [x] accept history as required parameter
  - [x] seq of documents by id (ordered by oldest to newest, i.e. timestamp ms asc)
  - [x] timestamp is optional if not given it's inferred from neighbouring "newer" documents with timestamp and fallback `now`
- [x] accept count as opt. query parameter

**References:**

- ~based on: #804~ (moved out of the PR stack)
- task: [ET-4012]
- story: [ET-4010]

[ET-4012]: https://xainag.atlassian.net/browse/ET-4012?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/web-api/openapi/front_office.yaml
+++ b/web-api/openapi/front_office.yaml
@@ -200,6 +200,12 @@ components:
           type: number
         properties:
           $ref: './schemas/document.yml#/DocumentProperties'
+    PersonalizedDocuments:
+      type: array
+      minItems: 0
+      maxItems: 100
+      items:
+        $ref: '#/components/schemas/PersonalizedDocumentData'
     StatelessPersonalizedDocumentsRequest:
       type: object
       required: [history]
@@ -217,15 +223,11 @@ components:
         warnings:
           type: array
           minItems: 0
-          maxItems: 100
+          maxItems: 200
           items:
             type: string
         documents:
-          type: array
-          minItems: 0
-          maxItems: 100
-          items:
-            $ref: '#/components/schemas/PersonalizedDocumentData'
+          $ref: '#/components/schemas/PersonalizedDocuments'
       example:
         warnings:
           - "historic document with id \"document_id1\" does not exist"
@@ -239,11 +241,7 @@ components:
       required: [documents]
       properties:
         documents:
-          type: array
-          minItems: 0
-          maxItems: 100
-          items:
-            $ref: '#/components/schemas/PersonalizedDocumentData'
+          $ref: '#/components/schemas/PersonalizedDocuments'
       example:
         documents:
           - id: 'document_id0'
@@ -255,11 +253,7 @@ components:
       required: [documents]
       properties:
         documents:
-          type: array
-          minItems: 0
-          maxItems: 100
-          items:
-            $ref: '#/components/schemas/PersonalizedDocumentData'
+          $ref: '#/components/schemas/PersonalizedDocuments'
       example:
         documents:
           - id: 'document_id0'


rustonaut:
spectral fails in without reporting an error in a line which doesn't seem to have an error :sweat:

janpetschexain:
> spectral fails in without reporting an error in a line which doesn't seem to have an error 😓

the non-reported error seem to be caused by the batch processing, if we change the `just` formula to
```
validate-openapi:
    spectral lint --verbose -F warn web-api/openapi/front_office.yaml; \
    spectral lint --verbose -F warn web-api/openapi/front_office.yaml
```
then the error is properly reported as
```
xayn_discovery_engine/web-api/openapi/schemas/document.yml
  36:8  error  oas3-schema          "type" property type must be string.                   History.type
 37:13  error  property-attributes  minItems should not be defined for a non-array schema  History.minItems
 38:13  error  property-attributes  maxItems should not be defined for a non-array schema  History.maxItems
```

justfile
```diff
@@ -203,7 +203,12 @@ install-openapi-validator:
       validator@${VALIDATOR_VERSION}
 
 validate-openapi:
-    spectral lint --verbose -F warn web-api/openapi/*.yaml
+    #!/usr/bin/env -S bash -eux -o pipefail
+    # We need to call it once per file, if we pass in multiple files it will
+    # has some bug where it does not report error correctly.
```
janpetschexain, line 208: :art: typo
```suggestion
    # have some bug where it does not report error correctly.
```

web-api/openapi/front_office.yaml
```diff
@@ -30,6 +30,43 @@ security:
...
+              schema:
+                $ref: '#/components/schemas/StatelessPersonalizedDocumentsResponse'
+        '400':
+           $ref: './responses/generic.yml#/BadRequest'
+        '409':
+          description: -|
+            impossible to create a personalized list for given context
+
+            This mainly can happen if all documents in the user history
+            do no longer exist (or where never ingested).
```
janpetschexain, line 64: :art: typo
```suggestion
            do no longer exist (or were never ingested).
```

web-api/openapi/front_office.yaml
```diff
@@ -165,6 +200,40 @@ components:
...
+    StatelessPersonalizedDocumentsRequest:
+      type: object
+      required: [history]
+      properties:
+        history:
+          $ref: './schemas/document.yml#/History'
+        count:
+          $ref: '#/components/schemas/Count'
+        published_after:
+          $ref: './schemas/time.yml#/PublicationDate'
```
janpetschexain, line 212: ❓ why do we take a different approach for this request and pass the `count` and `published_after` in the body instead of in the parameter list like we do for other similar endpoints? wouldn't it be better for a user to always have the same way of dealing with a specific parameter?

web-api/openapi/front_office.yaml
```diff
@@ -165,6 +200,40 @@ components:
...
+          minItems: 0
+          maxItems: 100
+          items:
+            type: string
+        documents:
+          type: array
+          minItems: 0
+          maxItems: 100
+          items:
+            $ref: '#/components/schemas/PersonalizedDocumentData'
```
janpetschexain, line 228: :art: we could define a schema for this array as well as it is used in multiple places.

web-api/openapi/front_office.yaml
```diff
@@ -165,6 +200,40 @@ components:
...
+      example:
+        warnings:
+          - "historic document with id \"document-id-1\" does not exist"
```
janpetschexain, line 231: :art: maybe like this to align it with the doc id example two lines below
```suggestion
          - "historic document with id \"document_id1\" does not exist"
```

janpetschexain (approved):
lgtm

PR #809, opened 2023-02-08
[Merged by Bors] - ET-4015 pre impl refactoring

rustonaut:
Factors out/refactors some parts which will be reused by `/personalized_documents` and personalized `/semantic_search`.

**References:**

- based on #806
- task: [ET-4015]
- story: [ET-4010]

[ET-4015]: https://xainag.atlassian.net/browse/ET-4015?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/coi/src/stats.rs
+++ b/coi/src/stats.rs
@@ -122,8 +122,12 @@ pub fn compute_coi_decay_factor(
 }
 
 /// Computes a weight distributions across cois based on their relevance.
-pub fn compute_coi_weights(cois: &[PositiveCoi], horizon: Duration) -> Vec<f32> {
-    let relevances = compute_coi_relevances(cois, horizon, system_time_now())
+pub fn compute_coi_weights(
+    cois: &[PositiveCoi],
+    horizon: Duration,
+    time: DateTime<Utc>,
+) -> Vec<f32> {
+    let relevances = compute_coi_relevances(cois, horizon, time)
         .into_iter()
         .map(|rel| 1.0 - (-3.0 * rel).exp())
         .collect_vec();
--- a/coi/src/stats.rs
+++ b/coi/src/stats.rs
@@ -15,6 +15,7 @@
 use std::time::Duration;
 
 use chrono::{DateTime, Utc};
+use itertools::Itertools;
 use serde::{Deserialize, Serialize};
 
 use crate::{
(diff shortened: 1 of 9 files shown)


coi/src/stats.rs
```diff
@@ -123,6 +124,31 @@ pub fn compute_coi_decay_factor(horizon: Duration, now: SystemTime, last_view: S
     ((horizon - days) / (horizon - 1. - f32::EPSILON)).max(0.)
 }
 
+/// Computes a weight distributions across cois based on their relevance.
+pub fn compute_coi_weights(cois: &[PositiveCoi], horizon: Duration) -> Vec<f32> {
```
rustonaut, line 128: moved here as it's kinda the same kind of function/deps situations as `compute_coi_relevance` (which is just above this function)

web-api/src/personalization/rerank.rs
```diff
@@ -0,0 +1,114 @@
...
+
+use crate::models::{DocumentId, DocumentTag, PersonalizedDocument};
+
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
+#[repr(transparent)]
+pub(super) struct Rank(usize);
+
+impl Rank {
+    pub(super) fn from_index(index: usize) -> Rank {
+        Rank(index + 1)
```
rustonaut, line 28: Informal feedback from pink was that ranks starting with `0` was an misunderstanding.

web-api/src/personalization/rerank.rs
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    pub(super) fn to_score(self) -> f32 {
+        #![allow(clippy::cast_precision_loss)]
+        1. / self.0 as f32
```
rustonaut, line 33: Arbitrary but when merging rankings we want to make sure that the score we return still has the characteristics of an score i.e.:

- bigger is better
- it's in `[0;1]`

This doesn't affect the outcome of the algorithm wrt. the resulting order. We just use it to compute the combined score which AFIK is currently not defined by pink. When it will be defined we will change it. But we need something now.

web-api/src/personalization/rerank.rs
```diff
@@ -0,0 +1,116 @@
...
+    documents.sort_unstable_by(|a, b| {
+        nan_safe_f32_cmp_desc(&a.score, &b.score).then_with(|| {
+            secondary_sorting_factor
+                .get(&a.id)
+                .cmp(&secondary_sorting_factor.get(&b.id))
+        })
+    });
+}
+
+fn pairs_to_rank_map<K, S>(
```
janpetschexain, line 103: :art: maybe sth like this?
```suggestion
fn rank_keys_by_scores<K, S>(
```

web-api/src/personalization/personalized_knn.rs
```diff
@@ -0,0 +1,161 @@
...
+    ) -> Result<Vec<PersonalizedDocument>, Error> {
+        let Search {
+            horizon,
+            max_cois,
+            count,
+            published_after,
+            interests,
+            excluded,
+            time,
+        } = self;
```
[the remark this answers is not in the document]
rustonaut: because then I didn't need to "touch" the other code and have less changes, well except because it was also moved into it's own file and the indention level changed it's not making a difference :see_no_evil:

web-api/src/personalization/personalized_knn.rs
```diff
@@ -0,0 +1,161 @@
...
+                }
+            }
+            Err(error) => {
+                error!("Error fetching documents: {error}");
+                errors.push(error);
+            }
+        };
+    }
+    (all_documents, errors)
+}
```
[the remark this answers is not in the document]
rustonaut: I just moved the code,

 but I would argue not using `fold` is more readable in this case as we don't benefit in any way from the folding as we anyway still modify both collections.
 
 I guess that falls again under the use native language constructs vs. functional combinators discussion.

janpetschexain (approved):
lgtm

rustonaut:
bors merge

PR #810, opened 2023-02-08
[Merged by Bors] - ET-4013/ET-4015 stateless personalization impl

rustonaut:
Implement the (user-)stateless `/personalized_documents` endpoint.

**References:**

- based on #809
- task: [ET-4015]
- story: [ET-4010]

[ET-4015]: https://xainag.atlassian.net/browse/ET-4015?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/web-api/src/personalization/routes.rs
+++ b/web-api/src/personalization/routes.rs
@@ -24,7 +24,8 @@ use actix_web::{
 use chrono::{DateTime, Utc};
 use itertools::Itertools;
 use serde::{Deserialize, Serialize};
-use xayn_ai_coi::CoiSystem;
+use tracing::{instrument, warn};
+use xayn_ai_coi::{CoiSystem, UserInterests};
 
 use super::{
     knn,
@@ -239,16 +240,16 @@ async fn stateless_personalize_documents(
         }
     }
 
-    let mut interests = Vec::new();
+    let mut interests = UserInterests::default();
     for embedding in embeddings.values() {
         /* TODO put in timestamp */
         state
             .coi
-            .log_positive_user_reaction(&mut interests, embedding);
+            .log_positive_user_reaction(&mut interests.positive, embedding);
     }
 
-    let documents = personalized_knn::Search {
-        interests: &interests,
+    let mut documents = personalized_knn::Search {
+        interests: &interests.positive,
         excluded: &[],
         horizon: state.coi.config().horizon(),
         max_cois: state.config.personalization.max_cois_for_knn,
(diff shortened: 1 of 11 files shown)


web-api/src/personalization/routes.rs
```diff
@@ -151,6 +167,154 @@ pub(crate) async fn update_interactions(
...
+impl StatelessPersonalizationRequest {
+    fn resolved_history(
+        self,
+        config: &PersonalizationConfig,
+        warnings: &mut Vec<Warning>,
+    ) -> Result<Vec<HistoryEntry>, Error> {
+        let max_history_len = config.max_stateless_history_size;
+        if self.history.len() > max_history_len {
+            warnings.push(format!("history truncated, max length is {max_history_len}").into());
+        }
```
janpetschexain, line 188: ❓ since we require in the spec that `history` is a non-empty array, don't we also have to check here that `!self.history.is_empty()`?

web-api/src/personalization/routes.rs
```diff
@@ -151,6 +167,154 @@ pub(crate) async fn update_interactions(
...
+    Json(request): Json<StatelessPersonalizationRequest>,
+) -> Result<impl Responder, Error> {
+    let mut warnings = vec![];
+    let published_after = request.published_after;
+    let count = request.document_count(state.config.as_ref())?;
+    let history = request.resolved_history(state.config.as_ref(), &mut warnings)?;
+    let ids = history.iter().map(|document| &document.id).collect_vec();
+    let time = history
+        .last()
+        .map_or_else(Utc::now, |entry| entry.timestamp);
```
janpetschexain, line 248: :art: since we need to guarantee that `history` is not empty in `resolved_history()`, we can also unwrap here
```suggestion
    let time = history.last().copied().unwrap(/* history has been checked */);
```

web-api/src/personalization/routes.rs
```diff
@@ -151,6 +167,154 @@ pub(crate) async fn update_interactions(
...
+struct StatelessPersonalizationRequest {
+    #[serde(default)]
+    count: Option<usize>,
+    #[serde(default)]
+    published_after: Option<DateTime<Utc>>,
+    history: Vec<UncheckedHistoryEntry>,
+}
+
+impl StatelessPersonalizationRequest {
+    fn resolved_history(
```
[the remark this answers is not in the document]
rustonaut: hm, I can change it but I intentionally didn't went with `check` because it does more then just checking.

web-api/src/personalization/routes.rs
rustonaut: yes, right there was one open question I had wrt. that:

- should it return a normal bad request or a not enough interactions error?

I guess a normal bad request?

web-api/src/personalization/routes.rs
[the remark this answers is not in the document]
rustonaut: it's also not just checking it but converting it, lets just call it `history ` I guess

web-api/src/personalization/routes.rs
janpetschexain: i'd say so too, if it doesn't conform to the spec, it's a bad request.

janpetschexain (approved):

janpetschexain:
bors merge

PR #811, opened 2023-02-08
[Merged by Bors] - ET-4011 allow to select interaction time

janpetschexain:
**Reference**

- [ET-4011]
- [ET-3952]

**Summary**

- replace `SystemTime` with `DateTime<Utc>`: `Utc::now()` anyway internally uses `SystemTime::now()`, but `DateTime` is also easily constructable for fixed values in contrast to `SystemTime`
- use `DateTime<Utc>` instead of `NaiveDateTime` for the in-memory storage to match it with the elastic/postgres storage
- expose the `time` argument in the internal apis up to the logic of the web-api routes, but not in the routes themselves, the routes & tests remain unchanged as they call the apis with `Utc::now()` for the time argument

[ET-4011]: https://xainag.atlassian.net/browse/ET-4011?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ET-3952]: https://xainag.atlassian.net/browse/ET-3952?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/coi/benches/benches.rs
+++ b/coi/benches/benches.rs
@@ -12,8 +12,9 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-use std::time::{Duration, SystemTime};
+use std::time::Duration;
 
+use chrono::Utc;
 use criterion::{black_box, criterion_group, BatchSize, Criterion};
 use itertools::Itertools;
 use rand::Rng;
@@ -37,24 +38,20 @@ fn create_positive_coi(n: usize, embedding_size: usize) -> Vec<PositiveCoi> {
 }
 
 fn bench_compute_coi_decay_factor(c: &mut Criterion) {
-    let horizon = black_box(Duration::new(3600 * 24 * 30, 0)); // 30 days
-    let now = black_box(SystemTime::now());
-    let last_view = black_box(
-        SystemTime::now()
-            .checked_sub(Duration::new(3600, 0))
-            .unwrap(),
-    );
+    let horizon = Duration::new(60 * 60 * 24 * 30, 0); // 30 days
+    let now = Utc::now();
+    let last_view = now - chrono::Duration::seconds(60 * 60);
 
     c.bench_function("compute_coi_decay_factor", |b| {
-        b.iter(|| compute_coi_decay_factor(horizon, now, last_view))
+        b.iter(|| black_box(compute_coi_decay_factor(horizon, now, last_view)))
     });
 }
 
 fn bench_compute_coi_relevance(c: &mut Criterion) {
     let count = [100, 500, 2000];
     let embedding_size = 128;
-    let horizon = black_box(Duration::new(3600 * 24 * 30, 0)); // 30 days
-    let now = black_box(SystemTime::now());
+    let horizon = Duration::new(60 * 60 * 24 * 30, 0); // 30 days
+    let now = Utc::now();
 
     let count_max: usize = *count.iter().max().unwrap();
     let positive_cois = create_positive_coi(count_max, embedding_size);
@@ -67,7 +64,7 @@ fn bench_compute_coi_relevance(c: &mut Criterion) {
         c.bench_function(&format!("compute_coi_relevance_{base_name}"), |b| {
             b.iter_batched(
                 || black_box(positive_cois),
-                |cois| compute_coi_relevances(cois, horizon, now),
+                |cois| black_box(compute_coi_relevances(cois, horizon, now)),
                 BatchSize::SmallInput,
             );
         });
--- a/coi/benches/benches.rs
+++ b/coi/benches/benches.rs
@@ -12,10 +12,10 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-use std::time::Duration;
+use std::{hint::black_box, time::Duration};
 
 use chrono::Utc;
-use criterion::{black_box, criterion_group, BatchSize, Criterion};
+use criterion::{criterion_group, BatchSize, Criterion};
 use itertools::Itertools;
 use rand::Rng;
 use rand_distr::Uniform;
@@ -44,7 +44,13 @@ fn bench_compute_coi_decay_factor(c: &mut Criterion) {
     let last_view = now - chrono::Duration::seconds(60 * 60);
 
     c.bench_function("compute_coi_decay_factor", |b| {
-        b.iter(|| black_box(compute_coi_decay_factor(horizon, now, last_view)))
+        b.iter(|| {
+            black_box(compute_coi_decay_factor(
+                black_box(horizon),
+                black_box(now),
+                black_box(last_view),
+            ))
+        })
     });
 }
 
@@ -65,7 +71,13 @@ fn bench_compute_coi_relevance(c: &mut Criterion) {
         c.bench_function(&format!("compute_coi_relevance_{base_name}"), |b| {
             b.iter_batched(
                 || black_box(positive_cois),
-                |cois| black_box(compute_coi_relevances(cois, horizon, now)),
+                |cois| {
+                    black_box(compute_coi_relevances(
+                        black_box(cois),
+                        black_box(horizon),
+                        black_box(now),
+                    ))
+                },
                 BatchSize::SmallInput,
             );
         });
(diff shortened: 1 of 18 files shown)


coi/src/context.rs
```diff
@@ -138,29 +135,33 @@ mod tests {
...
+        positive[0].stats.last_view = now
+            - chrono::Duration::from_std(Duration::from_secs_f32(0.5 * SECONDS_PER_DAY_F32))
+                .unwrap();
+        positive[1].stats.last_view = now
+            - chrono::Duration::from_std(Duration::from_secs_f32(1.5 * SECONDS_PER_DAY_F32))
+                .unwrap();
         let mut negative = create_neg_cois([[6., 61., 6.]]);
-        negative[0].last_view = epoch;
+        negative[0].last_view =
+            now - chrono::Duration::from_std(Duration::from_secs_f32(SECONDS_PER_DAY_F32)).unwrap();
```
rustonaut, line 148: :wrench: 

```suggestion
        negative[0].last_view = chrono::Duration::days(1);
```

coi/src/context.rs
```diff
@@ -138,29 +135,33 @@ mod tests {
...
         let cois = UserInterests { positive, negative };
 
+        let embedding = [1., 4., 4.].try_into().unwrap();
         let horizon = Duration::from_secs_f32(2. * SECONDS_PER_DAY_F32);
-
         let score = cois
-            .compute_score_for_embedding(&[1., 4., 4.].try_into().unwrap(), horizon, now)
+            .compute_score_for_embedding(&embedding, horizon, now)
```
rustonaut, line 154: :pushpin:  Given that we use use `chrono` now pretty much everywhere for `DateTime` we probably should also use it for `Duration`. With the drawback that `chrono` duration support negative duration..

coi/src/stats.rs
```diff
@@ -183,13 +180,16 @@ mod tests {
...
         let mut cois = create_pos_cois([[1., 2., 3.], [4., 5., 6.], [7., 8., 9.]]);
-        cois[0].stats.last_view -= Duration::from_secs_f32(0.5 * SECONDS_PER_DAY_F32);
-        cois[1].stats.last_view -= Duration::from_secs_f32(1.5 * SECONDS_PER_DAY_F32);
-        cois[2].stats.last_view -= Duration::from_secs_f32(2.5 * SECONDS_PER_DAY_F32);
+        cois[0].stats.last_view -=
+            chrono::Duration::from_std(Duration::from_secs_f32(0.5 * SECONDS_PER_DAY_F32)).unwrap();
+        cois[1].stats.last_view -=
+            chrono::Duration::from_std(Duration::from_secs_f32(1.5 * SECONDS_PER_DAY_F32)).unwrap();
+        cois[2].stats.last_view -=
+            chrono::Duration::from_std(Duration::from_secs_f32(2.5 * SECONDS_PER_DAY_F32)).unwrap();
```
rustonaut, line 188: :wrench: 

```suggestion
        cois[0].stats.last_view -= chrono::Duration::hours(12);
        cois[1].stats.last_view -= chrono::Duration::hours(36);
        cois[2].stats.last_view -= chrono::Duration::hours(60)
```

coi/src/stats.rs
```diff
@@ -202,16 +202,18 @@ mod tests {
...
+        let now = Utc::now();
+        let factor = compute_coi_decay_factor(horizon, now, now);
         assert_approx_eq!(f32, factor, 1.);
 
-        let now = epoch + Duration::from_secs_f32(5. * SECONDS_PER_DAY_F32);
-        let factor = compute_coi_decay_factor(horizon, now, epoch);
-        assert_approx_eq!(f32, factor, 0.585_914_5);
+        let last = now
+            - chrono::Duration::from_std(Duration::from_secs_f32(5. * SECONDS_PER_DAY_F32))
+                .unwrap();
```
rustonaut, line 211:
```suggestion
        let last = now - chrono::Duration::days(5);
```

web-api/src/storage/postgres.rs
```diff
@@ -315,7 +315,7 @@ impl Database {
                         .push_bind(update.point.to_vec())
                         .push_bind(update.stats.view_count as i32)
                         .push_bind(update.stats.view_time.as_millis() as i64)
-                        .push_bind(now);
+                        .push_bind(time.to_rfc3339());
```
rustonaut, line 318: :wrench: That should not be needed, `sqlx` has a chrono feature.

```suggestion
                        .push_bind(time);
```

web-api/src/storage/postgres.rs
```diff
@@ -353,7 +353,7 @@ impl Database {
                     builder
                         .push_bind(document_id)
                         .push_bind(user_id)
-                        .push_bind(now)
+                        .push_bind(time.to_rfc3339())
```
rustonaut, line 356: :wrench: 

```suggestion
                        .push_bind(time)
```

web-api/src/storage/postgres.rs
```diff
@@ -449,14 +449,15 @@ impl storage::Interaction for Storage {
...
+    async fn user_seen(&self, id: &UserId, time: DateTime<Utc>) -> Result<(), Error> {
         sqlx::query(
             "INSERT INTO users (user_id, last_seen)
-            VALUES ($1, Now())
+            VALUES ($1, $2)
             ON CONFLICT (user_id)
             DO UPDATE SET last_seen = EXCLUDED.last_seen;",
         )
         .bind(id)
+        .bind(time.to_rfc3339())
```
rustonaut, line 460: :wrench: 

```suggestion
        .bind(time)
```

rustonaut (commented):
I'm not sure if renaming `now` to `time` in all places was the best choice because from a business logic POV it's most times now (not in the constructor of CoiStats but in any of the logic on top of that, in storage, etc.). It's just that `now` might not be system-time `now `  in all situations (mainly tests).

coi/benches/benches.rs
```diff
@@ -31,30 +33,26 @@ fn create_positive_coi(n: usize, embedding_size: usize) -> Vec<PositiveCoi> {
...
-            .checked_sub(Duration::new(3600, 0))
-            .unwrap(),
-    );
+    let horizon = Duration::new(60 * 60 * 24 * 30, 0); // 30 days
+    let now = Utc::now();
+    let last_view = now - chrono::Duration::seconds(60 * 60);
 
     c.bench_function("compute_coi_decay_factor", |b| {
-        b.iter(|| compute_coi_decay_factor(horizon, now, last_view))
+        b.iter(|| black_box(compute_coi_decay_factor(horizon, now, last_view)))
```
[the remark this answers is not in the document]
janpetschexain: makes sense. i'll also switch to the `std::hint::black_box()` which finally got stablized with `1.66` (this could already be used in nightly in criterion via the `real_black_box` feature).

coi/src/context.rs
janpetschexain: in the decay computation we currently rely on non-negative durations as we `unwrap_or_default()` in that case. we could replace it but need an extra check for negative values then.

janpetschexain:
> I'm not sure if renaming `now` to `time` in all places was the best choice because from a business logic POV it's most times now (not in the constructor of CoiStats but in any of the logic on top of that, in storage, etc.). It's just that `now` might not be system-time `now ` in all situations (mainly tests).

but this is precisely the point, all the logic now works with some point in time and it doesn't care at all if this is "now" or sth else, all what matters is that the call time and the last_view time can be evaluated in relation to each other. only when the whole logic is called in the routes we decide to set the time to "now". this is the reason why i renamed it in the logic and only kept it at the routes and tests to hint at this difference in meaning.

coi/src/context.rs
janpetschexain: if we want to change this, i guess it's better to do this in a follow up pr. wdyt?

rustonaut: yes I think so too

rustonaut (approved):
lgtm

coi/src/context.rs
janpetschexain: i added [ET-4022] to the backlog

janpetschexain:
bors merge

PR #813, opened 2023-02-08
[Merged by Bors] - ET-3952 mockable interaction time

janpetschexain:
**Reference**

- [ET-3952]
- requires #811

**Summary**

- add a `time` to the `State`/`StateConfig` to give control over what "now" actually means in the personalization, this defaults to `Utc::now()`
- use individual times for each interaction, they are currently set to "now" which is the previous behavior and can be changed to meaningful values later on
- cleanup some docs

[ET-3952]: https://xainag.atlassian.net/browse/ET-3952?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/web-api/src/mind.rs
+++ b/web-api/src/mind.rs
@@ -19,7 +19,7 @@
 use std::{collections::HashMap, fs::File, io, path::Path};
 
 use anyhow::Error;
-use chrono::{DateTime, Utc};
+use chrono::{DateTime, Duration, Utc};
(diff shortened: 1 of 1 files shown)


PanicButtonPressed:
Out of curiosity, was something changed, how we could define the difference in datetime between the clicked results regarding the time relevance weight?

janpetschexain:
bors merge

PR #815, opened 2023-02-09
[Merged by Bors] - make ci more robust

rustonaut:
- Make sure `just web-dev-up` only returns after all db services are healthy and oneshots completed.
- Remove podman support this uses a feature which podman-compose does not yet support


janpetschexain (approved):
👍

rustonaut:
bors merge

PR #827, opened 2023-02-22
[Merged by Bors] - ET-4004 Return bad request if document does not exists

acrrd:
Return an error if some of the documents do not exist.
The documentation was mentioning that the endpoint does not return an error and then contradicts itself by stating that it could return 400. The part around the idempotency of the method was also wrong. Being idempotent does not mean that a method does not return an error if something does not exist.
The HTTP standard requires DELETE methods to be idempotent and it specify that it is in respect to the state of the server not the response. 
```
A request method is considered "idempotent" if the intended EFFECT ON THE SERVER of multiple identical requests with that method is the same as the effect for a single such request. Of the request methods defined by this specification, PUT, DELETE, and safe request methods are idempotent.
```

Returning an error makes `DELETE /documents/<id>` behaving like `GET /documents/<id>`  where we return 400 if the document does not exists.

### DIFF
--- a/web-api/src/storage.rs
+++ b/web-api/src/storage.rs
@@ -67,14 +67,10 @@ pub(crate) enum InsertionError {
 
 #[derive(Debug, From)]
 pub(crate) enum DeletionError {
+    #[from(types(sqlx::Error))]
     General(Error),
+    #[from]
     PartialFailure { errors: Vec<DocumentIdAsObject> },
-}
-
-impl From<sqlx::Error> for DeletionError {
-    fn from(e: sqlx::Error) -> Self {
-        DeletionError::General(e.into())
-    }
 }
 
 #[async_trait(?Send)]
--- a/web-api/src/storage.rs
+++ b/web-api/src/storage.rs
@@ -71,6 +71,12 @@ pub(crate) enum DeletionError {
     PartialFailure { errors: Vec<DocumentIdAsObject> },
 }
 
+impl From<sqlx::Error> for DeletionError {
+    fn from(e: sqlx::Error) -> Self {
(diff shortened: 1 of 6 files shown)


web-api/src/storage.rs
```diff
@@ -71,6 +71,12 @@ pub(crate) enum DeletionError {
     PartialFailure { errors: Vec<DocumentIdAsObject> },
 }
 
+impl From<sqlx::Error> for DeletionError {
+    fn from(e: sqlx::Error) -> Self {
+        DeletionError::General(e.into())
+    }
+}
```
janpetschexain, line 78: :art: we don't have to impl this conversion manually, instead we can annotate the variants of the `DeletionError` like
```rust
#[derive(Debug, From)]
pub(crate) enum DeletionError {
    #[from(types(sqlx::Error))]
    General(Error),
    #[from]
    PartialFailure { errors: Vec<DocumentIdAsObject> },
}
```

janpetschexain (approved):

web-api/src/storage.rs
```diff
@@ -71,6 +71,12 @@ pub(crate) enum DeletionError {
     PartialFailure { errors: Vec<DocumentIdAsObject> },
 }
 
+impl From<sqlx::Error> for DeletionError {
```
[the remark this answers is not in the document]
acrrd: Is not about moving the error to another place, the storage is still using it, and it depends on application errors that the storage should not know about.

web-api/src/storage.rs
acrrd: Nice, thanks

acrrd:
bors merge

PR #828, opened 2023-02-23
[Merged by Bors] - update embeddings for mind

janpetschexain:
**Summary**

- add in-memory storage utility to update the embeddings of documents, fails if documents don't exist yet
- if you change the embedding size, then be careful to not mix updated and non-updated documents afterwards, otherwise this can cause trouble if computations on embeddings of different sizes are performed


PanicButtonPressed (approved):
LGTM! Thanks for the fast implementation

janpetschexain:
bors merge

PR #826, opened 2023-02-22
[Merged by Bors] - ET-3898 get embeddings from postgres

janpetschexain:
**References**

- [ET-3898]
- followed by #829

**Summary**

- get the documents including the embeddings from postgres, only the score comes from elasticsearch if a knn search is done
- use iterators with `&T` items instead of `&[&T]` slices for better ergonomics and less collecting/cloning of large vectors
- fix partial deletion error leftovers from elastic & add test

[ET-3898]: https://xainag.atlassian.net/browse/ET-3898?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #829, opened 2023-02-27
[Merged by Bors] - ET-4072 update property endpoints

janpetschexain:
**Reference**

- [ET-4072]
- requires #826
- followed by #832

**Summary**

- move document properties & property db impls to the postgres side: get only from postgres, put/delete in postgres first and then in elastic
- add integration tests

[ET-4072]: https://xainag.atlassian.net/browse/ET-4072?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/web-api/src/storage.rs
+++ b/web-api/src/storage.rs
@@ -104,7 +104,7 @@ pub(crate) trait Document {
     ) -> Result<(), DeletionError>;
 }
 
-#[async_trait(?Send)]
+#[async_trait]
 pub(crate) trait DocumentProperties {
     async fn get(&self, id: &DocumentId) -> Result<Option<models::DocumentProperties>, Error>;
 
@@ -117,7 +117,7 @@ pub(crate) trait DocumentProperties {
     async fn delete(&self, id: &DocumentId) -> Result<Option<()>, Error>;
 }
 
-#[async_trait(?Send)]
+#[async_trait]
 pub(crate) trait DocumentProperty {
     async fn get(
         &self,
@@ -139,7 +139,7 @@ pub(crate) trait DocumentProperty {
     ) -> Result<Option<Option<()>>, Error>;
 }
 
-#[async_trait(?Send)]
+#[async_trait]
 pub(crate) trait Interest {
     async fn get(&self, user_id: &UserId) -> Result<UserInterests, Error>;
 }
@@ -170,7 +170,7 @@ pub(crate) trait Interaction {
     ) -> Result<(), Error>;
 }
 
-#[async_trait(?Send)]
+#[async_trait]
 pub(crate) trait Tag {
     async fn get(&self, user_id: &UserId) -> Result<HashMap<DocumentTag, usize>, Error>;
 }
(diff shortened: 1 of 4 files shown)


web-api/src/storage.rs
```diff
@@ -136,10 +136,10 @@ pub(crate) trait DocumentProperty {
         &self,
         document_id: &DocumentId,
         property_id: &DocumentPropertyId,
-    ) -> Result<Option<()>, Error>;
+    ) -> Result<Option<Option<()>>, Error>;
```
janpetschexain, line 139: this is similar to the changes in [ET-4004] such that the property delete endpoint properly signals a `DocumentPropertyIdNotFound`.

web-api/src/storage.rs
```diff
@@ -104,7 +104,7 @@ pub(crate) trait Document {
     ) -> Result<(), DeletionError>;
 }
 
-#[async_trait]
+#[async_trait(?Send)]
```
acrrd, line 107: Can't we do it only on the trait where we need it?

web-api/src/storage/elastic.rs
```diff
@@ -121,6 +121,13 @@ struct BulkResponse {
     items: Vec<HashMap<String, BulkItemResponse>>,
 }
 
+#[derive(Debug, Deserialize)]
+/// Deserializes from any map/struct dropping all fields.
+///
+/// This will not work with non self describing non schema
```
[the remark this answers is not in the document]
rustonaut: It means it isn't compatible with some formats e.g. bincode, but has no problems at all for json.

web-api/src/storage.rs
rustonaut: Did we start adding this due to adding more generics in async-trait traits?

web-api/src/storage/postgres.rs
```diff
@@ -622,19 +656,187 @@ impl Storage {
...
+    async fn get(&self, id: &DocumentId) -> Result<Option<DocumentProperties>, Error> {
+        let mut tx = self.postgres.pool.begin().await?;
+
+        let properties = sqlx::query_as::<_, QueriedDocumentProperties>(
+            "SELECT properties
+            FROM document
+            WHERE document_id = $1;",
+        )
+        .bind(id)
+        .fetch_optional(&mut tx)
```
rustonaut, line 675: :wrench:  You can just pass the pool instead of an transaction the always using of a transaction was only a thing due to a subtle bug of the sqlite driver of sqlx.

rustonaut: Same for some of the other endpoints below.

web-api/src/storage/elastic.rs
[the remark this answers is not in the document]
janpetschexain: also, this is old code, it just moved around.

web-api/src/storage.rs
janpetschexain: > Did we start adding this due to adding more generics in async-trait traits?

yes, the `Send` bounds were in conflict with some of the iterator arguments. since we can't enforce the `Send` bounds lint because of `actix-web`, we decided to relax it here as well.

> Can't we do it only on the trait where we need it?

i'll double check and only relax the bounds where needed.

PR #830, opened 2023-02-28
Modularize mind code

janpetschexain:
**Summary**

- split the mind code into submodules: `config`, `data` & `state`, the benchmarks themselves remain in the main module
- move some free functions into their type impls & improve caller ergonomics
- use `Panic` instead of `anyhow::Error` such that `.unwrap()`s  can be replaced with `?`

### DIFF
--- a/web-api/src/mind/config.rs
+++ b/web-api/src/mind/config.rs
@@ -18,7 +18,7 @@ use xayn_ai_coi::CoiConfig;
 
 use crate::personalization::PersonalizationConfig;
 
-#[derive(Clone, Debug, Serialize)]
+#[derive(Debug, Serialize)]
 pub(super) struct StateConfig {
     pub(super) coi: CoiConfig,
     pub(super) personalization: PersonalizationConfig,
@@ -35,6 +35,7 @@ impl Default for StateConfig {
     }
 }
 
+#[derive(Debug)]
 pub(super) struct GridSearchConfig {
     pub(super) thresholds: Vec<f32>,
     pub(super) shifts: Vec<f32>,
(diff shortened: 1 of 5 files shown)


web-api/src/mind/state.rs
```diff
@@ -0,0 +1,197 @@
...
+    }
+}
+
+/// The results of the saturation benchmark.
+#[derive(Debug, Default, Deref, DerefMut, Serialize)]
+pub(super) struct SaturationResult {
+    topics: Vec<SaturationTopicResult>,
+}
+
+impl SaturationResult {
```
PanicButtonPressed, line 191: We could move the result structs to `data.rs` as they hold data in some way but don't have a strong opinion about it

web-api/src/mind/config.rs
```diff
@@ -0,0 +1,134 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+use chrono::{DateTime, Utc};
+use serde::{Deserialize, Serialize};
+use xayn_ai_coi::CoiConfig;
+use xayn_test_utils::error::Panic;
```
PanicButtonPressed, line 18: I assume we use panic to get rid of the `unwraps`?

PanicButtonPressed (approved):
Looks good to me. Like the cleanup, increases the overview a lot

PanicButtonPressed:
FYI @krajewskiML

web-api/src/mind/config.rs
janpetschexain: exactly, as noted in the pr description

web-api/src/mind/state.rs
janpetschexain: i guess so, i neither have a strong opinion on that. i placed it there since it holds the state of the iteration outputs, but it could as well be moved into `data.rs`.

janpetschexain:
bors merge

PR #831, opened 2023-03-01
[Merged by Bors] - ET-4093 openapi: semantic search with user history

rustonaut:
Adds a `POST` variant of `/semantic_search/{document_id}` which allows specifying users either by id or by in-lining the user object (for now limited to `history`).

**References:**

- issue: [ET-4093]
- story: [ET-4092]

[ET-4093]: https://xainag.atlassian.net/browse/ET-4093?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ET-4092]: https://xainag.atlassian.net/browse/ET-4092?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/web-api/openapi/front_office.yaml
+++ b/web-api/openapi/front_office.yaml
@@ -239,7 +239,7 @@ components:
     SemanticSearchRequest:
       type: object
       properties:
-        document:
+        document_id:
           $ref: './schemas/document.yml#/DocumentId'
         count:
           $ref: '#/components/schemas/Count'
--- a/web-api/openapi/front_office.yaml
+++ b/web-api/openapi/front_office.yaml
@@ -246,15 +246,6 @@ components:
           $ref: '#/components/schemas/Count'
         published_after:
           $ref: './schemas/time.yml#/PublicationDate'
-        exclude_seen:
-          type: boolean
-          default: true
-          description: |
-            If true do not include documents the user has already seen as search candidate.
-
-            A trimmed version of the users history might be used for this filter.
-
-            This option is incompatible with not specifying a user.
         min_similarity:
           description: "Minimal similarity of a document to consider it as search candidate."
           type: number
@@ -262,8 +253,21 @@ components:
           minimum: 0
           maximum: 1
           default: 0
-        user:
-          $ref: './schemas/user.yml#/InputUser'
+        personalize:
+          description: Personalize the ranking of candidates based on a users preferences
+          type: object
+          properties:
+            exclude_seen:
+              type: boolean
+              default: true
+              description: |
+                If true do not include documents the user has already seen as search candidate.
+
+                A trimmed version of the users history might be used for this filter.
+
+                This option is incompatible with not specifying a user.
+            user:
+              $ref: './schemas/user.yml#/InputUser'
     SemanticSearchResponse:
       type: object
       required: [documents]
--- a/web-api/openapi/front_office.yaml
+++ b/web-api/openapi/front_office.yaml
@@ -157,8 +157,6 @@ paths:
           application/json:
             schema:
               $ref: '#/components/schemas/SemanticSearchRequest'
-      parameters:
-        - $ref: './parameters/path/document_id.yml'
       responses:
         '200':
           description: successful operation
--- a/web-api/openapi/front_office.yaml
+++ b/web-api/openapi/front_office.yaml
@@ -252,6 +252,8 @@ components:
           description: |
             If true do not include documents the user has already seen as search candidate.
 
+            A trimmed version of the users history might be used for this filter.
+
             This option is incompatible with not specifying a user.
         min_similarity:
           description: "Minimal similarity of a document to consider it as search candidate."
--- a/web-api/openapi/front_office.yaml
+++ b/web-api/openapi/front_office.yaml
@@ -246,6 +246,13 @@ components:
           $ref: '#/components/schemas/Count'
         published_after:
           $ref: './schemas/time.yml#/PublicationDate'
+        exclude_seen:
+          type: bool
+          default: true
+          description: |
+            If true do not include documents the user has already seen as search candidate.
+
+            This option is incompatible with not specifying a user.
         min_similarity:
           description: "Minimal similarity of a document to consider it as search candidate."
           type: number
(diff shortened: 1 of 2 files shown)


web-api/openapi/front_office.yaml
```diff
@@ -151,32 +151,12 @@ paths:
...
-            maximum: 1
-            default: 0
-        - name: personalize_for
-          in: query
-          description:  |-
-            A user for whom the documents will be personalized for.
-            If that user doesn't have enough interactions in the system, then the documents are returned unpersonalized.
-          required: false
-          schema:
-            $ref: './schemas/user.yml#/UserId'
```
rustonaut, line 179: If we anyway have query parameters in the body there is no reason to split them between the body and the url.

web-api/openapi/schemas/user.yml
```diff
@@ -2,3 +2,23 @@ UserId:
...
+
+    Be aware that currently using the `id` in combination with any
+    other field is not allowed.
+  type: object
+  properties:
+    id:
+      $ref: '#/UserId'
+    history:
+      $ref: './document.yml#/History'
+  # Disabled as our linter can't handle that correctly
```
acrrd, line 21: I don't think `oneOf` with that shape is valid at all w.r.t. the specification.

```yaml
    InputUserById:
      type: object
      required:
        - id
      properties:
        id:
          type: string
    InputUserByHistory:
      type: object
      required:
        - history
      properties:
        history:
          type: string
    InputUser:
      oneOf:
        - $ref: '#/components/schemas/InputUserById'
        - $ref: '#/components/schemas/InputUserByHistory'
```

This should do what we want.

web-api/openapi/schemas/user.yml
```diff
@@ -2,3 +2,23 @@ UserId:
   description: Id of a user.
   $ref: './id.yml#/Id'
   example: "user_id"
+
+InputUser:
+  description: |
+    Information about a user provided as input for an search
+
+    This can be anything from no user information, to a user id, to
```
acrrd, line 10: This should represent `no user information`, which is already represented by having the field `user` optional in `SemanticSearchrequest`.

web-api/openapi/schemas/user.yml
acrrd: We should also move `exclude_seen` under `user` since it does not make sense without it.

```yaml
InputUserById:
  type: object
    required:
      - id
    properties:
      id:
        type: string
InputUserByHistory:
  type: object
    required:
      - history
    properties:
      history:
        type: string
InputUserBy:
  oneOf:
    - $ref: '#/InputUserById'
    - $ref: '#/InputUserByHistory'

InputUser:
  type: object
  allOf:
    - $ref: '#/InputUserBy'
    - type: object
      properties:
        exclude_seen:
           type: bool
```

rustonaut: > I don't think oneOf with that shape is valid at all w.r.t. the specification.

No it is valid, it's just the tooling which doesn't work correctly.

Json schema is not based on describing data but describing _validation rules_, and `oneOf` is a keyword which "tests" the data against the object it's contained in with each of it's variants inlined into it once and makes sure it only matches once.

(The  relevant part is here https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-00#section-10.2, the first line "These keywords apply subschemas to the same location in the instance as the parent schema is being applied.")

I originally had a version similar to what you proposed but decided against it as it's semantically not quite right.

We don't have multiple different entities you can pass in as the user fields but just one (partial) inlined "user" entity. Of which we can include various fields, like the id, history, center of interest etc. Currently we are limited to passing in id and history and currently we allow only either of them. But that's  more an implementation detail then anything else. I just for now don't want to handle questions like "do you store the inline passed in history", "do you recreate CoIs from the inline history or do you update the users cois", "do you store the computed cois", etc.  So it's more an limitation due to  implementation details then a fundamental aspect of the endpoint.  

For the same reason I didn't include `exclude_seen` in the user entity/object, it's not an property of the user but the endpoint where we can pass in a user "inline".

rustonaut: You know what I will change it.

I originally had the idea that we could long term just provide an arbitrary subset of the user object and then act appropriately, i.e. if id fetch data from db and store it back, if inline history update/create cois from it, if inline cois use them as a base etc.

But just doing the oneOf like you wrote is simpler and more in line with how it's currently implemented.

rustonaut: Through I will still keep the `exclud_seen` out of the user struct.

This: 

```
InputUser:
  type: object
  allOf:
    - oneOf:
      - $ref: '#/UserById'
      - $ref: '#/UserByHistory'
    - type: object
      properties:
        exclude_seen:
          type: boolean
```

is nearly guaranteed to trip over some tools :see_no_evil:

acrrd: So I found out when I tried I put the `oneOf` in the wrong place and I was getting an error also from other openapi tools.

> We don't have multiple different entities you can pass in as the user fields but just one (partial) inlined "user" entity. Of which we can include various fields, like the id, history, center of interest etc.

I disagree with this. In the context of this request, it is totally valid to see the user as an entity with an id or an entity with history.
Most of the problem about the semantics seems to come from having the field called `user`.
We can call it `personalize`, and I think most of the problems will go away:
```yaml
personalize:
  type object
  properties:
    exclude_seen:
      type: boolean
    for: # other option for the name: with, using, user not a strong opinion on this
      type object
      oneOf:
        - $ref: '#/FromId'
        - $ref: '#/FromHistory'
```
What do you think about this?

rustonaut: okay

web-api/openapi/front_office.yaml
```diff
@@ -256,6 +236,38 @@ components:
...
+          properties:
+            exclude_seen:
+              type: boolean
+              default: true
+              description: |
+                If true do not include documents the user has already seen as search candidate.
+
+                A trimmed version of the users history might be used for this filter.
+
+                This option is incompatible with not specifying a user.
```
acrrd, line 268: :art: We can mark `user` as required, and we can remove this comment.

acrrd (approved):

PR #832, opened 2023-03-01
[Merged by Bors] - ET-3897 migrate documents from elastic to postgres

janpetschexain:
**Reference**

- [ET-3897]
- requires #829
- followed by #833

**Summary**

- add fallbacks to elastic for missing data in postgres while migration has not yet been finished
- implement data migration & add a `migration -n N -t T` binary (arguments are additionally to the ones we have for the back/frontoffice binaries)
  - `N`: number of docs per time intervall, defaults to `10`
  - `T`: time interval in seconds, defaults to `1`
- checked that the integration tests run with the feature flag enabled

**Test**

i tested the migration but i don't want to commit that code, hence i add it here for reference. therefore, we need some utility on the postgres side:
```rust
#[cfg(feature = "ET-3837")]
impl Database {
    pub(crate) async fn get_unmigrated(
        &self,
        ids: &[DocumentId],
    ) -> Result<Vec<UnmigratedDocument>, Error> {
        use sqlx::{postgres::PgRow, Row};

        QueryBuilder::new(
            "SELECT document_id, snippet, properties, tags, embedding
            FROM document
            WHERE document_id IN ",
        )
        .push_tuple(ids)
        .build()
        .try_map(|row: PgRow| {
            Ok(UnmigratedDocument {
                id: row.try_get("document_id")?,
                snippet: row.try_get("snippet")?,
                properties: row.try_get::<Json<_>, _>("properties")?.0,
                tags: row.try_get("tags")?,
                embedding: row.try_get("embedding")?,
            })
        })
        .fetch_all(&self.pool)
        .await
        .map_err(Into::into)
    }
}

#[cfg(feature = "ET-3837")]
impl Storage {
    pub(crate) async fn prepare_partially_migrated(&self) -> Result<(), Error> {
        use crate::{
            error::common::IngestingDocumentsFailed,
            models::DocumentProperty,
            storage::InsertionError,
        };

        let documents = (1..=10)
            .map(|i| {
                Ok((
                    IngestedDocument {
                        id: format!("d{i}").try_into()?,
                        snippet: format!("snippet {i}"),
                        properties: [("p".try_into()?, DocumentProperty(serde_json::json!(i)))]
                            .into(),
                        tags: vec![format!("t{i}").try_into()?],
                    },
                    xayn_ai_bert::Embedding1::from(vec![1.0; 128]).normalize()?,
                ))
            })
            .try_collect::<_, Vec<_>, Error>()?;
        storage::Document::insert(self, documents)
            .await
            .map_err(|error| match error {
                InsertionError::General(error) => error,
                InsertionError::PartialFailure {
                    failed_documents: documents,
                } => IngestingDocumentsFailed { documents }.into(),
            })?;
        let documents = (6..=10)
            .map(|i| {
                Ok((
                    IngestedDocument {
                        id: format!("d{i}").try_into()?,
                        snippet: String::default(),
                        properties: DocumentProperties::default(),
                        tags: Vec::default(),
                    },
                    xayn_ai_bert::Embedding1::default().normalize()?,
                ))
            })
            .try_collect::<_, Vec<_>, Error>()?;
        self.postgres.insert_documents(&documents).await
    }

    pub(crate) async fn check_state(&self) -> Result<(), Error> {
        let ids = (1..=10)
            .map(|i| format!("d{i}").try_into())
            .try_collect::<_, Vec<DocumentId>, _>()?;
        let documents = self.elastic.get_unmigrated(&ids).await?;
        info!(?documents, "Elastic");
        let documents = self.postgres.get_unmigrated(&ids).await?;
        info!(?documents, "Postgres");
        let is_migrated = *self.is_migrated.read().await;
        info!("DB is migrated: {}", is_migrated);
        Ok(())
    }
}
```
then we can use this utility in the binary to set up and check the db state. if we then run the enhanced `migration` binary
we will get the following expected output:
```
% cargo run --bin migration --release --features ET-3837 -- -n 2 -t 1

2023-03-07T13:51:00.217519Z  INFO setup_database{self=Config { base_url: "postgres://user:pw@localhost:3054/xayn", port: None, user: None, password: Secret([REDACTED alloc::string::String]), db: None, application_name: None, skip_migrations: false }}: xayn_web_api::storage::postgres: starting postgres setup
2023-03-07T13:51:00.241192Z  INFO setup_database{self=Config { base_url: "postgres://user:pw@localhost:3054/xayn", port: None, user: None, password: Secret([REDACTED alloc::string::String]), db: None, application_name: None, skip_migrations: false }}: sqlx::postgres::notice: relation "_sqlx_migrations" already exists, skipping    
2023-03-07T13:51:00.307727Z  INFO xayn_web_api::storage::postgres: Elastic documents=[UnmigratedDocument { id: DocumentId("d1"), snippet: "snippet 1", properties: {DocumentPropertyId("p"): DocumentProperty(Number(1))}, tags: [DocumentTag("t1")], embedding: NormalizedEmbedding(Embedding([0.088388346, ..., 0.088388346], shape=[128], strides=[1], layout=CFcf (0xf), const ndim=1)) }, UnmigratedDocument { id: DocumentId("d2"), snippet: "snippet 2", properties: {DocumentPropertyId("p"): DocumentProperty(Number(2))}, tags: [DocumentTag("t2")], embedding: NormalizedEmbedding(Embedding([0.088388346, ..., 0.088388346], shape=[128], strides=[1], layout=CFcf (0xf), const ndim=1)) }, UnmigratedDocument { id: DocumentId("d3"), snippet: "snippet 3", properties: {DocumentPropertyId("p"): DocumentProperty(Number(3))}, tags: [DocumentTag("t3")], embedding: NormalizedEmbedding(Embedding([0.088388346, ..., 0.088388346], shape=[128], strides=[1], layout=CFcf (0xf), const ndim=1)) }, UnmigratedDocument { id: DocumentId("d4"), snippet: "snippet 4", properties: {DocumentPropertyId("p"): DocumentProperty(Number(4))}, tags: [DocumentTag("t4")], embedding: NormalizedEmbedding(Embedding([0.088388346, ..., 0.088388346], shape=[128], strides=[1], layout=CFcf (0xf), const ndim=1)) }, UnmigratedDocument { id: DocumentId("d5"), snippet: "snippet 5", properties: {DocumentPropertyId("p"): DocumentProperty(Number(5))}, tags: [DocumentTag("t5")], embedding: NormalizedEmbedding(Embedding([0.088388346, ..., 0.088388346], shape=[128], strides=[1], layout=CFcf (0xf), const ndim=1)) }, UnmigratedDocument { id: DocumentId("d6"), snippet: "snippet 6", properties: {DocumentPropertyId("p"): DocumentProperty(Number(6))}, tags: [DocumentTag("t6")], embedding: NormalizedEmbedding(Embedding([0.088388346, ..., 0.088388346], shape=[128], strides=[1], layout=CFcf (0xf), const ndim=1)) }, UnmigratedDocument { id: DocumentId("d7"), snippet: "snippet 7", properties: {DocumentPropertyId("p"): DocumentProperty(Number(7))}, tags: [DocumentTag("t7")], embedding: NormalizedEmbedding(Embedding([0.088388346, ..., 0.088388346], shape=[128], strides=[1], layout=CFcf (0xf), const ndim=1)) }, UnmigratedDocument { id: DocumentId("d8"), snippet: "snippet 8", properties: {DocumentPropertyId("p"): DocumentProperty(Number(8))}, tags: [DocumentTag("t8")], embedding: NormalizedEmbedding(Embedding([0.088388346, ..., 0.088388346], shape=[128], strides=[1], layout=CFcf (0xf), const ndim=1)) }, UnmigratedDocument { id: DocumentId("d9"), snippet: "snippet 9", properties: {DocumentPropertyId("p"): DocumentProperty(Number(9))}, tags: [DocumentTag("t9")], embedding: NormalizedEmbedding(Embedding([0.088388346, ..., 0.088388346], shape=[128], strides=[1], layout=CFcf (0xf), const ndim=1)) }, UnmigratedDocument { id: DocumentId("d10"), snippet: "snippet 10", properties: {DocumentPropertyId("p"): DocumentProperty(Number(10))}, tags: [DocumentTag("t10")], embedding: NormalizedEmbedding(Embedding([0.088388346, ..., 0.088388346], shape=[128], strides=[1], layout=CFcf (0xf), const ndim=1)) }]
2023-03-07T13:51:00.310436Z  INFO xayn_web_api::storage::postgres: Postgres documents=[UnmigratedDocument { id: DocumentId("d1"), snippet: "snippet 1", properties: {DocumentPropertyId("p"): DocumentProperty(Number(1))}, tags: [DocumentTag("t1")], embedding: NormalizedEmbedding(Embedding([0.088388346, ..., 0.088388346], shape=[128], strides=[1], layout=CFcf (0xf), const ndim=1)) }, UnmigratedDocument { id: DocumentId("d2"), snippet: "snippet 2", properties: {DocumentPropertyId("p"): DocumentProperty(Number(2))}, tags: [DocumentTag("t2")], embedding: NormalizedEmbedding(Embedding([0.088388346, ..., 0.088388346], shape=[128], strides=[1], layout=CFcf (0xf), const ndim=1)) }, UnmigratedDocument { id: DocumentId("d3"), snippet: "snippet 3", properties: {DocumentPropertyId("p"): DocumentProperty(Number(3))}, tags: [DocumentTag("t3")], embedding: NormalizedEmbedding(Embedding([0.088388346, ..., 0.088388346], shape=[128], strides=[1], layout=CFcf (0xf), const ndim=1)) }, UnmigratedDocument { id: DocumentId("d4"), snippet: "snippet 4", properties: {DocumentPropertyId("p"): DocumentProperty(Number(4))}, tags: [DocumentTag("t4")], embedding: NormalizedEmbedding(Embedding([0.088388346, ..., 0.088388346], shape=[128], strides=[1], layout=CFcf (0xf), const ndim=1)) }, UnmigratedDocument { id: DocumentId("d5"), snippet: "snippet 5", properties: {DocumentPropertyId("p"): DocumentProperty(Number(5))}, tags: [DocumentTag("t5")], embedding: NormalizedEmbedding(Embedding([0.088388346, ..., 0.088388346], shape=[128], strides=[1], layout=CFcf (0xf), const ndim=1)) }, UnmigratedDocument { id: DocumentId("d6"), snippet: "", properties: {}, tags: [], embedding: NormalizedEmbedding(Embedding([], shape=[0], strides=[0], layout=CFcf (0xf), const ndim=1)) }, UnmigratedDocument { id: DocumentId("d7"), snippet: "", properties: {}, tags: [], embedding: NormalizedEmbedding(Embedding([], shape=[0], strides=[0], layout=CFcf (0xf), const ndim=1)) }, UnmigratedDocument { id: DocumentId("d8"), snippet: "", properties: {}, tags: [], embedding: NormalizedEmbedding(Embedding([], shape=[0], strides=[0], layout=CFcf (0xf), const ndim=1)) }, UnmigratedDocument { id: DocumentId("d9"), snippet: "", properties: {}, tags: [], embedding: NormalizedEmbedding(Embedding([], shape=[0], strides=[0], layout=CFcf (0xf), const ndim=1)) }, UnmigratedDocument { id: DocumentId("d10"), snippet: "", properties: {}, tags: [], embedding: NormalizedEmbedding(Embedding([], shape=[0], strides=[0], layout=CFcf (0xf), const ndim=1)) }]
2023-03-07T13:51:00.311310Z  INFO xayn_web_api::storage::postgres: DB is migrated: false
2023-03-07T13:51:04.336180Z  INFO xayn_web_api::storage::postgres: Elastic documents=[UnmigratedDocument { id: DocumentId("d1"), snippet: "snippet 1", properties: {DocumentPropertyId("p"): DocumentProperty(Number(1))}, tags: [DocumentTag("t1")], embedding: NormalizedEmbedding(Embedding([0.088388346, ..., 0.088388346], shape=[128], strides=[1], layout=CFcf (0xf), const ndim=1)) }, UnmigratedDocument { id: DocumentId("d2"), snippet: "snippet 2", properties: {DocumentPropertyId("p"): DocumentProperty(Number(2))}, tags: [DocumentTag("t2")], embedding: NormalizedEmbedding(Embedding([0.088388346, ..., 0.088388346], shape=[128], strides=[1], layout=CFcf (0xf), const ndim=1)) }, UnmigratedDocument { id: DocumentId("d3"), snippet: "snippet 3", properties: {DocumentPropertyId("p"): DocumentProperty(Number(3))}, tags: [DocumentTag("t3")], embedding: NormalizedEmbedding(Embedding([0.088388346, ..., 0.088388346], shape=[128], strides=[1], layout=CFcf (0xf), const ndim=1)) }, UnmigratedDocument { id: DocumentId("d4"), snippet: "snippet 4", properties: {DocumentPropertyId("p"): DocumentProperty(Number(4))}, tags: [DocumentTag("t4")], embedding: NormalizedEmbedding(Embedding([0.088388346, ..., 0.088388346], shape=[128], strides=[1], layout=CFcf (0xf), const ndim=1)) }, UnmigratedDocument { id: DocumentId("d5"), snippet: "snippet 5", properties: {DocumentPropertyId("p"): DocumentProperty(Number(5))}, tags: [DocumentTag("t5")], embedding: NormalizedEmbedding(Embedding([0.088388346, ..., 0.088388346], shape=[128], strides=[1], layout=CFcf (0xf), const ndim=1)) }, UnmigratedDocument { id: DocumentId("d6"), snippet: "snippet 6", properties: {DocumentPropertyId("p"): DocumentProperty(Number(6))}, tags: [DocumentTag("t6")], embedding: NormalizedEmbedding(Embedding([0.088388346, ..., 0.088388346], shape=[128], strides=[1], layout=CFcf (0xf), const ndim=1)) }, UnmigratedDocument { id: DocumentId("d7"), snippet: "snippet 7", properties: {DocumentPropertyId("p"): DocumentProperty(Number(7))}, tags: [DocumentTag("t7")], embedding: NormalizedEmbedding(Embedding([0.088388346, ..., 0.088388346], shape=[128], strides=[1], layout=CFcf (0xf), const ndim=1)) }, UnmigratedDocument { id: DocumentId("d8"), snippet: "snippet 8", properties: {DocumentPropertyId("p"): DocumentProperty(Number(8))}, tags: [DocumentTag("t8")], embedding: NormalizedEmbedding(Embedding([0.088388346, ..., 0.088388346], shape=[128], strides=[1], layout=CFcf (0xf), const ndim=1)) }, UnmigratedDocument { id: DocumentId("d9"), snippet: "snippet 9", properties: {DocumentPropertyId("p"): DocumentProperty(Number(9))}, tags: [DocumentTag("t9")], embedding: NormalizedEmbedding(Embedding([0.088388346, ..., 0.088388346], shape=[128], strides=[1], layout=CFcf (0xf), const ndim=1)) }, UnmigratedDocument { id: DocumentId("d10"), snippet: "snippet 10", properties: {DocumentPropertyId("p"): DocumentProperty(Number(10))}, tags: [DocumentTag("t10")], embedding: NormalizedEmbedding(Embedding([0.088388346, ..., 0.088388346], shape=[128], strides=[1], layout=CFcf (0xf), const ndim=1)) }]
2023-03-07T13:51:04.341466Z  INFO xayn_web_api::storage::postgres: Postgres documents=[UnmigratedDocument { id: DocumentId("d1"), snippet: "snippet 1", properties: {DocumentPropertyId("p"): DocumentProperty(Number(1))}, tags: [DocumentTag("t1")], embedding: NormalizedEmbedding(Embedding([0.088388346, ..., 0.088388346], shape=[128], strides=[1], layout=CFcf (0xf), const ndim=1)) }, UnmigratedDocument { id: DocumentId("d2"), snippet: "snippet 2", properties: {DocumentPropertyId("p"): DocumentProperty(Number(2))}, tags: [DocumentTag("t2")], embedding: NormalizedEmbedding(Embedding([0.088388346, ..., 0.088388346], shape=[128], strides=[1], layout=CFcf (0xf), const ndim=1)) }, UnmigratedDocument { id: DocumentId("d3"), snippet: "snippet 3", properties: {DocumentPropertyId("p"): DocumentProperty(Number(3))}, tags: [DocumentTag("t3")], embedding: NormalizedEmbedding(Embedding([0.088388346, ..., 0.088388346], shape=[128], strides=[1], layout=CFcf (0xf), const ndim=1)) }, UnmigratedDocument { id: DocumentId("d4"), snippet: "snippet 4", properties: {DocumentPropertyId("p"): DocumentProperty(Number(4))}, tags: [DocumentTag("t4")], embedding: NormalizedEmbedding(Embedding([0.088388346, ..., 0.088388346], shape=[128], strides=[1], layout=CFcf (0xf), const ndim=1)) }, UnmigratedDocument { id: DocumentId("d5"), snippet: "snippet 5", properties: {DocumentPropertyId("p"): DocumentProperty(Number(5))}, tags: [DocumentTag("t5")], embedding: NormalizedEmbedding(Embedding([0.088388346, ..., 0.088388346], shape=[128], strides=[1], layout=CFcf (0xf), const ndim=1)) }, UnmigratedDocument { id: DocumentId("d10"), snippet: "snippet 10", properties: {DocumentPropertyId("p"): DocumentProperty(Number(10))}, tags: [DocumentTag("t10")], embedding: NormalizedEmbedding(Embedding([0.088388346, ..., 0.088388346], shape=[128], strides=[1], layout=CFcf (0xf), const ndim=1)) }, UnmigratedDocument { id: DocumentId("d6"), snippet: "snippet 6", properties: {DocumentPropertyId("p"): DocumentProperty(Number(6))}, tags: [DocumentTag("t6")], embedding: NormalizedEmbedding(Embedding([0.088388346, ..., 0.088388346], shape=[128], strides=[1], layout=CFcf (0xf), const ndim=1)) }, UnmigratedDocument { id: DocumentId("d7"), snippet: "snippet 7", properties: {DocumentPropertyId("p"): DocumentProperty(Number(7))}, tags: [DocumentTag("t7")], embedding: NormalizedEmbedding(Embedding([0.088388346, ..., 0.088388346], shape=[128], strides=[1], layout=CFcf (0xf), const ndim=1)) }, UnmigratedDocument { id: DocumentId("d8"), snippet: "snippet 8", properties: {DocumentPropertyId("p"): DocumentProperty(Number(8))}, tags: [DocumentTag("t8")], embedding: NormalizedEmbedding(Embedding([0.088388346, ..., 0.088388346], shape=[128], strides=[1], layout=CFcf (0xf), const ndim=1)) }, UnmigratedDocument { id: DocumentId("d9"), snippet: "snippet 9", properties: {DocumentPropertyId("p"): DocumentProperty(Number(9))}, tags: [DocumentTag("t9")], embedding: NormalizedEmbedding(Embedding([0.088388346, ..., 0.088388346], shape=[128], strides=[1], layout=CFcf (0xf), const ndim=1)) }]
2023-03-07T13:51:04.342150Z  INFO xayn_web_api::storage::postgres: DB is migrated: true
2023-03-07T13:51:04.342752Z  INFO migration: return=()
```

[ET-3897]: https://xainag.atlassian.net/browse/ET-3897?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


web-api/src/ingestion/routes.rs
```diff
@@ -64,6 +72,10 @@ pub(super) fn configure_service(config: &mut ServiceConfig) {
                 .route(web::put().to(put_document_property.error_with_request_id()))
                 .route(web::delete().to(delete_document_property.error_with_request_id())),
         );
+    #[cfg(feature = "ET-3837")]
+    config.service(
+        web::resource("/migration").route(web::patch().to(migrate.error_with_request_id())),
```
janpetschexain, line 77: ❓ since this route is not available through the firewall it should only be accesible by us and therefore safe, right? otherwise we have to find another migration trigger.

web-api/src/storage/elastic.rs
```diff
@@ -225,6 +225,91 @@ impl Client {
         self.query_with_bytes(url, post_data).await
     }
 
+    pub(super) async fn get_interacted(
```
janpetschexain, line 228: this and the following functions just moved out of the trait impls below

web-api/src/storage/postgres.rs
```diff
@@ -649,11 +672,81 @@ impl Storage {
...
+        }
+
+        let mut tx = self.postgres.pool.begin().await?;
+
+        let ids = sqlx::query_as::<_, DocumentId>(
+            "SELECT document_id
+            FROM document
+            WHERE cardinality(embedding) = 0
+            LIMIT $1
+            FOR UPDATE;",
```
acrrd, line 702: This can block a large number of documents for the users for possibly a bit of time since ElasticSearch is involved.
It could be better do one document at a time and parallelize the work instead.

Also, what happens in this case if a write to a property was requested and we are in the case that ElasticSearch has not to be updated yet? I think the property update will overwrite the data with the new value. Is this correct?

acrrd (commented):
Having `is_migrated` in the configuration is prone to errors that could lead to data loss. 
It also forces to restart of the service with the changed configuration or call `/migrate` for each instance.
We could do a deployment where we enable only the functions that write to Postgres and Elastic search. We do the migration and then do another deployment with the feature flag removed.

I don't like the endpoint too much, it means that one instance is stuck doing the migration instead of serving the users. I'll prefer to have a separate binary that we can deploy, in one or multiple instances) alongside the instances that handle the migration. Any opinion on this @Robert-Steiner @joecks ?

web-api/src/storage/postgres.rs
janpetschexain: > This can block a large number of documents for the users for possibly a bit of time since ElasticSearch is involved.

`FOR UPDATE` will only lock `LIMIT` number of documents, the others won't be locked. this is why there are the parameters in the query to migrate `n` documents per time window `t`, eg we could migrate 10docs/1sec which would leave the number of locked documents quite small. this also doesn't mean that the docs are locked for the whole time window, but only for how long it takes to migrate them, basically one pg call to lock the ids, then one es call to get the doc data, then one pg call to insert the data.

> It could be better do one document at a time and parallelize the work instead.

parallelization could also be an issue if we have to many parallel calls to the dbs instead of just a few batched calls.

> Also, what happens in this case if a write to a property was requested and we are in the case that ElasticSearch has not to be updated yet? I think the property update will overwrite the data with the new value. Is this correct?

it is important to note that with the new deployment all insert/delete operations of documents/properties will be done to pg & es, not only es. the fallback only applies to data from before the new deployment. then we can have multiple scenarios:
- a write to a property of a doc which has been ingested after the new deployment or a doc which has already be migrated: the write will happen to pg & es since the migration won't touch it again
- a write to a property of a doc which has not yet been migrated and migration is running: the write can't happen because of the lock
- a write to a property of a doc which has not yet been migrated and migration is not yet running: here we can run into overwrite issues (with old data instead of new data), eg if the following order happens 1) write new property to pg, 2) lock & run doc migration which copies old property from es to pg, 3) write new property to es. i'll add locks to prevent that, we can remove them after the migration.

the last point is essentially an issue caused by the missing transaction system wrapping both es & pg. for example this can also happen for documents with no migration involved. image the following: 1) ingest document which writes to pg, 2) immediately call delete which deletes from pg, 3) delete tries at es but doesn't find the doc yet which is ok for es, 4) ingest doc to es happens to be slow this time and occurs after the delete. similar issues can be constructed for ingesting the doc twice with the same id but different data, which attempts to overwrite in our current implementation, then we could also end up with unmatching pg & es data. there are likely more of those cases, all of them hard to guard against without a transaction system.

janpetschexain:
> Having `is_migrated` in the configuration is prone to errors that could lead to data loss. It also forces to restart of the service with the changed configuration or call `/migrate` for each instance. We could do a deployment where we enable only the functions that write to Postgres and Elastic search. We do the migration and then do another deployment with the feature flag removed.
> 
> I don't like the endpoint too much, it means that one instance is stuck doing the migration instead of serving the users. I'll prefer to have a separate binary that we can deploy, in one or multiple instances) alongside the instances that handle the migration. Any opinion on this @Robert-Steiner @joecks ?

i'm happy to trigger the migration in a different way, i wasn't to sure how to do that best. what do you mean exactly by `We could do a deployment where we enable only the functions that write to Postgres and Elastic search.`? wouldn't that also lead to the situation where the user can't make any calls at all since no endpoints are enabled?

web-api/src/storage/postgres.rs
janpetschexain: pushed the lock addtions

acrrd: Ok, I'm convinced, but let's keep these numbers low.

Robert-Steiner:
From the infrastructure point of view: how will the migration process look like?
Afaiu we build a new image and deploy it. The services in that image can work with both the old and the already migrated data. Those services will handle the user requests.

For the migration part we spawn the same service in another instance that does not handle any user requests but only the migration. The service exposes a `/migration` endpoint that we (the infrastructure squad?) call (once via a GET?) to trigger the migration. Once the migration is done we remove this migration service. We build a new version of the service that removes the feature flag and deploy it. 
Is the above description correct?

> I don't like the endpoint too much, it means that one instance is stuck doing the migration instead of serving the users. I'll prefer to have a separate binary that we can deploy, in one or multiple instances) alongside the instances that handle the migration.

For us it would be easiest if we just get a docker image with a binary that does the migration. That means we deploy a one-off task that starts, performs the migration and exists. Would that be possible to do via a configuration flag or something else. If it requires to much work, we can also go with the endpoint. Maybe for later it makes sense to move the migration logic into its own binary.

Robert-Steiner:
Do we know how long a migration takes for let's say 10k documents?

rustonaut:
Why did we add `FOR UPDATE` to most endpoints?

I'm not sure if it's because I can't concentrate today, but I don't see how this is an improvement? @janpetschexain

janpetschexain:
> Why did we add `FOR UPDATE` to most endpoints?
> 
> I'm not sure if it's because I can't concentrate today, but I don't see how this is an improvement? @janpetschexain

because of this discussion above: https://github.com/xaynetwork/xayn_discovery_engine/pull/832#discussion_r1122045187

web-api/src/storage.rs
```diff
@@ -180,20 +182,31 @@ pub(crate) trait Tag {
 pub struct Config {
     elastic: elastic::Config,
     postgres: postgres::Config,
+    #[cfg(feature = "ET-3837")]
+    is_migrated: bool,
```
rustonaut, line 186: In general such information would be stored in the database instead of the config.

(Through not as a bool, but more like a migration version number/state.)

Instead of then using a endpoint you could automatically start the migration in the background when loading a binary/service containing the code, use the migration state to make sure that only n (here e.g. 1) instance slowly does the migration, have handling in case the instance with the migration disappears (in the future, not for now) and just poll the state every few minutes until the migrations are done to the expected state.

As the migrations only go one way you still might use a indicator in the app state to avoid additional database queries.

janpetschexain:
> For the migration part we spawn the same service in another instance that does not handle any user requests but only the migration. The service exposes a `/migration` endpoint that we (the infrastructure squad?) call (once via a GET?) to trigger the migration. Once the migration is done we remove this migration service. We build a new version of the service that removes the feature flag and deploy it. Is the above description correct?

yes, that's the current idea.

> For us it would be easiest if we just get a docker image with a binary that does the migration. That means we deploy a one-off task that starts, performs the migration and exists. Would that be possible to do via a configuration flag or something else. If it requires to much work, we can also go with the endpoint. Maybe for later it makes sense to move the migration logic into its own binary.

i'll have a look how this can be done.

> Do we know how long a migration takes for let's say 10k documents?

currently it's configurable to process n documents per t seconds, of course limited by the database throughput.

web-api/src/storage/postgres.rs
rustonaut: I don't thing the locks prevent that.

Stretching the transaction across the ES calls just makes sure that you can't have partial success/failure.

But it doesn't make sure that others reading data from elastic search do get the update before the postgres transaction concludes.

Furthermore to get what we currently get shouldn't it be enough to `FOR UPDATE` lock in the migration query before fetching the properties from elastic search instead of acquiring the lock in (nearly?) every place?

Also I'm not sure anyone used the property updates, wouldn't it have been simpler to deploy a version which stores all new documents in PG already and fetches the old way and has a switch to switch between old-property-updates, no-property-updates, new-property-updates, then disable updates temporary run a separate "scrip" to migrate all documents fastly and then switch over to new updates? Sure that would have been a bit of downtime for (property updates only) but AFIK with how things are currently used that would be much much simpler and reliable then a truly fluent gradual background update. 

@Robert-Steiner do we have statistics about how often the property update endpoints are used, if at all?

Robert-Steiner:
How urgent it this migration? I'm asking because we also need to prepare stuff on our side

Robert-Steiner:
How urgent it this migration? I'm asking because we also need to prepare stuff on our side

joecks:
An issue with a one-time off task, is the health checks on services. Normally when a service dies the health checks should restart it, which would cause kind of a loop.

janpetschexain:
> How urgent it this migration? I'm asking because we also need to prepare stuff on our side

not entirely sure, i guess not super urgent, but we need it at some point to only have a subset of docs in es (ET-4089). this doesn't block that though, as i'm just stacking one pr upon another.

> An issue with a one-time off task, is the health checks on services. Normally when a service dies the health checks should restart it, which would cause kind of a loop.

even if it gets restarted, it doesn't do the whole migration again, because the internal loop checks for unmigrated docs (based on missing data) and only attempts to migrate those. so if it gets restarted and migration has already finished before, then the loop should exit immediately successfully.

web-api/src/storage/postgres.rs
janpetschexain: i'd guess that the properties endpoint is not much used (if at all).

wdyt about the following: under the feature flag new data gets stored already to pg (in addition to es), fetching still behaves the old way from es. then we can run the migration/data copy, without the need to decide where to fetch from for each request. in the following pr which removes the feature flag we switch the fetching from pg to es. i'll push an update later and ping you.

rustonaut: seems good :+1:

Robert-Steiner: > @Robert-Steiner do we have statistics about how often the property update endpoints are used, if at all?
unfortunately, we currently have no insight into corresponding endpoints

janpetschexain: i pushed the simplifications, most of the removed code moved into the following pr #833, except for the fallback logic which was removed completely.

web-api/src/storage.rs
```diff
@@ -188,14 +190,23 @@ impl Config {
     pub(crate) async fn setup(&self) -> Result<Storage, SetupError> {
         let elastic = self.elastic.setup_client()?;
         let postgres = self.postgres.setup_database().await?;
-
-        Ok(Storage { elastic, postgres })
+        #[cfg(feature = "ET-3837")]
+        let is_migrated = RwLock::new(false);
```
rustonaut, line 194: AFIK this seem to be currently not doing anything as it's only set by one specif binary before exiting.

rustonaut: okay I need to check my network time sync settings, it just showed that your message _will_ be send in 14 seconds, with a nice count down to now :rofl:

rustonaut (approved):
lgtm

janpetschexain:
bors merge

PR #833, opened 2023-03-01
ET-4141 remove migration feature flag

janpetschexain:
**Reference**

- [ET-4141]
- requires #832

**Summary**

- remove feature flag `ET-3837` & dead code
- drop temporary postgres defaults


web-api/src/storage/postgres.rs
```diff
@@ -493,72 +555,235 @@ impl Database {
...
+    }
+
+    async fn delete(
+        &self,
+        document_id: &DocumentId,
+        property_id: &DocumentPropertyId,
+    ) -> Result<Option<Option<()>>, Error> {
+        let mut tx = self.postgres.pool.begin().await?;
+
+        let deleted = sqlx::query(
```
rustonaut, line 759: :pushpin: I have been wondering if delete shouldn't be more idempotent, i.e. deleting a property which doesn't exist should IMHO not fail (if the document doesn't exist it should).

rustonaut (approved):
lgtm, I thing we might want to make the deletion more idempotent but thats outside of the scope of this PR

janpetschexain:
bors merge

janpetschexain:
bors merge

PR #834, opened 2023-03-01
[Merged by Bors] - ET-4094 implement semantic search with inline user

rustonaut:
Implement the updated semantic search which can accept the user data inline instead of an user id.

The option to toggle exuding seen documents is in a follow up PR.

# References

- issue: [ET-4094]
- story: [ET-4092]
- based on: #831

[ET-4094]: https://xainag.atlassian.net/browse/ET-4094?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ET-4092]: https://xainag.atlassian.net/browse/ET-4092?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/web-api/src/personalization/knn.rs
+++ b/web-api/src/personalization/knn.rs
@@ -27,7 +27,7 @@ use crate::{
     Error,
 };
 
-/// KNN search based on Centers of Interest;
+/// KNN search based on Centers of Interest.
 pub(super) struct CoiSearch<I, J> {
     pub(super) interests: I,
     pub(super) excluded: J,
--- a/web-api/src/personalization/knn.rs
+++ b/web-api/src/personalization/knn.rs
@@ -27,7 +27,8 @@ use crate::{
     Error,
 };
 
-pub(super) struct Search<I, J> {
+/// KNN search based on Centers of Interest;
+pub(super) struct CoiSearch<I, J> {
     pub(super) interests: I,
     pub(super) excluded: J,
     pub(super) horizon: Duration,
@@ -37,7 +38,7 @@ pub(super) struct Search<I, J> {
     pub(super) time: DateTime<Utc>,
 }
 
-impl<'a, I, J> Search<I, J>
+impl<'a, I, J> CoiSearch<I, J>
 where
     I: IntoIterator,
     <I as IntoIterator>::IntoIter: Clone + Iterator<Item = &'a PositiveCoi>,
@@ -141,7 +142,7 @@ mod tests {
     #[tokio::test]
     async fn test_search_knn_documents_for_empty_cois() {
         let storage = Storage::default();
-        let documents = Search {
+        let documents = CoiSearch {
             interests: &[],
             excluded: &[],
             horizon: CoiConfig::default().horizon(),
--- a/web-api/src/personalization/routes.rs
+++ b/web-api/src/personalization/routes.rs
@@ -23,7 +23,7 @@ use chrono::{DateTime, Utc};
 use itertools::Itertools;
 use serde::{Deserialize, Serialize};
 use tracing::instrument;
-use xayn_ai_coi::CoiSystem;
+use xayn_ai_coi::{CoiConfig, CoiSystem};
 
 use super::{
     knn,
@@ -384,14 +384,16 @@ pub(crate) async fn personalize_documents_by(
 #[derive(Deserialize)]
 #[serde(untagged)]
 enum UnvalidatedInputUser {
-    Ref(String),
+    Ref {
+        id: String,
+    },
     Inline {
         history: Vec<UnvalidatedHistoryEntry>,
     },
 }
 
 enum InputUser {
-    Ref(UserId),
+    Ref { id: UserId },
     Inline { history: Vec<HistoryEntry> },
 }
 
@@ -402,7 +404,7 @@ impl UnvalidatedInputUser {
         warnings: &mut Vec<Warning>,
     ) -> Result<InputUser, Error> {
         Ok(match self {
-            Self::Ref(id) => InputUser::Ref(id.try_into()?),
+            Self::Ref { id } => InputUser::Ref { id: id.try_into()? },
             Self::Inline { history } => InputUser::Inline {
                 history: validate_history(history, config, warnings, Utc::now())?,
             },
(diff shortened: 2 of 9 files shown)


web-api/src/personalization/stateless.rs
```diff
@@ -0,0 +1,369 @@
...
+use super::PersonalizationConfig;
+use crate::{
+    error::{common::HistoryTooSmall, warning::Warning},
+    models::{DocumentId, DocumentTag},
+    storage::{self, TagWeights},
+    Error,
+};
+
+#[derive(Deserialize)]
+pub(super) struct UnvalidatedHistoryEntry {
```
frankpepermans, line 32: 💅 Just a suggestion, feel free to dismiss, unvalid is not a correct word (but we get the meaning of course).
Consider `CandidateHistoryEntry` maybe

rustonaut: The thing is it's not really a candidate it's a fixed entry we will take, it's just not yet clear if the user gave us malformed data.

Tbh. I'm not support happy about how things are handled wrt. validation  in general (not specific to my code).

The general problem is that when type safety and contextual enforcement of soft validity constraints overlap it can easily lead to very verbose code, and it did in our case.

In more detail: For many of the types like e.g. `DocumentId` we "soft" enforce the id's validity. What I mean with it is that if we get the id from a database (PG,ES) we take _whatever we get without validation_. Because there is just no sane way to handle the situation where a id in out system is "malformed", we can't just stop operating our service nor can we just randomly delete it or anything. (Side note: Because of stuff like that it's important to always handle encoding of queriers appropriately even if that shouldn't be needed given our constraints for what is a valid id.)

But when getting a id as input we want to always make sure it is valid.

The problem here is there is no (non magic) way to implicitly  pass the "context" under which the `DocumentId` is created when creating it as part of other structs through a serde `Deserializer` (or `FromRow` in PG, etc.).

In case of PG we kinda can just argue it's only used for internal databases (in our use-case) so the context is always the same so we can just implicitly assume it.

But for data from e.g. ES and inputs from JSON this doesn't work as they all go through serde/JSON.

So we had a few ways to change this:

- have some alternative serde impl for some of the interfaces using a wrapper and "external" annotations and mirror types

- parse a different type and validate by hand (easier the the first point, it's what we did)

- use some magic to pass in a context in some way

Additionally the validation is dependent on the config, too in some cases.

And while we went with the second method I'm not sure it was the best idea,  through also there is no off the shelf solution I'm aware of for the 3rd solution and creating and maintaining one isn't the best investment of time for the current situation :sweat_smile:

web-api/src/personalization/knn.rs
```diff
@@ -27,7 +27,8 @@ use crate::{
     Error,
 };
 
-pub(super) struct Search<I, J> {
+/// KNN search based on Centers of Interest;
```
janpetschexain, line 30: :art: punctuation
```suggestion
/// KNN search based on Centers of Interest.
```

web-api/src/personalization/stateless.rs
```diff
@@ -0,0 +1,369 @@
...
+}
+
+pub(super) struct LoadedHistoryEntry {
+    pub(super) id: DocumentId,
+    pub(super) timestamp: DateTime<Utc>,
+    pub(super) embedding: NormalizedEmbedding,
+    pub(super) tags: Vec<DocumentTag>,
+}
+
+/// Given an iterator over the history from _newest_ to oldest calculates user interests and tag weights.
```
janpetschexain, line 118: ❓ shouldn't it be the opposite? if we have an iterator from newest to oldest history, then the shifts to the cois in `log_positive_user_reaction()` for newer interactions will be overwritten by shifts for older interactions.

web-api/src/personalization/routes.rs
```diff
@@ -464,27 +382,119 @@ pub(crate) async fn personalize_documents_by(
 }
 
 #[derive(Deserialize)]
-struct SemanticSearchQuery {
+struct UnvalidatedInputUser {
+    #[serde(default)]
+    id: Option<String>,
+    #[serde(default)]
+    history: Option<Vec<UnvalidatedHistoryEntry>>,
+}
```
janpetschexain, line 390: :art: since all fields are deserialized as defaults, we can just annotate it for the struct instead
```suggestion
#[derive(Default, Deserialize)]
#[serde(default)]
struct UnvalidatedInputUser {
    id: Option<String>,
    history: Option<Vec<UnvalidatedHistoryEntry>>,
}
```

web-api/src/personalization/routes.rs
```diff
@@ -464,27 +382,119 @@ pub(crate) async fn personalize_documents_by(
...
+        warnings: &mut Vec<Warning>,
+    ) -> Result<InputUser, Error> {
+        Ok(match (self.id, self.history) {
+            (Some(id), None) => InputUser::Ref { id: id.try_into()? },
+            (None, Some(history)) => InputUser::Inline {
+                history: validate_history(history, config, warnings, Utc::now(), true)?,
+            },
+            _ => {
+                return Err(BadRequest::from(
+                    "personalize.user must haver _either_ an `id` or a `history` field",
```
janpetschexain, line 410: :art: typo
```suggestion
                    "personalize.user must have _either_ an `id` or a `history` field",
```

web-api/src/personalization/routes.rs
```diff
@@ -495,29 +505,29 @@ struct SemanticSearchResponse {
 
 async fn semantic_search(
     state: Data<AppState>,
-    document_id: Path<String>,
-    query: Query<SemanticSearchQuery>,
+    query: Json<UnvalidatedSemanticSearchQuery>,
```
janpetschexain, line 508: :art: we could destructure it already here to avoid the `.into_inner()` call below, like we do in other routes
```suggestion
    Json(query): Json<UnvalidatedSemanticSearchQuery>,
```

janpetschexain (commented):
looks good, some questions left

web-api/src/personalization/stateless.rs
rustonaut: yes you are right, this was already wrong in the previous version :see_no_evil:

rustonaut: turns out it was only a bug in the documentation :tada:

janpetschexain (approved):
lgtm

rustonaut:
bors merge

PR #837, opened 2023-03-06
[Merged by Bors] - ET-4105 document candidates

janpetschexain:
**Reference**

- [ET-4105]
- https://github.com/xaynetwork/xayn_discovery_engine/pull/827#discussion_r1114723437
- requires #833
- followed by #845

**Summary**

- add story feature flag `ET-4089`
- add optional `is_candidate` to the `document` table & `IngestedDocument`
- simplify occurences of `(IngestedDocument, NormalizedEmbedding)` to `IngestedDocumet` by moving the embedding into the doc, the tuple was a relict from before the split of the storage types and the engine types; this also allows us to easily extend the doc type, for example with the `is_candidate` field
- separate storage errors from application data errors by returning failed ids in the storage layer and handling errors in the route logic layer; follows from this comment: https://github.com/xaynetwork/xayn_discovery_engine/pull/827#discussion_r1114723437, as i touched the coded in this pr anyways i just made the change but if required we can also split this out into a separate pr

[ET-4105]: https://xainag.atlassian.net/browse/ET-4105?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


web-api/src/storage/elastic.rs
```diff
@@ -109,12 +114,48 @@ struct BulkItemResponse {
     error: Value,
 }
 
+impl BulkItemResponse {
+    fn is_success_status(&self, allow_not_found: bool) -> bool {
```
janpetschexain, line 118: this and the following function just moved up from the free-floating functions below

web-api/src/storage/elastic.rs
```diff
@@ -219,6 +260,130 @@ impl Client {
         self.query_with_bytes(url, post_data).await
     }
 
+    pub(super) async fn get_by_embedding<'a>(
```
janpetschexain, line 263: this and the following functions just moved out of the trait impl

web-api/src/storage.rs
```diff
@@ -93,15 +72,14 @@ pub(crate) trait Document {
         params: KnnSearchParams<'a, impl IntoIterator<Item = &'a DocumentId>>,
     ) -> Result<Vec<PersonalizedDocument>, Error>;
 
-    async fn insert(
-        &self,
-        documents: Vec<(IngestedDocument, NormalizedEmbedding)>,
-    ) -> Result<(), InsertionError>;
+    /// Inserts the documents and reports failed ids.
+    async fn insert(&self, documents: Vec<IngestedDocument>) -> Result<Vec<DocumentId>, Error>;
```
rustonaut, line 76: :wrench:  Documents failing is an error. Why do we return it now as a non error? Also why return it as a vec which looks like it could e.g. be the list of actually inserted documents or something like that?

web-api/src/storage.rs
```diff
@@ -93,15 +72,14 @@ pub(crate) trait Document {
...
 
+    /// Deletes the documents and reports failed ids.
     async fn delete(
         &self,
         ids: impl IntoIterator<IntoIter = impl Clone + ExactSizeIterator<Item = &DocumentId>>,
-    ) -> Result<(), DeletionError>;
+    ) -> Result<Vec<DocumentId>, Error>;
```
rustonaut, line 82: :wrench:  Same as the previous comment.

web-api/src/storage/memory.rs
```diff
@@ -321,50 +323,49 @@ impl storage::Document for Storage {
         Ok(documents)
     }
 
-    async fn insert(
-        &self,
-        documents_embeddings: Vec<(IngestedDocument, NormalizedEmbedding)>,
-    ) -> Result<(), InsertionError> {
-        if documents_embeddings.is_empty() {
-            return Ok(());
+    async fn insert(&self, new_documents: Vec<IngestedDocument>) -> Result<Vec<DocumentId>, Error> {
```
rustonaut, line 326: :wrench: Same as before, why make an error a Ok value? One which looks like something else, too.

web-api/src/storage/memory.rs
```diff
@@ -321,50 +323,49 @@ impl storage::Document for Storage {
...
-        embeddings.retain(|id, _| !ids.contains(id));
-        documents.1 = Embeddings::borrowed(embeddings);
+        let mut ids = ids.into_iter().collect::<HashSet<_>>();
         interactions.retain(|_, interactions| {
             interactions.retain(|(id, _)| !ids.contains(id));
             !interactions.is_empty()
         });
+        documents.0.retain(|id, _| !ids.contains(id));
+        let mut embeddings = mem::take(&mut documents.1).into_heads().map;
+        embeddings.retain(|id, _| !ids.remove(id));
```
rustonaut, line 365: :wrench: 

```suggestion
        embeddings.retain(|id, _| !ids.contains(id));
```

If we anyway drop `ids` a moment later we should use `contains` here I think as it more represents what we do and earlier removing is not an optimization either. Or did I miss something?

web-api/src/storage/postgres.rs
```diff
@@ -235,10 +263,11 @@ impl Database {
         Ok(())
     }
 
+    #[cfg(not(feature = "ET-4089"))]
     pub(super) async fn delete_documents(
         &self,
         ids: impl IntoIterator<IntoIter = impl Clone + ExactSizeIterator<Item = &DocumentId>>,
-    ) -> Result<(), DeletionError> {
+    ) -> Result<Vec<DocumentId>, Error> {
```
rustonaut, line 270: :wrench:  same as with the other Result error/Ok changes

web-api/src/storage/memory.rs
janpetschexain: we don't drop them, but return the remaining ids a few lines later. the leftover ids are the ones which are not present in the db but were requested to be deleted, which aligns it with the es/pg db impl.

rustonaut (approved):
lgtm

frankpepermans:
Bors merge

rustonaut (approved):

janpetschexain:
bors merge

janpetschexain:
bors ping

janpetschexain:
bors merge

janpetschexain:
bors merge

janpetschexain:
we did a manual merge because it's currently impossible to `bors merge` a reopened pr a second time after it has already been merged once.

PR #838, opened 2023-03-06
[Merged by Bors] - ET-4185 Return an empty Result-set when k=0

frankpepermans:
We encountered an issue with KNN-search, if the K-value is zero, then the storage provider throws an exception.
Therefore, simply return an empty Result instead.


rustonaut (approved):
lgtm

frankpepermans:
bors merge

PR #839, opened 2023-03-07
ET-4104 input user history limits

rustonaut:
- [x] bump history limit to 200
- a new limit for how much of the history is used for CoI computation
  - [x] add new config
  - [x] use new config
  
**References:**

- Issue: [ET-4104]
- Story: [Et-4092]
- Based On: #834


janpetschexain (approved):
lgtm

btw it's a bit annoying that the following pr has already been merged into this pr, as it makes it harder to review and also the merge history of the two prs will be erased into one commit once this goes into `main`.

PR #843, opened 2023-03-07
ET-XXXX-make-document-id-required

rustonaut:
When the text search is added we can no longer mark it as required as the way to mark either required will make our tooling fall over.

But until then we can mark `id` as required.


rustonaut:
became irrelevant through ET-4149/#846

PR #845, opened 2023-03-07
[Merged by Bors] - ET-4106 set document candidates

janpetschexain:
**Reference**

- [ET-4106]
- [ET-4107]
- [ET-4126]
- requires #837
- followed by #849

**Summary**

- define api spec for setting candidates
- impl setting candidates for storages & add route to the backoffice

[ET-4106]: https://xainag.atlassian.net/browse/ET-4106?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ET-4107]: https://xainag.atlassian.net/browse/ET-4107?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ET-4126]: https://xainag.atlassian.net/browse/ET-4126?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/web-api/openapi/back_office.yaml
+++ b/web-api/openapi/back_office.yaml
@@ -93,6 +93,25 @@ paths:
         '400':
           $ref: './responses/generic.yml#/BadRequest'
 
+  /documents/candidates:
+    put:
+      tags:
+        - back office
+      summary: Set the documents considerred for recommendations.
+      description: Set the documents considerred for recommendations.
+      operationId: replaceDocumentCandidates
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/DocumentCandidatesRequest'
+      responses:
+        '204':
+          description: successful operation
+        '400':
+          $ref: './responses/generic.yml#/BadRequest'
+
   /documents/{document_id}:
     parameters:
       - $ref: './parameters/path/document_id.yml'
@@ -307,3 +326,19 @@ components:
           maxItems: 1000
           items:
             $ref: './schemas/document.yml#/DocumentId'
+    DocumentCandidate:
+      type: object
+      required: [id]
+      properties:
+        id:
+          $ref: './schemas/document.yml#/DocumentId'
+    DocumentCandidatesRequest:
+      type: object
+      required: [documents]
+      properties:
+        documents:
+          type: array
+          minItems: 0
+          maxItems: 1e+6
+          items:
+            $ref: '#/components/schemas/DocumentCandidate'
--- a/web-api/openapi/back_office.yaml
+++ b/web-api/openapi/back_office.yaml
@@ -97,8 +97,8 @@ paths:
     put:
       tags:
         - back office
-      summary: Set the documents considerred for recommendations.
-      description: Set the documents considerred for recommendations.
+      summary: Set the documents considered for recommendations.
+      description: Set the documents considered for recommendations.
       operationId: replaceDocumentCandidates
       requestBody:
         required: true
@@ -339,6 +339,5 @@ components:
         documents:
           type: array
           minItems: 0
-          maxItems: 1e+6
           items:
             $ref: '#/components/schemas/DocumentCandidate'
--- a/web-api/src/ingestion/routes.rs
+++ b/web-api/src/ingestion/routes.rs
@@ -25,6 +25,8 @@ use tokio::time::Instant;
 use tracing::{error, info, instrument};
 
 use super::AppState;
+#[cfg(feature = "ET-4089")]
+use crate::error::common::FailedToSetSomeDocumentCandidates;
 use crate::{
     error::{
         application::WithRequestIdExt,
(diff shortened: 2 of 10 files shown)


web-api/src/storage/postgres.rs
```diff
@@ -372,6 +461,118 @@ impl Database {
...
+                builder
+                    .reset()
+                    .push_tuple(ids.by_ref().take(Self::BIND_LIMIT))
+                    .push(" RETURNING document_id;")
+                    .build()
+                    .try_map(|row| DocumentId::from_row(&row))
+                    .fetch_all(&mut tx)
+                    .await?,
+            );
+        }
```
rustonaut, line 553: :wrench: 

the unsetting part will not work as you can't split a `NOT IN` into multiple queies as on each split documents from the other splits will be "not in this split".

we probably could do `UPDATE document SET is_candidate = FALSE` followed by  a split `UPDATE document SET is_candidate = TRUE WHERE document_id IN (?...)`.

Besides that if the number of documents is much smaller then the number of documents which are candidates we probably should create a INDEX over all candidates and then delete all from index and set new through I don't know for sure I would need to check out how postges handles this kind of operations  performance wise.

web-api/openapi/back_office.yaml
```diff
@@ -303,3 +326,19 @@ components:
...
+        id:
+          $ref: './schemas/document.yml#/DocumentId'
+    DocumentCandidatesRequest:
+      type: object
+      required: [documents]
+      properties:
+        documents:
+          type: array
+          minItems: 0
+          maxItems: 1e+6
```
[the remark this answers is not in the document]
janpetschexain: exactly, the linter forced me to add a `maxItems`. i set it to `1e6` mostly arbitrarily, since json doesn't allow for infinity. is there a way to ignore the linter for this item? or is there a way to say "unlimited but bound by general size limits"?

web-api/src/ingestion/routes.rs
```diff
@@ -37,17 +39,22 @@ use crate::{
...
 
 pub(super) fn configure_service(config: &mut ServiceConfig) {
+    config.service(
+        web::resource("/documents")
+            .route(web::post().to(new_documents.error_with_request_id()))
+            .route(web::delete().to(delete_documents.error_with_request_id())),
+    );
+    #[cfg(feature = "ET-4089")]
+    config.service(
+        web::resource("/documents/candidates")
```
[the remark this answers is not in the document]
janpetschexain: i would say so too. we could do a query on the tenants to be sure that nobody uses a document id called `"candidates"`, but that's quite unlikely. after the merge it's the user's fault if they use reserved key words for the document ids. should we add this to the regex to make it a hard error?

web-api/src/storage/postgres.rs
janpetschexain: good point, damn bind limits. i'll add the index and update the query.

janpetschexain: i also changed the es delete to a retain, this simplifies the code overall and also keeps es in a cleaner state.

web-api/src/storage/postgres.rs
```diff
@@ -453,6 +463,88 @@ impl Database {
...
+        &self,
+        ids: impl IntoIterator<IntoIter = impl Clone + ExactSizeIterator<Item = &DocumentId>>,
+    ) -> Result<(Vec<DocumentId>, Vec<IngestedDocument>, Warning<DocumentId>), Error> {
+        let mut tx = self.pool.begin().await?;
+
+        let mut ingestable = ids.into_iter().collect::<HashSet<_>>();
+        let unchanged = sqlx::query_as::<_, DocumentId>(
+            "UPDATE document SET
+            is_candidate = FALSE
+            RETURNING document_id;",
```
rustonaut, line 477: :wrench:  If you want to use it with `RETURNING` you still need to add a where clause, this will always return all documents.

```suggestion
            "UPDATE document SET
            is_candidate = FALSE
            WHERE is_candidate
            RETURNING document_id;",
```

rustonaut: :wrench: Also that now would return _all_ document which previously had been candidates  independent of weather or not they still are candidates.

So you still would need to set diff that with the set of new candidates.

janpetschexain: no, this only returns ids where the flag actually changed, as the [pg docs state](https://www.postgresql.org/docs/current/sql-update.html):
> The optional RETURNING clause causes UPDATE to compute and return value(s) based on each row actually updated.

janpetschexain: > So you still would need to set diff that with the set of new candidates.

this already happens in the filtermap a few lines below the query.

rustonaut: > The optional RETURNING clause causes UPDATE to compute and return value(s) based on each row actually updated.

You update all row, some already having the same value you update it to and postgres internally being able to optimize the writes away due to an index doesn't change the semantics of the query/outcome.

janpetschexain: i see, i misunderstood the semantics in the docs, my bad, i'll update it.

PR #846, opened 2023-03-09
[Merged by Bors] - ET-4149 Free text semantic search impl

frankpepermans:
Add query as a new InputDocument type.
Creates embedding on the fly, if we don't get an existinf Document, but rather a free text query.


web-api/src/personalization/routes.rs
```diff
@@ -465,12 +465,24 @@ impl UnvalidatedSemanticSearchQuery {
...
+                "either id or query must be present in the request, but both were found",
+            )
+            .into())
+        } else if let Some(id) = self.id {
+            Ok(InputDocument::Ref(id.try_into()?))
+        } else if let Some(query) = self.query {
+            Ok(InputDocument::Query(query))
+        } else {
+            Err(BadRequest::from("either id or query must be present in the request").into())
+        }
```
rustonaut, line 485: :art:  for this kind a if let matching a `match` is more appropriate, it has the nice side effect that it enforces at compiler time that you match on all combinations.

```suggestion
        match (self.id, self.query) {
            (Some(_), Some(_)) => Err(BadRequest::from(
                "either id or query must be present in the request, but both were found",
            )
            .into()),
            (None, Some(query)) => Ok(InputDocument::Query(query)),
            (Some(id), None) => Ok(InputDocument::Ref(id.try_into()?)),
            (None, None) => {
                Err(BadRequest::from("either id or query must be present in the request").into())
            }
        }
```

What the appropriate order is on the other hand is preference, some prefer a ordering based on the match patter, some would order the non error branches above the error branches and some would do that the other way around. Ordering based on the match pattern has the benefit that it can be applied in all contexts, so I often prefer that.

web-api/openapi/schemas/document.yml
```diff
@@ -32,6 +32,14 @@ DocumentTag:
   maxLength: 256
   example: "tag"
 
+DocumentSearchQuery:
+  description: A search query can be any non-empty, UTF-8-encoded string which doesn't contain a zero byte.
+  type: string
+  pattern: ".+"
+  minLength: 1
+  maxLength: 512
+  example: "what are the latest developments surrounding transformer models?"
```
rustonaut, line 41: :question: I'm not sure how well this kind of questions work with the embedding? Id not so well maybe we should use a different example.

.github/workflows/web-service.yml
```diff
@@ -91,4 +98,4 @@ jobs:
 
       - name: docker push
         id: docker-push
-        run: docker push "${IMAGE_NAME}"
+        run: docker push "${IMAGE_NAME}"
```
rustonaut, line 101: :wrench: white space error

web-api/src/personalization/routes.rs
frankpepermans: Yup!
I'll change to your suggestion

web-api/openapi/schemas/document.yml
frankpepermans: Was thinking about a plain "Lorem ipsum...", but updated the example to be more like a headline now

.github/workflows/web-service.yml
frankpepermans: Windows!

rustonaut (approved):
lgtm

PR #859, opened 2023-03-22
[Merged by Bors] - FIX: drop integration test db with force

janpetschexain:
**Summary**

- use `FORCE` to avoid race-condition-ish errors while dropping test dbs


janpetschexain:
bors merge

PR #860, opened 2023-03-23
ET-4186 tenant id extension

rustonaut:
Adds a call context middleware replacing the tracing/logging middleware which

- extracts the TenantId from a header (currently `X-Tenenat-Id` that is prone to needing change)
- like the previous middleware generates a request id and setups up the logging context (including the tenant id)
- adds the context (a struct containing `TenantId` and `RequestId` as a middleware extension)
- adds the context as a task local storage variable
  -  see comment in file about why, but it mainly allows us to have much simpler code
- adds a `call_context` configuration scope which allows specifying a `legacy_tenant`, if done so the tenant id header is no longer required and will default to the legacy tenant and if given must be the same as the legacy tenant ( this is for migration purpose)

**References:**

- Issue: [ET-4186]
- Story: [ET-4172]


integration-tests/src/lib.rs
```diff
@@ -256,6 +278,7 @@ async fn setup_web_dev_test_context(
         id,
         postgres: postgres.unwrap(),
         elastic_search: elastic_search.unwrap(),
+        tenant_id: Uuid::new_v4().to_string(),
```
rustonaut, line 281: Follow up PRs will set-up a tenant db (schema) etc.

web-api/src/error/application.rs
```diff
@@ -122,116 +106,3 @@ pub(crate) struct Unimplemented {
 }
 
 impl_application_error!(Unimplemented => INTERNAL_SERVER_ERROR);
-
```
rustonaut, line 125: All of this is no longer needed with the api call context in task local storage, we could also later one start using `#[get("/path/")]` from actix_web.

web-api/src/middleware/call_context.rs
```diff
@@ -0,0 +1,220 @@
...
+
+pub(crate) struct ApiCallContext {
+    #[allow(unused)]
+    pub(crate) tenant_id: TenantId,
+    pub(crate) request_id: RequestId,
+}
+
+tokio::task_local! {
+    static API_CALL_CONTEXT: Arc<ApiCallContext>;
+}
```
rustonaut, line 45: I originally thought we need this for the db handling but it turns out we do not.

Still it does:

- remove all the hard to understand "request id" injection type magic
- mages the (follow up) db code much simpler (saving 500+ line changes or so)

There is no drawback (with how we do things) for the error handling usage, but there is
one drawback for the Database usage, `tokio::spawn` doesn't carry over the context so
if we want to clone storage and pass it to a spawned task and use it there we need to take
additional care to also carry over the context, this is currently not the case anywhere in our 
code and writing a `spawn_with_context` function isn't too hard.

Given how much task locals makes things simpler I decided to go with it for now and we might
later one change the DB handling to work more seamless with spawn if we  need it.

rustonaut: I just realized I have to implement the more complex DB version earlier (due to the details of the connection pools `before_acquire` error handling). I still would keep it in task local storage anyway.

web-api/src/middleware/call_context.rs
```diff
@@ -0,0 +1,227 @@
...
+        Self(Uuid::nil())
+    }
+}
+
+/// Configuration for handling tenants.
+#[derive(Debug, Default, Deserialize, Serialize)]
+#[serde(default)]
+pub struct Config {
+    /// Mocks the tenant id for legacy single tenant deployments
+    pub(crate) legacy_tenant: Option<TenantId>,
```
acrrd, line 150: I'm not convinced by this. This approach means that all the changes to the database and configuration in our environments must be done before this is merged.
We can just do nothing if the tenant id `None`

acrrd (commented):
I gave only a quick look.
I was quite confused by the name `call_context`, many web frameworks use `RequestContext` or `ReqContext` and I think we should use one of these as well.

web-api/src/middleware/call_context.rs
rustonaut: > This approach means that all the changes to the database and configuration in our environments must be done before this is merged.

No it doesn't, it only means that a single config option has to be set per-tenant before merging this.

We can also set it initially to an arbitrary id and later set it to ids we care about, this won't be a problem.

The follow up PR creates a role for the legacy tenant on the fly on startup if it's missing, but in difference to normal multi tenant that role will still use the `public` schema.

This also means that for legacy tenants we still can change the ids on the fly as they don't appear anywhere in the data for now (through they do appear in the logs).

rustonaut: done now config defaults to legacy tenant with an  `Uuid::nill()` as tenant id

other ids still can be set, legacy tenant can be disabled by setting `use_legacy_tenant = false` (`use_legacy_tenant = true` is the default and `use_legacy_tenant = "uuid...."` is for when we start giving non-multi-tenant tenants properly tenant ids and future migrations).

rustonaut:
I renamed it to `RequestContext`.

For now I will keep the `use_legacy_tenant` config option in the `[request_context]` config section, it's doesn't anymore perfectly fit in there. I will change it  in the PR which add the tenant management/migration rust API as it should be part of that config. But that config doesn't exist yet and I don't know yet how exactly it will be named, how it will be structured etc.

acrrd:
> For now I will keep the `use_legacy_tenant` config option in the `[request_context]` config section, it's doesn't anymore perfectly fit in there. I will change it in the PR which add the tenant management/migration rust API as it should be part of that config. But that config doesn't exist yet and I don't know yet how exactly it will be named, how it will be structured etc.

We shouldn't have any configuration for the legacy tenant at this point. We keep a `Option<TenantId>` in the context and if it is `None` we return a storage that works like the current one (so we do nothing), while when we have the id we setup the store to use it. This code must be backward compatible without the need to modify the current infrastructure or configuration

rustonaut:
> This code must be backward compatible without the need to modify the current infrastructure or configuration

I think there is some misunderstanding, after the last push the PR is now fully backward compatible with the current infrastructure without changing any configuration options.

> We shouldn't have any configuration for the legacy tenant at this point. 

It's for now more a config options to disable legacy teant mode, i.e. to enable multi tenant mode.

- currently `use_legacy_tenant` defaults to `true`
- deployments of multi tenant would set `use_legacy_tenant` to `false`
- once we use the same kind of infrastructure for legacy tenants as we do for multi tenant setup (i.e. control plane etc.) we would set it to `use_legacy_tenant = "uuid"` and that would "just" work (as long as you don't idk. try to create additional tenants in the silo of a legacy tenant)  
- if that works all well and fine we then could migrate the db, if we want to, and remove the `use_legacy_tenant` option

Not separating between legacy tenant setups and multi tenant setups is IMHO not a good idea as it can lead to all kind of painful bugs, and only knowing when a request comes in that there is a legacy tenant would also introduce a bunch of complications to other code later down the line.

rustonaut:
As a side note the reason why there is a `use_legacy_teant = true/false/uuid` option is because TOML doesn't have null, which from a config language design POV makes a lot of sense but means I couldn't just default `legacy_tenant` to some UUID and then set it to `null` on a multi tenant setup :face_exhaling:

rustonaut:
Now it's only `use_legacy_tenant = true/false`.

rustonaut:
> In the first PR, in the discussion about task_local, can you add which functions will need it [..]

TL;DR:

- original use case gone
- still a better use case for the request id error handling problem then the type magic we had before
- for that use case it doesn't come with any risk as it will be used before we even enter our handler and e.g. could spawn a task
- it's not affecting our current tests
- hypothetical future "only storage layer" integration tests can use `RequestContext::wrap_future` put a future into the context of an request

---

Originally I planed to use it with sqlx Pool `before_acquire`, `after_connect` handlers. This methods would have allowed me to put the connections returned from the pool into the right context on the fly, but because they
are set once when creating the pool I needed to side step normal parameter passing.

This would have been _really_ convenient making the follow up PR just a few lines. Sadly this didn't work out due to how sqlx Pool handles errors in this callbacks (not directly propagating them). Given how "fatal" not having a tenant id is I considered panicing, but due to very bad experience with panic-resume due some dev's not understanding the constraints around panics (i.e. diesel) I decided against it and instead used a combination of  `StorageBuilder`, actix extractors, wrapping the DB pool etc. leading to a much bigger PR :sob: 

The only thing which still uses it now is some of the code which injects the request id into the json error response.

Doing so allowed me to remove a big chunk  of complicated type magic, **tbh. in retrospective I should have used task local storage for this from the get to go instead of using the type magic**. Sure implicit parameters are stylistically not grate. But type magic like that is often just way worse of a choice as it tends to be hard to maintain or adapt for anyone not familiar with using such tricks. It also allows us to use the recommended actix macros like `#[get(...)]`.

I could split it out into a separate PR I guess, but that wouldn't reduce the lines needing review by a lot as it's just a few lines add which need to be reviewed + deletion of a huge chunk of type magic.

rustonaut:
I will split it out into a separate PR.

integration-tests/src/lib.rs
```diff
@@ -197,6 +217,9 @@ where
 
         [embedding]
         directory = "../assets/smbert_v0003"
+
+        [request_context]
+        disable_legacy_tenant = false
```
janpetschexain, line 222: 🔧 i assume this comes from an older iteration of the code, but since the naming is different now this won't have any effect. i'd also assume that `disabled=false` would become `enabled=true`, right? (i guess it didn't crash in the tests, because the default is set to `enabled=true`)
```suggestion
        [tenants]
        enable_legacy_tenant = true
```

web-api/src/middleware/request_context.rs
```diff
@@ -0,0 +1,184 @@
...
+        .map(TenantId::try_from)
+        .transpose()?;
+
+    match header_value {
+        //FIXME in follow up PR this ID will be fetched from the database
+        //      during startup/storage initialization.
+        None if config.enable_legacy_tenant => Ok(TenantId::missing()),
+        None => Err(anyhow!("{TENANT_ID_HEADER} header missing")),
+        Some(passed_value) => Ok(passed_value),
+    }
```
acrrd, line 183: :wrench: We don't have to check the header if we are not multi-tenant
```suggestion
  if config.enable_legacy_tenant {
        return Ok(TenantId::missing());
    }

    request
        .headers()
        .get(TENANT_ID_HEADER)
        .map(TenantId::try_from)
        .ok_or_else(||  anyhow!("{TENANT_ID_HEADER} header missing"))?
```

web-api/src/middleware/request_context.rs
```diff
@@ -0,0 +1,184 @@
...
+        Ok(id) => id,
+        Err(error) => {
+            let response = middleware_failure(
+                "setup_request_context",
+                request,
+                Some(request_id),
+                None,
+                error,
+            );
+            return Either::Left(future::ok(response));
```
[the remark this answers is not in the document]
rustonaut: We should not spend any though on weather we use Right or Left in a context where we return a `impl Trait`.

The whole idea of `Either` compared to `Result`/a custom enum  is that while both sides have different types they have the same kind of API and are API wise roughly equivalent.  Anything beyond  is an implementation detail (e.g. the left future always resolves to an error). At least in rust where very intentionally types like `Option` and `Result` where added instead of having a more generic `Either`. 

Convention wise using `Left` first then `Right` is the only common convention. Which is also why e.g. the `either` library maps `Result<R,L> to Either<L, R>` as in proper (rust) usages of `Either` the the error case most times happen as a form of "early return" and in turn befor the normal result. I.e. it happens in a place people tend to use `Left` for.

But anyway this discussion is kinda pointless as the `Either` parameter order isn't part of the API due to `impl Trait`.

web-api/src/middleware/request_context.rs
```diff
@@ -0,0 +1,184 @@
...
+    request: ServiceRequest,
+    service: &S,
+) -> impl Future<Output = Result<ServiceResponse<BoxBody>, actix_web::Error>> + 'static
+where
+    A: AsRef<tenants::Config> + 'static,
+    S: Service<ServiceRequest, Response = ServiceResponse<BoxBody>, Error = actix_web::Error>,
+    S::Future: 'static,
+{
+    let request_id = RequestId::generate();
+    let config = request.app_data::<Data<A>>().unwrap().get_ref().as_ref();
```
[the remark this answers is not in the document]
rustonaut: Not sure as it's a fundamental bug which makes any request impossible to process, but the follow up PR has changes which make this decision irrelevant so I would just not change it here.

web-api/tests/legacy_tenant.rs
```diff
@@ -0,0 +1,88 @@
...
+        |_client, url, _| async move {
+            // don't use injected "X-Tenant-Id" header
+            let client = Client::new();
+            send_assert(
+                &client,
+                client
+                    .post(url.join("/documents")?)
+                    .json(&json!({
+                        "documents": [
+                            { "id": "d1", "snippet": "once in a spring there was a fall" },
```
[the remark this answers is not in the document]
rustonaut: but no trailing commas invite more merge conflicts

web-api/tests/cmd/cli_overrides.auto.toml
```diff
@@ -37,6 +37,9 @@ stdout = """
   "embedding": {
     "directory": "assets",
     "token_size": 250
+  },
+  "tenants": {
+    "enable_legacy_tenant": true
```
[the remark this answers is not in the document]
rustonaut: It's not really much of an issue, we rarely add options and the tests very easily point out what exactly need to change. The additional overhead needed to fix this isn't worth the benefit it would get us IMHO.

integration-tests/src/lib.rs
rustonaut: It's no a `single_tenant_mode` it's an option which allows having a legacy tenant. If we want in parallel with other tenants (not that we need that but we get that for free).

Furthermore in the long run we likely will use the normal deployment templates we have for multi tenant for single tenant, too (to massively reduce maintenance overhead).  Hence this option isn't just about single tenant, but about "old"  deployments we don't do this way anymore.

Hence `legacy`.

Also given that it must work when deploying to production without changing the config means that if the config option is set it is set to disable the legacy tenant, as the default needs to be having it enabled.

Hence `disable` .

And there is no reason why we should even switch the default of the option, that would just be unnecessary work, so the next time the option is touched is when removing it.

rustonaut: > i assume this comes from an older iteration of the code, but since the naming is different now this won't have any effect.

yes it doesn't have any effect, just is explicit.

web-api/src/middleware/request_context.rs
```diff
@@ -0,0 +1,184 @@
...
+use thiserror::Error;
+use tracing::{info_span, instrument, trace, Instrument};
+use uuid::Uuid;
+
+use crate::{error::early_failure::middleware_failure, tenants};
+
+pub(crate) struct RequestContext {
+    #[allow(unused)]
+    pub(crate) tenant_id: TenantId,
+    pub(crate) request_id: RequestId,
```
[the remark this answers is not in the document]
rustonaut: > which mode we are without having to refer back to the configuration.

There is no `mode`.

Any tenant including legacy ones have a id (in case of legacy tenants it's generated on the fly if it doesn't exist yet)..

> This will also enable us in separating the request_id part from the tenant part, since they are not really related.

They are deeply related as they are both part of the request context (we will pass the request id through a header instead of generating it in the future).  And both must be given when setting up the logging span.
Splitting it apart without any need for this split is IMHO just making things unnecessary complicated.

web-api/src/middleware/request_context.rs
```diff
@@ -0,0 +1,184 @@
...
+    request: &ServiceRequest,
+) -> Result<TenantId, anyhow::Error> {
+    let header_value = request
+        .headers()
+        .get(TENANT_ID_HEADER)
+        .map(TenantId::try_from)
+        .transpose()?;
+
+    match header_value {
+        //FIXME in follow up PR this ID will be fetched from the database
```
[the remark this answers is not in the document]
rustonaut: > should continue to work as is 

It does.

Like written it will fetch the id for the legacy tenant form the database.  The first time the new rust code is started  (of the follow up PR) it will randomly generate the id and insert it there (and rename the `public` schema to `t:{legacy_tenant_id}`).

This approach is easier to implement then any of the alternatives AFIK, as it makes the whole thing about having different `modes` and other special cases through the code unnecessary and renaming a schema is pretty trivial (one line of sql).

It also has the neat side effect of making migrating legacy tenants to non-legacy single tenants or multi tenants much easier or allow us to add additional tenants on a single tenant mode deployment. Non of this played a role for why I choose to implement it this way, but they are nice anyway.

The reason I didn't purse this approach from the beginning was that I though  there would be certain problems with this which turned out to be non-issues.

web-api/src/middleware/request_context.rs
rustonaut: we do like mentioned above there is no multi tenant mode, just legacy tenants which will be used by default if no tenant id is given explicitly and the legacy tenant is enabled.

integration-tests/src/lib.rs
acrrd: We don't have an old system. The current deployment will continue to live as it is. Each of the current tenant will stay in their own infrastructure, they will run in parallel by definition, and they don't need an id. 
What we need to make that work is avoiding to changing the role if we are in single tenant mode and don't look for an header.

web-api/src/middleware/request_context.rs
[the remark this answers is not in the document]
acrrd: This is more complicated simply because we are doing work that is not needed. Generating a random id here is logically wrong because the id must come from the control panel and the infrastructure not our code.

> It also has the neat side effect of making migrating legacy tenants to non-legacy single tenants or multi tenants much easier or allow us to add additional tenants on a single tenant mode deployment

The only use case we will have, much later in the future, will be to move the current tenants onto the multi-tenant infrastructure. Till then we will have two separate infrastructures and we might deploy customers on one or the other depending on their needs.

web-api/src/middleware/request_context.rs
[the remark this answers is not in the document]
acrrd: We don't need tenants to have their id. The code for multi-tenant is fortunately very contained.
If we are in a multi-tenant we check for the header and use `SET ROLE` other we do nothing. 

> future deployments of single tenant setups will use the same template as our multi tenant setup, not doing so would add massive maintenance overhead.

No, the two infrastructures will run in parallel till we have the resource to merge them.  It will not incur any overhead because we are going to use what we already have for a single tenant and we are going to deploy one multi-tenant environment only for customers to test the platform, not production. 
Any work that involves trying to partially migrate the current tenant to a multi-tenant-like setup is a waste of resources. We don't need it for our current goals, it is a middle-term or long target that we can address in the future with dedicated stories.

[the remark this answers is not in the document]
acrrd: > No, the two infrastructures will run in parallel till we have the resource to merge them.  It will not incur in any overhead because we are going to use what we already have for single tenant.

The two are completely orthogonal to each other, they have been deeply connected in the implementation but one can leave without the other. Is not an optimization that is refactoring the code so that each part works only on one concept. I didn't mark it correctly and I'm also fine with leaving it as it is now.

web-api/src/middleware/request_context.rs
acrrd: So we can have the header and `enable_legacy_tenant= true`? That seems an error.

rustonaut: yes, means less hassle when migrating much later down the line and forbidding it is additional work with doesn't help us in any way.

So less work and less problems further down the line seems like a win-win for me.

janpetschexain (approved):
lgtm

PR #864, opened 2023-03-30
[Merged by Bors] - Check downloaded data before moving it to assets/

acrrd:
If the checksum fails the data is still in the correct place to be used by the programs, moreover, future downloading of that assets will be skipped.

With this PR we first check that the data is not corrupted and then we moved it to the correct place.
One main difference from before is that the archive file is not retained.
This can be easily changed if you think is better to keep it.


acrrd:
bors merge

PR #865, opened 2023-03-30
correct knn results of hybrid search

rustonaut:
Normally the way ES sums up scores is not a problem but due to how we combine them for reranking we have to account for how many factors go into the final score.

### DIFF
--- a/web-api/src/storage/elastic.rs
+++ b/web-api/src/storage/elastic.rs
@@ -318,7 +318,12 @@ impl Client {
         if let Some(min_similarity) = params.min_similarity {
             body.insert("min_score".to_string(), json!(min_similarity));
         }
+        // This need to account for each (sub)query contributing to
+        // the score corrected by boost. (this includes multiple score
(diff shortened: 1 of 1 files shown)


web-api/src/storage/elastic.rs
```diff
@@ -334,7 +339,7 @@ impl Client {
                     .hits
                     .hits
                     .into_iter()
-                    .map(|hit| (hit.id, hit.score))
+                    .map(|hit| (hit.id, hit.score / multi_search_correction_factor))
```
acrrd, line 342: :question: How can this bring the result from ES back to `(0..1)`? I think we need to rescale them. We should also consider that if the score range from BM25 is higher than knn query, then, we are only scoring based on that.

acrrd: I don't see much value in applying this. Having 5 as a score instead of 10 will not make it sensibly better and we are confusing the code.
I'll prefer to tackle this properly in the future.

PR #866, opened 2023-04-03
Add loading of model with one output that is already the embedding

acrrd:
Allows to load a model that has only one layer and the layer is already the embedding.


acrrd:
Not needed atm.

PR #867, opened 2023-04-03
[Merged by Bors] - Add prepare_data.sh

acrrd:
Port to this repository a helper script from `xayn_ai`.
The scripts can prepare and upload an archive with the model using the correct naming format that we expect.

### DIFF
--- a/.github/scripts/prepare_data.sh
+++ b/.github/scripts/prepare_data.sh
@@ -1,64 +1,59 @@
-#!/bin/bash
+#!/usr/bin/env -S bash -eux -o pipefail
+
 # This script takes as input a directory, the name of the archive and a version.
 # It creates an archive in the correct format in the current directory
 # and adds the necessary information to verify its content.
 # The archive will contain the directory name and the provided version.
 # If the option --upload is provided the script will upload the archive to the s3 bucket.
 #
-# ./prepare_data.sh ./bert_models rubert v0 will generate an archive rubert_v0.tgz
-# with one directory rubert_v0 that contains the files that are present in ./bert_models.
-# 
-# If --upload is passed as an option then the file is also uploaded on the S3 bucket.
+# prepare_data.sh <model_dir> <model_name> <version_tag> [--upload]
+# It creates an archive <model_name>_<version>.tgz which contains one directory <model_name>_<version>
+# that contains the files that are present in <model_dir>.
 
 # directory to prepare for upload
-DIR_PATH=$1
+DIR_PATH="$1"
 shift
-NAME=$1
+NAME="$1"
 shift
-VERSION=$1
+VERSION="$1"
 shift
 
 while [ $# -gt 0 ]; do
   opt="$1"
   shift
 
-  case $opt in
+  case "$opt" in
     --upload)
     UPLOAD=true
     ;;
   esac
 done
 
-DIR_PATH=$(pwd)/$DIR_PATH
-DIR_NAME=$(basename $DIR_PATH)
+DIR_PATH="$(pwd)/$DIR_PATH"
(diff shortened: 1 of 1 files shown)


.github/scripts/prepare_data.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
+# This script takes as input a directory, the name of the archive and a version.
```
rustonaut, line 2: :art: It would be nice if it had a clear usage doc comment like

`Usage: prepare_data.sh <models_dir> <model_name> <version_tag>  [--usage]`

.github/scripts/prepare_data.sh
```diff
@@ -0,0 +1,64 @@
...
+while [ $# -gt 0 ]; do
+  opt="$1"
+  shift
+
+  case $opt in
+    --upload)
+    UPLOAD=true
+    ;;
+  esac
+done
```
rustonaut, line 30: :art: I don't think it's worth changing it here but in general flags come before the fixed arguments i.e.:

`Usage: prepare_data.sh [--upload] <models_dir> <model_name> <version_tag>`

.github/scripts/prepare_data.sh
```diff
@@ -0,0 +1,64 @@
...
+
+DIR_PATH=$(pwd)/$DIR_PATH
+DIR_NAME=$(basename $DIR_PATH)
+ARCHIVE_BASENAME="${NAME}_$VERSION"
+ARCHIVE_NAME="$ARCHIVE_BASENAME.tgz"
+URL="s3://xayn-yellow-bert/$NAME/$ARCHIVE_NAME"
+CHECKSUM_FILE="sha256sums"
+
+CURRENT_DIR=$(pwd)
+cd $DIR_PATH
```
rustonaut, line 40: :wrench: I'm not sure why this is here when we then switch to a temp dir which is normally out of tree (e.g. in /tmp)

```suggestion
```

.github/scripts/prepare_data.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
```
rustonaut, line 1: :wrench:  make the script fail if things go wrong

```suggestion
#!/usr/bin/env -S bash -eux -o pipefail
```

.github/scripts/prepare_data.sh
```diff
@@ -0,0 +1,64 @@
...
+TMP_DIR=$(mktemp -d)
+cd $TMP_DIR
+cp -r $DIR_PATH .
+if [ $DIR_NAME != $ARCHIVE_BASENAME ]; then
+  mv $DIR_NAME $ARCHIVE_BASENAME
+fi
+TO_ARCHIVE="$TMP_DIR/$ARCHIVE_BASENAME"
+
+# compute checksum file
+cd $ARCHIVE_BASENAME
```
rustonaut, line 52: :art: we can make it much simpler

```suggestion
cp -r "$DIR_PATH" "$TMP_DIR/$ARCHIVE_BASENAME"

# compute checksum file
cd  "$TMP_DIR/$ARCHIVE_BASENAME"
```

rustonaut (commented):
some changes, and quoting is basically missing everywhere

.github/scripts/prepare_data.sh
acrrd: There is an instantiated example a few lines below. I can make it more general

.github/scripts/prepare_data.sh
acrrd: Having it after makes it slightly easier to parse and use. If you decide to upload the archive after you generate it you can go to the previous command and add ` --upload` there without moving.

acrrd:
bors merge

PR #868, opened 2023-04-04
[Merged by Bors] - Remove naming conflict in the api

acrrd:
`/documents/candidates`  can conflict with `/documents/<id>`. At the moment there isn't such a conflict because of different methods but this will forbid us from adding methods to `/documents/<id>` and in general can confusing for the end user and us.


acrrd:
bors merge

PR #869, opened 2023-04-04
[Merged by Bors] - Use sparse only on the CI

acrrd:
Forcing the use of sparse with `just` conflicts with local development where people are forced to change the configuration of their tools to comply or face needless recompilation.
IMHO this should have never been put in `.env`.


acrrd:
bors merge

PR #871, opened 2023-04-09
[Merged by Bors] - simplify request id injection

rustonaut:
This replaces complex type magic used to inject the request ID into an error with a simple task local storage injection.

IMHO this is what I should have done from the beginning.

Sure "magic implicit parameters" in form of task local data aren't grate but:

- we anyway use a extension system which doesn't enforce the existence of the value through the type system. In case of the type magic we use the `HttpRequest` extensions, which if we forget to put the request id in will fail at runtime and is kinda like a implicit parameter. In case of TLS we use TLS and fail at runtime if we forget to put it there the same way (but with an error message which better points to the source of failure).  **So the type magic doesn't provide any additional type safety, only complexity.**

- the type magic is complicated an hostile to changes, error messages in it are also worse due to being less specific

- the type magic breaks actix macros like `#[get('/...')]` through if we want to use that is a different question

- as extractors are run as part of actix there is no risk of accidentally calling it in a `tokio::spawn`ed sub-task 

- if we for some reason run it outside of the context of the actix request handler we can always inject  it using `wrap_future` , but for integration tests this is not necessary and for unit tests outside of unit testing the request id injection unlikely to ever matter.

### DIFF
--- a/web-api/src/ingestion/routes.rs
+++ b/web-api/src/ingestion/routes.rs
@@ -27,18 +27,13 @@ use tracing::{error, info, instrument};
 
 use super::AppState;
 use crate::{
-    app::TenantState,
-    error::{
-        application::WithRequestIdExt,
-        common::{
-            BadRequest,
-            DocumentIdAsObject,
-            DocumentNotFound,
-            DocumentPropertyNotFound,
-            FailedToDeleteSomeDocuments,
-            FailedToSetSomeDocumentCandidates,
-            IngestingDocumentsFailed,
-        },
+    error::common::{
+        BadRequest,
+        DocumentNotFound,
+        DocumentPropertyNotFound,
+        FailedToDeleteSomeDocuments,
+        FailedToSetSomeDocumentCandidates,
+        IngestingDocumentsFailed,
     },
     models::{self, DocumentId, DocumentProperties, DocumentProperty, DocumentTag},
     storage,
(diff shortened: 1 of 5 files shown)


web-api/src/lib.rs
```diff
@@ -23,7 +22,8 @@
     clippy::pedantic,
     noop_method_call,
     rust_2018_idioms,
-    unused_qualifications
+    unused_qualifications,
+    unsafe_op_in_unsafe_fn
```
rustonaut, line 26: There is a false positive related to macro hygenics and encapsulation AFIKT  this falls over a `#[deny(unsafe_op_in_unsafe_fn)]` which fails as `deny` is less strict then `forbid` and it's set in
the `std::thread_local!` macro (`src/std/thread/local.rs.html#185`) used by `tokio::task_local!`.

web-api/src/ingestion/routes.rs
```diff
@@ -46,34 +43,31 @@ pub(super) fn configure_service(config: &mut ServiceConfig) {
     config
```
janpetschexain, line 43: :art: since we already touch the code here, we could make it a bit less verbose, instead of `config.service(resource::route())` we could use the `config.route()` alias, like
```rust
config
    .route("/candidates", web::get().to(get_document_candidates))
    .route("/candidates", web::put().to(set_document_candidates))
    // etc
```
similarly for the personalization routes.

acrrd:
We can merge this as soon as is ready to simplify the other PR.

web-api/src/middleware/request_context.rs
```diff
@@ -103,6 +108,21 @@ impl RequestId {
...
+        F: 'static + Future,
+    {
+        CURRENT_REQUEST_ID.scope(self, future)
+    }
+
+    pub(crate) fn extract_from_task_local_storage() -> Result<RequestId, AccessError> {
+        CURRENT_REQUEST_ID
+            .try_with(|id| *id)
+            .map_err(|_| AccessError {
+                method: "extract_from_task_local_storage",
```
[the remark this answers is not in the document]
rustonaut: Through we extract the task local data from the task local storage. 

It's the task variant of thread local storage.

But I can rename it anyway. I just removed the `storage` part :rocket:

web-api/src/ingestion/routes.rs
rustonaut: I'm not sure that is simpler as it duplicates all the paths. If we do that we could also use the `#[get(...)]` etc. macros.

rustonaut:
bors merge

PR #872, opened 2023-04-09
[Merged by Bors] - ET-4229 silo tenant managment and db migration api

rustonaut:
Add the full silo management rust API.

As it is long term meant to be used as it's own tool it is in it's own crate. But for now we just import and run it on startup.

Legacy tenant (if enabled) is handled as following:

- on early startup: if there is a public schema with tenant tables we rename it to the legacy tenants schema and create a new public schema
- on early startup: if enabled but not setup a random id will be generated and a tenant with that id is created, we store the id in the management schema 
- on startup if enabled load the legacy tenant id from the management schema once and cache it
- if enabled and no tenant id is given with an request the loaded/cached legacy tenant id will be used 

with that except the mapping of no tenant id to one loaded from the db no special handling is given to legacy tenants during request handling and in the future this helps with a lot of future changes as legacy tenants have different ids (so merging silos is easier) and can be refferred to by no or their internal id (so seamless migration to using a tenant id for them)

**References:**

- issue: [ET-4229]
- story: [ET-4172]
- based on: #860, #862

[ET-4229]: https://xainag.atlassian.net/browse/ET-4229?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


web-api-shared/src/postgres.rs
```diff
@@ -16,10 +16,98 @@ use std::{fmt::Display, str::FromStr};
...
+            //TODO check how this might be used, the default we set here
+            //     before didn't really work but might have been used
+            application_name: None,
+            skip_migrations: false,
+        }
+    }
+}
+
+impl Config {
+    pub fn to_connection_options(&self) -> Result<PgConnectOptions, sqlx::Error> {
```
rustonaut, line 75: just moved

web-api-shared/src/postgres.rs
```diff
@@ -16,10 +16,98 @@ use std::{fmt::Display, str::FromStr};
...
 use thiserror::Error;
 
+use crate::{request::TenantId, serde::serialize_redacted};
+
+pub type Client = Pool<Postgres>;
+
+/// Configuration for connection to postgres
+#[derive(Debug, Clone, Deserialize, Serialize)]
+#[serde(default)]
+pub struct Config {
```
rustonaut, line 31: just moved

web-api-db-ctrl/src/lib.rs
```diff
@@ -0,0 +1,524 @@
...
+            tx.execute_many(query.as_str())
+                .try_for_each(|_| future::ready(Ok(())))
+                .await?;
+        }
+        Ok(legacy_tenant_id)
+    }
+
+    /// Allows using the admin user as `web-api-mt` user.
+    #[instrument(skip(self), err)]
+    pub async fn admin_as_mt_user_hack(&self) -> Result<(), Error> {
```
rustonaut, line 197: Just to avoid problems: This has nothing to do with the legacy tenant handling.

This exist because the code already assumes there are two users:

1. `admin` which owns the schemas/structure of the db and is e.g. used to create tenants and run schema migrations
2. `web-api-mt` which only is allowed to use schemas

I'm not sure if I like this hack, but managing permissions for this later one is a bit of a pain. And changing our deployment in the future to have two users from the get to go shouldn't be too hard :thinking:

web-api-db-ctrl/src/lib.rs
```diff
@@ -0,0 +1,531 @@
...
+                if err.to_string().contains("tuple concurrently updated") && count > 0 {
+                    continue;
+                } else {
+                    return Err(err);
+                }
+            }
+            Ok(true)
+        } else {
+            Ok(false)
+        };
```
rustonaut, line 407: This loop/retry logic wasn't triggered a single time when I ran the test multiple times.

I'm not completely sure if it is because it can't fail due to how postgres works or if it's just very unlikely,
so just to be safe I keep this in.

Also this function only is used by the `admin_as_mt_user_hack` so once we remove that we can also remove this function.

web-api/src/middleware/request_context.rs
```diff
@@ -162,20 +134,36 @@ where
...
 }
+
+fn trim_ascii(ascii: &[u8]) -> &[u8] {
+    let start = ascii
+        .iter()
+        .position(|byte| !byte.is_ascii_whitespace())
+        .unwrap_or_default();
+    let end = ascii
+        .iter()
+        .rposition(|byte| !byte.is_ascii_whitespace())
```
janpetschexain, line 166: 🔧 `.rposition()` returns the index, that means if we look for the rightmost byte which is not a whitespace, then we have to include it in the returned slice: `&ascii[start..=end]`, which means we have to update the unwrapping as well. or we have to add 1 to the rposition index.

we probably should add a small test for this function.

janpetschexain (commented):
looks good, some questions and suggestions

web-api-db-ctrl/src/lib.rs
```diff
@@ -0,0 +1,535 @@
...
+
+/// Unlocks an id locked with [`lock_id_until_unlock()`].
+///
+/// This *can not* be used to unlock ids locked with [`lock_id_until_end_of_transaction()`].
+async fn unlock_lock_id(
+    tx: impl Executor<'_, Database = Postgres>,
+    lock_id: i64,
+) -> Result<(), sqlx::Error> {
+    debug!({ lock_id }, "pg_advisory_unlock");
+    sqlx::query("SELECT pg_advisory_unlock($1)")
```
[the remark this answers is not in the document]
rustonaut: I think the only think we could/should do when getting a unexpected bool value is warn , so the sqlx warning should be enough.

I guess the main difference is that we would warn on an error level :thinking:

web-api-db-ctrl/src/lib.rs
rustonaut: I ran into this now :tada: 

But it doesn't work :disappointed: .

I forgot that one an sql statement runs into an error in a transaction it will make any further statements in the transaction fail.

So this needs to be done outside of the transaction :thinking: 

But user creation and similar are still transaction in postgres, which we would lose here, which I guess is fine given for what we use it but means it a not fully general function.

rustonaut:
- [ ] the post rebase fixes are mostly done except that we currently don't derive a per-tenant lock id anymore

web-api-db-ctrl/src/lib.rs
rustonaut: fixed :rocket:

rustonaut:
refactored to move postgres code from `lib.rs` to `postgres.rs`.

rustonaut:
There seems to be a reproducible "slow test + cancellation" which only happens in CI:

test test_store_user_history_disabled has been running for over 60 seconds

I have to look into it.

web-api-db-ctrl/src/postgres.rs
```diff
@@ -0,0 +1,521 @@
...
+    let (lock_was_held,) = sqlx::query_as::<_, (bool,)>("SELECT pg_advisory_unlock($1)")
+        .bind(lock_id)
+        .fetch_one(tx)
+        .await?;
+    if lock_was_held {
+        debug!({ lock_id }, "pg_advisory_unlock");
+    } else {
+        error!(
+            { lock_id },
+            "spurious pg_advisory_unlock which wasn't locket"
```
janpetschexain, line 499: :art: typo
```suggestion
            "spurious pg_advisory_unlock which wasn't locked"
```

web-api-db-ctrl/src/postgres.rs
```diff
@@ -0,0 +1,521 @@
...
+
+    //TODO we need to decide how to handle partial failure
+    if !failures.is_empty() {
+        for (tenant_id, error) in failures {
+            error!({ %tenant_id, %error }, "migration failed");
+        }
+        bail!("all tenant migrations failed");
+    }
+
+    conn.close().await?;
```
janpetschexain, line 130: ❓ shouldn't we move this above l.122 before the error return branching? such that we close the connection no matter what result is returned from the function.

janpetschexain (approved):
👍

rustonaut:
bors merge

rustonaut:
bors m-

rustonaut:
bors merge-

rustonaut:
It's best to merge the follow up PR into this and then merge it into main, as a build based on this PR without the follow up PR has some issues (if it's used with this PR and then switches to the follow up PR, if we switch from before this PR to with-the-followup-PR it's not an issue)

rustonaut:
bors merge

PR #879, opened 2023-04-13
[Merged by Bors] - FIX: candidates setter

janpetschexain:
**Summary**

- avoid overly many update operations in PG by switching the order of comparison and update


janpetschexain:
bors merge

PR #880, opened 2023-04-13
[Merged by Bors] - Set high limit for request size

acrrd:
Set the limit for the request high. The default is two low for us and we want to handle this limits only in the infrastructure.


joecks (approved):

acrrd:
bors merge

PR #883, opened 2023-04-17
[Merged by Bors] - ET-4332 impl do not override is candidate

rustonaut:
If `default_is_candidate` is given instead of `is_candidate` we will use any existing `is_candidate` setting from the database for given document id and only use the value of `default_is_candidate` if the document is new.

Specifying both `default_is_candidate` and `is_candidate` is an error handled the same way as a malformed document id, tag name or similar.

**References:**

- story: [ET-4330]
- task: [ET-4332]

[ET-4330]: https://xainag.atlassian.net/browse/ET-4330?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ET-4332]: https://xainag.atlassian.net/browse/ET-4332?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/web-api/src/ingestion/routes.rs
+++ b/web-api/src/ingestion/routes.rs
@@ -242,7 +242,7 @@ async fn upsert_documents(
         .map(|document| {
             let (snippet, is_candidate) = existing_data
                 .get(&document.id)
-                .map(|(s, c)| (s, *c))
+                .map(|(snippet, is_candidate)| (snippet, *is_candidate))
                 .unzip();
 
             let new_is_candidate = document.is_candidate_op.resolve(is_candidate);
--- a/web-api/src/ingestion/routes.rs
+++ b/web-api/src/ingestion/routes.rs
@@ -19,6 +19,7 @@ use actix_web::{
     HttpResponse,
     Responder,
 };
+use anyhow::bail;
 use itertools::Itertools;
 use serde::{de, Deserialize, Deserializer, Serialize};
 use tokio::time::Instant;
@@ -92,10 +93,6 @@ where
     }
 }
 
-const fn default_is_candidate() -> bool {
-    true
-}
-
 #[derive(Debug, Deserialize)]
 struct UnvalidatedIngestedDocument {
     id: String,
@@ -105,8 +102,10 @@ struct UnvalidatedIngestedDocument {
     properties: HashMap<String, DocumentProperty>,
     #[serde(default)]
     tags: Vec<String>,
-    #[serde(default = "default_is_candidate")]
-    is_candidate: bool,
+    #[serde(default)]
+    is_candidate: Option<bool>,
+    #[serde(default)]
+    default_is_candidate: Option<bool>,
 }
 
 #[derive(Debug)]
@@ -115,7 +114,48 @@ struct IngestedDocument {
     snippet: String,
     properties: DocumentProperties,
     tags: Vec<DocumentTag>,
-    is_candidate: bool,
+    is_candidate_op: IsCandidateOp,
+}
+
+#[derive(Clone, Debug, Copy)]
+enum IsCandidateOp {
+    SetTo(bool),
+    DefaultTo(bool),
+}
+
+impl IsCandidateOp {
+    /// Returns the new value and `has_existing && is_changed_value`
+    fn resolve(self, existing: Option<bool>) -> NewIsCandidate {
+        match self {
+            IsCandidateOp::SetTo(new) => NewIsCandidate {
+                value: new,
+                existing_and_has_changed: existing
+                    .map(|previous| previous != new)
+                    .unwrap_or_default(),
+            },
+            IsCandidateOp::DefaultTo(default) => NewIsCandidate {
+                value: existing.unwrap_or(default),
+                existing_and_has_changed: false,
+            },
+        }
+    }
+}
+
+#[derive(Clone, Copy)]
+#[cfg_attr(test, derive(PartialEq, Debug))]
+struct NewIsCandidate {
+    value: bool,
+    existing_and_has_changed: bool,
+}
+
+impl NewIsCandidate {
+    fn has_changed_to_false(self) -> bool {
+        self.existing_and_has_changed && !self.value
+    }
+
+    fn has_changed_to_true(self) -> bool {
+        self.existing_and_has_changed && self.value
+    }
 }
 
 impl UnvalidatedIngestedDocument {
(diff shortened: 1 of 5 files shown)


web-api/src/ingestion/routes.rs
```diff
@@ -134,50 +152,77 @@ async fn new_documents(
...
+                    let is_candidate_op = match (is_candidate, default_is_candidate) {
+                        (Some(_), Some(_)) => {
+                            bail!("You can only use either of is_candidate or default_is_candidate")
+                        }
+                        (Some(value), None) => IsCandidateOp::SetTo(value),
+                        (None, Some(value)) => IsCandidateOp::DefaultTo(value),
+                        (None, None) => IsCandidateOp::SetTo(true),
+                    };
+                    //FIXME fetch snippet with is_candidate below and only calculate/set
+                    //      the embedding if the snippet changed.
```
janpetschexain, line 183: :art: this is part of [ET-4237], not sure if we need an extra fixme comment here.

web-api/src/ingestion/routes.rs
```diff
@@ -134,50 +152,77 @@ async fn new_documents(
 
     let start = Instant::now();
 
-    let (documents, mut failed_documents) = body
+    let (preprocessed, mut failed_documents): (Vec<_>, Vec<_>) = body
```
janpetschexain, line 155: :art: it's nicer to annotate the types at the end of the call like `.partition_result::<Vec<_>, Vec<_>, _, _>()` such that the IDE can show the full type hint for the assigned variables.

janpetschexain (commented):
looks good, i think it should be too hard to rebase #882 afterwards as the changes align quite well.

web-api/src/ingestion/routes.rs
rustonaut: I would argue it's the other way around as it's also clearer what parts the type hints are exactly for.

And at least in my case the IDE is showing type hints properly :thinking:

web-api/src/storage.rs
```diff
@@ -83,6 +83,11 @@ impl<T> FromIterator<T> for Warning<T> {
 
 #[async_trait(?Send)]
 pub(crate) trait Document {
+    async fn get_candidate_status(
```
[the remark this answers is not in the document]
rustonaut: I moved it.

web-api/src/ingestion/routes.rs
janpetschexain: exactly, sry i should have said that i mean the inlay hints. tool tips always show properly.

web-api/src/ingestion/routes.rs
```diff
@@ -115,7 +114,48 @@ struct IngestedDocument {
...
+                existing_and_has_changed: existing
+                    .map(|previous| previous != new)
+                    .unwrap_or_default(),
+            },
+            IsCandidateOp::DefaultTo(default) => NewIsCandidate {
+                value: existing.unwrap_or(default),
+                existing_and_has_changed: false,
+            },
+        }
+    }
```
rustonaut, line 141: This is slightly more complex then in the previous version but allows us to only add/remove is candidate where needed for changed documents.

web-api/src/ingestion/routes.rs
```diff
@@ -177,65 +226,87 @@ async fn upsert_documents(
...
+    let (changed_documents, new_documents) = documents
+        .into_iter()
+        .map(|document| {
+            let (snippet, is_candidate) = existing_data
                 .get(&document.id)
-                .map(|snippet| document.snippet == *snippet)
-                .unwrap_or_default()
-        });
+                .map(|(s, c)| (s, *c))
+                .unzip();
```
rustonaut, line 246: We could squeeze the logic into the map statement, but I think with unzip it's a bit more readable.

web-api/src/ingestion/routes.rs
```diff
@@ -177,65 +226,87 @@ async fn upsert_documents(
...
+            let new_is_candidate = document.is_candidate_op.resolve(is_candidate);
+            let unchanged_snippet = snippet
+                .map(|snippet| snippet == &document.snippet)
+                .unwrap_or_default();
+
+            (document, new_is_candidate, unchanged_snippet)
+        })
+        .partition::<Vec<_>, _>(|(_, _, unchanged_snippet)| *unchanged_snippet);
 
-    // checked above that all changed documents exist, hence no extended failed documents
```
rustonaut, line 197: :pushpin: I just realized that in case of race conditions between some delete documents operation and an update this isn't necessary true. But outside of the scope of this PR.

web-api/src/ingestion/routes.rs
```diff
@@ -115,7 +114,48 @@ struct IngestedDocument {
...
+}
+
+#[derive(Clone, Debug, Copy)]
+enum IsCandidateOp {
+    SetTo(bool),
+    DefaultTo(bool),
+}
+
+impl IsCandidateOp {
+    /// Returns the new value and `has_existing && is_changed_value`
```
janpetschexain, line 127: :art: the docs seem to be slightly outdated as the field names changed a bit.

web-api/src/ingestion/routes.rs
```diff
@@ -177,65 +226,87 @@ async fn upsert_documents(
...
+    // Hint: Documents which have a changed snippet are also in `new_documents`.
+    let (changed_documents, new_documents) = documents
+        .into_iter()
+        .map(|document| {
+            let (snippet, is_candidate) = existing_data
                 .get(&document.id)
-                .map(|snippet| document.snippet == *snippet)
-                .unwrap_or_default()
-        });
+                .map(|(s, c)| (s, *c))
```
janpetschexain, line 245: :art: unzipping is fine, i would just not abbreviate the names to make it more readable
```suggestion
                .map(|(snippet, is_candidate)| (snippet, *is_candidate))
```

web-api/tests/document_candidates.rs
```diff
@@ -181,3 +181,50 @@ async fn test_candidates_warning() {
     })
     .await;
 }
+
+#[tokio::test]
+async fn test_candidates_reingestion() {
```
janpetschexain, line 186: :art: this is quite similar to the test in `tests/ingestion.rs` l.135. i didn't take a full look, is this a duplication maybe?

janpetschexain (approved):
👍

rustonaut:
bors merge

PR #885, opened 2023-04-19
[Merged by Bors] - explict example for ingestion

rustonaut:
The auto generated examples do contain both `is_candidate` and `default_is_candidate` which is invalid.

But we can't (easily/cleanly) add constraints that only either of them is allowed as depending on how we do so this either trips over the linter or the API generator.


acrrd (approved):

rustonaut:
bors merge

PR #886, opened 2023-04-19
[Merged by Bors] - fix problem with generated code

rustonaut:
don't annotate the default value for is_candidate

if we do so there are some cases in which `is_candidate` will be unexpectedly set to the default value, instead of not being set using the server default.


acrrd (approved):

rustonaut:
bors merge

PR #887, opened 2023-04-19
[Merged by Bors] - ET-4314 normalize hybrid search score

janpetschexain:
**Reference**

- [ET-4314]

**Summary**

- normalize the ES score for hybrid search: in case where a document only has a bm25 score but no knn score and the distribution of bm25 scores in that query is bad, the normalized score is likely to be bad as well (but we can't do much better without expensive multi-query strategies). in all other cases the normalization is likely to be good enough.
- there is one main caveat though currently: the normalization happens for each search call, so if we eg make a search call for multiple cois and then combine the returned documents, then the score normalization is applied to each subset of documents returned for each coi instead of for all documents together. this can lead to skewed scores if the score distribution for each subset varies much. if we want to normalize over multiple search calls, we would need to return the intermediate normalization components instead of the normalized scores. i guess if this is required, it should be done in a follow up pr.

[ET-4314]: https://xainag.atlassian.net/browse/ET-4314?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/web-api/src/storage/elastic.rs
+++ b/web-api/src/storage/elastic.rs
@@ -254,6 +254,7 @@ impl Client {
         self.query_with_bytes(url, post_data).await
     }
 
+    #[allow(clippy::too_many_lines)]
     pub(super) async fn get_by_embedding<'a>(
         &self,
         params: KnnSearchParams<'a, impl IntoIterator<Item = &'a DocumentId>>,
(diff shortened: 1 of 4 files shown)


rustonaut (approved):
lgtm

janpetschexain:
bors merge

PR #891, opened 2023-04-20
[Merged by Bors] - ET-4304 tenant id str instead of UUID

rustonaut:
I copied the final version of TenantId (from the db management PR) and then changed it to use `Arc<str>` instead of `UUID`.

This means there is a bit of other changes:

- a InvalidTenantId error type
- additional derives needed in later PRs
  - including sqlx::Type/sqlx::transparent 
- parsing from `ascii` instead of `HeaderValue` 
  - (because it will get moved into `web-api-shared`, which doesn't depend on actix_web and it's the actix version of `HeaderValue`)
  
For now I arbitrarily constrained the TenantId to   `^[a-zA-Z0-9_:@.-]{1,50}$`  we could bump the length to 61 and allow any chars the quoted postgres identifier supports. But I think having it more constrained for now is a good idea.

### DIFF
--- a/web-api/src/middleware/request_context.rs
+++ b/web-api/src/middleware/request_context.rs
@@ -218,19 +218,15 @@ fn trim_ascii(ascii: &[u8]) -> &[u8] {
 fn trim_ascii_start(ascii: &[u8]) -> &[u8] {
     ascii
         .iter()
-        .position(is_not_ascii_whitespace)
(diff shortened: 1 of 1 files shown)


janpetschexain (approved):
lgtm

rustonaut:
bors merge

PR #892, opened 2023-04-20
[Merged by Bors] - ET-4345 prepare es for multi tenant

rustonaut:
Prepares for changes in follow up PRs:

- split elastic Client  into core web-api independent client code and our requests/endpoints
  - the generic parts get moved into elastic/client, later we will move them to `web-api-shared`  
- create a `ClientBuilder`
- `SegmentableUrl` abstraction
- use `Arc` as appropiate 
- don't depend on `DocumentId` in `elastic/client.rs`
- have custom error type in `elastic/client.rs`

### DIFF
--- a/web-api/src/storage/elastic/client.rs
+++ b/web-api/src/storage/elastic/client.rs
@@ -12,7 +12,7 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-use std::{collections::HashMap, fmt::Debug, sync::Arc};
+use std::{collections::HashMap, fmt::Debug, str::FromStr, sync::Arc};
 
 use reqwest::{
     header::{HeaderMap, HeaderValue, CONTENT_TYPE},
@@ -62,11 +62,9 @@ pub(crate) struct Client {
 
 impl Client {
     pub(crate) fn builder(config: &Config) -> Result<ClientBuilder, SetupError> {
-        let mut base_url = config.url.parse::<SegmentableUrl>()?;
-
         Ok(ClientBuilder {
             config: Arc::new(config.clone()),
-            base_url: Arc::new(base_url),
+            base_url: Arc::new(config.url.parse::<SegmentableUrl>()?),
             client: reqwest::Client::new(),
         })
     }
@@ -83,7 +81,10 @@ impl ClientBuilder {
     pub(crate) fn build(&self) -> Client {
         Client {
             config: self.config.clone(),
-            url_to_index: self.base_url.with_segments([&self.config.index_name]),
+            url_to_index: self
+                .base_url
+                .with_segments([&self.config.index_name])
+                .into(),
             client: self.client.clone(),
         }
     }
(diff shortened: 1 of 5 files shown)


web-api/src/storage/elastic/client.rs
```diff
@@ -0,0 +1,297 @@
...
+            url_to_index: self
+                .base_url
+                .with_segments([&self.config.index_name])
+                .into(),
+            client: self.client.clone(),
+        }
+    }
+}
+
+#[derive(Debug, Serialize)]
```
rustonaut, line 93: This is mostly just moved except that:

- we a generic over `DocumentId` (it will be moved to `web-api-shard` where there is no document id
- has a custom error type instead of `InternalError` (**to be done**)

web-api/src/storage/elastic/client.rs
```diff
@@ -0,0 +1,315 @@
...
+}
+
+impl TryFrom<Url> for SegmentableUrl {
+    type Error = anyhow::Error;
+
+    fn try_from(mut url: Url) -> Result<Self, Self::Error> {
+        url.path_segments_mut()
+            .map_err(|()| anyhow::anyhow!("non segmentable url"))?;
+        Ok(Self(url))
+    }
```
janpetschexain, line 306: 🔧 since we don't actually need the segments here, but instead just want to test for it, we could do the test right away:
```suggestion
    fn try_from(url: Url) -> Result<Self, Self::Error> {
        if url.cannot_be_a_base() {
            Err(anyhow::anyhow!("non segmentable url"))
        } else {
            Ok(Self(url))
        }
    }
```

web-api/src/storage/elastic/client.rs
```diff
@@ -0,0 +1,315 @@
...
+                .with_segments([&self.config.index_name])
+                .into(),
+            client: self.client.clone(),
+        }
+    }
+}
+
+#[derive(Debug, Serialize)]
+#[serde(rename_all = "lowercase")]
+pub(super) enum BulkInstruction<'a, D> {
```
janpetschexain, line 100: :art: i guess the `D` comes from `DocumentId`, but since we made it generic to be independent of the actual type, should we name the generic parameter sth like `I` or `ID` to make clear that it stands for the id?

web-api/src/storage/elastic/client.rs
```diff
@@ -0,0 +1,315 @@
...
+
+        self.query_with_bytes(url, post_data).await
+    }
+}
+
+#[derive(Debug, Error, displaydoc::Display, From)]
+pub(crate) enum Error {
+    /// Transmitting a request or receiving the response failed: {0}
+    Transport(reqwest::Error),
+    /// Elastic Search failed, status={status}, url={url}, \nbody={error}
```
janpetschexain, line 271: :art: since this will eventually appear in some server logs we should probably avoid newlines here, iirc this was already an issue in some other part of the code.
```suggestion
    /// Elastic Search failed, status={status}, url={url}, body={error}
```

rustonaut: I would argue you do want newlines in long error messages, it's supper common in all kind of software including e.g. what gets logged to Linux system logs.

rustonaut: :pushpin: outside of the scope of this PR

janpetschexain: iirc handling newlines wasn't an issue, it was more about the visual presentation of the logs.

janpetschexain (approved):
lgtm

rustonaut:
bors merge

PR #893, opened 2023-04-24
[Merged by Bors] - Unify release build

acrrd:
Unify the two ci to build the image of the service in one. Remove unused workflow to build other images.
When you create an image both services will be built, they will have a different image name but the same tag.
The workflow allows you to specify:
 * platform to build for,
 * model name and version to use in the image

It automatically creates a tag that describes what is in the image, when it was created and from which commit. It still allows overriding the default with a custom tag.


acrrd:
Build for arm64: https://github.com/xaynetwork/xayn_discovery_engine/actions/runs/4788721240
Build for amd64: https://github.com/xaynetwork/xayn_discovery_engine/actions/runs/4788724437

.github/workflows/web-service.yml
```diff
@@ -30,96 +45,105 @@ jobs:
...
-          tar -xvf web-service.tar
-          cd ./web-service
           TAG=${{ github.event.inputs.tag }}
           if [ -z "$TAG" ]; then
-            TAG="branch-$(git rev-parse --abbrev-ref HEAD)"
-            if [[ "${TAG}" == "branch-HEAD" ]]; then
-              TAG="$(git rev-parse --short HEAD)"
-            fi
+            TAG="$(git rev-parse --abbrev-ref HEAD)"
+            TAG="$TAG-$(date +"%y%W%u%H%M%S")"
```
janpetschexain, line 111: ❓ this creates sth like `23171163035` which is not immediately clear to read to me. wdyt about a `YMDHMS` timestamp instead?

.github/workflows/web-service.yml
```diff
@@ -30,96 +45,105 @@ jobs:
...
+            TAG="$TAG-$(git rev-parse --short HEAD)"
```
janpetschexain, line 112: ❓ even though both git parses use different options, it will usually just result in the same string, like `unify_release_build-23171163035-unify_release_build`. do we actually need both or shouldn't the first be enough?

.github/workflows/web-service.yml
```diff
@@ -30,96 +45,105 @@ jobs:
...
-        uses: actions/download-artifact@9bc31d5ccc31df68ecc42ccf4149144866c47d8a # v3.0.2
+      - name: Set up QEMU
+        if: inputs.platform == 'arm64'
+        uses: docker/setup-qemu-action@e81a89b1732b9c48d79cd809d8d81d79c4647a18 # v2.1.0
         with:
-          name: web-service
+          platforms: ${{ inputs.platform }}
 
-      - name: Create docker image
+      - name: Images name
```
janpetschexain, line 106: :art: typo
```suggestion
      - name: Image names
```

.github/workflows/web-service.yml
```diff
@@ -30,96 +45,105 @@ jobs:
...
-      - name: Setup job
-        uses: ./.github/actions/setup-job-docker
-
-      - name: Build aarch64 artifacts
-        env:
-          CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER: aarch64-linux-gnu-gcc
-          CC_aarch64_unknown_linux_gnu: aarch64-linux-gnu-gcc
-          CXX_aarch64_unknown_linux_gnu: aarch64-linux-gnu-g++
+      - name: Set up aarch64 environment
+        if: inputs.platform == 'arm64'
```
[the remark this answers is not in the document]
acrrd: The doc do not uses it, if you check the build is working correctly.
https://docs.github.com/en/actions/using-jobs/using-conditions-to-control-job-execution

.github/workflows/web-service.yml
acrrd: I wanted to have it short because the tag can is limited to 128. I we can go with yMDHMS this will correspond to
230425180410 that is actually only on char longer.

.github/actions/release_artifact/action.yml
```diff
@@ -0,0 +1,72 @@
...
+          echo "Unsupported platform ${{ inputs.platform }}"
+          exit 1
+        fi
+
+        echo "build_target=$BUILD_TARGET" >> $GITHUB_ENV
+
+    - name: Build service
+      shell: bash
+      run: |
+        cargo build --release --bin ${{ inputs.bin_name }} --target ${{ env.build_target }}
```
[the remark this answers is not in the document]
acrrd: I like it more like it is, it is uniform with the other and it will produce a smaller diff if we need to change it.

.github/workflows/web-service.yml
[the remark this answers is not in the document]
janpetschexain: good to know 👍 maybe i looked at an older doc version before.

janpetschexain (approved):
lgtm

acrrd:
bots merge

PR #896, opened 2023-04-25
[Merged by Bors] - ET-4380 Move summarizer here from https://github.com/xaynetwork/xayn_summarizer

acrrd:
[ET-4380]

Move the summarizer crate in this repository.
Only minimal change to comply with the lineters have been applied.

[ET-4380]: https://xainag.atlassian.net/browse/ET-4380?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #899, opened 2023-04-26
[Merged by Bors] - ET-4382 summarize during ingestion

janpetschexain:
**Reference**

- [ET-4382]
- requires #896

**Summary**

- summarize document snippets during ingestion if the flag is set

[ET-4382]: https://xainag.atlassian.net/browse/ET-4382?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/web-api/src/ingestion/routes.rs
+++ b/web-api/src/ingestion/routes.rs
@@ -108,7 +108,7 @@ struct UnvalidatedIngestedDocument {
 #[derive(Debug)]
 struct IngestedDocument {
     id: DocumentId,
-    snippet: Snippet,
+    snippet: String,
     properties: DocumentProperties,
     tags: Vec<DocumentTag>,
     is_candidate_op: IsCandidateOp,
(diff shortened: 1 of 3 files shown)


web-api/src/ingestion/routes.rs
```diff
@@ -159,6 +193,11 @@ impl UnvalidatedIngestedDocument {
     fn validate(self) -> Result<IngestedDocument, DocumentIdAsObject> {
         let validate = || -> anyhow::Result<_> {
             let id = self.id.as_str().try_into()?;
+            let snippet = if self.summarize {
+                Snippet::Raw(self.snippet)
```
acrrd, line 197: The summarizer is returning `Result<String>` atm but it cannot really fail. I can change that and I would be more to call it here than having `Snippet`. In the end this is already a transformation and not only validation.
wdyt?

janpetschexain: i wasn't sure if we should put the summarization into the validation, but since you're ok with that i'll change it as it simplifies the code.

acrrd: Would be better to have validation and transformation as separate steps but it will add overhead. Since the summarizer cannot trigger an error I think is better to have it here than to be forced to always have mut document to be able to read the snippet.

acrrd (approved):

janpetschexain:
bors merge

janpetschexain:
bors merge

PR #901, opened 2023-04-27
[Merged by Bors] - Make pool size configurable

acrrd:
Make pool size configurable. At the moment only the minimum size is configurable and it is also the maximum since we don't have any automatic mechanism that close a connection.

The default value has been incremented from 10 to 25.


Robert-Steiner (approved):

acrrd:
bors merge

PR #909, opened 2023-05-03
[Merged by Bors] - fix problem with not setup legacy tenant

rustonaut:
We should never hit that in practice with how we currently do things.

But better safe then sorry.

### DIFF
--- a/web-api-db-ctrl/src/elastic.rs
+++ b/web-api-db-ctrl/src/elastic.rs
@@ -13,7 +13,7 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 use once_cell::sync::Lazy;
-use reqwest::Method;
+use reqwest::{Method, StatusCode};
 use serde_json::{json, Value};
 use xayn_web_api_shared::{elastic::Client, request::TenantId};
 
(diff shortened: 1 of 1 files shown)


web-api-db-ctrl/src/elastic.rs
```diff
@@ -48,21 +48,35 @@ pub(crate) async fn setup_legacy_tenant(
...
-              "add": {
-                "index": default_index,
-                "alias": tenant_id,
-              }
-            }
-          ]
-        }))
+    let response = elastic
+        .with_index(default_index)
+        .request(Method::GET, [], [])
```
janpetschexain, line 53: ❓ should we use a `Method::HEAD` request instead to have an even more light-weight "ping" response?
```suggestion
        .request(Method::HEAD, [], [])
```

janpetschexain (approved):
lgtm

web-api-db-ctrl/src/elastic.rs
rustonaut: during some manual testing this didn't work correctly once for me as elastic did hung on the response for whatever reason :see_no_evil: 

also this is done once during startup, the size difference of the response really doesn't matter

rustonaut:
bors merge

PR #914, opened 2023-05-03
[Merged by Bors] - ET-4097 use appropriate log levels

janpetschexain:
**Reference**

- [ET-4097]
- followed by #920

**Summary**

- raise some of the log levels
- assign a log level for each application error
- improve some log messages

[ET-4097]: https://xainag.atlassian.net/browse/ET-4097?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


web-api/src/error/application.rs
```diff
@@ -46,16 +45,30 @@ where
...
 
+/// Workaround to use a non-constant `tracing::Level` in the `tracing::event!` macro.
+macro_rules! application_event {
+    ($level:expr, $($fields:tt)*) => {{
+        match $level {
+            ::tracing::Level::TRACE => ::tracing::trace!($($fields)*),
+            ::tracing::Level::DEBUG => ::tracing::debug!($($fields)*),
+            ::tracing::Level::INFO => ::tracing::info!($($fields)*),
+            ::tracing::Level::WARN => ::tracing::warn!($($fields)*),
+            ::tracing::Level::ERROR => ::tracing::error!($($fields)*),
```
janpetschexain, line 56: unfortunately we can't use non-constant levels in the `event!` macros. we also can't turn the `ApplicationError::level()` method into an associated constant because of trait object safety.

i wrapped this in a macro to avoid to much bloat, if someone has a nicer solution then i'm happy to change this.

web-api/src/personalization/routes.rs
```diff
@@ -542,7 +543,7 @@ async fn semantic_search(
         documents: documents.into_iter().map_into().collect(),
     }));
 
-    tracing::info!(elapsed_time = request_start.elapsed().as_millis());
+    debug!(elapsed_time = request_start.elapsed().as_millis());
```
acrrd, line 546: This should stay `info` to avoid having to change the level of the whole application just to get some data from this. It will be removed.

integration-tests/src/lib.rs
```diff
@@ -237,7 +237,7 @@ where
     let config = config::load_with_args([0u8; 0], args);
 
     start::<A>(config)
-        .instrument(info_span!("test", test_id = %services.id))
+        .instrument(error_span!("test", test_id = %services.id))
```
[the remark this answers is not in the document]
rustonaut: it's not logging an error,

it' adding a span (i.e. information which will be included when logging an event) with the error log level

the reason we do so is because we want the span always be enabled, e.g. if it is `info` and we only log errors when logging an error it would not include the span

I did the same change in the draft PR I did to do some additional testing for the multi tenant code.

The reason we do add the test_id as a span is to know to which tests logs belong, but it just works soso as spans don't span across spawned futures.

web-api/src/error/common.rs
```diff
@@ -22,101 +22,99 @@ use derive_more::From;
...
+impl_application_error!(HistoryTooSmall => BAD_REQUEST, INFO);
 
 /// Custom error for 400 Bad Request status code.
 #[derive(Debug, Error, Display, Serialize, From)]
 pub(crate) struct BadRequest {
     pub(crate) message: Cow<'static, str>,
 }
 
-impl_application_error!(BadRequest => BAD_REQUEST);
+impl_application_error!(BadRequest => BAD_REQUEST, INFO);
```
rustonaut, line 117: :question:  is info the right choice for this kind of wrong API usage errors, normally you would log them at a debug level

web-api/src/net.rs
```diff
@@ -125,7 +125,7 @@ where
     // `spawn` it on the tokio runtime. This hands off the responsibility to poll the server to
     // tokio. At the same time we keep the `JoinHandle` so that we can wait for the server to
     // stop and get it's return value (we don't have to await `term_handle`).
-    let span = info_span!(parent: None, "web-api");
+    let span = error_span!(parent: None, "web-api");
     span.follows_from(Span::current());
```
rustonaut, line 129: :pushpin:  while this the correct way to do it, with how we currently log things this won't get logged in any usefull way :see_no_evil: 

This means we e.g. can't look up the test_id of the "current" span when logging event in the server.

web-api/src/error/common.rs
janpetschexain: just discussed this with @acrrd , we decided to keep it at `info` level for better visibility in case some customer uses the api in a wrong way again.

rustonaut (approved):
lgtm

janpetschexain:
bors merge

PR #917, opened 2023-05-03
[Merged by Bors] - Update ci image

acrrd:
This was created to install `lld` but since is a bit of a pain to update the image I fitted in it multiple changes.

* Use `bookworm-20230502-slim` instead of debian stable. Bookworm is now in hard freeze and it will become the new stable in summer. It allows us to install awscli from the package manager and to have an up-to-date toolchain that is more similar to what we have on our machine. We were already using bookworm for the image that we use to deploy. 
* Install `lld` so that we can use it later on
* Install the packages and target that are needed to compile for `arm64`
* Update to rustup 1.26 and rust 1.69
* Move the code that generates the image to a justfile task to be able to test it more easily

### DIFF
--- a/.github/docker/Dockerfile.ci-image
+++ b/.github/docker/Dockerfile.ci-image
@@ -1,4 +1,4 @@
-FROM debian:11.6-slim
+FROM debian:bookworm-20230502-slim
 
 ARG rust_version
 ARG just_version
@@ -13,13 +13,10 @@ RUN set -eux; \
     apt-get update; \
     apt-get install -y --no-install-recommends \
       ca-certificates wget curl git zstd gcc libc6-dev \ 
-      nodejs npm postgresql-client python3.9 python3-pip \
-      unzip; \
+      nodejs npm postgresql-client python3.11 python3-pip \
+      unzip lld awscli g++-aarch64-linux-gnu libc6-dev-arm64-cross \
+      ; \
     rm -rf /var/lib/apt/lists/*;
-
-RUN curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip" && \
-    unzip awscliv2.zip && \
-	./aws/install
 
 # Begin: Rust base
 # Taken from https://github.com/rust-lang/docker-rust/blob/f558c4c660fd925e278da432640342de993f2bc7/1.68.0/bullseye/slim/Dockerfile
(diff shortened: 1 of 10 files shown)


.github/docker/Dockerfile.ci-image
```diff
@@ -13,14 +13,11 @@ RUN set -eux; \
     apt-get update; \
     apt-get install -y --no-install-recommends \
       ca-certificates wget curl git zstd gcc libc6-dev \ 
-      nodejs npm postgresql-client python3.9 python3-pip \
-      unzip; \
+      nodejs npm postgresql-client python3.11 python3-pip \
+      unzip lld awscli g++-aarch64-linux-gnu libc6-dev-arm64-cross \
+      ; \
```
janpetschexain, line 18: :art: unnecessary newline
```suggestion
      unzip lld awscli g++-aarch64-linux-gnu libc6-dev-arm64-cross; \
```

.github/docker/Dockerfile.ci-image
```diff
@@ -13,14 +13,11 @@ RUN set -eux; \
...
     rm -rf /var/lib/apt/lists/*;
 
-RUN curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip" && \
-    unzip awscliv2.zip && \
-	./aws/install
-
 # Begin: Rust base
 # Taken from https://github.com/rust-lang/docker-rust/blob/f558c4c660fd925e278da432640342de993f2bc7/1.68.0/bullseye/slim/Dockerfile
```
janpetschexain, line 22: 🔧 we should update the link to the blob for the current version

.github/docker/Dockerfile.ci-image
acrrd: I put it on a new line to avoid having to move it every time we add something. To me is like having a closing parenthesis on a new line but I don't have a strong opinion on this

.github/docker/Dockerfile.ci-image
acrrd: I was sure I did it :see_no_evil:

.github/docker/Dockerfile.ci-image
janpetschexain: i see, up to you to decide

acrrd:
> wrench we should also bump the `rust-version` in the workspace's `Cargo.toml`

I don't want to bump it because this code still works perfectly with 1.68. Bumping means that our own project that we didn't update yet cannot use the last version of the `bert` crate.
I'm more to remove it or to add a check that the compiler in the version specified there can actually compile the project.

janpetschexain (approved):
lgtm

acrrd:
bors merge

PR #925, opened 2023-05-04
[Merged by Bors] - ET-4418 rename X-Tenant-Id => X-Xayn-Tenant-id

rustonaut:
Rename header to avoid conclusions.

**References:**

- story [ET-4411]
- issue [ET-4418]

[ET-4411]: https://xainag.atlassian.net/browse/ET-4411?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ET-4418]: https://xainag.atlassian.net/browse/ET-4418?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/integration-tests/src/lib.rs
+++ b/integration-tests/src/lib.rs
@@ -168,7 +168,7 @@ pub async fn test_two_apps<A1, A2, F>(
 fn build_client(services: &Services) -> Arc<Client> {
     let mut default_headers = HeaderMap::default();
     default_headers.insert(
-        "X-Tenant-Id",
+        "X-Xayn-Tenant-Id",
         services.tenant_id.as_str().try_into().unwrap(),
     );
     Arc::new(
--- a/web-api/src/middleware/request_context.rs
+++ b/web-api/src/middleware/request_context.rs
@@ -209,7 +209,7 @@ where
     )
 }
 
-const TENANT_ID_HEADER: &str = "X-Tenant-Id";
+const TENANT_ID_HEADER: &str = "X-Xayn-Tenant-Id";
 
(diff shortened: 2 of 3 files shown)


rustonaut:
> It will avoid confusion in the code

It's not about confusion in the code but about middleware in out pipeline setting this header with a slightly different meaning. Except depending on various decisions our Infra Team  might make or not make it could not be a different meaning in the future for multi tenant deployments, we will have to see :shrug: 

So we might need to change it in the future again, we will see.

But it's a tenant id so we probably want to keep that naming part as far as possible.

Multi-app setups are a different approach (in my understanding) which can also be used to archive multi tenancy but AFIK we are currently going in a different direction so I would want to avoid such naming.

rustonaut:
bors merge

acrrd:
> It's not about confusion in the code but about middleware in out pipeline setting this header with a slightly different meaning. Except depending on various decisions our Infra Team might make or not make, it could not be a different meaning in the future for multi tenant deployments. We will have to see shrug
 
 We have to read it in the code and the debug request so it can create confusion for us.

rustonaut:
> We have to read it in the code and the debug request so it can create confusion for us.

we want a tenant id it gives us a tenant id where is the confusions?

rustonaut:
bors merge

PR #926, opened 2023-05-04
Do not give tags score to documents that do not have it

acrrd:
The tag score of a document is decided by its position in a sorted (by tag weight) array.
When we have documents with tag weight `0` (that means that they do not match the user's tag) or if two documents have the exact count, we are "randomly" and arbitrarily associate different scores to them based on their position in the input.

This PR tries to mitigate the case in which documents have `0` weight. In this case, the no-weight documents are assigned `0` as a score instead of a score based on their position. As a side-effect of this means that if one document has a weight and the others don't, it will be significantly boosted. For this reason, I reduced the weight of the tag score from `0.5` to  `0.3`.

### DIFF
--- a/web-api/src/personalization.rs
+++ b/web-api/src/personalization.rs
@@ -101,7 +101,7 @@ impl Default for PersonalizationConfig {
             default_number_documents: 10,
             // FIXME: what is a default value we know works well with how we do knn?
             max_cois_for_knn: 10,
-            score_weights: [0.5, 0.5, 0.0],
+            score_weights: [0.7, 0.3, 0.0],
             store_user_history: true,
             max_stateless_history_size: 200,
             max_stateless_history_for_cois: 20,
--- a/web-api/src/personalization/rerank.rs
+++ b/web-api/src/personalization/rerank.rs
@@ -63,17 +63,28 @@ fn rerank_by_tag_weight(
     documents: &[PersonalizedDocument],
     tag_weights: &HashMap<DocumentTag, usize>,
 ) -> HashMap<DocumentId, f32> {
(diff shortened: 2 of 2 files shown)


web-api/src/personalization.rs
```diff
@@ -101,7 +101,7 @@ impl Default for PersonalizationConfig {
             default_number_documents: 10,
             // FIXME: what is a default value we know works well with how we do knn?
             max_cois_for_knn: 10,
-            score_weights: [0.5, 0.5, 0.0],
+            score_weights: [0.7, 0.3, 0.0],
```
janpetschexain, line 104: ❓ i guess we should reduce the default tag score weight in the `SemanticSearchConfig` as well for the same reason?

web-api/src/personalization/rerank.rs
```diff
@@ -63,17 +63,28 @@ fn rerank_by_tag_weight(
...
+        })
+        .filter(|(_, w)| *w != 0);
+
+    let ranked_documents = rank_keys_by_score(weighted_documents, |w1, w2| w1.cmp(w2).reverse());
 
-    rank_keys_by_score(weighted_documents, |w1, w2| w1.cmp(w2).reverse())
+    documents
+        .iter()
+        .fold(ranked_documents, |mut ranked_documents, doc| {
+            ranked_documents.entry(doc.id.clone()).or_insert(0.0);
```
janpetschexain, line 84: :art:
```suggestion
            ranked_documents.entry(doc.id.clone()).or_default();
```

janpetschexain (approved):
good catch 👍

PR #929, opened 2023-05-08
[Merged by Bors] - ET-4443 add unit tests for rerank module

janpetschexain:
**Reference**

- [ET-4443]

**Summary**

- add unit tests for rerank module

[ET-4443]: https://xainag.atlassian.net/browse/ET-4443?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/web-api/src/personalization/rerank.rs
+++ b/web-api/src/personalization/rerank.rs
@@ -219,11 +219,11 @@ mod tests {
         PositiveCoi { id, point, stats }
     }
 
-    fn sort_ids(scores: HashMap<DocumentId, f32>) -> Vec<String> {
+    fn ids_by_score(scores: &HashMap<DocumentId, f32>) -> Vec<&str> {
         scores
-            .into_iter()
+            .iter()
             .sorted_by(|(_, s1), (_, s2)| nan_safe_f32_cmp_desc(s1, s2))
-            .map(|(document, _)| document.to_string())
+            .map(|(document, _)| document.as_ref().as_str())
             .collect()
     }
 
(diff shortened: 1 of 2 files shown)


web-api/src/personalization/rerank.rs
```diff
@@ -163,3 +163,190 @@ pub fn bench_rerank<S>(
...
+        assert_eq!(
+            keys[0..=1].iter().copied().collect::<HashSet<_>>(),
+            [4, 5].into(),
+        );
+        assert!(keys[0] == 4 || keys[0] == 5);
+        assert!(keys[1] == 4 || keys[1] == 5);
+        assert_eq!(keys[2], 3);
+        assert!(keys[3] == 0 || keys[3] == 1 || keys[3] == 2);
+        assert!(keys[4] == 0 || keys[4] == 1 || keys[4] == 2);
+        assert!(keys[5] == 0 || keys[5] == 1 || keys[5] == 2);
```
rustonaut, line 252: :art: I think using the hash sets is good enough

```suggestion
        assert_eq!(keys[2], 3);
        assert_eq!(
            keys[3..=5].iter().copied().collect::<HashSet<_>>(),
            [0, 1, 2].into(),
        );
```

or collect_vec + sorted?

also while it doesn't matter much here I think a `assert_eq!(keys.len(), 6);` at the end would be nice

web-api/src/personalization/rerank.rs
```diff
@@ -163,3 +163,190 @@ pub fn bench_rerank<S>(
...
+    }
+
+    #[test]
+    fn test_rank_key_by_different_scores() {
+        let scores = (0..=6).enumerate().map(|(score, key)| (key, score));
```
rustonaut, line 257: :art: this is fine but a bit confusing, shouldn't the enumerate be the key instead of the score, and arent they the same :thinking:  If we just want arbitrary pairs I think  using iter + map would be a bit more clear then using enumerate.

web-api/src/personalization/rerank.rs
```diff
@@ -163,3 +163,190 @@ pub fn bench_rerank<S>(
...
+        let stats = CoiStats {
+            view_count: i + 1,
+            view_time: Duration::ZERO,
+            last_view: time,
+        };
+
+        PositiveCoi { id, point, stats }
+    }
+
+    fn sort_ids(scores: HashMap<DocumentId, f32>) -> Vec<String> {
```
rustonaut, line 222: :art:  I noticed that when reading the tests I was thinking "if I sort the id don't I always get the same order?"

Maybe renaming it  it to `ids_sorted_by_score` or `ids_by_score` or similar would be a bit better :thinking:

rustonaut (commented):
looks mostly good to me just minor things

rustonaut (approved):
lgtm

PR #934, opened 2023-05-09
Set http request timeout

acrrd:


web-api/src/storage/elastic/client.rs
```diff
@@ -45,6 +46,14 @@ pub(crate) struct Config {
...
     password: Secret<String>,
     index_name: String,
+
+    /// Request timeout in seconds.
+    #[serde(with = "serde_duration_as_secs", default = "default_timeout")]
+    timeout: Duration,
+}
+
+fn default_timeout() -> Duration {
+    Duration::from_secs(2)
```
rustonaut, line 56: :wrench:  Just set the default in `impl Default` the `serde(default)` container attribute will grab it from there

web-api/src/utils.rs
```diff
@@ -53,3 +53,22 @@ pub(crate) fn serialize_to_ndjson(
     }
     Ok(body)
 }
+
+pub(crate) mod serde_duration_as_secs {
```
rustonaut, line 57: :wrench:  Why rename this just to save 4 characters in name which anyway gets auto-completed.

rustonaut (changes requested):

web-api/src/utils.rs
acrrd: Is shorter, faster to read, and uniform with `as_secs`/`from_secs` from `Duration`.

rustonaut: `as_secs`/`from_secs` are like `FromStr::Err` named in a way they likely wouldn't be named anymore, just around rust ~1.3(?) days there was much less clearity about naming schema.

And just because other places use bad style doesn't mean we have to do so, too.

rustonaut: E.g. chrono does use seconds, nanoseconds, etc. (except in one helper trait for rounding time :shrug: )

acrrd: I'll change it I don't care about it too much.
I don't think is a bad convention, it is clear and completely unambiguous to everyone, with no drawbacks. Since it is used in a literal string with serde is not touched by rust-analyzer and we have to write more for no reason.

rustonaut: rust analizer works for string literals in macros which are turned in idents, or at least in VSCode it does

rustonaut (approved):
lgtm

web-api/src/utils.rs
acrrd: rust-analyzer 1.69.0 (84c898d 2023-04-16) is not able to rename or autocomplete in those. Do you use any particular option?

PR #907, opened 2023-05-02
[Merged by Bors] - ET-4294 tenant specific testing

rustonaut:
- [x] by default run tests with created tenants
- [x] allow configuring the usage of legacy tenant for tests
- [x] test legacy migration, but not to the degree the follow up PR does
- [x] better logging
- [x] no longer setup/tear down test dbs in justfile
- [x] better CI/in-container detection
- [x] use `web-api-db-ctrl::Silo` to setup/teardown test env   
  
**References:**

- story [ET-4344]
- issue [ET-4294]
- based on #919 based on #941
- followed by #942

[ET-4344]: https://xainag.atlassian.net/browse/ET-4344?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ET-4294]: https://xainag.atlassian.net/browse/ET-4294?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #941, opened 2023-05-10
[Merged by Bors] - ET-4413 (1/2) preperation for migration tests  

rustonaut:
- [X] decouples starting applications from logging initializing
  - currently we use a `static Once` to make sure logging is not initialized twice
  - but if we include two different versions of the application into one integration test we now have two Once cells but still only one logging facility, so it will fail :  
    -  we always want to init the logging with the newest dispatcher setup, never with the one from the older version we pull in for migration tests 
  - using `try_init()` for logging also has some issues, e.g. when writing logs to files you might truncate multiple times instead of once (as this is done before the `try_init()`)
  - additionally it is now a bit more clean in the sense that before you had the illusion you could configure logging for each application in a test, but it was once for all applications across all tests, so configuring any `[logging]` parameter in a test could lead to unstable arbitrary feeling results
  - a even better solution would be to use scoped logging dispatchers instead of global ones, that would not only allow us to have the logs setup per test and properly not printed if the test doesn't fail but also to change logging config per test. But it's outside of the scope of this PR.

- [x] include the ES schema in the `web-api` lib so that we can use it in migration tests
  - the build release binaries should strip it out through dead code analysis

**References:**

- issue [ET-4413]
- story [ET-4411]
- followed by #919 then #907 then #942

[ET-4413]: https://xainag.atlassian.net/browse/ET-4413?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,7 @@ members = [
 ]
 
 [workspace.package]
-version = "0.1.0"
+version = "0.1.1"
 edition = "2021"
 rust-version = "1.68.0"
 license = "AGPL-3.0-only"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,7 +35,7 @@ thiserror = "1.0.39"
 tokio = { version = "1.26.0", default-features = false }
 toml = "0.7.3"
 tracing = "0.1.37"
-tracing-subscriber = { version = "0.3.16", features = ["env-filter", "json"] }
+tracing-subscriber = { version = "0.3.17", features = ["env-filter", "json"] }
 uuid = { version = "1.3.0", features = ["serde", "v4"] }
 
 [profile.dev.package.tract-core]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,6 +35,7 @@ thiserror = "1.0.39"
 tokio = { version = "1.26.0", default-features = false }
 toml = "0.7.3"
 tracing = "0.1.37"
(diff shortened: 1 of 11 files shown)


Cargo.toml
```diff
@@ -35,6 +35,7 @@ thiserror = "1.0.39"
 tokio = { version = "1.26.0", default-features = false }
 toml = "0.7.3"
 tracing = "0.1.37"
+tracing-subscriber = { version = "0.3.16", features = ["env-filter", "json"] }
```
janpetschexain, line 38: :art: new version
```suggestion
tracing-subscriber = { version = "0.3.17", features = ["env-filter", "json"] }
```

web-api/src/logging.rs
```diff
@@ -70,16 +70,19 @@ impl Default for Config {
...
-pub(crate) fn init_tracing(log_config: &Config) {
-    INIT_TRACING.call_once(|| {
-        init_tracing_once(log_config);
-        init_panic_logging();
-    });
+/// Initializes the logging.
+///
+/// This should be called before [`crate::start()`] is called.
+///
+/// Even through this returns an error if logging was already initialized you
```
janpetschexain, line 77: :art: typo
```suggestion
/// Even though this returns an error if logging was already initialized you
```

rustonaut:
bors merge

PR #942, opened 2023-05-11
ET-4413 added migration integration test

rustonaut:
Add a migration integration test which first start up a older version of this library runs some ingestion and checks it then stops the older version, revoke permissions from public and start the current version checking if all migrations work as expected.

**References:**

- issue [ET-4413]
- story [ET-4344]


web-api/tests/legacy_tenant_migration.rs
```diff
@@ -132,3 +142,163 @@ async fn test_if_the_initializations_works_correct_for_not_setup_legacy_tenants(
...
+
+    let args = &[
+        "integration-test",
+        "--bind-to",
+        "127.0.0.1:0",
+        "--config",
+        &format!("inline:{config}"),
+    ];
+
+    dbg!(&args);
```
janpetschexain, line 194: ❓ is this a leftover from local testing?

rustonaut: yesn't if the test fails we very likely want to have  this and if it doesn't fail the test output capturing will capture it

janpetschexain (approved):
lgtm

web-api/tests/legacy_tenant_migration.rs
rustonaut: I removed it but doesn't your IDE by default collapse tests which pass?  I mean IDE always displaying all output of all tests no matter weather the test passes or not seems quite bad design IMHO.

rustonaut:
bors merge

PR #949, opened 2023-05-22
[Merged by Bors] - ET-4486 per integration test logging

rustonaut:
This PR does mainly two things:

-  use `TestWriter` to make log capturing work again
- use per-test non-global `Dispatch` 

As a side effect of this the creation of the async runtime is moved into the `test_app`/`test_two_app` scaffold, sadly this leads to ~3 line changes per integration test :sweat: 

**References:**

- task [ET-4486]
- story [ET-4485]
- followed by #950 

[ET-4486]: https://xainag.atlassian.net/browse/ET-4486?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ET-4485]: https://xainag.atlassian.net/browse/ET-4485?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #950, opened 2023-05-22
ET-4487 per-test temp dir

rustonaut:
- have a temp dir per-test which is cleared if the test doesn't fail
- allow writing logs as JSON to a file in the temp-dir (on a per-test basis)
- allow different filter for stdout and file logging (filter using `EnvFilter` directives)
- make sure the log => tracing facade works correctly

**References:**

- issue: [ET-4487]
- story: [Et-4485]
- based on #949


integration-tests/src/lib.rs
```diff
@@ -140,55 +188,153 @@ where
...
-            .init()
+        let env_filter = env_opt_var_os(XAYN_TEST_FALLBACK_LOG)
+            .unwrap_or_else(|| "warn".into())
+            .parse::<EnvFilter>()
+            .expect("XAYN_TEST_FALLBACK_LOG has invalid EnvFilter directives");
+
+        let max_level = env_filter.max_level_hint();
+
+        // WARNING: You can not use `.init()` here as it will setup the log=>tracing
+        //          facade in a way which yields incorrect results.
```
rustonaut, line 199: `.init()` setup the the `log` => `tracing` facade with the max level of the global logger, but we need to use the max level of _any_ global or non global  filter or log entries will be missing, especially given that we default to global fallback logger to `"warn"`

integration-tests/src/lib.rs
```diff
@@ -140,55 +188,153 @@ where
...
+        XAYN_TEST_LOG,
+    )
 });
 
-pub fn initialize_local_test_logging(_test_id: &str) -> Dispatch {
+static LOG_STDOUT_ENV_FILTER: Lazy<Option<(String, Option<LevelFilter>)>> =
+    Lazy::new(|| additional_env_filter(XAYN_TEST_STDOUT_LOG, "true", &LOG_ENV_FILTER.0));
+
+static LOG_FILE_ENV_FILTER: Lazy<Option<(String, Option<LevelFilter>)>> =
+    Lazy::new(|| additional_env_filter(XAYN_TEST_FILE_LOG, "false", &LOG_ENV_FILTER.0));
```
rustonaut, line 291: The next PR also adds `LOG_FLAME_ENV_FILTER`, because e.g. you might want to collect `trace` logs for flame graphs or charts.

rustonaut:
This turned out more use-full then the flame graph follow up PR, as the problem didn't show up in the flame graph but in some (trivial) manual analysis of the events logged to a file using `XAYN_TEST_FILE_LOG` and `XAYN_TEST_KEEP_TEMP_DIRS`.

web-api/tests/document_candidates.rs
```diff
@@ -19,10 +19,9 @@ use reqwest::{Client, StatusCode, Url};
 use serde::Deserialize;
 use serde_json::{json, Value};
 use xayn_integration_tests::{send_assert, send_assert_json, test_app, UNCHANGED_CONFIG};
-use xayn_test_utils::error::Panic;
 use xayn_web_api::Ingestion;
 
-async fn ingest(client: &Client, url: &Url) -> Result<(), Panic> {
+async fn ingest(client: &Client, url: &Url) -> Result<(), anyhow::Error> {
```
janpetschexain, line 24: :art: reuse via import, similarly for the other occurences
```suggestion
use anyhow::Error;

async fn ingest(client: &Client, url: &Url) -> Result<(), Error> {
```

integration-tests/src/lib.rs
```diff
@@ -140,55 +227,157 @@ where
...
-    F: Future,
+    F: Future<Output = Result<(), Error>>,
 {
-    let subscriber = initialize_local_test_logging(test_id);
+    let test_id = TestId::generate();
+    let subscriber = initialize_local_test_logging(&test_id);
 
     dispatcher::with_default(&subscriber, || {
-        let body = body.instrument(error_span!(parent: None, "test", %test_id));
+        let _guard = DeleteTemptDirIfNoPanic {
```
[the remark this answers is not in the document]
rustonaut: scoping rules require that the variables lives until the end of the scope,  through I would love to add a explicit `drop` to avoid confusion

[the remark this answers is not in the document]
rustonaut: one exception is `_` which simplified is a way to say "don't bind this to any variable"

janpetschexain (approved):
lgtm

PR #957, opened 2023-05-25
[Merged by Bors] - ET-4531 Use xaynia2 as default model

acrrd:
Use Xaynia 2 as the default model of the model.
Unfortunately changing the token size will be a breaking change for some deployments so it cannot be done in this PR but we need to wait for each deployment to configure that explicitly. Later we should also remove defaults that can easily break backward compatibility.


rustonaut (approved):
lgtm

acrrd:
bors merge

acrrd:
bors merge

PR #958, opened 2023-05-25
[Merged by Bors] -  ET-4521 removed old test index setup

rustonaut:
Removes the old method of setting up the ES mapping for testing implicitly through an additional service in the compose file.

We don't use it anymore and it can hide issues with tests.


janpetschexain (approved):
lgtm

rustonaut:
bors merge

PR #967, opened 2023-06-06
ET-4538 refactor ES  search code

rustonaut:
Refecators the ES search code to be better readable.

There are a number of TODOs where I wasn't clear if something was a bug or intended behavior.

### DIFF
--- a/web-api/src/personalization/knn.rs
+++ b/web-api/src/personalization/knn.rs
@@ -28,9 +28,9 @@ use crate::{
 };
 
 /// KNN search based on Centers of Interest.
-pub(super) struct CoiSearch<I, J> {
+pub(super) struct CoiSearch<'a, I> {
     pub(super) interests: I,
-    pub(super) excluded: J,
+    pub(super) excluded: &'a [DocumentId],
     pub(super) horizon: Duration,
     pub(super) max_cois: usize,
     pub(super) count: usize,
@@ -38,12 +38,10 @@ pub(super) struct CoiSearch<I, J> {
     pub(super) time: DateTime<Utc>,
 }
 
-impl<'a, I, J> CoiSearch<I, J>
+impl<'a, I> CoiSearch<'a, I>
 where
     I: IntoIterator,
     <I as IntoIterator>::IntoIter: Clone + Iterator<Item = &'a Coi>,
-    J: IntoIterator,
-    <J as IntoIterator>::IntoIter: Clone + ExactSizeIterator<Item = &'a DocumentId>,
 {
     /// Performs an approximate knn search for documents similar to the user interests.
     pub(super) async fn run_on(
@@ -60,7 +58,6 @@ where
 
         let weights_sum = cois.iter().map(|(_, w)| w).sum::<f32>();
 
-        let excluded = self.excluded.into_iter();
         let document_futures = cois
             .iter()
             .map(|(coi, weight)| async {
@@ -79,7 +76,7 @@ where
                 storage::Document::get_by_embedding(
                     storage,
                     KnnSearchParams {
-                        excluded: excluded.clone(),
+                        excluded: self.excluded,
                         embedding: &coi.point,
                         count,
                         num_candidates,
--- a/web-api/src/personalization/knn.rs
+++ b/web-api/src/personalization/knn.rs
@@ -23,7 +23,7 @@ use xayn_ai_coi::{compute_coi_weights, Coi};
 use crate::{
     error::common::InternalError,
     models::{DocumentId, PersonalizedDocument},
-    storage::{self, KnnSearchParams},
+    storage::{self, KnnSearchParams, SearchStrategy},
     Error,
 };
 
@@ -85,7 +85,7 @@ where
                         num_candidates,
                         published_after: self.published_after,
                         min_similarity: None,
-                        query: None,
+                        strategy: SearchStrategy::Knn,
                     },
                 )
                 .await
--- a/web-api/src/storage.rs
+++ b/web-api/src/storage.rs
@@ -55,8 +55,24 @@ pub(crate) struct KnnSearchParams<'a, I> {
     pub(crate) num_candidates: usize,
     pub(crate) published_after: Option<DateTime<Utc>>,
     pub(crate) min_similarity: Option<f32>,
-    /// An additional query which will be run in parallel with the KNN search.
-    pub(super) query: Option<&'a str>,
+    pub(super) strategy: SearchStrategy<'a>,
+}
+
+#[derive(Copy, Clone)]
+pub(crate) enum SearchStrategy<'a> {
+    Knn,
+    HybridWeighted {
+        /// An additional query which will be run in parallel with the KNN search.
+        query: &'a str,
+    },
+    #[allow(dead_code)]
+    HybridLocalRrf {
+        query: &'a str,
+    },
+    #[allow(dead_code)]
+    HybridEsRrf {
+        query: &'a str,
+    },
 }
 
 #[derive(Debug, Deref, DerefMut, From)]
--- a/web-api/src/storage.rs
+++ b/web-api/src/storage.rs
@@ -46,8 +46,8 @@ use crate::{
     Error,
 };
 
-pub(crate) struct KnnSearchParams<'a, I> {
-    pub(crate) excluded: I,
+pub(crate) struct KnnSearchParams<'a> {
+    pub(crate) excluded: &'a [DocumentId],
     pub(crate) embedding: &'a NormalizedEmbedding,
     /// The number of documents which will be returned if there are enough fitting documents.
     pub(crate) count: usize,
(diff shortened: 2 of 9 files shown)


web-api/src/storage/elastic.rs
```diff
@@ -49,103 +55,115 @@ impl Client {
...
+            knn_object,
+            generic_parameters,
+            inner_filter,
+        } = create_common_knn_search_parts(params);
+
+        let knn_request = merge_json_objects([knn_object, generic_parameters.clone()]);
+
+        // TODO[pmk/now] the code originally didn't normalize the KNN request, that seems wrong
+        //      and will likely strongly discount it
+        let knn_scores = normalize_scores(self.search_request(knn_request).await?);
```
rustonaut: The code for hybrid search originally did not normalize the `knn_scores` and them merged them with a weightes sum like method with normalized bm25 scores. Given how values are very differently distributed in the 0..1 range of KNN and normalized bm25 this seems nearly guaranteed to not yield good results/implicitly penaltize KNN, at least with common values I had in the past. But idk. if this is a bug or intentional.

web-api/src/storage/elastic.rs
```diff
@@ -339,3 +357,160 @@ struct IngestedDocument<'a> {
...
+fn normalize_scores<K>(mut scores: HashMap<K, f32>) -> HashMap<K, f32>
+where
+    K: Eq + Hash,
+{
+    let max_score = scores
+        .values()
+        .max_by(|l, r| l.total_cmp(r))
+        .copied()
+        .unwrap_or_default();
+    // TODO[pmk/now] the original code did a .max(1.0); this looked like bug, verify
```
rustonaut, line 445: For some reason the max score clause for bm25 normalization had a `max(1.0)`, this seems strange as I would thing especially with values >1 we would want to normalize bm25.

web-api/src/storage/elastic.rs
```diff
@@ -339,3 +357,160 @@ struct IngestedDocument<'a> {
...
+}
+
+fn merge_scores_weighted<K>(
+    scores: impl IntoIterator<Item = (f32, HashMap<K, f32>)>,
+) -> HashMap<K, f32>
+where
+    K: Eq + Hash,
+{
+    // TODO[pmk/now] this originally didn't apply weights to scores where there was only one hit
+    //               this seemed like a bug and was removed, verify if that is correct
```
rustonaut, line 461: This isn't exactly the same algorithm as before.

- now: weight each score by 1/2
- before: weight each score by 1/2 if there are scores for both values, but don't weight it at all if there is only a score for one value

AFIK we want to favor documents appearing in both results noticable over such appearing in only one, so this seemed wrong. :question:

web-api/src/storage/elastic.rs
rustonaut: (it says outdate because I backported a refactoring from the follow up PR after making the comment, but my comment/question is not outdated)

web-api/src/storage/elastic.rs
```diff
@@ -339,3 +331,181 @@ struct IngestedDocument<'a> {
...
+
+    for score in scores.values_mut() {
+        *score /= max_score;
+    }
+
+    scores
+}
+
+fn merge_scores_weighted<K>(
+    scores: impl IntoIterator<Item = (f32, HashMap<K, f32>)>,
```
janpetschexain, line 455: ❓ should we enforce bounds on the weights (like each weight must be in `[0, 1]` and they must sum up to `1.`). maybe it's too much runtime effort to enforce it, but we should at least document it, otherwise the merged scores may not be normalized anymore.

web-api/src/storage/elastic.rs
```diff
@@ -339,3 +331,181 @@ struct IngestedDocument<'a> {
...
+where
+    K: Eq + Hash,
+{
+    scores
+        .into_iter()
+        .fold(HashMap::new(), |mut acc, (key, value)| {
+            acc.entry(key).or_default().add_assign(value);
+            acc
+        })
+}
```
janpetschexain, line 500: :art: we could make this an extension trait on `Iterator` to be able to use it more ergonomically in the functional iterator style.

web-api/src/storage/elastic.rs
```diff
@@ -339,3 +331,181 @@ struct IngestedDocument<'a> {
...
+
+fn take_highest_n_scores<K>(n: usize, scores: HashMap<K, f32>) -> HashMap<K, f32>
+where
+    K: Eq + Hash,
+{
+    scores
+        .into_iter()
+        .sorted_unstable_by(|(_, s1), (_, s2)| s1.total_cmp(s2).reverse())
+        .take(n)
+        .collect()
```
janpetschexain, line 510: :art: we should return early if `scores.len() <= n`, otherwise we sort and allocate in vain.

acrrd:
:wrench:
Can we move all the changes to the scoring system to another PR?
Refactoring shouldn´t change the business logic code especially when is something as important as the scoring system.
It is quite annoying and confusing to see the two discussions and changes intertwined.

rustonaut:
> Can we move all the changes to the scoring system to another PR?

sure

the main reason it is the way it is is because I wanted to get feedback first about  what is an intententional and what is accidental behavior before spending additional time with detangling  the code which is already a bit of a pain wrt. the follow up PR

web-api/src/storage/elastic.rs
rustonaut: I for some reason did read `value.max(1)` as "value maximally 1". :facepalm:  It's kinda an interesting case as when having `std::cmp::max` in mind you would never mistake it but also reading it a "value maximally 1" doesn't seem wrong at all either.

> in light of the other discussions/question though, we probably should apply max-normalization uniformly for all 

probably, or use rrf

web-api/src/storage/elastic.rs
rustonaut: > what i've seen so far in hybrid ranking schemes is that the weight factor is only applied if there are multiple scores to be merged

I mainly have seen rank fusion using sum+boost (e.g. average/weighted sum/etc.), RRF and some trained fusion algorithm.

web-api/src/storage/elastic.rs
rustonaut: I don't think such constraints should be part of this function as it's not a constraint of this function but the environment we might call it in at best.

web-api/src/personalization/knn.rs
```diff
@@ -23,14 +23,14 @@ use xayn_ai_coi::{compute_coi_weights, PositiveCoi};
...
+    storage::{self, KnnSearchParams, SearchStrategy},
     Error,
 };
 
 /// KNN search based on Centers of Interest.
-pub(super) struct CoiSearch<'a, I, J> {
+pub(super) struct CoiSearch<'a, I> {
     pub(super) interests: I,
-    pub(super) excluded: J,
+    pub(super) excluded: &'a [DocumentId],
```
janpetschexain, line 33: ❓ why make it less flexible? for example now we can't use iterators on hashsets anymore without reallocation.

web-api/src/storage.rs
```diff
@@ -47,20 +47,48 @@ use crate::{
     Error,
 };
 
-pub(crate) struct KnnSearchParams<'a, I> {
-    pub(crate) excluded: I,
+pub(crate) struct KnnSearchParams<'a> {
+    pub(crate) excluded: &'a [DocumentId],
```
janpetschexain, line 51: ❓ why make it less flexible? for example now we can't use iterators on hashsets anymore without reallocation.

janpetschexain (commented):
unfortunately this has quite some conflicts with the removal of the newer-than-now filter which need to be resolved first.

web-api/src/personalization/knn.rs
rustonaut: 1. Because only then can we attach some of the method you wanted to have attached to `KnnSearchParams` to it without involving additional clones of which you can't be sure to be cheap in the place they are done.

2. Because we don't use the generics but they add a bunch of complexity all through the code.

web-api/src/storage.rs
rustonaut: See other comment.

web-api/src/storage/elastic.rs
```diff
@@ -322,3 +326,186 @@ struct IngestedDocument<'a> {
...
+        KnnSearchParts {
+            knn_object,
+            generic_parameters,
+            inner_filter,
+        }
+    }
+
+    fn create_search_filter(&self) -> JsonObject {
+        let mut filter = JsonObject::new();
+        if self.excluded.is_empty().not() {
```
janpetschexain, line 360: ❓ this is quite odd, is it on purpose? otherwise we can write it in the usual way with the operator instead and also remove the `Not` import.
```suggestion
        if !self.excluded.is_empty() {
```

janpetschexain (approved):
👍

web-api/src/personalization/knn.rs
rustonaut: >  reduction of complexity across the code could also be done via a type alias.

Don't work for most of the complexity introduced here as it involves many `impl Trait`'s and complex where bounds.

I mean we went from `impl IntoIterator<Item=&'a DocumentId>` to `impl IntoIterator<IntoIter = impl ExactSizeIterator<Item = &'a DocumentId> + Clone>` .

For a situation where in all places in the code we use only things which we for "free" (deref,ref) can turn into a `&[DocumentId]` this is rather a lot of complexity in a situation where it not even clear that if we maybe use some other input the conversion to `&[DocumentId]` is more expensive then the cost of additional collects in lower level code.

PR #969, opened 2023-06-06
[Merged by Bors] - Add github merge queue

acrrd:
Add the possibility to use merge queue from github. The workflow for bors will be removed in another pr so we can merge this and we can still merge between the merging of this pr and the settings that we need to modify.
To run the CI in the merge queue is it necessary to enable in the setting the status check. For this reason `ci-ok` has been added to be able to check that the workflow executes correctly. For the same reason, we cannot have two separate workflows anymore; we can still decide to run a job only in case of `merge_queue` with some complications on `ci-ok`. At the moment we don't have anything that requires so much time that we want to run only during the merge.

After this has been merged we can enable to merge queue and set the status check to make this work, after this changes using bors will fail with a timeout since the required status check will not pass.

The check that the PR doesn't modify any past migration has been removed since it appears to be completely broken, see #971.

`CI.yml` is capital because otherwise, github ignores it.


acrrd:
Oh the new file is not even run. I'll try to understand why.

acrrd:
This is ready to be reviewed again

jorge-cr-13 (approved):

acrrd:
bors merge

PR #972, opened 2023-06-07
ET-3349 remove negative cois

janpetschexain:
**Reference**

- [ET-3349]

**Summary**

- remove `NegativeCoi`, `CoiPoint`, `UserInterests`, `UserInteractionType` and related dead code
- rename `PositiveCoi` to `Coi` etc and align names in the `coi` crate
- remove unused columns from postgres

[ET-3349]: https://xainag.atlassian.net/browse/ET-3349?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


coi/src/config.rs
```diff
@@ -30,8 +30,7 @@ use crate::{
 pub struct Config {
     shift_factor: f32,
     threshold: f32,
-    min_positive_cois: usize,
-    min_negative_cois: usize,
+    min_cois: usize,
```
janpetschexain, line 33: the corresponding env vars have never been set for any tenant so far, hence we can change the names without issues.

coi/src/context.rs
janpetschexain, line 1: all dead code was removed. the remaining logic moved directly into `coi::system::System::score()`.

web-api-db-ctrl/postgres/tenant/20230607110100_coi.sql
```diff
@@ -0,0 +1,16 @@
...
+--  This program is distributed in the hope that it will be useful,
+--  but WITHOUT ANY WARRANTY; without even the implied warranty of
+--  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+--  GNU Affero General Public License for more details.
+--
+--  You should have received a copy of the GNU Affero General Public License
+--  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+ALTER TABLE center_of_interest DROP COLUMN is_positive;
+ALTER TABLE interaction DROP COLUMN user_reaction;
```
janpetschexain, line 16: we only ever stored positive cois/interactions, hence we don't need any data migration.

web-api/src/storage/postgres.rs
```diff
@@ -665,32 +653,27 @@ impl Database {
...
-                            .push_bind(time)
-                            .push_bind(*interaction as i16);
+                            .push_bind(time);
                     },
                 )
-                .push(
-                    " ON CONFLICT (doc_id, user_id, time_stamp) DO UPDATE SET
-                    user_reaction = EXCLUDED.user_reaction;",
-                )
+                .push(" ON CONFLICT DO NOTHING;")
```
janpetschexain, line 676: we should probably store the newer timestamp on conflict. but this should be handled in a follow up pr, this pr keeps the logic as before.

web-api/src/personalization/routes.rs
```diff
@@ -565,23 +549,25 @@ async fn personalize_knn_search_result(
             storage::Tag::get(storage, &id).await?,
         ),
         InputUser::Inline { history } => {
-            let config: &PersonalizationConfig = config.as_ref();
-            let history = trim_history(history, config.max_stateless_history_for_cois);
+            let history = trim_history(
+                history,
+                AsRef::<PersonalizationConfig>::as_ref(config).max_stateless_history_for_cois,
```
rustonaut, line 554: :art: I do not thing that is stylisitcally or readability wise a good idea in general (same below).

I would argue `Trait::<generics>::method_with_self(value)` is for rust beginners somewhat of an advanced and confusing syntax feature and should be avoided (outside of  some exceptions where it makes things way simpler).

rustonaut (approved):
lgtm (through see one comment above)

web-api/src/personalization/routes.rs
janpetschexain: it might not be very beginner friendly, but how exactly does this prevent us from using this syntax? there are no rust beginners working on the code and if that ever happens, then people can still ask questions.

PR #976, opened 2023-06-08
ET-4570 remove newer-than-now filter

janpetschexain:
**Reference**

- [ET-4570]

**Summary**

- remove filtering of future dates
- avoid empty filter in request body
- update spec

[ET-4570]: https://xainag.atlassian.net/browse/ET-4570?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/web-api/openapi/schemas/time.yml
+++ b/web-api/openapi/schemas/time.yml
@@ -7,14 +7,30 @@ PublicationDate:
   maxLength: 40
   example: "2000-05-14T20:22:50Z"
   description: |
-    A RFC3339 compatible date-time
-
+    A RFC3339 compatible date-time:
     - can be in the future
     - will be converted to and then stored as UTC
-    - sub-second resolution is not guaranteed.
-    - while `properties.publication_date` is in the future the document will not be
-      returned for personalized documents, once it is no longer in the future it will
-      automatically be considered for personalization from then on
+    - sub-second resolution is not guaranteed
+
+PublishedAfter:
+  type:
+    $ref: '#/PublicationDate/type'
+  format:
+    $ref: '#/PublicationDate/format'
+  minLength:
+    $ref: '#/PublicationDate/minLength'
+  maxLength:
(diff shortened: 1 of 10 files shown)


rustonaut:
why, that filter is very usefull for news companies and similar which seem to be our main customers?

web-api/openapi/schemas/time.yml
```diff
@@ -7,14 +7,30 @@ PublicationDate:
   maxLength: 40
   example: "2000-05-14T20:22:50Z"
   description: |
-    A RFC3339 compatible date-time
+    A RFC3339 compatible date-time:
+    - can be in the future
```
rustonaut, line 11: If we remove the filter maybe we should not accept documents with a publication data in the future anymore? (with some tolerances around clock drive;  Or at least prevent them from being set as candidate or similar.)

Because it would mean you can get recommendations for a scheduled to be released but not yet released document.

janpetschexain: good question, i'll raise it in the daily

janpetschexain:
> just out interest why, performance?
> 
> That filter is very useful for news companies and similar which seem to be our main customers?

i guess mainly for two reasons:
- there is no opt-out and its implicit, contrary to the `published_after` filter, so if we want this we probably should make it opt-in like a `published_before` filter
- logical not combinators are not generally supported in all dbs, depending on which solution we pick for the vector index we may not be able to support it in the current form

@acrrd ?

acrrd:
I would add that is not that useful, no one ever asked for it and no one is using it. Newspapers aren't our main target. The implementation of only `published_date`  was more of a workaround to avoid having to implement proper filtering in properties.
Now that we are going to do that, `published_date`  should stop being special and behave like any other property.

rustonaut (approved):
lgtm

PR #978, opened 2023-06-12
ET-4572  set track_total_hits to false

rustonaut:
Could be rebased onto main, but would make already nasty merge conflicts for #967 worse.

**References:**

- [ET-4572]

[ET-4572]: https://xainag.atlassian.net/browse/ET-4572?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/web-api-shared/src/elastic.rs
+++ b/web-api-shared/src/elastic.rs
@@ -283,6 +283,7 @@ impl Client {
             return Ok(HashMap::new());
         }
         body.insert("_source".into(), json!(false));
+        body.insert("track_total_hits".into(), json!(false));
         self.query_with_json::<_, SearchResponse<I, NoSource>>(
             self.create_url(["_search"], None),
             Some(body),


web-api-shared/src/elastic.rs
```diff
@@ -283,6 +283,7 @@ impl Client {
             return Ok(HashMap::new());
         }
         body.insert("_source".into(), json!(false));
+        body.insert("track_total_hits".into(), json!(false));
```
janpetschexain, line 286: ❓ the es docs only list `track_total_hits` as a query param, but not as a body key. does this still has the desired effect and is not just silently discarded?

rustonaut: Which documentation?

Like in the documentation linked in the issue ([track_total_hits](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-your-data.html) it's a body parameter :thinking:

janpetschexain: i see, i was referring to the [docs of the _search endpoint](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-search.html). looks like they forgot to update their docs properly.

PR #983, opened 2023-06-13
ET-4591 no tenant middleware for silo management

rustonaut:
Do not run tenant/app specific middleware for the silo management endpoint.

Mainly do not require the `X-Xayn-Tenant-Id` header.

**References:**

- [ET-4591]

[ET-4591]: https://xainag.atlassian.net/browse/ET-4591?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/web-api/src/net.rs
+++ b/web-api/src/net.rs
@@ -79,7 +79,9 @@ impl Default for Config {
 pub(crate) fn start_actix_server(
     net_config: Config,
     legacy_tenant: Option<TenantId>,
+    attach_state: impl Fn(&mut ServiceConfig) + Send + Clone + 'static,
     attach_app: impl Fn(&mut ServiceConfig) + Send + Clone + 'static,
+    attach_ops: impl Fn(&mut ServiceConfig) + Send + Clone + 'static,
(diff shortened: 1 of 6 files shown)


web-api/src/net.rs
```diff
@@ -93,9 +95,16 @@ pub(crate) fn start_actix_server(
                     .route(web::get().to(HttpResponse::Ok))
                     .wrap(Cors::default()),
             )
+            .service(
+                web::scope("/_ops")
+                    .configure(&attach_state)
+                    .configure(&attach_ops)
+                    .wrap(Cors::default()),
```
rustonaut, line 102: This changes `/_silo_management` => `/_ops/silo_management`.

The problem is that a wild card scope `scope("")` is too greedy and will prevent routing the normal app.
Similar only ingestion has `_silo_management` but this is code used by both. Lastly we might have other
ad-hoc operations/management endpoints in the future and would have the same problem. So by having
a `_ops` scope we don't have the issue. I already informed ops.

janpetschexain (approved):
lgtm

PR #985, opened 2023-06-14
ET-4594 validate tags

janpetschexain:
**Reference**

- [ET-4594]

**Summary**

- update spec
- validate number of tags


rustonaut (approved):
lgtm

PR #986, opened 2023-06-14
ET-4595 restrict ids

janpetschexain:
**Reference**

- [ET-4595]

**Summary**

- update spec
- disallow dots in property ids
- disallow underscore prefix in all ids

[ET-4595]: https://xainag.atlassian.net/browse/ET-4595?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/web-api/openapi/back_office.yaml
+++ b/web-api/openapi/back_office.yaml
@@ -14,14 +14,13 @@ info:
 
     ## Document
     By document, we refer to a cohesive text, for example a complete news article.
-    However, we only require a simple representation for our system to work. Consisting just of a unique ID, a text snippet and optional properties.
+    However, we only require a simple representation for our system to work. Consisting just of a unique id, a text snippet and optional properties.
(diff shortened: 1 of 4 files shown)


rustonaut:
One thing I just have been wondering about is, isn't it very reasonable to want to use an URL as document id if you have a full REST system, in which case it's pretty much guaranteed to have a `.`.

And dot's shouldn't be a problem for document ids for us AFIK.

janpetschexain:
> One thing I just have been wondering about is, isn't it very reasonable to want to use an URL as document id if you have a full REST system, in which case it's pretty much guaranteed to have a `.`.
> 
> And dot's shouldn't be a problem for document ids for us AFIK.

i'm not sure what you are trying to say? dots are still valid in doc ids, they have just been removed from property ids.

PR #987, opened 2023-06-15
ET-4629 remove min_similarity

janpetschexain:
**Reference**

- [ET-4629]

**Summary**

- remove the `min_similarity` option from spec and code


rustonaut (approved):
lgtm

PR #988, opened 2023-06-15
ET-4630-fix-dynamic-setting

rustonaut:
This PR does two things:

- set `dynamic` to `strict` for the outer most object as it allows us to find bugs in our code 
  - add a definition for `tags` to the `mapping` 
- ~set `dynamic` to `runtime` for property fields so that they can be indexed in the future, this is fixing problems with a new filter by properties functionality~

**References:** 

- [ET-4630]

[ET-4630]: https://xainag.atlassian.net/browse/ET-4630?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/web-api-db-ctrl/elasticsearch/mapping.json
+++ b/web-api-db-ctrl/elasticsearch/mapping.json
@@ -11,6 +11,9 @@
                 "index": true,
                 "similarity": "dot_product"
             },
+            "tags": {
+                "type": "keyword"
+            },
             "properties": {
                 "dynamic": false,
                 "properties": {
--- a/web-api-db-ctrl/elasticsearch/mapping.json
+++ b/web-api-db-ctrl/elasticsearch/mapping.json
@@ -12,7 +12,7 @@
                 "similarity": "dot_product"
             },
             "properties": {
-                "dynamic": "runtime",
+                "dynamic": false,
                 "properties": {
                     "publication_date": {
                         "type": "date"
--- a/web-api-db-ctrl/elasticsearch/mapping.json
+++ b/web-api-db-ctrl/elasticsearch/mapping.json
@@ -1,5 +1,6 @@
 {
     "mappings": {
+        "dynamic": "strict",
         "properties": {
             "snippet": {
                 "type": "text"
@@ -11,7 +12,7 @@
                 "similarity": "dot_product"
             },
             "properties": {
-                "dynamic": false,
+                "dynamic": "runtime",
                 "properties": {
                     "publication_date": {
                         "type": "date"


rustonaut:
Posting here because I think it's good to know for all people:

In elasticsearch the dynamic setting can be set to `true` , `false`, `runtime` and `strict` as defined [here]( https://www.elastic.co/guide/en/elasticsearch/reference/current/dynamic.html).

This determines weather elasticsearch will try to infer the type of added fields (which can be very brittle).

But there are a few tricky parts to it:

- if you use `runtime` then you can use the fields for filtering, but it's potentially so slow that it's impossible to use with the normal sync elastic search API. *Oversimplified* it's  like finding all results ignoring the value of runtime fields and then applying a dump `flat_map` which scans through all the results and applies filters of runtime fields.

- for `"dynamic": false` it says: 

  >  New fields are ignored. These fields will not be indexed or searchable, but will still appear in the _source field of returned hits. These fields will not be added to the mapping, and new fields must be added explicitly. 
  
  what this fails to state is that they will **not** be added to current and **future mappings** :see_no_evil: 
  
  What this means is if e.g. all fields in `foo` are `dynamic: false` then even if you add an mapping/index for `foo.bar` later one all existing documents will be treated as if they don't have a `foo.bar` value and only _new_ documents will be searchable/filterable (you can also use update to set the `foo.bar` to the value `foo.bar` currently has e.g. using `"script":  "ctx._source.foo.bar = ctx._source.foo.bar"` on every existing document)
  
  The later part is a problem for us for upgrading current deployments to support property based filtering.

acrrd:
:question: Few questions:
Do we have any guarantee that with `runtime` older fields will be indexed?
Is the code in the silo management api reading this file or we need to also modify some code?
Should we update the current deployment with this? How switching `dynamic`  from `false` to `runtime` will affect the documents that are in the database?

rustonaut:
>  Do we have any guarantee that with runtime older fields will be indexed?

I just have been looking into it.

The documentation states that yes you can later add an index.

But the results seem to be ... mixed. Like literally.

While after adding an index it's enforced that any new documents match the type, documents already in the database which have a invalid type (e.g. using one of the rfc3339 but not twitch compatible time stamps) are still there and there is no warning or similar.

In case of ingesting some malformed timestamps and then setting the schema they still where returned in some situations :confused:  And I'm generally not sure if inferred runtime type string  => explicit type date works fully as expected.

Lastly I'm not sure if there are performance problems, I though I could just profile it  but the problem is that requests are dominated by a combination of generic local ES overhead and clever ad-hoc optimizations/caching ES does :face_exhaling:

rustonaut:
I have read up more on the topic:

- you can switch `dynamic` from `false` to `inherit` and it will "see" old documents 
  - the runtime mappings from `inherit` aren't indexed they are more similar like computed properties using a script like `ctx._source["property_name"]`  which is why you now can
  "see" them
- you can create explicit index for values previously having only runtime mappings 
  - but this will **not** delete the runtime mapping
  - and the runtime mapping will shadow the explicit mapping when querying
  - but the constraints from the explicit mapping is still checked when posting new documents
  - this can lead to very confusing results, and can create the illusion that the old documents are in the index **they are not**.
  
Also no matter how you turn it when you go from an property not being indexed at all and it being indexed the database _has to_ scan over all documents checking all values of this properties across all documents to create the index there is no technical way around it and the only question is how much overhead in form of e.g. network traffic and similar can be avoided.

The proper way to  add new indexed properties is to update the mapping and then call 

```
POST <index_name>/_update_by_query?conflicts=proceed
```

Which will internally touch all document and update/create any new indices as needed.

So `dynamic` stays `false`.

This endpoint has some neat parameters, like e.g. an async form where we can run it in the background and get a handle to check if it completed as well as an option to throttle it. Similar we probably want to set the `refresh` option appropriately.

This fit's in line with the documented behavior of "it might take a while and queries might see in-btween state until completed" from the pending spec on confluence.

acrrd (approved):

PR #990, opened 2023-06-19
ET-4632 only include properties if requested

janpetschexain:
**Reference**

- [ET-4632]

**Summary**

- only include document properties in the fronoffice responses if they are requested via the `include_properties` flag
- default the flag to `true` for now, it will be changed to false after some time

[ET-4632]: https://xainag.atlassian.net/browse/ET-4632?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/web-api/src/personalization/routes.rs
+++ b/web-api/src/personalization/routes.rs
@@ -131,11 +131,17 @@ pub(crate) async fn update_interactions(
     Ok(())
 }
 
+const fn default_true() -> bool {
+    true
+}
+
 /// Represents personalized documents query params.
 #[derive(Debug, Deserialize)]
 struct PersonalizedDocumentsQuery {
     count: Option<usize>,
     published_after: Option<DateTime<Utc>>,
+    #[serde(default = "default_true")]
+    include_properties: bool,
 }
 
 impl PersonalizedDocumentsQuery {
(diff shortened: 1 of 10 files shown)


web-api/src/personalization/routes.rs
```diff
@@ -443,15 +461,11 @@ impl UnvalidatedInputDocument {
 
 #[derive(Debug, Deserialize)]
 struct UnvalidatedPersonalize {
-    #[serde(default = "true_fn")]
+    #[serde(default = "default_true")]
     exclude_seen: bool,
     user: UnvalidatedInputUser,
 }
 
-fn true_fn() -> bool {
```
acrrd, line 451: ` default_exclude_seen`

web-api/src/personalization/routes.rs
```diff
@@ -131,11 +131,17 @@ pub(crate) async fn update_interactions(
     Ok(())
 }
 
+const fn default_true() -> bool {
```
acrrd, line 134: Usually, we call them with the name of the property, not the value. `default_include_properties`.

PR #995, opened 2023-06-21
Return original text if summarization fails

acrrd:
If there isn't enough information, the summarizer returns an empty string, and we create an embedding for that.
If we can't summarize we return the original text.


janpetschexain (approved):
good catch 👍 also with the new validation this would have caused a lot of bad requests (instead of being silently accepted).

PR #999, opened 2023-06-22
ET 4677 null is ok

acrrd:
Null is as a value in the properties and it is thread as if the property has no value.

### DIFF
--- a/web-api/openapi/schemas/document.yml
+++ b/web-api/openapi/schemas/document.yml
@@ -22,16 +22,18 @@ DocumentProperty:
   description: | 
     Arbitrary property that can be attached to a document.
     When a property is set to `null` it is treated as though that property has no values.
-  type:
-    nullable: true
-    oneOf:
-      - type: boolean
-      - type: number
-      - $ref: '#/DocumentPropertyString'
-      - type: array
-        minItems: 0
-        items:
-          $ref: '#/DocumentPropertyString'
+  oneOf:
+    - type: boolean
+      nullable: true
+    - type: number
+      nullable: true
+    - $ref: '#/DocumentPropertyString'
+      nullable: true
+    - type: array
+      nullable: true
+      minItems: 0
+      items:
+        $ref: '#/DocumentPropertyString'
   example: "News title"
 
 DocumentProperties:
--- a/web-api/openapi/schemas/document.yml
+++ b/web-api/openapi/schemas/document.yml
@@ -19,7 +19,10 @@ DocumentPropertyString:
   maxLength: 2048
 
 DocumentProperty:
-  description: Arbitrary property that can be attached to a document.
+  description: | 
+    Arbitrary property that can be attached to a document.
+    When a property is set to `null` it is treated as though that property has no values.
+  nullable: true
   oneOf:
     - type: boolean
     - type: number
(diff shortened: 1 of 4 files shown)


web-api/openapi/schemas/document.yml
```diff
@@ -19,13 +19,19 @@ DocumentPropertyString:
   maxLength: 2048
 
 DocumentProperty:
-  description: Arbitrary property that can be attached to a document.
+  description: | 
+    Arbitrary property that can be attached to a document.
+    When a property is set to `null` it is treated as though that property has no values.
```
Robert-Steiner, line 24:
```suggestion
    When a property is set to `null`, it is treated as if that property had no values.
```

Robert-Steiner (approved):

janpetschexain (approved):

web-api/src/models.rs
```diff
@@ -181,14 +179,16 @@ pub(crate) struct DocumentProperties(HashMap<DocumentPropertyId, DocumentPropert
...
-        properties: HashMap<DocumentPropertyId, DocumentProperty>,
+        mut properties: HashMap<DocumentPropertyId, DocumentProperty>,
         size: usize,
     ) -> Result<Self, InvalidDocumentProperties> {
-        if size <= 2_560 {
-            Ok(Self(properties))
-        } else {
-            Err(InvalidDocumentProperties { size })
+        if size > 2_560 {
+            return Err(InvalidDocumentProperties { size });
```
rustonaut, line 186: This size will include nullable properties we do remove AFIK, but we can still merge it and then fix that later.

web-api/openapi/schemas/document.yml
```diff
@@ -19,12 +19,18 @@ DocumentPropertyString:
   maxLength: 2048
 
 DocumentProperty:
-  description: Arbitrary property that can be attached to a document.
+  description: | 
+    Arbitrary property that can be attached to a document.
+    When a property is set to `null`, it is treated as if that property had no values.
```
rustonaut, line 24: :wrench: A property existing and having no value and it not existing for a specific document isn't guaranteed to be the same, so we should change the formulation to something like `as if the document doesn't have that property` or similar. Especially wrt. to filters this could matter.

web-api/src/models.rs
```diff
@@ -146,7 +146,7 @@ impl TryFrom<Value> for DocumentProperty {
 
     fn try_from(mut property: Value) -> Result<Self, Self::Error> {
         match &mut property {
-            Value::Bool(_) | Value::Number(_) => {}
+            Value::Bool(_) | Value::Number(_) | Value::Null => {}
```
rustonaut, line 149: It would be better to have this as a filter in the deserialization step because internally we still do not support null values. But this can be done in a followup PR.

rustonaut (approved):
there are some things which should be changed, but it's enough to do so in a follow up PR AFIK

web-api/openapi/schemas/document.yml
acrrd: It is the same exact formulation the ES uses.

web-api/src/models.rs
rustonaut: Additionally this implicitly makes it now possible to set a properties value to null using the `PUT /documents/{document_id}/properties/{property}` endpoint, which is not correct and should either fail or implicitly transform it into a `DELETE /documents/{document_id}/properties/{property}`.

PR #1000, opened 2023-06-23
Empty strings are valid properties and other changes

acrrd:


joecks (approved):

joecks (approved):

.spectral.js
```diff
@@ -50,6 +50,8 @@ module.exports = {
     },
     'ibm-major-version-in-path': 'off',
     'ibm-schema-description': 'off',
+    // spectral has is own rules to check schema that support v3.1
```
janpetschexain, line 53: :art: typos
```suggestion
    // spectral has its own rules to check schemas that support v3.1
```

web-api/openapi/schemas/document.yml
```diff
@@ -11,29 +11,32 @@ DocumentPropertyId:
...
+  items:
+    $ref: '#/DocumentPropertyString'
+
 DocumentProperty:
   description: | 
-    Arbitrary property that can be attached to a document.
-    When a property is set to `null`, it is treated as if that property had no values.
+    Arbitrary data that can be attached to a document. A subset of JSON is supported.
+    When a property is set to null, or an empty array it is treated as though that property has no values.
+    This means that they will not appear in searches with filter on this properties.
```
janpetschexain, line 32: :art: phrasing
```suggestion
    When a property is set to null or an empty array, it is treated as if that property had no values.
    This means that they won't appear in searches with filters on these properties.
```

web-api/openapi/schemas/document.yml
```diff
@@ -11,29 +11,32 @@ DocumentPropertyId:
 DocumentPropertyString:
   description: |
     Arbitrary string property that can be attached to a document.
-    Enclosing whitespace will be trimmed.
     The length constraints are in bytes, not characters.
   type: string
   pattern: '^[^\x00]+$'
-  minLength: 1
+  minLength: 0
```
janpetschexain, line 17: ❓ why should we accept empty strings at all? if we get an empty string, this is most likely a logic error on the customer side. and it also doesn't bring any benefit to our filter logic, on the contrary.

❓ also, why shouldn't we trim them, or if we don't want to change the input, why don't we at least validate if they are trimmed?

web-api/openapi/schemas/document.yml
acrrd: I should trust less the phrasing from ES documentation :grimacing:

PR #1003, opened 2023-06-26
ET-4599 array of strings filter

janpetschexain:
**Reference**

- [ET-4599]
- requires #1002
- followed by #1010

**Summary**

- declare array of strings filters in the spec & update changelog
- add containment operator to the compare filter serde
- update compare filter in elasticsearch calls

[ET-4599]: https://xainag.atlassian.net/browse/ET-4599?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.spectral.js
+++ b/.spectral.js
@@ -60,6 +60,7 @@ module.exports = {
       files: [
         "web-api/openapi/back_office.yaml#/components/schemas/DocumentCandidatesRequest",
         "web-api/openapi/back_office.yaml#/components/schemas/DocumentCandidatesResponse",
+        "web-api/openapi/schemas/document.yml#/DocumentPropertyArrayString",
       ],
       rules: {
         "ibm-array-attributes": "off",
--- a/web-api/openapi/front_office.yaml
+++ b/web-api/openapi/front_office.yaml
@@ -170,6 +170,8 @@ components:
         properties:
           $eq:
             $ref: ./schemas/document.yml#/DocumentPropertyString
+          $in:
+            $ref: ./schemas/document.yml#/DocumentPropertyArrayString
         minProperties: 1
         maxProperties: 1
     Filter:
--- a/web-api/openapi/front_office.yaml
+++ b/web-api/openapi/front_office.yaml
@@ -81,7 +81,7 @@ paths:
               schema:
                 $ref: '#/components/schemas/PersonalizedDocumentsResponse'
         '400':
-           $ref: './responses/generic.yml#/BadRequest'
+          $ref: './responses/generic.yml#/BadRequest'
         '409':
           description: impossible to create a personalized list for the user.
           content:
@@ -145,7 +145,7 @@ paths:
               schema:
                 $ref: '#/components/schemas/SemanticSearchResponse'
         '400':
-            $ref: './responses/generic.yml#/BadRequest'
+          $ref: './responses/generic.yml#/BadRequest'
 
 components:
   securitySchemes:
@@ -172,6 +172,7 @@ components:
             $ref: ./schemas/document.yml#/DocumentPropertyString
           $in:
             $ref: ./schemas/document.yml#/DocumentPropertyArrayString
+            maxItems: 10
         minProperties: 1
         maxProperties: 1
     Filter:
@@ -242,7 +243,10 @@ components:
           type: boolean
           default: false
         filter:
-          $ref: '#/components/schemas/Filter'
+          description:
+            $ref: '#/components/schemas/Filter/description'
+          oneOf:
+            - $ref: '#/components/schemas/FilterCompare'
     SemanticSearchResponse:
       type: object
       required: [documents]
--- a/web-api/tests/filter.rs
+++ b/web-api/tests/filter.rs
@@ -87,7 +87,7 @@ fn test_filter_string() {
             index(
                 &client,
                 &ingestion_url,
-                json!({ "p": { "type": "keyword" }, "q": { "type": "keyword" } }),
+                json!({ "p1": { "type": "keyword" }, "p2": { "type": "keyword" } }),
             )
             .await?;
             ingest(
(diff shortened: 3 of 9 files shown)


.spectral.js
```diff
@@ -67,6 +67,7 @@ module.exports = {
       files: [
         "web-api/openapi/back_office.yaml#/components/schemas/DocumentCandidatesRequest",
         "web-api/openapi/back_office.yaml#/components/schemas/DocumentCandidatesResponse",
+        "web-api/openapi/schemas/document.yml#/DocumentPropertyArrayString",
```
janpetschexain, line 70: does anybody has an idea how to silence this lint, such that it doesn't produce the following error?
```
/__w/xayn_discovery_engine/xayn_discovery_engine/web-api/openapi/schemas/document.yml
 20:29  warning  ibm-array-attributes  Array schemas should define a numeric 'minItems' field  DocumentPropertyArrayString
 20:29  warning  ibm-array-attributes  Array schemas should define a numeric 'maxItems' field  DocumentPropertyArrayString
```
i also tried it with `"web-api/openapi/front_office.yaml#/components/schemas/FilterCompare/additionalProperties/properties/$in",` but to no effect 🤷

rustonaut: I think we do want to define `minItems` & `maxItems` here :thinking: 

For `minItems` we want to be clear if it's 0+ or 1+.

For `maxItems` we can't just accept arbitrary long arrays AFIK, especially if we want to hold the option open to not use elasticsearch in the future. The really hard question is what to chooses for `maxItems`?

web-api/tests/filter.rs
```diff
@@ -140,3 +140,226 @@ fn test_filter_string() {
...
+            .await;
+            assert!(documents.is_empty());
+
+            let documents = send_assert_json::<SemanticSearchResponse>(
+                &client,
+                client
+                    .post(personalization_url.join("/semantic_search")?)
+                    .json(&json!({
+                        "document": { "query": "zero" },
+                        "filter": { "q": { "$in": ["this", "that", "other"] } }
```
rustonaut, line 235: :art:   maybe name it not `q` but something which is more obvious not `p`

web-api/tests/filter.rs
```diff
@@ -140,3 +140,226 @@ fn test_filter_string() {
...
+        },
+    );
+}
+
+#[test]
+fn test_filter_array_string_multiple() {
+    test_two_apps::<Ingestion, Personalization, _>(
+        UNCHANGED_CONFIG,
+        UNCHANGED_CONFIG,
+        Some(&[("p", "keyword"), ("q", "keyword")]),
```
rustonaut, line 253:
```suggestion
        Some(&[("p", "keyword[]"), ("q", "keyword[]")]),
```

.spectral.js
janpetschexain: > For minItems we want to be clear if it's 0+ or 1+.

one of the previous changes allowed empty arrays as valid, so we have to go with 0. but then `minItems: 0` is redundant, because without `minItems` this always holds according to the openapi docs.

> For maxItems we can't just accept arbitrary long arrays AFIK, especially if we want to hold the option open to not use elasticsearch in the future. The really hard question is what to chooses for maxItems?

we limit the overall byte size of the properties, so this automatically limits how much a specific property can hold, hence we don't have to put a `maxItems` limit.

we could differentiate the types though, one for `[string]` properties and one for `$in` filter arguments and could put a limit on the latter. not sure if this is too much though. the question would still be what limit.

web-api/tests/filter.rs
rustonaut: (same for the other test below)

rustonaut (commented):
some small changes

and the comment from a previous  short review wrt. us probably wanting to specify array bounds instead of disabling the lint (before I knew we can't disable the lint).

web-api/tests/filter.rs
janpetschexain: this worked before because it went directly to es. after the rebase i had to change it like you suggested.

.spectral.js
janpetschexain: i've set it to `minItems: 0` and `maxItems: 100` for now. even though the upper limit is a breaking change it should be way more than enough for now and we can also increase it in the future.

regarding the linter, we may be able to implement our custom resolver which doesn't only inline `$ref`s from the root file but also all other `$ref`s from the non-root files, then the overrides should apply (maybe, i didn't get a fully clear picture from their docs). this needs to be tried by someone who knows javascript though. or we need to look for another linter.

rustonaut: I think we should make `$in` have a different limit as both have very different performance properties, I also would say it makes sense to set some limit for both, through potential quite high limits.

I'm not sure if 100 is too small of an limit, we know of at least one use-case which wants a higher limit but probably so high that it might outstrip any limit.

web-api/openapi/front_office.yaml
```diff
@@ -240,7 +243,10 @@ components:
           type: boolean
           default: false
         filter:
-          $ref: '#/components/schemas/Filter'
+          description:
+            $ref: '#/components/schemas/Filter/description'
+          oneOf:
+            - $ref: '#/components/schemas/FilterCompare'
```
janpetschexain, line 249: i just noticed this issue in my latest pr but backported the fix here. for whatever reason the documentation generation fails to resolve the ref completely if used in a request body, eg the properties of the ref still get resolved but the description and type gets lost. interestingly resolving the ref works if used in a query param like for the other endpoint.

it would be nice to just use the ref, but i couldn't figure out if this is fixable and how to fix this, so i preferred a complete documentation over duplication. the issue seems to be rooted in the `oneOf` of the referenced schema, we could just remove it in this pr as it only contains a single item, but it will become a problem immediately in the next prs when we add more items. anybody has an idea how to make this work with just the ref?

web-api/openapi/front_office.yaml
```diff
@@ -172,6 +172,7 @@ components:
             $ref: ./schemas/document.yml#/DocumentPropertyString
           $in:
             $ref: ./schemas/document.yml#/DocumentPropertyArrayString
+            maxItems: 10
```
rustonaut, line 175: Let's point this out to @acrrd not that the limit is directly conflicting with one of the customers use-cases.

rustonaut (approved):
lgtm

web-api/openapi/front_office.yaml
janpetschexain: i'll bring it up once he is back

PR #1004, opened 2023-06-27
Use json to print logs

acrrd:
Flattens the fields in the root to ease access and improve readability.
The current span is already present in `spans`.


acrrd:
We are testing this to see how it shows in cloud watch

janpetschexain (approved):
lgtm

PR #1008, opened 2023-06-28
ET-4689-check-es-index-on-startup

rustonaut:
- [X] on startup check the ES index, mainly check if the embedding length is as expected
- [X] make sure that in a single tenant setup a bad tenant migration will kill the whole service
  - this is  already the current behavior, :pushpin: we need to handle this better for multi tenant

**References:**

- issue: [ET-4689]

[ET-4689]: https://xainag.atlassian.net/browse/ET-4689?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/web-api-db-ctrl/src/elastic.rs
+++ b/web-api-db-ctrl/src/elastic.rs
@@ -14,7 +14,7 @@
 
 use std::fmt::Debug;
 
-use anyhow::{anyhow, bail};
+use anyhow::bail;
 use once_cell::sync::Lazy;
 use reqwest::{Method, StatusCode};
 use serde_json::{json, Value};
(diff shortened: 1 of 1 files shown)


rustonaut:
(many) :pushpin:s: I originally wanted to place more into this PR but realized that it would blow up, additionally there are some parts which are obvious to big for this PR: 

- integration test for the silo crate independent of the rest of the code base (not much new code, but we want to probably reuse some parts of the test utils, but not all)
- check all schema properties for compatibility
- update schema where necessary (e.g. when we added `document.tags` )
  - handle changeable mapping properties, like `ignore_malformed`
- extend checks to also check `document.properties` against the indexed properties schema
  - requires to move (some) code for index properties schema handling into the silo management crate
  - in context of that we might slightly change some ES update/PG transaction interactions, I realized that there is a subtle error race condition (race with network errors)  which can lead to a situation where a property definition is updated in ES but doesn't get updated in PG. Due to technical limitations we always have the problem in one direction. But if it's not updated in ES but always in PG we can as part of self-checks and/or start up schema migrations repair it, but not the other way around
- add a silo management  function which does some DB self-checks, including all the ones above, through most self-checks are better placed on the web-api independent of the silo management, so maybe not

web-api-db-ctrl/src/elastic.rs
```diff
@@ -78,16 +78,41 @@ pub(crate) async fn migrate_tenant_index(
...
-    //      incremental application of the change is possible
+    Ok(())
+}
+
+fn check_mapping_compatibility(
+    existing_mapping: &Value,
+    base_mapping: &Value,
+) -> Result<(), Error> {
+    let existing_embeddings = &existing_mapping["mappings"]["properties"]["embedding"];
+    let expected_embeddings = &base_mapping["mappings"]["properties"]["embedding"];
```
janpetschexain, line 99: :art: we probably should introduce constants for these reused indexing strings to better guard against typos.

janpetschexain (approved):
lgtm

PR #1009, opened 2023-06-28
Use env to execute bash

acrrd:
Use `env` to execute bash to better support systems where bash is not in `/bin` (like my nixos).
We are already doing this in the justfile.

`wait_for_it.sh` is not used anywhere so is removed. Was probably used in the compose file or docker image.

### DIFF
--- a/.github/scripts/build_ci_image.sh
+++ b/.github/scripts/build_ci_image.sh
@@ -1,6 +1,4 @@
-#!/bin/bash
-
-set -e
+#!/usr/bin/env -S bash -e
 
 source .env
 RUST_VERSION=$(perl -ne 'print $1 if /channel = \"(.*)\"/' rust-toolchain.toml)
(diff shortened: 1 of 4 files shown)


janpetschexain:
> `wait_for_it.sh` is not used anywhere so is removed. Was probably used in the compose file or docker image.

❓ this was meant to be used with the python `discovery_engine_api_tests`. if we want to remove this script, then we should also decide what to do with the tests. they never went into the ci and by now they are also partially redundant with the rust integration tests. should we just check if all python tests are covered in rust, add the missing and the remove the python tests all together?

acrrd:
> > `wait_for_it.sh` is not used anywhere so is removed. Was probably used in the compose file or docker image.
> 
> question this was meant to be used with the python `discovery_engine_api_tests`. if we want to remove this script, then we should also decide what to do with the tests. they never went into the ci and by now they are also partially redundant with the rust integration tests. should we just check if all python tests are covered in rust, add the missing and the remove the python tests all together?

We will port the missing tests and delete the python one.

janpetschexain (approved):

PR #1010, opened 2023-06-29
ET-4605 and or filter

janpetschexain:
**Reference**

- [ET-4605]
- requires #1003
- followed by #1016

**Summary**

- declare and/or filters in the spec & update changelog
- impl combine filter serde
- add combine filter in elasticsearch calls

[ET-4605]: https://xainag.atlassian.net/browse/ET-4605?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/web-api/src/personalization/filter.rs
+++ b/web-api/src/personalization/filter.rs
@@ -164,6 +164,8 @@ pub(crate) enum CombineOp {
 pub(crate) struct Filters(Vec<Filter>);
 
 impl Filters {
+    const MAX_LEN: usize = 10;
+
     fn is_below_level(&self, max: usize) -> bool {
         (max > 0)
             .then(|| self.iter().all(|filter| filter.is_below_level(max - 1)))
@@ -178,17 +180,13 @@ impl<'de> Deserialize<'de> for Filters {
     {
         struct FiltersVisitor;
 
-        impl FiltersVisitor {
-            const MAX_LEN: usize = 10;
-        }
-
         impl<'de> Visitor<'de> for FiltersVisitor {
             type Value = Filters;
 
             fn expecting(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
                 formatter.write_str(&format!(
                     "a json array with at most {} combination arguments",
-                    Self::MAX_LEN,
+                    Self::Value::MAX_LEN,
                 ))
             }
 
@@ -197,13 +195,13 @@ impl<'de> Deserialize<'de> for Filters {
                 A: SeqAccess<'de>,
             {
                 let size = seq.size_hint().unwrap_or_default();
-                if size > Self::MAX_LEN {
+                if size > Self::Value::MAX_LEN {
                     return Err(A::Error::invalid_length(size, &Self));
                 }
 
                 let mut filters = Vec::with_capacity(size);
                 while let Some(filter) = seq.next_element()? {
-                    if filters.len() < Self::MAX_LEN {
+                    if filters.len() < Self::Value::MAX_LEN {
                         filters.push(filter);
                     } else {
                         return Err(A::Error::invalid_length(filters.len() + 1, &Self));
@@ -224,6 +222,10 @@ pub(crate) struct Combine {
     pub(crate) filters: Filters,
 }
 
+impl Combine {
+    const MAX_LEVEL: usize = 2;
+}
+
 impl<'de> Deserialize<'de> for Combine {
     fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
     where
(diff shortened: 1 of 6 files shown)


web-api/src/storage/elastic/filter.rs
```diff
@@ -12,36 +12,81 @@
...
+                *filter = json!([filter.take(), clause]);
+            }
+            _ => unreachable!(/* filter and clause are array or object */),
+        }
+    } else {
+        filter.insert(occurence.to_string(), clause);
+    }
+}
+
+fn extend_filter(filter: &mut JsonObject, clause: &Filter, is_not_root: Option<&'static str>) {
```
rustonaut, line 43: I will approve it for now but I feel the readability and maintainability of this code is pretty bad.

Some ideas:

- maybe we probably should not special case "is root"/"is not root"
- avoid or change how we do some of the micro optimizations like the `"should": [{...}]` as `"should": {...}` part

web-api/tests/filter.rs
```diff
@@ -391,3 +391,162 @@ fn test_filter_array_string_multiple() {
...
+            let documents = send_assert_json::<SemanticSearchResponse>(
+                &client,
+                client
+                    .post(personalization_url.join("/semantic_search")?)
+                    .json(&json!({ "document": { "query": "zero" }, "filter": { "$or": [] } }))
+                    .build()?,
+                StatusCode::OK,
+            )
+            .await;
+            assert_eq!(documents.ids(), ["d1", "d2", "d3"].into());
```
rustonaut, line 513: :wrench:  There is a bug somewhere here I think.

`"$or": []` should always be false, i.e. not match any documents

I mean it should map to `{ "should":  [], "minimum_should_match": 1 }` which can never evaluate to have enough matches :thinking: 

(Also logically `$and: [ (...),*]` is like `true (&& ...)*` and `$or: [...,*]` like `false (|| ...)*` )

web-api/src/personalization/filter.rs
```diff
@@ -147,10 +148,142 @@ impl<'de> Deserialize<'de> for Compare {
...
+                    } else {
+                        return Err(A::Error::invalid_length(filters.len() + 1, &Self));
+                    }
+                }
+
+                Ok(Filters(filters))
+            }
+        }
+
+        deserializer.deserialize_seq(FiltersVisitor)
```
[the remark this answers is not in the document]
janpetschexain: but this opens up for performance issues. if a request contains a very large filter (eg as much as the byte limit of the request allows) and we deserialize it completely and only afterwards check the length and discard it, then a lot of work was wasted.
the overhead for a custom visitor on the other hand is not too much as it is only a single additional function.

web-api/src/storage/elastic/filter.rs
janpetschexain: > maybe we probably should not special case "is root"/"is not root"

i agree with that, but it requires further refactoring in the es modules, because we would also need to apply the `{"bool": {"filter": object}}` structure to the root level, but currently we carry only the `object` thingy around. hence i didn't make these changes yet, but we definitely should follow up on this.

> avoid or change how we do some of the micro optimizations like the "should": [{...}] as "should": {...} part

i was wondering how es handles that internally, does it make a difference performance wise? or is it handled as array of objects anyways?

web-api/tests/filter.rs
janpetschexain: this is forwarded as is to es, so we probably have to decide if es's edge case handling is what we want or not.

web-api/src/personalization/filter.rs
```diff
@@ -147,10 +148,142 @@ impl<'de> Deserialize<'de> for Compare {
...
+
+        deserializer.deserialize_map(CombineVisitor)
+    }
+}
+
 #[derive(Clone, Debug, Deserialize, PartialEq)]
 #[serde(untagged)]
 pub(crate) enum Filter {
     Compare(Compare),
+    Combine(Combine),
```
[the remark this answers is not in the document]
rustonaut: it's not about performance but about making sure we never parse `$and` or `$or` accidentally as field names.

Like when adding `$document.tags` or similar .

web-api/src/personalization/filter.rs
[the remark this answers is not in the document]
rustonaut: For preventing performance issue we have size limits on the JSON body, and wrt. the generated code it's much much more then just one function.

Through we can keep it that way, still I think we should in general prefer writing the simpler code and only do such optimizations if there are performance issues.

web-api/tests/filter.rs
rustonaut: ES :face_exhaling: 

I guess that becomes a [:pushpin: issue](https://xainag.atlassian.net/browse/ET-4738).

web-api/src/personalization/filter.rs
```diff
@@ -147,10 +148,142 @@ impl<'de> Deserialize<'de> for Compare {
...
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub(crate) struct Combine {
+    pub(crate) operation: CombineOp,
+    pub(crate) filters: Filters,
+}
+
+impl Combine {
+    const MAX_LEVEL: usize = 2;
```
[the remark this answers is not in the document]
rustonaut: you didn't push any changes

web-api/src/personalization/filter.rs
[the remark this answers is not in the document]
janpetschexain: from a user input this can never happen as we don't allow `$` as a char in ids. if we on our side decide to add some special ids using forbidden chars, then we need to very carefully evaluate if and what to use, we probably need a list of reserved special ids then. until then we can keep it as is.

rustonaut (approved):
it would be nice to still change the `Combine::MAX_LEVEL` to a more clear name like `MAX_DEPTH`.

web-api/src/storage/elastic/filter.rs
janpetschexain: i simplified the code such that all clauses must be arrays now, this removes a lot of the array/object handling.

rustonaut (approved):

PR #1012, opened 2023-06-29
Update semantic_search's doc

acrrd:
Describe all the other ways this endpoint can be sued.

### DIFF
--- a/web-api/openapi/front_office.yaml
+++ b/web-api/openapi/front_office.yaml
@@ -118,7 +118,9 @@ paths:
       description: |-
         This endpoint allows you to search through the documents in different ways.
         When a document id is used, the system will return documents that are similar to the one in input.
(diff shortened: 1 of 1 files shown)


web-api/openapi/front_office.yaml
```diff
@@ -114,9 +114,11 @@ paths:
     post:
       tags:
         - front office
-      summary: Returns documents similar to the given document.
+      summary: Search through documents.
       description: |-
-        Returns a list of documents that are semantically similar to the one given as input.
+        This endpoint allows you to search through the documents in different ways.
+        When a document id is used, the system will return documents that are similar to the one in input.
+        When a query is provided, the system will return documents that are similar to the query. If `enable_hybrid_search` is passed that the system will also perform keyword matching between the query and the documents. It is possible to personalise the result by passing a user id or history; in this case, the system will consider the user's interests to rank the documents.
```
janpetschexain, line 121: :art: phrasing
```suggestion
        When a query is provided, the system will return documents that are similar to the query.
        If `enable_hybrid_search` is passed, then the system will also perform keyword matching between the query and the documents.
        It is possible to personalize the result by passing a user id or history. In this case, the system will consider the user's interests to rank the documents.         
```

janpetschexain (approved):
lgtm

PR #1015, opened 2023-07-04
ET-4504 retry es once for network error

rustonaut:
Add a retry police which is used when requests to elasticsearch fail due to network errors.

- [x] retry elasticsearch requests in web-api if they fail due to network errors
- [x] retry elasticsearch requests for silo api
- [x] check timeouts applies at higher level places

**References:**

- issue: [ET-4504]

[ET-4504]: https://xainag.atlassian.net/browse/ET-4504?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/web-api-shared/src/elastic.rs
+++ b/web-api-shared/src/elastic.rs
@@ -379,7 +379,8 @@ impl Client {
             let error = String::from_utf8_lossy(&body).into_owned();
             Err(Error::Status { status, url, error })
         } else {
-            Ok(response.json().await?)
+            let body = response.bytes().await?;
+            Ok(serde_json::from_slice(&body)?)
         }
     }
 
--- a/web-api-shared/src/elastic.rs
+++ b/web-api-shared/src/elastic.rs
@@ -70,8 +70,8 @@ impl Default for Config {
             index_name: "test_index".into(),
             timeout: Duration::from_secs(2),
             retry_policy: ExponentialJitterRetryPolicyConfig {
-                max_retries: 2,
-                step_size: Duration::from_millis(150),
+                max_retries: 3,
+                step_size: Duration::from_millis(300),
                 max_backoff: Duration::from_millis(1000),
             },
         }
--- a/web-api-shared/src/elastic.rs
+++ b/web-api-shared/src/elastic.rs
@@ -247,24 +247,21 @@ impl<I> BulkResponse<I> {
 }
 
 #[derive(Debug, Deserialize)]
-struct Hit<I, T> {
+struct Hit<I> {
     #[serde(rename = "_id")]
     id: I,
-    #[allow(dead_code)]
-    #[serde(rename = "_source")]
-    source: T,
     #[serde(rename = "_score")]
     score: f32,
 }
 
 #[derive(Debug, Deserialize)]
-struct Hits<I, T> {
-    hits: Vec<Hit<I, T>>,
+struct Hits<I> {
+    hits: Vec<Hit<I>>,
 }
 
 #[derive(Debug, Deserialize)]
-struct SearchResponse<I, T> {
-    hits: Hits<I, T>,
+struct SearchResponse<I> {
+    hits: Hits<I>,
 }
 
 /// Deserializes from anything discarding any response.
@@ -282,9 +279,23 @@ impl<'de> Deserialize<'de> for SerdeDiscard {
         D: Deserializer<'de>,
     {
         use serde::de;
-        // Hint: We can't just return Ok(Self) this will trigger some checks
-        //       making sure the whole body is parsed for some deserializers
-        //       in some contexts.
+        // Hint: a "discarding" serde type is more tricky then it might seem
+        // 1. just returning `Ok(SerdeDiscard)` without calling the deserializer can fail
+        //    in some edge cases due to the input which is supposed to be parsed not being
+        //    parsed at all. Instead of being parsed and discarded.
+        //    - this also can lead to a bunch of other unexpected behavior, e.g. we parsed a
+        //      `_source` field even after we disabled `_source` inclusion and blindly returning
+        //      `Ok(SerdeDiscard)` as a value in a map deserialized form json somehow didn't raise
+        //      and error (even through it should have as only the value is discarded, which still means
+        //      the fields itself is required if you don't also use `serde(default)`)
+        // 2. not parsing the input can lead to two errors 1st the deserializer raising an
+        //    error because something is wrong, 2nd we don't realize if the server has some
+        //    major issues and send us partial bodies or similar
+        // 3. to parse and then discard `deserialize_any` is required, or else the deserializer can't proceed at all
+        // 4. if a derserializer is used which might call `visit_enum` (i.e. not json) it might still
+        //    fail as there is no equivalent of deserialize_any for the enum variant
+        // 5. Just using the serde derive on a empty `struct Foo { }` will only work with json object
+        //    responses, but not arrays, string, etc.
         struct DiscardingVisitor;
 
         macro_rules! impl_simpl_sink {
@@ -293,7 +304,7 @@ impl<'de> Deserialize<'de> for SerdeDiscard {
                 where
                     E: de::Error,
                 {
-                    Ok(DiscardResponse)
+                    Ok(SerdeDiscard)
                 }
             )*);
         }
@@ -409,7 +420,7 @@ impl Client {
         }
         body.insert("_source".into(), json!(false));
         body.insert("track_total_hits".into(), json!(false));
-        self.query_with_json::<_, SearchResponse<I, SerdeDiscard>>(
+        self.query_with_json::<_, SearchResponse<I>>(
             Method::POST,
             self.create_url(["_search"], None),
             Some(body),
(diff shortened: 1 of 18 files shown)


web-api-shared/src/elastic.rs
```diff
@@ -65,6 +68,11 @@ impl Default for Config {
             password: String::from("changeme").into(),
             index_name: "test_index".into(),
             timeout: Duration::from_secs(2),
+            retry_policy: ExponentialJitterRetryPolicyConfig {
+                max_retries: 2,
+                step_size: Duration::from_millis(150),
+                max_backoff: Duration::from_millis(1000),
+            },
```
rustonaut, line 75: No idea what good values are.

In a app we probably would use 1s or similar for step size, but because it's internal requests going with something lower should be preferable.

web-api-shared/src/serde.rs
```diff
@@ -55,6 +55,48 @@ pub mod serde_duration_as_seconds {
     }
 }
 
+pub mod serde_duration_in_config {
```
rustonaut, line 58: I took a glance as crate providing something like that. 

We could use some crate for this the problem is every crate a took a glance at had some issues.

Like bad code, being too flexible (we only want to have 1 way to e.g. specify 100ms, not 10 different formattings), or using a too-unusual format (e.g. partial rfc3339 duration like `PT10M20S ` it also happens to forget to define sub-seconds for duration, through a lot of code assumes they exist ), strange whitespace handling, etc.

So just dumping this minimal implementation here was faster.

E.g. `duration-string` in general does what we want, except for it's whitespace handling (ignore them so `10 00 00    s` is valid) and it's code is ... safe, correct, but not very good.

web-api-shared/src/elastic.rs
```diff
@@ -104,9 +115,25 @@ impl Client {
...
+        &self,
+        error_filter: impl Fn(&E) -> bool,
+        code: impl FnMut() -> F,
+    ) -> Result<T, E>
+    where
+        F: Future<Output = Result<T, E>>,
+        E: Display,
+    {
+        let policy = ExponentialJitterRetryPolicy::new(self.retry_policy.clone())
+            .with_retry_filter(error_filter);
```
rustonaut, line 132: The retry filter is call context specific and might be affected by the generic parameters of the functions it's called in, so in turn we always want to configure it "around the places" where we use the policy.

web-api-shared/src/elastic.rs
```diff
@@ -219,35 +247,144 @@ impl<I> BulkResponse<I> {
 }
 
 #[derive(Debug, Deserialize)]
-struct Hit<I, T> {
+struct Hit<I> {
     #[serde(rename = "_id")]
     id: I,
-    #[allow(dead_code)]
-    #[serde(rename = "_source")]
-    source: T,
```
rustonaut, line 227: This one was a "fun" bug.

We do not include `_source` anymore in any request, so no matter what `T` serializes to de-serializing this should always have failed. 

But it didn't.

It turns out that somehow the combinations of `serde_json` with a `T::deserialize` which never calls the deserializers somehow "eats" the bug that the filed is fully missing.

Be aware that serde requires the deserializer to be called, so this wasn't a bug in `serde`/`serde_json` but our code.

web-api-shared/src/elastic.rs
```diff
@@ -219,35 +247,144 @@ impl<I> BulkResponse<I> {
...
 
+/// Deserializes from anything discarding any response.
+///
+/// Requires the `Deserializer` implementation to support `deserialize_any` and
+/// might not work for cases where `visitor.visit_enum` is called.
+///
+/// This means it does work without restrictions for formats like `json`.
 #[derive(Debug)]
-struct NoSource;
+pub struct SerdeDiscard;
```
rustonaut, line 274: I combined `NoSource` with `IgnoreResponse` and made it a re-usable utility.

web-api-shared/src/elastic.rs
```diff
@@ -219,35 +247,144 @@ impl<I> BulkResponse<I> {
...
+            where
+                A: de::EnumAccess<'de>,
+            {
+                //Hint: This can have the same issues as a deserialize `Ok(DiscardResponse)`,
+                //      but we can't really do anything about it there is no `EnumAccess.variant_any`.
+                Ok(SerdeDiscard)
+            }
+        }
+
+        deserializer.deserialize_any(DiscardingVisitor)
```
rustonaut, line 387: This is a bit of a pre-mature optimizations we could replace it with `Value::derserialize(deserializer)?; Ok(SerdeDiscard)` (with the drawback of a bunch of unnecessary alloc/frees).

web-api-shared/src/elastic.rs
```diff
@@ -304,16 +440,39 @@ impl Client {
         })
     }
 
-    pub async fn query_with_bytes<B, T>(
+    pub async fn query_with_bytes<T>(
+        &self,
+        method: Method,
+        url: Url,
+        post_data: Option<(HeaderMap<HeaderValue>, Bytes)>,
```
rustonaut, line 447: We change this to `Bytes` to make sure cloning the body between retries is cheap.

web-api-shared/src/elastic.rs
```diff
@@ -304,16 +440,39 @@ impl Client {
...
+        )
+        .await
+    }
+
+    async fn query_with_bytes_without_retrying<B, T>(
         &self,
         method: Method,
         url: Url,
         post_data: Option<(HeaderMap<HeaderValue>, B)>,
-    ) -> Result<Option<T>, Error>
```
rustonaut, line 312: It turns out this was very error prone, we did had some subtle bugs where a 404 might get ignored in some places where it should yield an error.

So I replaced it with `T` and a extension trait allowing converting a `Result<T, Error>` into a `Result<Option<T>, Error>` based on weather the error is `Error:ResourceNotFound`.

web-api-shared/src/elastic.rs
```diff
@@ -323,14 +482,15 @@ impl Client {
...
+            Err(Error::ResourceNotFound(path))
         } else if !status.is_success() {
             let url = response.url().clone();
             let body = response.bytes().await?;
             let error = String::from_utf8_lossy(&body).into_owned();
             Err(Error::Status { status, url, error })
         } else {
-            Ok(Some(response.json().await?))
+            let body = response.bytes().await?;
+            Ok(serde_json::from_slice(&body)?)
```
rustonaut, line 493: This was a bug as it would report (de-)serialization errors as `Error::Transport` which is bad as we don't want to retry if there is an (de-)serialization error (through we do want to retry if awaiting the body fails, like we still do).

web-api/src/storage/elastic.rs
```diff
@@ -47,13 +50,6 @@ use crate::{
     Error,
 };
 
-/// Deserializes from any map/struct dropping all fields.
-///
-/// This will not work with non self describing non schema
-/// formats like bincode.
-#[derive(Debug, Deserialize)]
-struct IgnoredResponse {/* Note: The {} is needed for it to work correctly. */}
```
rustonaut, line 55: got combined with similar struct into `SerdeDiscard`

web-api-shared/src/elastic.rs
```diff
@@ -367,10 +527,24 @@ pub enum Error {
         url: Url,
         error: String,
     },
-    /// Failed to serialize a requests or deserialize a response.
+    /// Failed to serialize a requests or deserialize a response: {0}
     Serialization(serde_json::Error),
     /// Given endpoint was not found: {0}
```
janpetschexain, line 532: :art: we should update the error message as well
```suggestion
    /// Given resource was not found: {0}
```

web-api-shared/src/elastic.rs
```diff
@@ -367,10 +527,24 @@ pub enum Error {
...
+pub trait NotFoundAsOptionExt<T> {
+    fn not_found_as_option(self) -> Result<Option<T>, Error>;
+}
+
+impl<T> NotFoundAsOptionExt<T> for Result<T, Error> {
+    fn not_found_as_option(self) -> Result<Option<T>, Error> {
+        match self {
+            Err(Error::ResourceNotFound(_)) => Ok(None),
+            Ok(value) => Ok(Some(value)),
+            Err(error) => Err(error),
```
janpetschexain, line 545: ❓ shouldn't the happy path come first in a match?
```suggestion
        match self {
            Ok(value) => Ok(Some(value)),
            Err(Error::ResourceNotFound(_)) => Ok(None),
            Err(error) => Err(error),
```

web-api-shared/src/elastic.rs
```diff
@@ -219,35 +247,144 @@ impl<I> BulkResponse<I> {
...
 
-impl<'de> Deserialize<'de> for NoSource {
-    fn deserialize<D>(_: D) -> Result<Self, D::Error>
+impl<'de> Deserialize<'de> for SerdeDiscard {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
     where
         D: Deserializer<'de>,
     {
-        Ok(Self)
+        use serde::de;
```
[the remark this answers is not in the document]
rustonaut: I can (will) do so, but the imported `de` doesn't have much meaning outside of the serde blog.

janpetschexain (approved):
lgtm

PR #1016, opened 2023-07-05
ET-4600 date filter

janpetschexain:
**Reference**

- [ET-4600]
- requires #1010
- followed by #1017

**Summary**

- declare date filter in the spec & update changelog
- impl date filter serde
- add date filter in elasticsearch calls including date range merging
- deprecate `published_after` request option & make it a filter if present

[ET-4600]: https://xainag.atlassian.net/browse/ET-4600?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


web-api/src/personalization/routes.rs
```diff
@@ -138,61 +139,96 @@ const fn default_include_properties() -> bool {
 
 /// Represents personalized documents query params.
 #[derive(Debug, Deserialize)]
-struct PersonalizedDocumentsQuery {
+struct UnvalidatedPersonalizedDocumentsQuery {
     count: Option<usize>,
     published_after: Option<DateTime<Utc>>,
+    filter: Option<String>,
```
janpetschexain, line 145: we need to deserialize the filter as an `Option<String>` first, because the query params can be url encoded, which would be break deserializing directly as a `Option<Filter>`. hence i adopted the validation approach from the other endpoints here.

maybe in a follow up, we should also deprecate the query params and move them to a request body.

web-api/src/ingestion/routes.rs
```diff
@@ -51,21 +51,10 @@ use crate::{
         DocumentTags,
     },
     storage::{self, property_filter::IndexedPropertiesSchemaUpdate},
+    utils::deprecate,
     Error,
 };
 
-// https://datatracker.ietf.org/doc/html/draft-dalal-deprecation-header-00
```
janpetschexain, line 57: the macro moved into the `utils` module

web-api/src/personalization/filter.rs
```diff
@@ -288,6 +308,74 @@ impl Filter {
             Self::Combine(combine) => combine.filters.is_below_depth(max),
         }
     }
+
+    pub(crate) fn insert_published_after(
```
janpetschexain, line 312: we only need to keep this around until the deprecated request option gets removed.

web-api/src/personalization/filter.rs
```diff
@@ -288,6 +308,74 @@ impl Filter {
...
+            };
+
+            Some(filter)
+        } else {
+            filter
+        }
+    }
+}
+
+impl<'de> Deserialize<'de> for Filter {
```
janpetschexain, line 355: i know

web-api/src/personalization/filter.rs
```diff
@@ -288,6 +308,74 @@ impl Filter {
...
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        // this is essentially what #[serde(untagged)] generates, but with better error messages
+        // than the rather inexpressive "data did not match any variant of untagged enum Filter"
+        use serde::__private::de::{Content, ContentRefDeserializer};
```
janpetschexain, line 362: i know this is not very nice, but i tripped multiple times over the issue that `#[serde(untagged)]` generates a pretty useless error message. if there is any better way to bubble up the errors of the variants, then i'm happy to change this.

web-api/src/storage/elastic/filter.rs
```diff
@@ -117,27 +168,106 @@ mod tests {
...
+        assert_eq!(
+            filter,
+            vec![
+                json!({}),
+                json!({ RANGE: { "a": { "gt": "d", "lt": "c" } } }),
+                json!({}),
+                json!({}),
+                json!({}),
+            ],
+        );
```
rustonaut, line 224: :wrench:  This tests are not correct.

We can not assume if someone writes:

```
{ "$and": [
  { "a": { "$gt":  "b" } },
  { "a": { "$lt":  "c" } },
  { "a": { "$gt":  "d" } },
] }
```

That they mean range `(d;c)`.

It would imply that it's range `(x;c)` where x is the larger date time between `b` and `d` (be aware that this isn't the same as the higher ordered string due to timezones).

web-api/src/storage/elastic/filter.rs
```diff
@@ -14,37 +14,96 @@
...
+const BOOL: &str = "bool";
 const FILTER: &str = "filter";
 const MINIMUM_SHOULD_MATCH: &str = "minimum_should_match";
+const RANGE: &str = "range";
 const SHOULD: &str = "should";
 const TERM: &str = "term";
 const TERMS: &str = "terms";
 
-fn extend_value(filter: &mut JsonObject, occurence: &'static str, clause: Value) {
+fn merge_range(clause: &mut Vec<Value>) {
```
rustonaut, line 32: :wrench:  maybe indicate in the name that this can only be used for conjunctions but not disjunctions of clauses.

web-api/src/storage/elastic/filter.rs
```diff
@@ -14,37 +14,96 @@
...
+                .unwrap(/* clause[RANGE] is object */)
+                .keys()
+                .next(/* clause[RANGE] contains exactly one key */)
+                .map(String::as_str)
+            })
+    }
+
+    let mut i = 0;
+    while i < clause.len() {
+        if let Some(field) = is_range(&clause[i]).map(ToString::to_string) {
```
rustonaut, line 49: :art: `ToString` is used as a short hand for display formatting, but here we create a owned string from a `&str` so using `ToOwned` or `String::from` would be a bit better stylistically.

web-api/src/storage/elastic/filter.rs
```diff
@@ -14,37 +14,96 @@
...
+                        unreachable!(/* clause[j][RANGE][field] is object */);
+                    };
+                    clause[i]
+                        .as_object_mut()
+                        .unwrap(/* clause[i] is object */)[RANGE]
+                        .as_object_mut()
+                        .unwrap(/* clause[i][RANGE] is object */)[&field]
+                        .as_object_mut()
+                        .unwrap(/* clause[i][RANGE][field] is object */)
+                        .extend(range);
```
rustonaut, line 75: :wrench:  See below we can't just override operators of the same kind.

web-api/src/storage/elastic/filter.rs
```diff
@@ -14,37 +14,96 @@
...
+                .as_object()
+                .unwrap(/* clause[RANGE] is object */)
+                .keys()
+                .next(/* clause[RANGE] contains exactly one key */)
+                .map(String::as_str)
+            })
+    }
+
+    let mut i = 0;
+    while i < clause.len() {
```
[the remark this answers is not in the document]
janpetschexain: but this will panic. in the inner while loop the `clause` gets shorter everytime we `swap_remove()` a range clause. so once we are again in the outer loop, `clause.len()` may have changed and `clause[i]` can panic.

web-api/src/storage/elastic/filter.rs
janpetschexain: good catch, i'll update the range merging and also the tests.

PR #1018, opened 2023-07-05
ET-4707 add generic huggingface tokenizer

petricevich:


bert/src/tokenizer/huggingface.rs
```diff
@@ -0,0 +1,198 @@
...
+            .iter()
+            .take(20)
+            .copied()
+            .collect::<Vec<_>>();
+        assert!(
+            first_ids
+                == vec![
+                    2, 4538, 2128, 8561, 1, 6541, 69469, 2762, 5, 3, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0
+                ]
+        );
```
janpetschexain, line 116: :art: iterating and collecting into a vector is definitely a possibility to compare, we can improve readability for example via slicing though.
```suggestion
        assert_eq!(
            encoding.token_ids.slice(s![0, ..20]),
            arr1(&[2, 4538, 2128, 8561, 1, 6541, 69469, 2762, 5, 3, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]),
        );
```
the slicing returns an array, that's why we need to wrap the expected values in one instead of a vec. the items can be imported via `use ndarray::{arr1, s};`. similarly for the other tests.

.github/scripts/download_assets.sh
```diff
@@ -49,9 +49,8 @@ if [ $# -gt 0 ]; then
...
 else
-    download xaynia v0002
-    download smbert v0003
-    download smbert_mocked v0003
-    download sjbert v0003
-    download smroberta_tokenizer v0000
+    download qasmbert v0002
+    download smbert v0004
+    download smbert_mocked v0004
+    download e5_mocked v0000
```
janpetschexain, line 55: ❓ it looks like this contains the regular e5 model, is that correct?

the mocked smbert model is a remnant from the times where we developed for mobile targets where we needed fast dummy models for the tests. the mocked smbert just takes the same inputs as the original smbert, then does nothing and returns a constant output in the same format as the original one. we should probably clean this up a bit once we have time.

if we can take the original e5 model for tests anyways, then there is no need to create a mocked e5 model. we should also not name the real model as mocked.

bert/src/config.rs
```diff
@@ -37,27 +37,16 @@ use crate::{
...
-/// path = "mecab"
-///
-/// # the path is always `vocab.txt`
+/// # the path is always `tokenizer.json`
 /// [tokenizer]
-/// cleanse-accents = true
-/// cleanse-text = true
-/// lower-case = false
-/// max-chars = 100
+/// add_special_tokens = true
```
janpetschexain, line 42: 🔧 this contradicts the implementation where `-` are used in the name of this option instead of `_`.

either way is fine for me, we should just apply the casing consistently to all config option names.

bert/src/lib.rs
```diff
@@ -57,8 +57,5 @@ pub use crate::{
     },
 };
 
-/// A Bert pipeline with an average pooler.
-pub type AvgBert = Pipeline<crate::tokenizer::bert::Tokenizer, AveragePooler>;
-
-/// A Roberta pipeline with an average pooler.
-pub type AvgRoberta = Pipeline<crate::tokenizer::roberta::Tokenizer, AveragePooler>;
+/// A Transformer pipeline with an average pooler.
+pub type AvgE5 = Pipeline<crate::tokenizer::huggingface::Tokenizer, AveragePooler>;
```
janpetschexain, line 61: ❓ the pipeline is not really limited to e5, since we are able to load any transformer model. wdyt about `AvgTransformer` or `AvgEmbedder` or sth similar?

web-api/src/models.rs
```diff
@@ -125,13 +125,13 @@ fn is_valid_property_id(value: &str) -> bool {
 fn is_valid_string(value: &str, len: usize) -> bool {
     static RE: Lazy<Regex> = Lazy::new(|| Regex::new(r"^[^\x00]+$").unwrap());
 
-    (1..=len).contains(&value.len()) && RE.is_match(value)
+    RE.is_match(value)
```
janpetschexain, line 128: :art: this is the quick fix you mentioned in the daily, right? this is not supposed to be merged later on.

janpetschexain (commented):
looks already good, there are some remaining issues.

i also left some suggestions regarding rust conventions.

the ci also complains about some lints, it's probably good to fix them such that the new tests which you added are also run in the ci.

.github/scripts/download_assets.sh
petricevich: This is a "mocked" model in that it is a tiny version of the model that otherwise matches the architecture and tokenizer. The idea was to reduce download size and increase speed of tests (it has 2 layers and embedding size of 96 I believe). If that is not a concern, I could put in the real model

janpetschexain: i see, if we already have it, we can stay with the mocked model for tests. faster tests are always nice to have.

janpetschexain (approved):
lgtm 👍

PR #1019, opened 2023-07-10
increase filter limits

rustonaut:
for filter `$in` 10  => 500

be aware the JSON body size limits still apply

### DIFF
--- a/web-api/src/personalization/filter.rs
+++ b/web-api/src/personalization/filter.rs
@@ -41,6 +41,10 @@ pub(crate) enum CompareOp {
     Lte,
 }
 
+impl CompareOp {
+    const MAX_VALUES_PER_IN: usize = 500;
+}
+
 #[derive(Clone, Debug, PartialEq)]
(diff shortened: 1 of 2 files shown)


web-api/src/personalization/filter.rs
```diff
@@ -185,7 +185,7 @@ pub(crate) enum CombineOp {
 pub(crate) struct Filters(Vec<Filter>);
 
 impl Filters {
-    const MAX_LEN: usize = 10;
+    const MAX_LEN: usize = 500;
```
Robert-Steiner, line 188: not sure if this is the right value 
in the pr in which we implement the `in` filter we don't touch that value but instead have the value defined here https://github.com/xaynetwork/xayn_discovery_engine/blob/2a7f03e88083c03b920a8f5cf242366923dc81cd/web-api/src/personalization/filter.rs#L99

Robert-Steiner (approved):
can you please build the images and send me the docker tag

PR #1020, opened 2023-07-11
ET-4772 improve generated docs

janpetschexain:
**Reference**

- [ET-4772]

**Summary**

it is probably best to visually review these changes via the `just generate-openapi-doc` command:
- group routes into sections
- improve phrasing and layout

[ET-4772]: https://xainag.atlassian.net/browse/ET-4772?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/web-api/openapi/back_office.yaml
+++ b/web-api/openapi/back_office.yaml
@@ -1,4 +1,5 @@
 openapi: 3.1.0
+
 info:
   title: Back Office API
   version: 2.3.0
@@ -32,9 +33,35 @@ info:
     Properties are optional data for documents, which are usually needed to properly show the document back to the user, when it returns as a personalized document.
     If for example, you'd wish to integrate a type of carousel view, listing a total of 10 personalized documents in a "for you"-section, then you might choose to display each document as an image and title, with a url to link the user to when pressed.
     For this, you would need three document properties: `image`, `link` and `title`.
+  x-logo:
+    url: https://uploads-ssl.webflow.com/5ea197660b956f76d26f0026/637f7edd68c1ae2f12d4689e_Xayn%20Logo%202022%20-%20Footer.svg
+    altText: Xayn
 tags:
   - name: back office
     description: System management, including documents and properties.
+    x-traitTag: true
+  - name: documents
+    x-displayName: Documents
+  - name: candidates
+    x-displayName: Document candidates
+  - name: properties
+    x-displayName: Document properties
+  - name: property
+    x-displayName: Document property
+  - name: property indexing
+    x-displayName: Document property indexing
+x-tagGroups:
+  - name: Documents
+    tags:
+      - documents
+  - name: Document candidates
+    tags:
+      - candidates
+  - name: Document properties
+    tags:
+      - properties
+      - property
+      - property indexing
 
 security:
   - ApiKeyAuth: []
@@ -44,6 +71,7 @@ paths:
     post:
       tags:
         - back office
+        - documents
       summary: Add documents to the system.
       description: |
         Add documents to the system.
@@ -75,6 +103,7 @@ paths:
     delete:
       tags:
         - back office
+        - documents
       summary: Delete all listed documents.
       description: Delete all documents listed in the request body.
       operationId: deleteDocuments
@@ -94,6 +123,7 @@ paths:
     get:
       tags:
         - back office
+        - candidates
       summary: Get the documents considered for recommendations.
       description: Get the documents considered for recommendations.
       operationId: listDocumentCandidates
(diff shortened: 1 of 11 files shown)


web-api/openapi/back_office.yaml
```diff
@@ -44,11 +75,10 @@ paths:
     post:
       tags:
         - back office
-      summary: Add documents to the system.
-      description: |
-        Add documents to the system.
-
-        The system will create a representation of the document that will be used to match it against the preferences of a user.
+        - documents
+      summary: Ingest documents
```
janpetschexain, line 79: i changed all `summary` items to short titles/headlines, as they appear as such in the navigation overview. the `description` items contain full sentences/paragraphs.

web-api/openapi/back_office.yaml
```diff
@@ -32,9 +33,39 @@ info:
...
+    x-displayName: Documents
+  - name: candidates
+    x-displayName: Document candidates
+  - name: properties
+    x-displayName: Document properties
+  - name: property
+    x-displayName: Document property
+  - name: property indexing
+    x-displayName: Document property indexing
+x-tagGroups:
```
janpetschexain, line 57: these `x-` extensions are redocly specific, if we change the doc generator we need to adjust them accordingly.

web-api/openapi/parameters/path/document_id.yml
janpetschexain, line 1: this file moved into `parameters/path/id.yml`.

web-api/openapi/parameters/path/user_id.yml
janpetschexain, line 1: this file moved into `parameters/path/id.yml`.

web-api/openapi/schemas/document.yml
```diff
@@ -1,15 +1,15 @@
 DocumentId:
-  description: Id of a document.
+  description:
+    $ref: './id.yml#/Id/description'
   $ref: './id.yml#/Id'
-  example: "document_id"
```
janpetschexain, line 5: in general syblings to items already defined in a `$ref` are not allowed and ignored in openapi. there are exceptions to `summary` and `description` which overwrite the `$ref`'s one.

hence i took the `$ref`ed description which is much more descriptive and also holds important information about constraints (we have to explicitly reference it here, otherwise references of references fail to resolve in the validator).

other items like `example` are ignored anyways, so i removed them. the generator will pick up the `example` of the reference.

web-api/openapi/schemas/document.yml
```diff
@@ -28,8 +28,8 @@ DocumentPropertyArrayString:
   maxItems: 100
 
 DocumentProperty:
-  description: |
-    Arbitrary data that can be attached to a document. A subset of JSON is supported.
+  description: |-
+    Mostly arbitrary data that can be attached to a document. A subset of JSON is supported.
     When a property is set to null or an empty array, it is treated as if that property had no values.
     This means that they won't appear in searches with filters on these properties.
```
janpetschexain, line 34: ❗ this description doesn't show up anywhere in the generated docs (likely because of the `oneOf`) and i didn't find a way to display it, also not via extensions. so we have to figure out if we can find a solution for that issue later on, since it contains important information.

joecks (approved):
Muy bien

web-api/openapi/front_office.yaml
```diff
@@ -1,20 +1,21 @@
...
   version: 2.3.0
   description: |-
     # Front Office
     The front office is typically used within front-end apps, for example a website or a mobile application.
-    With this API, you can handle interactions with documents and request a list of personalized documents.
+    With this API, you can handle interactions with documents and search for personalized documents.
 
     ## Format
-    All the request bodies in this API are JSON encoded and their `Content-Type` headers should be set to `application/json`.
+    All the request bodies in this API are JSON encoded and their `content-type` header should be set to `application/json`.
```
rustonaut, line 12: :art: Http headers are generally referenced in their capitalized form if you look at RFCs and similar.

```suggestion
    All the request bodies in this API are JSON encoded and their `Content-Type` header should be set to `application/json`.
```

rustonaut (approved):
one small thing otherwise lgtm

web-api/openapi/front_office.yaml
janpetschexain: the issue is that upper case characters in header names are invalid in http/2 and http/3, see for example https://www.rfc-editor.org/rfc/rfc7540#section-8.1.2 for http/2. even though we shouldn't rule out http/1, we should still stick to the conventions of the newer versions.

PR #1027, opened 2023-07-13
ET-4818 handle filter edge cases

janpetschexain:
**Reference**

- [ET-4818]
- requires #1026

**Summary**

- add tests for null & empty string array property filtering


web-api/tests/filter.rs
```diff
@@ -842,6 +853,121 @@ fn test_filter_date() {
...
+            let documents = send_assert_json::<SemanticSearchResponse>(
+                &client,
+                client
+                    .post(personalization_url.join("/semantic_search")?)
+                    .json(&json!({
+                        "document": { "query": "zero" },
+                        "filter": { "p2": { "$eq": true } }
+                    }))
+                    .build()?,
+                StatusCode::OK,
```
rustonaut, line 941: :art: maybe we add an annotation that it should fail once we validate against the schema?

rustonaut (approved):
lgtm

PR #1028, opened 2023-07-17
ET-4777 merge range filter

janpetschexain:
**Reference**

- [ET-4777]
- requires #1027
- followed by #1029

**Summary**

- logically merge elasticsearch ranges for the same property in `$and` and `$or` filters: merge `$gt` and `$gte` into a greater range as well as `$lt` and `$lte` into a less range
- for `$and` filters additionally merge the greater and less ranges into one range

[ET-4777]: https://xainag.atlassian.net/browse/ET-4777?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/web-api/src/personalization/filter.rs
+++ b/web-api/src/personalization/filter.rs
@@ -79,10 +79,10 @@ impl<'de> Deserialize<'de> for CompareWith {
                         Unexpected::Unsigned(value)
                     } else if let Some(value) = value.as_i64() {
                         Unexpected::Signed(value)
+                    } else if let Some(value) = value.as_f64() {
+                        Unexpected::Float(value)
                     } else {
-                        Unexpected::Float(value.as_f64().unwrap(
-                            /* always possible without feature `arbitrary_precision` */
-                        ))
+                        Unexpected::Other("number")
                     }
                 }
 
--- a/web-api/src/personalization/filter.rs
+++ b/web-api/src/personalization/filter.rs
@@ -79,10 +79,10 @@ impl<'de> Deserialize<'de> for CompareWith {
                         Unexpected::Unsigned(value)
                     } else if let Some(value) = value.as_i64() {
                         Unexpected::Signed(value)
-                    } else if let Some(value) = value.as_f64() {
-                        Unexpected::Float(value)
                     } else {
-                        Unexpected::Other("number")
+                        Unexpected::Float(value.as_f64().unwrap(
+                            /* always possible without feature `arbitrary_precision` */
+                        ))
                     }
                 }
 
--- a/web-api/src/storage/elastic/filter.rs
+++ b/web-api/src/storage/elastic/filter.rs
@@ -73,26 +73,18 @@ impl Serialize for Terms<'_> {
 
 fn cmp_range(this: &Value, other: &Value) -> Option<Ordering> {
     match (this, other) {
-        (Value::Number(this), Value::Number(value)) => {
-            // unwraps are always possible without feature `arbitrary_precision`
-            Some(this.as_f64().unwrap().total_cmp(&value.as_f64().unwrap()))
-        }
-        (Value::String(this), Value::String(value)) => {
-            // unwraps are always possible since range strings are valid dates
-            Some(
-                DateTime::parse_from_rfc3339(this)
-                    .unwrap()
-                    .cmp(&DateTime::parse_from_rfc3339(value).unwrap()),
-            )
-        }
-        // TODO: remove Option<_> and change to unreachable! once filters are validated against indexed properties schema
+        (Value::Number(this), Value::Number(value)) => this
+            .as_f64()
+            .and_then(|this| value.as_f64().map(|value| this.total_cmp(&value))),
+        (Value::String(this), Value::String(value)) => DateTime::parse_from_rfc3339(this)
+            .and_then(|this| DateTime::parse_from_rfc3339(value).map(|value| this.cmp(&value)))
+            .ok(),
         _ => None,
-        // _ => unreachable!(/* values are both numbers or dates */),
     }
 }
 
 #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq, Serialize)]
-#[serde(rename_all = "lowercase")]
+#[serde(rename_all = "snake_case")]
 enum GreaterRangeOp {
     Gt,
     Gte,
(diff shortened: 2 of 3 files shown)


web-api/src/storage/elastic/filter.rs
```diff
@@ -70,34 +71,141 @@ impl Serialize for Terms<'_> {
...
 }
 
+#[derive(Debug)]
+struct LessRange<'a> {
+    operation: LessRangeOp,
+    value: &'a DocumentProperty,
+}
+
+impl LessRange<'_> {
+    fn and(&mut self, less: Self) {
```
rustonaut, line 149: :art: ? minor nit-pick by convention of implementing operator like function the parameter containing the other value used with the operator is most times called `other`.  This has the benefit or not carrying any potentially accident meaning. Like here `less` not necessary being less then self.

web-api/src/storage/elastic/filter.rs
```diff
@@ -249,6 +357,110 @@ enum Clause<'a> {
     MustNot(MustNot<'a>),
 }
 
+fn fully_merge_range(mut clause: Vec<Clause<'_>>) -> Vec<Clause<'_>> {
```
rustonaut, line 360: :wrench: 

```suggestion
fn merge_ranges_in_and_clause(mut clause: Vec<Clause<'_>>) -> Vec<Clause<'_>> {
```

(name is just an arbitrary proposal not what I want it to be)

This is only correctly usable iff applied to a clauses from a `and` statement, I think it's important for the function name to reflect this. At least for me `fully` doesn't do so at all.

web-api/src/storage/elastic/filter.rs
```diff
@@ -249,6 +357,110 @@ enum Clause<'a> {
...
+                } else {
+                    j += 1;
+                }
+            }
+        }
+        i += 1;
+    }
+
+    clause
+}
```
rustonaut, line 400: sorting (i.e. semi-partitioning) by the tuple of `(field, operand, operand value type)` using an existing library function probably would yield more readable code, especially if we fold/fold-in-place for the merging step

but we can keep it like it is for now

web-api/src/storage/elastic/filter.rs
```diff
@@ -249,6 +357,110 @@ enum Clause<'a> {
...
+
+fn partly_merge_range(mut clause: Vec<Clause<'_>>) -> Vec<Clause<'_>> {
```
rustonaut, line 402: :wrench: 

see previous comment

potential fix:

```suggestion
fn merge_ranges_in_or_clause(mut clause: Vec<Clause<'_>>) -> Vec<Clause<'_>> {
```

web-api/src/storage/elastic/filter.rs
```diff
@@ -249,6 +357,110 @@ enum Clause<'a> {
...
+    fn is_less_range<'a>(clause: &Clause<'a>) -> Option<&'a DocumentPropertyId> {
+        if let Clause::Range(range) = clause {
+            range.less.is_some().then_some(range.field)
+        } else {
+            None
+        }
+    }
+
+    let mut i = 0;
+    while i < clause.len() {
```
rustonaut, line 420: :wrench: 

```suggestion
    for i in 0..clause.len() {
```

web-api/src/personalization/filter.rs
```diff
@@ -79,10 +79,10 @@ impl<'de> Deserialize<'de> for CompareWith {
                         Unexpected::Unsigned(value)
                     } else if let Some(value) = value.as_i64() {
                         Unexpected::Signed(value)
-                    } else if let Some(value) = value.as_f64() {
-                        Unexpected::Float(value)
                     } else {
-                        Unexpected::Other("number")
+                        Unexpected::Float(value.as_f64().unwrap(
+                            /* always possible without feature `arbitrary_precision` */
```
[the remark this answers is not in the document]
janpetschexain: fair point, i'll replace the unwrapping

web-api/src/storage/elastic/filter.rs
```diff
@@ -249,6 +357,110 @@ enum Clause<'a> {
...
+    fn is_range<'a>(clause: &Clause<'a>) -> Option<&'a DocumentPropertyId> {
+        if let Clause::Range(range) = clause {
+            Some(range.field)
+        } else {
+            None
+        }
+    }
+
+    let mut i = 0;
+    while i < clause.len() {
```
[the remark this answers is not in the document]
janpetschexain: as said before, this will panic because the condition in the for loop is only checked once at loop initialization. as we remove elements the vec shrinks and we will get out of bounds panic. that's why we need to increment the counter manually and check the condition at every iteration.

web-api/src/storage/elastic/filter.rs
janpetschexain: this panics, see https://github.com/xaynetwork/xayn_discovery_engine/pull/1028#discussion_r1269126269

web-api/src/storage/elastic/filter.rs
```diff
@@ -70,34 +71,141 @@ impl Serialize for Terms<'_> {
...
+        (Value::String(this), Value::String(value)) => {
+            // unwraps are always possible since range strings are valid dates
+            Some(
+                DateTime::parse_from_rfc3339(this)
+                    .unwrap()
+                    .cmp(&DateTime::parse_from_rfc3339(value).unwrap()),
+            )
+        }
+        // TODO: remove Option<_> and change to unreachable! once filters are validated against indexed properties schema
+        _ => None,
```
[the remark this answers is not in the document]
janpetschexain: makes sense, i'll keep the option and make it a bit more robust

web-api/src/storage/elastic/filter.rs
```diff
@@ -70,34 +71,141 @@ impl Serialize for Terms<'_> {
     }
 }
 
+fn cmp_range(this: &Value, other: &Value) -> Option<Ordering> {
```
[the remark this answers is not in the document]
janpetschexain: it turns out that this is not completely possible, but at least to a certain extend. the issue is that the secondary ordering depends on the `and`/`or` case, but if we put the primary and secondary ordering into one `PartialOrd for GreaterRange`, then we can't differentiate based on the two cases anymore (we would either need a differentiator field within the `GreaterRange` or two distinct `GreaterRangeAnd`/`GreaterRangeOr` or sth similar, all of which would create quite some boilerplate).

but we can leverage `PartialOrd` at least a bit, so what i did is to move the `range_cmp()` into a `PartialOrd for DocumentProperty` and to derive `PartialOrd for GreaterRangeOp/LessRangeOp`. this simplifies the `and()`/`or()` methods of `GreaterRange`/`LessRange` and makes them more readable. and it still allows us two have different secondary orderings depending on the method as required.

rustonaut (approved):
lgtm

PR #1029, opened 2023-07-18
ET-4820 wrap clause type

janpetschexain:
**Reference**

- [ET-4820]
- requires #1028

**Summary**

- impl `Clauses` as the root filter clauses wrapper, don't convert it into a json object inside the filter module

[ET-4820]: https://xainag.atlassian.net/browse/ET-4820?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/web-api/src/storage/elastic.rs
+++ b/web-api/src/storage/elastic.rs
@@ -21,7 +21,7 @@ pub(crate) use client::{Client, ClientBuilder};
 use itertools::Itertools;
 use reqwest::Method;
 use serde::{Deserialize, Serialize};
-use serde_json::json;
+use serde_json::{json, Value};
 use tracing::info;
 use xayn_ai_bert::NormalizedEmbedding;
 pub(crate) use xayn_web_api_shared::elastic::{BulkInstruction, Config};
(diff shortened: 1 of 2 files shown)


web-api/src/storage/elastic.rs
```diff
@@ -497,7 +498,11 @@ struct KnnSearchParts {
 
 impl KnnSearchParams<'_> {
     fn create_common_knn_search_parts(&self) -> KnnSearchParts {
-        let inner_filter = self.create_search_filter();
+        let Ok(Value::Object(inner_filter)) =
+            serde_json::to_value(Clauses::new(self.filter, self.excluded))
+        else {
+            unreachable!(/* filter clauses is valid json object */);
+        };
```
janpetschexain, line 505: 📌 in a follow up, we should also apply the same ideas to `KnnSearchParts`, such that we have typed it out completely and the json serialization only happens at the request body creation.

rustonaut (approved):
lgtm

PR #1032, opened 2023-07-18
ET-4756 implement minimal base implementation for document splitting

rustonaut:
**Limitations:**

- only very simple (but not naive) sentence splitting algorithm
- only english support
- scoring (rerank by interest) with documents with multiple splits needs to be decided on (current implementation might be fine)
- interacting on documents with multiple embedding needs improvements (currently strongly biased for documents the more splits they have), might need fundamental changes depending on customer use-case
- personalized `/semantic_search`  with stateless user history  has the same issue as normal interacting 
- handling `/semantic_search` based on finding documents similar documents with splits needs a full rework (currently it won't fall over but will not  work well) 
- when matching multiple snippets of the same document we currently sum-up the scores and keep track of which snippets where matched (internally, returned with dev option)
  - with the option which return `snippet` we also should (also? instead?) return the slice of the split
  - customers might want to know not just which split was found but ranking between the found splits
  - customers (we?) might want allow reacting to splits (potentially multiple at once) instead of just documents, through we probably still want allow to also react to the document as a whole even if split 

**References:**

- issue [ET-4756]
- story [ET-4755]

### DIFF
--- a/coi/src/document.rs
+++ b/coi/src/document.rs
@@ -22,7 +22,7 @@ pub trait Document {
     fn id(&self) -> &Self::Id;
 
     /// Gets the embedding of the document.
-    fn embedding(&self) -> &NormalizedEmbedding;
+    fn embeddings<'a>(&'a self) -> Box<dyn Iterator<Item = &'a NormalizedEmbedding> + 'a>;
 }
 
 #[cfg(test)]
@@ -46,14 +46,14 @@ pub(crate) mod tests {
 
     pub(crate) struct TestDocument {
         pub(crate) id: DocumentId,
-        pub(crate) embedding: NormalizedEmbedding,
+        pub(crate) embeddings: Vec<NormalizedEmbedding>,
     }
 
     impl TestDocument {
         pub(crate) fn new(id: usize, embedding: NormalizedEmbedding) -> Self {
             Self {
                 id: DocumentId::mocked(id),
-                embedding,
+                embeddings: vec![embedding],
             }
         }
     }
@@ -65,8 +65,8 @@ pub(crate) mod tests {
             &self.id
         }
 
-        fn embedding(&self) -> &NormalizedEmbedding {
-            &self.embedding
+        fn embeddings<'a>(&'a self) -> Box<dyn Iterator<Item = &'a NormalizedEmbedding> + 'a> {
+            Box::new(self.embeddings.iter())
         }
     }
 }
--- a/web-api-db-ctrl/elasticsearch/mapping.json
+++ b/web-api-db-ctrl/elasticsearch/mapping.json
@@ -14,6 +14,9 @@
             "tags": {
                 "type": "keyword"
             },
+            "parent": {
+                "type": "keyword"
+            },
             "properties": {
                 "dynamic": false,
                 "properties": {
--- a/web-api-db-ctrl/postgres/tenant/20230712141144_document.sql
+++ b/web-api-db-ctrl/postgres/tenant/20230712141144_document.sql
@@ -1,0 +1,31 @@
+-- Copyright 2023 Xayn AG
+--
+-- This program is free software: you can redistribute it and/or modify
+-- it under the terms of the GNU Affero General Public License as
+-- published by the Free Software Foundation, version 3.
+--
+-- This program is distributed in the hope that it will be useful,
+-- but WITHOUT ANY WARRANTY; without even the implied warranty of
+-- MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+-- GNU Affero General Public License for more details.
+--
+-- You should have received a copy of the GNU Affero General Public License
+-- along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+CREATE TYPE preprocessing_step AS ENUM (
+    'none',
+    'split',
+    'summarize'
+);
+
+ALTER TABLE document
+    ALTER COLUMN is_summarized DROP DEFAULT,
+    ALTER COLUMN is_summarized TYPE preprocessing_step
+        USING CASE
+            WHEN is_summarized THEN 'summarize'::preprocessing_step
+            ELSE 'none'::preprocessing_step
+        END,
+    ALTER COLUMN is_summarized SET DEFAULT 'none';
+
+ALTER TABLE document
+    RENAME COLUMN is_summarized TO preprocessing_step;
--- a/web-api-shared/src/elastic.rs
+++ b/web-api-shared/src/elastic.rs
@@ -420,12 +420,9 @@ impl Client {
             .await
     }
 
-    pub async fn search_request<I>(
-        &self,
-        mut body: JsonObject,
-    ) -> Result<HashMap<I, (f32, HashSet<usize>)>, Error>
+    pub async fn search_request<Id>(&self, mut body: JsonObject) -> Result<ScoreMap<Id>, Error>
     where
-        I: FromStr + Eq + Hash,
+        Id: FromStr + Eq + Hash,
     {
         if body.get("size") == Some(&json!(0)) {
             return Ok(HashMap::new());
@@ -440,7 +437,7 @@ impl Client {
             )
             .await?;
 
-        let mut map = HashMap::default();
+        let mut map = HashMap::<_, (f32, HashSet<usize>)>::default();
         for hit in response.hits.hits.into_iter() {
             if let Some(suffix) = hit.id.strip_prefix("_child.") {
                 let (idx, parent_id) = suffix.split_once('.').ok_or_else(|| {
@@ -529,7 +526,7 @@ impl Client {
         method: Method,
         url: Url,
         body: Option<B>,
-    ) -> Result<Option<T>, Error>
+    ) -> Result<T, Error>
     where
         B: Serialize,
         T: DeserializeOwned,
@@ -546,6 +543,8 @@ impl Client {
         self.query_with_bytes(method, url, post_data).await
     }
 }
+
+pub type ScoreMap<Id> = HashMap<Id, (f32, HashSet<usize>)>;
 
 #[derive(Debug, Error, displaydoc::Display, From)]
 pub enum Error {
--- a/web-api-shared/src/elastic.rs
+++ b/web-api-shared/src/elastic.rs
@@ -13,7 +13,7 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 use std::{
-    collections::HashMap,
+    collections::{HashMap, HashSet},
     fmt::{Debug, Display},
     future::Future,
     hash::Hash,
@@ -22,6 +22,7 @@ use std::{
     time::Duration,
 };
 
+use anyhow::anyhow;
 use bytes::Bytes;
 use derive_more::From;
 use futures_retry_policies::tokio::retry;
@@ -186,14 +187,18 @@ impl Client {
 
 #[derive(Debug, Serialize)]
 #[serde(rename_all = "lowercase")]
-pub enum BulkInstruction<'a, I> {
+pub enum BulkInstruction<I> {
     Index {
         #[serde(rename = "_id")]
-        id: &'a I,
+        id: I,
     },
     Delete {
         #[serde(rename = "_id")]
-        id: &'a I,
+        id: I,
+    },
+    Update {
+        #[serde(rename = "_id")]
+        id: I,
     },
 }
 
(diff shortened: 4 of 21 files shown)


acrrd:
@rustonaut Can you please take a look at the conflicts?

web-api-shared/src/elastic.rs
```diff
@@ -415,33 +420,51 @@ impl Client {
...
-        })
+        let response = self
+            .query_with_json::<_, SearchResponse<String>>(
+                Method::POST,
+                self.create_url(["_search"], None),
+                Some(body),
+            )
+            .await?;
+
+        let mut map = HashMap::<_, (f32, HashSet<usize>)>::default();
```
rustonaut, line 440: :pushpin:  for now we track only which splits we found and with a dev option for tests return that but we probably need/want something better (see PR description)

web-api-db-ctrl/elasticsearch/mapping.json
```diff
@@ -14,6 +14,9 @@
             "tags": {
                 "type": "keyword"
             },
+            "parent": {
+                "type": "keyword"
+            },
```
rustonaut, line 19: :wrench:  I will change this as part of some out of band review feedback.

rustonaut: I decided against it  the degree of complexity this adds to the code in ways which are prone to have bugs later down the line makes it not worth it to do so AFIK.

The reason this is a field instead of just using the part in the id of the document is so that it's indexed and we can delete by-parent id.

This is needed due upserts no longer working with split documents, as the number of childs in a upsert can have changed.

But at the point of time we do the upserts we don't always know which children the document had _before_ we started to reingest it.  So we have the choice between
either doing additiona PG queries and then sending a long list to the server, or just have a by-parent index and use delete by query in ES which is easier to write, less complex and probably more performant/cheap, too.

web-api/src/storage/elastic.rs
```diff
@@ -186,39 +192,94 @@ impl Client {
...
-                [
-                    serde_json::to_value(BulkInstruction::Index { id: &document.id })
-                        .map_err(Into::into),
-                    serde_json::to_value(IngestedDocument {
+        let pre_deletions = documents
+            .iter()
+            .filter_map(|document| (document.embeddings.len() > 1).then_some(&document.id))
+            .collect_vec();
+
+        self.delete_by_parents(pre_deletions).await?;
```
rustonaut, line 206: :thinking:  We could make `delete_by_parents` non-blocking by setting both `wait_for_completion=false` and `conflicts=proceed`.

Due to how `delete_by_parents` uses the ES versioning system it would lead to only deleting documents which haven't been changed since the start of the delete operation. If they where changed since then (through the operation below) it would fail but due to `conflict=proceed` this would be fine, too. 

Through not sure if we should.

web-api/src/storage/elastic.rs
```diff
@@ -186,39 +192,94 @@ impl Client {
...
         Ok(response.failed_documents("index", false).into())
     }
 
+    pub(super) async fn delete_by_parents(&self, parents: Vec<&DocumentId>) -> Result<(), Error> {
+        let url = self.create_url(["_delete_by_query"], [("refresh", None)]);
+        let body = json!({
+            "query": {
+                "terms": {
+                    "parent": parents,
+                }
```
rustonaut, line 261: :thinking:  In all cases we do have the number of childs the new document has. We could use this to instead of running "delete where `parent={parent}`" run "delete where `parent={parent} && child_idx >= {child_count}`.

Not sure if it's worth the additional index field.

web-api/src/storage/elastic.rs
```diff
@@ -535,65 +596,67 @@ impl KnnSearchParams<'_> {
...
-    for score in scores.values_mut() {
+    for (score, _) in scores.values_mut() {
         *score /= max_score;
     }
 
     scores
 }
 
-fn normalize_scores_if_max_gt_1<K>(mut scores: HashMap<K, f32>) -> HashMap<K, f32>
+fn normalize_scores_if_max_gt_1<K, M>(mut scores: HashMap<K, (f32, M)>) -> HashMap<K, (f32, M)>
```
janpetschexain, line 614: :art: we could reuse the `ScoreMap` type alias, it is a bit more restrictive that what is currently there, but we actually don't need the flexibility of the added generics.

:art: we can also rename the `K` to `Id` as you did in another function below
```suggestion
fn normalize_scores_if_max_gt_1<Id>(mut scores: ScoreMap<Id>) -> ScoreMap<Id>
```

web-api/src/storage/postgres.rs
```diff
@@ -121,13 +151,15 @@ impl Database {
                 .push_values(
                     documents.by_ref().take(Self::BIND_LIMIT / 6),
                     |mut builder, document| {
+                        let embeddings: Vec<_> = document.embeddings.iter().map(Json).collect();
```
janpetschexain, line 154: :art: we have `itertools` at our disposal
```suggestion
                        let embeddings = document.embeddings.iter().map(Json).collect_vec();
```

web-api/src/storage/postgres.rs
```diff
@@ -165,10 +197,10 @@ impl Database {
                 builder
                     .reset()
                     .push_tuple(ids.by_ref().take(Self::BIND_LIMIT))
-                    .push(" RETURNING document_id, is_candidate;")
+                    .push(" RETURNING document_id, is_candidate, coalesce(array_length(embeddings, 1), 0) as embedding_count;")
```
janpetschexain, line 200: :art: in the [postgres docs this function name is capital case](https://www.postgresql.org/docs/15/functions-conditional.html#FUNCTIONS-COALESCE-NVL-IFNULL). similarly below.
```suggestion
                    .push(" RETURNING document_id, is_candidate, COALESCE(array_length(embeddings, 1), 0) as embedding_count;")
```

web-api/tests/split_documents.rs
```diff
@@ -0,0 +1,448 @@
...
+            ingest(&client, &ingestion_url).await?;
+
+            let res = query("cars", 2).await?;
+            assert_eq!(res.id_ranking(), vec!["d2"]);
+            assert_eq!(res.documents[0].splits, [0, 2].into_iter().collect());
+
+            let res = query("color", 3).await?;
+            assert_eq!(res.id_ranking(), vec!["d1", "d2"]);
+            assert_eq!(res.documents[0].splits, [0, 1].into_iter().collect());
+            assert_eq!(res.documents[1].splits, [1].into_iter().collect());
```
janpetschexain, line 189: :art: the set can directly be collected like this. similarly below.
```suggestion
            assert_eq!(res.documents[0].splits, [0, 1].into());
            assert_eq!(res.documents[1].splits, [1].into());
```

web-api/tests/split_documents.rs
```diff
@@ -0,0 +1,448 @@
...
+                    .build()?,
+                StatusCode::NO_CONTENT,
+                false,
+            )
+            .await;
+            send_assert_json::<SearchResponse>(
+                &client,
+                client
+                    .post(personalization_url.join("/users/u1/personalized_documents")?)
+                    .json(&json!({}))
```
janpetschexain, line 383: :art: the request body for that endpoint is optional since all keys are optional, hence i implemented the serde of that endpoint such that an empty request body can also be omitted
```suggestion
```

web-api/src/storage/elastic.rs
rustonaut: > as we still have "refresh" in every es call, we should also keep this call waiting for completion. we can reconsider this once we deal with "refresh" again.

we did strip `refresh` already out from some calls and it also doesn't have anything to do directly with `refresh` as such call could always being before a refresh is done and complete after the refresh, i.e. refresh isn't really a synchronization primitive, it just makes some background updates run earlier and maybe less in the background AFIK

rustonaut:
> unwrapping and the empty case handling

there shouldn't be any unwrapping or empty case special handling

any instances where there is only because of parts which would have been too much for this PR to also contain

we could newtype `Embeddings` to forbid the empty case, but otherwise it we can't really attach any other logic to it as the rest is specific to where `Embeddings` is used and not `Embeddings` itself

coi/src/document.rs
```diff
@@ -22,7 +22,7 @@ pub trait Document {
     fn id(&self) -> &Self::Id;
 
     /// Gets the embedding of the document.
-    fn embedding(&self) -> &NormalizedEmbedding;
+    fn embeddings<'a>(&'a self) -> Box<dyn Iterator<Item = &'a NormalizedEmbedding> + 'a>;
```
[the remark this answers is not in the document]
rustonaut: it's a trait, using `impl Trait` in traits isn't supported without using hacks which hinder rust-analyzer and internally produce the same as the code above just with more complexity.

IMHO we should in general use `dyn` more where usable as it can make code more compact, faster to build and any performance drawback of (reliable branch predicted) vtable calls _tends_ to be armortized by e.g. more cache friendly code

web-api-db-ctrl/postgres/tenant/20230712141144_document.sql
```diff
@@ -0,0 +1,31 @@
...
+    ALTER COLUMN is_summarized DROP DEFAULT,
+    ALTER COLUMN is_summarized TYPE preprocessing_step
+        USING CASE
+            WHEN is_summarized THEN 'summarize'::preprocessing_step
+            ELSE 'none'::preprocessing_step
+        END,
+    ALTER COLUMN is_summarized SET DEFAULT 'none';
+
+ALTER TABLE document
+    RENAME COLUMN is_summarized TO preprocessing_step;
```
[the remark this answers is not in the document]
rustonaut: no they are different name spaces, you can e.g. name a column  pretty much anything as long as it's not a keyword, through if there is a bad overlap with e.g. aggregation functions depending on SQL syntax it can be confusion for a human

web-api/src/ingestion/routes.rs
```diff
@@ -202,20 +208,24 @@ impl UnvalidatedIngestedDocument {
...
         let snippet = DocumentSnippet::new(self.snippet, config.max_snippet_size)?;
-        let summary = self.summarize.then(|| {
-            summarize(
-                &Summarizer::Naive,
-                &Source::PlainText {
-                    text: snippet.as_str().to_owned(),
-                },
-                &Config::default(),
+
+        //FIXME this should be done post-validation
```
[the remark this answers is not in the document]
rustonaut: it's a leftover from where `summarize` was directly called here and `PreprocessingStep` didn't exist :sweat:

web-api/src/ingestion/routes.rs
```diff
@@ -433,6 +447,81 @@ async fn upsert_documents(
...
+) -> Result<Vec<DocumentEmbedding>, Error> {
+    let snippet = &document.snippet;
+    Ok(match document.preprocessing_step {
+        PreprocessingStep::None => embed_whole(embedder, snippet)?,
+        PreprocessingStep::Split => embed_with_splitting(embedder, snippet)?,
+        PreprocessingStep::Summarize => embed_with_summarizer(embedder, snippet)?,
+    })
+}
+
+fn embed_whole(embedder: &Embedder, snippet: &str) -> Result<Vec<DocumentEmbedding>, Error> {
```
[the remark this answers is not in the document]
rustonaut: IMHO It's not part of the `Embedder` it's a preprocessing step fully independent of the embedding calculation.

web-api/src/ingestion/routes.rs
```diff
@@ -433,6 +447,81 @@ async fn upsert_documents(
...
+}
+
+fn embed_with_splitting(
+    embedder: &Embedder,
+    snippet: &str,
+) -> Result<Vec<DocumentEmbedding>, Error> {
+    cutters::cut(snippet, cutters::Language::English)
+        .into_iter()
+        .map(|split| {
+            // FIXME use something more elegant and robust
```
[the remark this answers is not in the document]
rustonaut: for german support we can do so

through it's not just something, the approach `cutters` uses fundamentally doesn't scale to our needs long term

and we probably  also need a language detection library

web-api/src/personalization/routes.rs
```diff
@@ -649,12 +680,13 @@ async fn semantic_search(
...
         InputDocument::Ref(id) => {
-            let embedding = storage::Document::get_embedding(&storage, &id)
+            let mut embeddings = storage::Document::get_embeddings(&storage, &id)
                 .await?
                 .ok_or(DocumentNotFound)?;
             excluded.push(id);
             (
-                embedding,
+                // TODO[pmk/soon] how to handle by document search with split documents
+                embeddings.pop().unwrap().embedding,
```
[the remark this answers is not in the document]
rustonaut: > panicking if the embeddings are empty?

they can't be empty

But this code is also anyway fundamentally wrong so I didn't want to put any more work into it and as far as I can tell the simplest way forward would be to run one search per embedding in which case empty doesn't matter anymore (empty => empty merged response list). Only issue the cost when a document was e.g. split into 200 sections.

> maybe with a wrapper type DocumentEmbeddings 

yes

web-api/src/ingestion/routes.rs
```diff
@@ -433,6 +447,81 @@ async fn upsert_documents(
     }
 }
 
+fn calculate_embeddings(
```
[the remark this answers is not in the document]
rustonaut: I prefer not to do so as it IMHO is more readable due to clearly keeping the embedding calculation out of the already way to long upsert documents function.

web-api/src/storage/postgres.rs
rustonaut: I still prefer the other style, but I also normally do not use inlay hints as I find them too verbose. I changed it.

web-api/src/ingestion/routes.rs
```diff
@@ -717,4 +806,30 @@ mod tests {
...
+            (valid.end..valid.end, valid.len()..valid.len()),
+            (3..4, 1..2),
+        ] {
+            let got =
+                derive_range(&full_string[valid.clone()], &full_string[r2.clone()]).map_err(|_| ());
+            assert_eq!(
+                got.clone(),
+                Ok(expected.clone()),
+                "for {r2:?} in {valid:?} we get {got:?} but expect Ok({expected:?})"
+            );
```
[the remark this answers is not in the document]
rustonaut: I did it that way because I preferred the uniform error message, but it doesn't matter much either way.

web-api/src/storage/elastic.rs
```diff
@@ -228,148 +289,141 @@ impl Client {
         Ok(response.failed_documents("delete", true).into())
     }
 
-    pub(super) async fn retain_documents(
+    async fn document_property_update(
```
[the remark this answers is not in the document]
rustonaut: Used `document_update` and also documented it to clarify that the number of splits can not be changed.

rustonaut:
Another thing which became clear today and should be changed in this PR:

- instead of having one snippet and byte offsets for splits have multiple snippets

This probably should be done as part of this PR due to how it affects storage of `embeddings`, but we can put it into a follow up PR, too.

The original design was based on the idea that (if ES did support it) we would "just" have multiple embedding per document and keep the rest the same

now we don't quite have this and a more fitting descriptions would be to have multiple pairs of text+embedding. For this we  also probably want to  the storage to allow having multiple "text snippets" per document in PG and ES. We happen to already have this in ES (due to how we implemented splits) but not yet in PG.

So I would add:

1. `snippet TEXT` => `snippet TEXT[]`
2. `embedding jsonb[]` to be internal just the embedding i.e. `[...]` instead of `{ "embedding": [...], "range": [<start>,<end>] }` (or maybe just remove the `"range"`)? (we still need `jsonb[]` as workaround for sqlx not supporting `FLOAT[][]` for now)

implicit constraint that array length of `snippet` == that of `embedding` (we could make that a sql check, but they can be annoying to maintain)

any thoughts on this?

janpetschexain:
yes makes sense. regarding 2. i would just go with the embedding internally instead of an boject with an embedding field. also the implicit length constraint should be fine, as with other items we don't validate invariants if stuff comes from the database as we make sure that we only store valid stuff in the db by prior validation.

PR #1034, opened 2023-07-19
ET-4828 post personalized documents

janpetschexain:
**Reference**

- [ET-4828]

**Summary**

- deprecate `GET /users/{id}/personalized_documents` with params in favor of `POST` with body
- add missing deprecation tests for document candidates

[ET-4828]: https://xainag.atlassian.net/browse/ET-4828?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/web-api/openapi/front_office.yaml
+++ b/web-api/openapi/front_office.yaml
@@ -2,7 +2,7 @@ openapi: 3.1.0
 
 info:
   title: Front Office API
-  version: 2.3.0
+  version: 2.3.1
   description: |-
     # Front Office
     The front office is typically used within front-end apps, for example a website or a mobile application.
@@ -48,11 +48,14 @@ security:
 
 paths:
   /users/{user_id}/personalized_documents:
-    get:
+    parameters:
+      - $ref: './parameters/path/id.yml#/UserId'
+
+    post:
       tags:
         - front office
         - search
-      summary: Get personalized documents for the user
+      summary: Personalize documents for the user
       description: |-
         Get a list of documents personalized for the given `user_id`.
         Each document contains the id and the score, which is a value between 0 and 1 where a higher value means that the document matches the preferences of the user better.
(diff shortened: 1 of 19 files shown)


web-api/src/ingestion/routes.rs
```diff
@@ -68,6 +68,18 @@ pub(super) fn configure_service(config: &mut ServiceConfig) {
                 .route(web::get().to(get_document_candidates))
                 .route(web::put().to(set_document_candidates)),
         )
+        .service(
+            // this resource is deprecated and undocumented and will be removed in the future
+            web::resource("/documents/candidates")
```
janpetschexain, line 73: this had to move up as it was actually unreachable below the `/documents/{id}` routes. i also added tests for the regression.

web-api/openapi/front_office.yaml
```diff
@@ -48,20 +48,51 @@ security:
 
 paths:
   /users/{user_id}/personalized_documents:
-    get:
+    parameters:
+      - $ref: './parameters/path/id.yml#/UserId'
+
+    post:
```
rustonaut, line 54: One interesting observation:

We always "brake" the http standard, either it's a post request which semantically shouldn't be one or it's a get request with a body (which given the standard should have no semantic meaning (but must still be passed on)).

The question is if we shouldn't go with get + body (like e.g. ES).

In the past (many many years ago) some (rare) middle ware messed with that but by now this shouldn't matter and the standard had been updated to put it into a "not stylistically recommended but okay" situation AFIK. 

Main benefit would be that browser caching get requests might still work with that :thinking: .

web-api/tests/personalized_documents.rs
```diff
@@ -260,10 +319,41 @@ fn personalization_include_properties(include_properties: bool) {
...
+
+#[test]
+fn test_deprecated_personalization_include_properties() {
+    personalization_include_properties(true, true);
+}
+
+#[test]
+fn test_deprecated_personalization_exclude_properties() {
+    personalization_include_properties(false, true);
+}
```
rustonaut, line 338: Integration tests are slow to start up, so where viable we likely want to have one test which covers more part instead of many tests which cover one part.

Through only if viable and no inference is expected/likely/possible.

But what we do here is somewhat one the line of unit testing with integration tests.

Sadly this is one of the many things where rust build-in testing lack support (I mean it's also on of the things where theoretical preference of maximal isolation diverges from the reality  of having limited resources and time.

web-api/openapi/front_office.yaml
acrrd: The problem that I see is also with that the API from the browser seems not to allow adding body to GET.
https://xhr.spec.whatwg.org/#the-send()-method
https://fetch.spec.whatwg.org/#dom-request (point 35).

This endpoint is not dissimilar from `/semantic_search` and we already use POST there.

rustonaut (approved):
lgtm

web-api/tests/personalized_documents.rs
janpetschexain: i'll set the tests up in a more compact way, this should solve it.

PR #1037, opened 2023-07-24
ET-4846 validate max number candidates dev option against requested count

janpetschexain:
**Reference**

- [ET-4846]

**Summary**

- validate the `max_num_candidates` dev option against the requested `count`
- move other dev option validations from the route block into the new validation block


web-api/src/personalization/routes.rs
```diff
@@ -429,8 +436,26 @@ struct DevOption {
 }
 
 impl DevOption {
-    fn is_some(&self) -> bool {
-        self.hybrid.is_some() || self.max_number_candidates.is_some()
+    fn validate(&self, enable_dev: bool, count: usize) -> Result<(), Error> {
+        if !enable_dev && (self.hybrid.is_some() || self.max_number_candidates.is_some()) {
+            // notify the caller instead of silently discarding the dev option
+            return Err(ForbiddenDevOption::DevDisabled.into());
+        }
```
rustonaut, line 443: :question:/:pushpin: This is quite prone to bugs when adding new dev options . Maybe we find a better way. But either way outside of this PR.

web-api/src/personalization/routes.rs
```diff
@@ -429,8 +436,26 @@ struct DevOption {
...
+        if let Some(DevHybrid::EsRrf { .. }) = self.hybrid {
+            // not available because of the es license, return a 403 instead of forwarding a 500
+            return Err(ForbiddenDevOption::EsRrfUnlisenced.into());
+        }
```
rustonaut, line 447: :pushpin:  should we then just remove the code?

web-api/src/personalization/routes.rs
```diff
@@ -497,26 +526,34 @@ impl UnvalidatedSemanticSearchRequest {
...
-        let count = validate_count(
+        let tenants_config: &tenants::Config = config.as_ref();
+
+        let document = document.validate()?;
+        let count = count.unwrap_or(semantic_search_config.default_number_documents);
+        dev.validate(tenants_config.enable_dev, count)?;
+        let num_candidates = dev
+            .max_number_candidates
+            .unwrap_or(semantic_search_config.max_number_candidates);
+        validate_count(
```
rustonaut, line 538: :wrench:  We now have the  `count < max_number_candidates` validation twice.

Once as part of the dev options and once outside of the dev options, this is prone to introducing confusions and bugs when changing things.

Given that `dev.validate`  validates if the dev options are valid it should not check `count`  I think. 

Through it could error if `dev.max_number_candidates` has some absurd value, like `0` or some very very high number (if we want to do the later depends on weather we will give dev options to anyone but Dazzle or if it was just a super rare exception IMHO).

web-api/src/personalization/routes.rs
```diff
@@ -497,26 +526,34 @@ impl UnvalidatedSemanticSearchRequest {
...
         let is_deprecated = published_after.is_some();
 
         Ok(SemanticSearchRequest {
             document,
             count,
+            num_candidates,
             personalize,
             enable_hybrid_search,
-            dev,
+            dev_hybrid_search,
```
rustonaut, line 556: :wrench:  I would not strip out the dev options here as it means doing way more changes when adding any other dev options. Especially given what dev options are for this seems like a bad idea.

web-api/src/personalization/routes.rs
```diff
@@ -591,19 +629,14 @@ struct Personalize {
...
-    let count = count.unwrap_or(default);
-
-    if (1..=max.min(candidates)).contains(&count) {
-        Ok(count)
-    } else {
-        Err(InvalidDocumentCount { count, min: 1, max })
+fn validate_count(count: usize, max: usize, candidates: usize) -> Result<(), InvalidDocumentCount> {
+    let min = 1;
+    let max = max.min(candidates);
+    if !(min..=max).contains(&count) {
```
rustonaut, line 635: :art: 

```suggestion
    let max = max.min(candidates);
    if !(1..=max).contains(&count) {
```

web-api/src/personalization/routes.rs
janpetschexain: there is probably no nice solution. we should remove dev options once they are not needed anymore (for example if we move a feature to production) to keep the complexity of the dev options low.

web-api/src/personalization/routes.rs
janpetschexain: the issue i see is that without splitting the dev option into its components, we carry around parts of the dev options into the route block, for example `dev.max_number_candidates`, which are not supposed to be used there since it was already used in the request validation where it got updated.

so i guess it's an argument of boilerplate vs logic bugs, where i tend to go with the former. wdyt?

rustonaut: in the end it's at the messy boarder between input parsing, validation and pre-processing so I guess we can keep it the way you changed it to

rustonaut (approved):
lgmt

web-api/src/personalization/routes.rs
acrrd: yes

janpetschexain: i added [ET-4859] to the backlog

PR #1040, opened 2023-07-25
ET-4799 length limit and error reporting in ingestion

rustonaut:
- report per-document errors for ingestion
- update errors to be more informative/less confusing
  - be more explicit about what failed (e.g. syntax vs. length limits)
  - be more explicit about where it failed (e.g. which property)
  - make some errors more clear (e.g. `InvalidDocumentProperties`)
  - be clear about what is expected (wrt. the failure)
  - same uniform error for all "string" validation types
  - do nolonger contain the malformed value (both returned and logged)
    - because the malformed value could e.g. be a 2MiB DocumentId with null bytes and ascii controll codes, nothing we ever want to log
    - through for this point I need some feedback, maybe we should still include it in some normalized and cut/shortened form :thinking:  

**References:**

- issue: [ET-4799]

[ET-4799]: https://xainag.atlassian.net/browse/ET-4799?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/web-api/src/error/common.rs
+++ b/web-api/src/error/common.rs
@@ -72,12 +72,11 @@ pub(crate) struct InvalidDocumentPropertyId {
 
 impl_application_error!(InvalidDocumentPropertyId => BAD_REQUEST, INFO);
 
-/// Malformed property {document}/{property}, {invalid_reason}: {invalid_value}
+/// Malformed property {property}, {invalid_reason}: {invalid_value}
 #[derive(Debug, Error, Display, Serialize)]
 // there are some false positives with clippy and displaydoc
 #[allow(clippy::doc_markdown)]
 pub(crate) struct InvalidDocumentProperty {
-    pub(crate) document: DocumentId,
     pub(crate) property: DocumentPropertyId,
     pub(crate) invalid_value: Value,
     pub(crate) invalid_reason: InvalidDocumentPropertyReason,
@@ -125,6 +124,7 @@ pub(crate) struct InvalidDocumentTags {
 impl_application_error!(InvalidDocumentTags => BAD_REQUEST, INFO);
 
 #[derive(Debug, Error, Display, Serialize)]
+#[serde(rename_all = "snake_case")]
 pub(crate) enum InvalidDocumentSnippet {
     /// Malformed document snippet.
     Value { value: String },
--- a/web-api/src/error/common.rs
+++ b/web-api/src/error/common.rs
@@ -52,7 +52,7 @@ impl_application_error!(DocumentPropertyNotFound => BAD_REQUEST, INFO);
 #[cfg_attr(test, derive(PartialEq))]
 #[serde(rename_all = "snake_case")]
 pub(crate) enum InvalidString {
-    /// Invalid size. Got {got}, expected {min}..={max}.
+    /// Invalid byte size. Got {got}, expected {min}..={max}.
     Size { got: usize, min: usize, max: usize },
     /// Invalid syntax, expected: {expected}
     Syntax { expected: &'static str },
@@ -81,7 +81,7 @@ pub(crate) struct InvalidDocumentPropertyId(pub(crate) InvalidString);
 
 impl_application_error!(InvalidDocumentPropertyId => BAD_REQUEST, INFO);
 
-/// Malformed property {property}, {invalid_reason}: {invalid_value}
+/// Malformed document property {property}, {invalid_reason}: {invalid_value}
 #[derive(Debug, Error, Display, Serialize)]
 // there are some false positives with clippy and displaydoc
 #[allow(clippy::doc_markdown)]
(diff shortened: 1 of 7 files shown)


web-api/src/models.rs
```diff
@@ -77,13 +99,15 @@ macro_rules! string_wrapper {
                 type Error = $error;
 
                 fn try_from(mut value: String) -> Result<Self, Self::Error> {
+                    use ::std::ops::RangeInclusive;
+
```
janpetschexain, line 103: :art: since the macro is not exported, we can just use the module import declared above and don't have to import it again
```suggestion
```

web-api/src/error/common.rs
```diff
@@ -48,36 +48,44 @@ pub(crate) struct DocumentPropertyNotFound;
...
 
-/// Malformed user id.
-#[derive(Debug, Error, Display, Serialize)]
-pub(crate) struct InvalidUserId {
-    pub(crate) value: String,
+#[derive(Debug, Display, Serialize)]
+#[cfg_attr(test, derive(PartialEq))]
+#[serde(rename_all = "snake_case")]
+pub(crate) enum InvalidString {
+    /// Invalid size. Got {got}, expected {min}..={max}.
```
janpetschexain, line 55: :art:
```suggestion
    /// Invalid byte size. Got {got}, expected {min}..={max}.
```

web-api/src/models.rs
```diff
@@ -182,26 +182,31 @@ pub(crate) struct DocumentProperty(Value);
...
         property_id: &DocumentPropertyId,
         mut value: Value,
     ) -> Result<Self, InvalidDocumentProperty> {
+        let check_string_property_value = |s: &str| {
+            if s.len() <= 2_048 && GENERIC_STRING_SYNTAX.is_match(s) {
+                Ok(())
+            } else {
+                Err(())
+            }
+        };
```
janpetschexain, line 194: 🔧 we can reuse the `validate_string()` function
```suggestion
        let validate_string =
            |value: &str| validate_string(value, 0..=2_048, &GENERIC_STRING_SYNTAX);
```

web-api/src/models.rs
```diff
@@ -182,26 +182,31 @@ pub(crate) struct DocumentProperty(Value);
...
+
         match &mut value {
             Value::Bool(_) | Value::Number(_) | Value::Null => {}
             Value::String(string) => {
-                if !is_valid_string_empty_ok(string, 2_048) {
+                let Ok(()) = check_string_property_value(string) else {
```
janpetschexain, line 199: :wrench: a `if is_err()` seems to transport the intention a bit better than `let Ok(())`, with the reused `validate_string()` we can write it as
```suggestion
                if validate_string(string).is_err() {
```

web-api/src/models.rs
```diff
@@ -219,19 +223,17 @@ impl DocumentProperty {
                         });
                     };
                     trim(string);
-                    if !is_valid_string(string, 2_048) {
+                    let Ok(()) = check_string_property_value(string) else {
```
janpetschexain, line 226: :wrench: a `if is_err()` seems to transport the intention a bit better than `let Ok(())`, with the reused `validate_string()` we can write it as
```suggestion
                if validate_string(string).is_err() {
```

web-api/src/models.rs
```diff
@@ -182,26 +182,31 @@ pub(crate) struct DocumentProperty(Value);
...
+                let Ok(()) = check_string_property_value(string) else {
                     return Err(InvalidDocumentProperty {
-                        document: document_id.clone(),
                         property: property_id.clone(),
-                        invalid_value: value.clone(),
+                        invalid_value: value,
                         invalid_reason: InvalidDocumentPropertyReason::InvalidString,
                     });
-                }
+                };
```
janpetschexain, line 205: :art: semicolon can be removed if above's suggestion is used
```suggestion
                }
```

web-api/src/models.rs
```diff
@@ -219,19 +223,17 @@ impl DocumentProperty {
...
-                    if !is_valid_string(string, 2_048) {
+                    let Ok(()) = check_string_property_value(string) else {
                         return Err(InvalidDocumentProperty {
-                            document: document_id.clone(),
                             property: property_id.clone(),
                             invalid_value: value.clone(),
                             invalid_reason: InvalidDocumentPropertyReason::InvalidString,
                         });
-                    }
+                    };
```
janpetschexain, line 232: :art: semicolon can be removed if above's suggestion is used
```suggestion
                }
```

janpetschexain (commented):
looks good, some suggestions

janpetschexain:
> do nolonger contain the malformed value (both returned and logged)
>    because the malformed value could e.g. be a 2MiB DocumentId with null bytes and ascii controll codes, nothing we ever want to log
>    through for this point I need some feedback, maybe we should still include it in some normalized and cut/shortened form 🤔

i'd say it's better to not include it. we should help users with detecting input errors, but only to an extend where it doesn't hurt us.

janpetschexain (approved):
lgtm

PR #1041, opened 2023-07-25
Et 4299 try onnx runtime

krajewskiML:
Implements script responsible for running onnxruntime and tract comparison in the examples folder

### DIFF
--- a/bert/Cargo.toml
+++ b/bert/Cargo.toml
@@ -26,7 +26,7 @@ xayn-test-utils = { path = "../test-utils" }
 criterion = { workspace = true }
 csv = { workspace = true }
 indicatif = "0.17.5"
-onnxruntime = {version = ">=0.0.2", features = ["model-fetching"] }
+# onnxruntime = {version = ">=0.0.2", features = ["model-fetching"] }
 ort = {version = "1.15.2", features = ["download-binaries"] }
 rand = "0.8.5"
 
--- a/bert/Cargo.toml
+++ b/bert/Cargo.toml
@@ -14,6 +14,7 @@ lindera-core = "0.25.0"
 lindera-dictionary = "0.25.0"
 lindera-tokenizer = "0.25.0"
 ndarray = { workspace = true, features = ["serde"] }
+onnxruntime = {version = ">=0.0.2", features = ["model-fetching"] }
 serde = { workspace = true }
 serde_json = { workspace = true }
 sqlx = { workspace = true, optional = true }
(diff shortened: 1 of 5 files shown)


bert/Cargo.toml
```diff
@@ -14,6 +14,7 @@ lindera-core = "0.25.0"
 lindera-dictionary = "0.25.0"
 lindera-tokenizer = "0.25.0"
 ndarray = { workspace = true, features = ["serde"] }
+onnxruntime = {version = ">=0.0.2", features = ["model-fetching"] }
```
rustonaut, line 17: This can be a `dev-dependency` examples, tests, benchmarks have access to dev dependencies.

bert/examples/onnx_runtime.rs
```diff
@@ -0,0 +1,169 @@
...
+
+    let start = std::time::Instant::now();
+    for text in texts.into_vec() {
+        let tokens = tokenizer.encode(text, false).unwrap();
+        // reshape input tensor
+        let token_ids = token_reshaper(tokens.get_ids());
+        let attention_mask = token_reshaper(tokens.get_attention_mask());
+        let type_ids = token_reshaper(tokens.get_type_ids());
+        let in_tensor = vec![token_ids, attention_mask, type_ids];
+        let outputs: Vec<OrtOwnedTensor<f32, _>> = session.run(in_tensor)?;
```
rustonaut, line 78: depending internal details of  `OrtOwnedTensor<f32, _>` and `session.run(in_tensor)?` the compiler could theoretically be able to deduce that `output` is not used `session` (besides costing time) has no side effects and then end up eliminating the while loop :upside_down_face: 

practically this is impossible to happen as `session` will call C-FFI so the compiler doesn't know anymore that it doesn't have side effects etc.

still because of this you want to to pass unused output of benchmarks through  `std::hint::black_box` (similar const input values, but we don't have them here)

rustonaut:
Wrt. the issues with running it and `so`/`dll` loading:

The rust crate `onnxruntime` only provides rust bindings to the microsoft onnxruntime, it (for good reasons) does not bundle the runtime, nor does the runtime need to even be available to build it.

What that means is that the runtime must be preinstalled in the computer where you want to run it in a compatible version.

E.g. when build on my system `libonnxruntime.so.1.7.0` i.e. ABI version 1.7.0 of the runtime.

rustonaut:
I think you can find a prebuild blob of the microsoft onnxruntime here: https://github.com/microsoft/onnxruntime/releases/tag/v1.15.1

rustonaut:
was integrated in different PR AFIK

PR #1043, opened 2023-07-26
Et 4760 engine with sagemaker

joecks:


web-api/src/embedding.rs
```diff
@@ -12,68 +12,103 @@
...
 }
 
+#[derive(Debug, Deserialize, Serialize)]
+struct Response {
+    embeddings: Vec<NormalizedEmbedding>,
+}
+
 pub(crate) struct Embedder {
-    embedder: AvgEmbedder,
+    embedding_dim: usize,
```
Robert-Steiner, line 49: not sure if we can remove this

janpetschexain, line 49: :art: we usually call these kind of params with a `_size` suffix. `_dim` would be ok as well, we should just be consistent across the code base.

the main issue with `_dim` is that it is ambiguous, it could stand for dimensionality (which is what we actually refer to, like `384`), but it could also stand for dimension (which we don't refer to as it is the index of the model output array in which the embeddings lie, usually `2` for our models).

janpetschexain (commented):
> that it might make sense to add another runtime type and depending on the type load the embedder

if we can't run the sagemaker setup locally or in the ci, then this would be required.

acrrd:
@Robert-Steiner can we close this? If yes please close it and delete the branch.

PR #1044, opened 2023-07-27
ET-4775 check filter against property schema

janpetschexain:
**Reference**

- [ET-4775]

**Summary**

- impl filter validation wrt the indexed properties schema
- improve filter serde tests & add filter validation tests

[ET-4775]: https://xainag.atlassian.net/browse/ET-4775?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/web-api/src/personalization/filter.rs
+++ b/web-api/src/personalization/filter.rs
@@ -583,7 +583,7 @@ mod tests {
         ] {
             for number in [json!(42_u64), json!(-42_i64), json!(42_f64)] {
                 assert_eq!(
-                    serde_json::from_str::<CompareWith>(&format!(r#"{{ "{op}": {number} }}"#))
+                    serde_json::from_str::<CompareWith>(&json!({ op: number }).to_string())
                         .unwrap(),
                     CompareWith {
                         operation,
@@ -601,7 +601,7 @@ mod tests {
     #[test]
     fn test_compare_with_date() {
         assert_eq!(
-            serde_json::from_str::<CompareWith>(&format!(r#"{{ "$eq": "{DATE}" }}"#)).unwrap(),
+            serde_json::from_str::<CompareWith>(&json!({ "$eq": DATE }).to_string()).unwrap(),
             CompareWith {
                 operation: CompareOp::Eq,
                 value: json!(DATE).try_into().unwrap(),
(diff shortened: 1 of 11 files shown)


web-api/src/personalization/filter.rs
```diff
@@ -444,227 +519,114 @@ mod tests {
...
             ("$lt", CompareOp::Lt),
             ("$lte", CompareOp::Lte),
         ] {
             for number in [json!(42_u64), json!(-42_i64), json!(42_f64)] {
                 assert_eq!(
-                    serde_json::from_str::<CompareWith>(
-                        &json!({ operation.0: number }).to_string()
-                    )
-                    .unwrap(),
+                    serde_json::from_str::<CompareWith>(&format!(r#"{{ "{op}": {number} }}"#))
```
rustonaut, line 586: :wrench:  Why not just use `json`? 

(here and below where you sometimes explicitly changed away from it).

rustonaut (commented):
looks mostly good to me

the only thing is we check `(operator, value)` and `(value, property_type)` disjoint with implicit assumptions about the other. 

This is fine for now as it works and changing it is a major rework.

But in the end which `(value, property_type)` combinations are valid is dependent on `operator`.

web-api/src/personalization/filter.rs
janpetschexain: i'm a bit undecided about this, in the end the outcome is of course the same if we use `format!()` or `json!().to_string()`. currently i'm leaning more towards the former as this seems a bit closer to what we would get in a request written by a user. but i also the that the additional escaping in the formatting is not nice. should we go back to the latter, wdyt?

rustonaut: I think  `json!` is more readable and guarantees that the json itself is always well formed and as such should be preferred in this context as we are not testing the json parsing itself.

Any performance differences are in this use-case irrelevant AFIK.

janpetschexain:
> looks mostly good to me
> 
> the only thing is we check `(operator, value)` and `(value, property_type)` disjoint with implicit assumptions about the other.
> 
> This is fine for now as it works and changing it is a major rework.
> 
> But in the end which `(value, property_type)` combinations are valid is dependent on `operator`.

yes, in the end it is the same line of thought like that we do additional validation on a property with `schema.validaty_property()`, just applied to filters this time. the `schema.validate_filter()` is of course not enough/complete on its own.

rustonaut (approved):
lgtm

PR #1046, opened 2023-07-27
ET-4883 rescale es knn score

janpetschexain:
**Reference**

- [ET-4883]
- followed by #1047

**Summary**

- rescale es knn scores back to similarity scores
- use `ScoreMap` type alias consistently
- normalize es knn scores in cases where we use reranking with interest & tag scores, until we change how we combine the three scores we need to normalize them again
- update docs for scores

[ET-4883]: https://xainag.atlassian.net/browse/ET-4883?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/web-api/openapi/front_office.yaml
+++ b/web-api/openapi/front_office.yaml
@@ -58,7 +58,7 @@ paths:
       summary: Personalize documents for the user
       description: |-
         Get a list of documents personalized for the given `user_id`.
-        Each document contains the id and the score, where a higher value means that the document matches the preferences of the user better.
+        Each document contains the id and the score, where a higher value means that the document matches the preferences of the user better. Scores can be compared only with other scores that belong to the same request; comparing scores of documents that have been obtained through different requests can lead to unexpected results.
         The documents also contain their properties if this is requested and the properties are not empty.
         Documents that have been interacted with by the user are filtered out from the result.
         Note that you can request personalized documents for a specific `user_id`, only after that same `user_id` has made enough interactions via our system.
--- a/web-api/openapi/front_office.yaml
+++ b/web-api/openapi/front_office.yaml
@@ -178,7 +178,7 @@ paths:
         When a `query` is provided, the system will return documents that are similar to the query.
         If `enable_hybrid_search` is passed, then the system will also perform keyword matching between the query and the documents.
         It is possible to personalize the result by passing a user id or history. In this case, the system will consider the user's interests to rank the documents.
-        Each document contains the `id` and the `score`, where a higher value means that the document is more similar to the input.
+        Each document contains the `id` and the `score`, where a higher value means that the document is more similar to the input. Scores can be compared only with other scores that belong to the same request; comparing scores of documents that have been obtained through different requests can lead to unexpected results.
         The documents also contain their `properties` if this is requested and the properties are not empty.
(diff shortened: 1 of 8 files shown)


web-api/openapi/front_office.yaml
```diff
@@ -58,7 +58,7 @@ paths:
       summary: Personalize documents for the user
       description: |-
         Get a list of documents personalized for the given `user_id`.
-        Each document contains the id and the score, which is a value between 0 and 1 where a higher value means that the document matches the preferences of the user better.
+        Each document contains the id and the score, where a higher value means that the document matches the preferences of the user better.
```
acrrd, line 61:
```suggestion
        Each document contains the id and the score, where a higher value means that the document matches the preferences of the user better. Scores can be compared only with other scores that belong to the same request; comparing scores of documents that have been obtained through different requests can lead to unexpected results.
```

web-api/openapi/front_office.yaml
```diff
@@ -178,7 +178,7 @@ paths:
         When a `query` is provided, the system will return documents that are similar to the query.
         If `enable_hybrid_search` is passed, then the system will also perform keyword matching between the query and the documents.
         It is possible to personalize the result by passing a user id or history. In this case, the system will consider the user's interests to rank the documents.
-        Each document contains the `id` and the `score`, which is a value between 0 and 1 where a higher value means that the document is more similar to the input.
+        Each document contains the `id` and the `score`, where a higher value means that the document is more similar to the input.
```
acrrd, line 181:
```suggestion
        Each document contains the `id` and the `score`, where a higher value means that the document is more similar to the input. Scores can be compared only with other scores that belong to the same request; comparing scores of documents that have been obtained through different requests can lead to unexpected results.
```

acrrd (commented):
@rustonaut I left my comment as suggestions so we can merge them if you are ok with them.

web-api/src/personalization/routes.rs
```diff
@@ -684,6 +685,9 @@ async fn semantic_search(
     .await?;
 
     if let Some(personalize) = personalize {
+        if matches!(strategy, SearchStrategy::Knn) {
+            normalize_knn_scores(&mut documents);
+        }
```
rustonaut, line 690: :wrench:  The documents already have been normalized  with `rescale_knn_scores` in `knn_search` called by `get_by_embedding`.

rustonaut, line 690: I'm a bit confused.

We ourself move (-1;1) ranges to (0;1) in most situations.

But now we do undo that shift but then in some cases redo it.

Why not always take that scores from ES as we get them, that should not be a bug given the guarantees about values we can give.

Like for personalization we anyway want `(0;1)` general in pretty much all places in the code we assumed we get `(0;1)` ranges from ES instead of "correct (-1;1) ranges".

web-api/openapi/front_office.yaml
rustonaut: looks good

rustonaut (approved):
lgtm, please also commit your suggestions before merging

PR #1048, opened 2023-08-01
ET-4918 dev.include_snippet [1/n]

rustonaut:
Needed for testing, previously we had dev.include_splits.

This implements most of [ET-4678] but is not quite the same, mainly it's a dev options so it doesn't yet need to lock itself into a specific format. For [ET-4678] we also need to consider the difference between raw document splits and snippets and maybe we want to compact `include_FIELD` into something like `include: Includes(HashSet<OptionalField>)` or similar.

**References:**

- issue: [ET-4756]
- story: [ET-4755]
- related: [ET-4678]

[ET-4678]: https://xainag.atlassian.net/browse/ET-4678?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ET-4678]: https://xainag.atlassian.net/browse/ET-4678?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ET-4678]: https://xainag.atlassian.net/browse/ET-4678?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/web-api/src/personalization/routes.rs
+++ b/web-api/src/personalization/routes.rs
@@ -12,8 +12,6 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-use std::collections::HashMap;
-
 use actix_web::{
     http::StatusCode,
     web::{self, Data, Json, Path, Query, ServiceConfig},
@@ -49,7 +47,14 @@ use crate::{
         common::{BadRequest, DocumentNotFound, ForbiddenDevOption, InvalidDocumentCount},
         warning::Warning,
     },
-    models::{DocumentId, DocumentProperties, DocumentQuery, PersonalizedDocument, UserId},
+    models::{
+        DocumentId,
+        DocumentProperties,
+        DocumentQuery,
+        DocumentSnippet,
+        PersonalizedDocument,
+        UserId,
+    },
     storage::{self, KnnSearchParams, MergeFn, NormalizationFn, SearchStrategy},
     tenants,
     utils::deprecate,
@@ -252,6 +257,7 @@ async fn personalized_documents(
         },
         Utc::now(),
         include_properties,
+        false,
     )
     .await?
     {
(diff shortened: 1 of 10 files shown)


web-api/src/personalization/routes.rs
```diff
@@ -274,8 +280,10 @@ async fn personalized_documents(
 struct PersonalizedDocumentData {
     id: DocumentId,
     score: f32,
-    #[serde(skip_serializing_if = "HashMap::is_empty")]
-    properties: DocumentProperties,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    properties: Option<DocumentProperties>,
```
janpetschexain, line 284: 🔧 this creates an issue as it will not skip serialization for empty properties `Some(json!({}))`, but only if `include_properties = false`. this also contradicts what we say in the frontoffice docs: https://github.com/xaynetwork/xayn_discovery_engine/blob/main/web-api/openapi/front_office.yaml#L182

so we could change the docs to account for the new behavior, but i think it would be better to change the serialization check to `option is_none || inner is_empty`.

web-api/tests/semantic_search.rs
```diff
@@ -351,3 +375,46 @@ fn test_semantic_search_with_dev_option_candidates() {
         },
     );
 }
+
+#[test]
+fn test_semantic_search_with_snippets() {
```
janpetschexain, line 380: :art: we should indicate in the name that this is a dev option like we do above
```suggestion
fn test_semantic_search_with_dev_option_snippet() {
```

PR #1049, opened 2023-08-01
ET-4756-2 es refactorings [2/n]

rustonaut:
Various refactoring related to ES done for other changes we do later one.

- ES bulk: 
  - take `I` instead of `&'a I` as `I` can still be a `&'a DocumentId`
  - support `Update`
- `ScoreMap<Id>` type alias
  - using this everywhere makes it easier to change the underlying type and more consistently conveys intention 
  - in difference to existing draft PR doesn't jet change the underlying type
- added ES `fn document_update(id, script_body)`
  - used by many endpoints which all do the same, i.e. update the document `_source` through a mini script
  - this logic will become more complicated in followup PRs
- fix a subtle bug of `query_with_json` which happened to not have caused any issue yet
  
**References:**

- issue: [ET-4756]
- story: [ET-4755]


web-api/src/storage/elastic.rs
```diff
@@ -259,64 +245,62 @@ impl Client {
...
-            },
+            }),
+        )
+        .await
+    }
+
+    async fn document_update(
+        &self,
+        document_id: &DocumentId,
+        update_script: Value,
```
janpetschexain, line 298: :art: we could force it to take a `JsonObject` instead of a `Value` to catch some potential misuse at compile time

```suggestion
        update_script: JsonObject,
```

web-api/src/storage/elastic.rs
```diff
@@ -623,8 +603,10 @@ impl NormalizationFn {
     }
 }
 
+type DynMergeFn = dyn Fn(ScoreMap<DocumentId>, ScoreMap<DocumentId>) -> ScoreMap<DocumentId>;
+
 impl MergeFn {
-    fn to_fn(self) -> Box<dyn Fn(ScoreMap, ScoreMap) -> ScoreMap> {
+    fn to_fn(self) -> Box<DynMergeFn> {
```
janpetschexain, line 609: ❓ since the type alias is not reused elsewhere we could just inline it here. or is the linter complaining about type complexity?

```suggestion
impl MergeFn {
    fn to_fn(self) -> Box<dyn Fn(ScoreMap<DocumentId>, ScoreMap<DocumentId>) -> ScoreMap<DocumentId>> {
```

janpetschexain (approved):
lgtm

web-api/src/storage/elastic.rs
rustonaut: > or is the linter complaining about type complexity?

yes

PR #1056, opened 2023-08-07
ET-4923 multi split support in postgres [6/n]

rustonaut:
Add support for multi-snippet documents in postgres (and only there).

- support for interacting with multi-snippet document is still only semi-supported as there are a bunch of open questions currently it will convert all "interacted with document id" occurrences with "interacted with first snippet of document" but also for search exclusions "exclude all documents with any interaction" 
- memory/mind db are still assuming single snippet documents, changing this would require rewriting huge portions of the memory impl.

**References:**

- story: [ET-4755]
- issue: [ET-4756]
- based on #1055 

### DIFF
--- a/web-api-db-ctrl/postgres/tenant/20230713173212_document.sql
+++ b/web-api-db-ctrl/postgres/tenant/20230713173212_document.sql
@@ -25,5 +25,7 @@ CREATE TABLE snippet (
 INSERT INTO snippet(document_id, sub_id, snippet, embedding)
     SELECT document_id, 0, snippet, embedding FROM document;
 
-ALTER TABLE document RENAME COLUMN snippet TO original;
+ALTER TABLE document ALTER COLUMN snippet TYPE BYTEA USING sha256(snippet::bytea);
+ALTER TABLE document RENAME COLUMN snippet TO original_sha256;
+
 ALTER TABLE document DROP COLUMN embedding;
--- a/web-api-db-ctrl/postgres/tenant/20230713173212_document.sql
+++ b/web-api-db-ctrl/postgres/tenant/20230713173212_document.sql
@@ -1,0 +1,29 @@
+-- Copyright 2023 Xayn AG
+--
+-- This program is free software: you can redistribute it and/or modify
+-- it under the terms of the GNU Affero General Public License as
+-- published by the Free Software Foundation, version 3.
+--
+-- This program is distributed in the hope that it will be useful,
+-- but WITHOUT ANY WARRANTY; without even the implied warranty of
+-- MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+-- GNU Affero General Public License for more details.
+--
+-- You should have received a copy of the GNU Affero General Public License
+-- along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+CREATE TABLE snippet (
+    document_id TEXT NOT NULL
+        REFERENCES document(document_id) ON DELETE CASCADE,
+    sub_id INTEGER NOT NULL,
+    snippet TEXT NOT NULL,
+    embedding FLOAT4[] NOT NULL,
+
+    PRIMARY KEY (document_id, sub_id)
+);
+
+INSERT INTO snippet(document_id, sub_id, snippet, embedding)
+    SELECT document_id, 0, snippet, embedding FROM document;
+
+ALTER TABLE document RENAME COLUMN snippet TO raw_document;
+ALTER TABLE document DROP COLUMN embedding;
--- a/web-api-db-ctrl/postgres/tenant/20230713173212_document.sql
+++ b/web-api-db-ctrl/postgres/tenant/20230713173212_document.sql
@@ -25,5 +25,5 @@ CREATE TABLE snippet (
 INSERT INTO snippet(document_id, sub_id, snippet, embedding)
     SELECT document_id, 0, snippet, embedding FROM document;
 
-ALTER TABLE document RENAME COLUMN snippet TO raw_document;
+ALTER TABLE document RENAME COLUMN snippet TO original;
 ALTER TABLE document DROP COLUMN embedding;
--- a/web-api-db-ctrl/postgres/tenant/20230807103000_interaction.sql
+++ b/web-api-db-ctrl/postgres/tenant/20230807103000_interaction.sql
@@ -19,4 +19,4 @@ ALTER TABLE interaction
     ADD COLUMN snippet_idx INTEGER NOT NULL DEFAULT 0,
     DROP CONSTRAINT interaction_pkey,
     ADD PRIMARY KEY (document_id, snippet_idx, user_id, time_stamp),
-    ADD FOREIGN KEY (document_id, snippet_idx) REFERENCES snippet(document_id, snippet_idx);
+    ADD FOREIGN KEY (document_id, snippet_idx) REFERENCES snippet(document_id, snippet_idx) ON DELETE CASCADE;
--- a/web-api-db-ctrl/postgres/tenant/20230807103000_interaction.sql
+++ b/web-api-db-ctrl/postgres/tenant/20230807103000_interaction.sql
@@ -16,7 +16,7 @@ ALTER TABLE interaction
     RENAME COLUMN doc_id TO document_id;
 
 ALTER TABLE interaction
-    ADD COLUMN snippet_idx INTEGER NOT NULL DEFAULT 0,
+    ADD COLUMN sub_id INTEGER NOT NULL DEFAULT 0,
     DROP CONSTRAINT interaction_pkey,
-    ADD PRIMARY KEY (document_id, snippet_idx, user_id, time_stamp),
-    ADD FOREIGN KEY (document_id, snippet_idx) REFERENCES snippet(document_id, snippet_idx) ON DELETE CASCADE;
+    ADD PRIMARY KEY (document_id, sub_id, user_id, time_stamp),
+    ADD FOREIGN KEY (document_id, sub_id) REFERENCES snippet(document_id, sub_id) ON DELETE CASCADE;
--- a/web-api-db-ctrl/postgres/tenant/20230807103000_interaction.sql
+++ b/web-api-db-ctrl/postgres/tenant/20230807103000_interaction.sql
@@ -1,0 +1,22 @@
+-- Copyright 2023 Xayn AG
+--
+-- This program is free software: you can redistribute it and/or modify
+-- it under the terms of the GNU Affero General Public License as
+-- published by the Free Software Foundation, version 3.
+--
+-- This program is distributed in the hope that it will be useful,
+-- but WITHOUT ANY WARRANTY; without even the implied warranty of
+-- MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+-- GNU Affero General Public License for more details.
+--
+-- You should have received a copy of the GNU Affero General Public License
+-- along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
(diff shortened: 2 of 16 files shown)


web-api-db-ctrl/postgres/tenant/20230713173212_document.sql
```diff
@@ -0,0 +1,29 @@
...
+
+CREATE TABLE snippet (
+    document_id TEXT NOT NULL
+        REFERENCES document(document_id) ON DELETE CASCADE,
+    snippet_idx INTEGER NOT NULL,
+    snippet TEXT NOT NULL,
+    embedding FLOAT4[] NOT NULL,
+
+    PRIMARY KEY (document_id, snippet_idx)
+);
```
rustonaut, line 23: I'm quite split between having a `snippets TEXT[]` + `embedding jsonb[]` fields in document instead of another table as in general we always add/replace/delete all snippets when adding/replacing/deleting a document.

I went for this here as it allows us to only fetch the snippets we need instead of always returning all snippets, which can be potentially very many with embedding being potentially >1KiB per embedding.

web-api-db-ctrl/postgres/tenant/20230713173212_document.sql
```diff
@@ -0,0 +1,29 @@
...
+
+INSERT INTO snippet(document_id, snippet_idx, snippet, embedding)
+    SELECT document_id, 0, snippet, embedding FROM document;
+
+ALTER TABLE document RENAME COLUMN snippet TO raw_document;
```
rustonaut, line 28: We have to properly decide how the handle "raw documents" (which long term could be e.g. PDF), especially wrt. upsertion, applied size limits and billing.

web-api-db-ctrl/postgres/tenant/20230807103000_interaction.sql
```diff
@@ -0,0 +1,22 @@
...
+-- along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+ALTER TABLE interaction
+    RENAME COLUMN doc_id TO document_id;
+
+ALTER TABLE interaction
+    ADD COLUMN snippet_idx INTEGER NOT NULL DEFAULT 0,
+    DROP CONSTRAINT interaction_pkey,
+    ADD PRIMARY KEY (document_id, snippet_idx, user_id, time_stamp),
+    ADD FOREIGN KEY (document_id, snippet_idx) REFERENCES snippet(document_id, snippet_idx);
```
rustonaut, line 22: This will handle interaction with snippets just fine, but we still have to decide how to handle interactions with a "full document" in many places. Majorly we can't have the rust equivalent of `(document_id, Option<snippet_id>)` in SQL :face_exhaling: (due to NULL != NULL and PRIMARY KEY implies NOT NULL)

web-api/src/storage.rs
```diff
@@ -143,10 +143,10 @@ impl<T> FromIterator<T> for Warning<T> {
 
 #[async_trait(?Send)]
 pub(crate) trait Document {
-    async fn get_interacted(
+    async fn get_snippets_for_interaction(
```
rustonaut, line 146: Why did we call it interacted? It's not past tense? 

As we might have interaction on both snippets and documents long term I added the `snipepts` part.

web-api/src/storage/postgres.rs
```diff
@@ -105,45 +100,38 @@ impl Database {
...
-        documents: impl IntoIterator<Item = &IngestedDocument>,
+        documents: &[DocumentForIngestion],
     ) -> Result<(), Error> {
         let mut tx = self.begin().await?;
 
         let mut builder = QueryBuilder::new(
             "INSERT INTO document (
                 document_id,
-                snippet,
+                raw_document,
```
rustonaut, line 110: :pushpin: Depending how we will handle `raw_documents` we might want to consider returning a hash of the document instead (it's used for deciding if an upsert is a property update or full document change).

web-api-db-ctrl/postgres/tenant/20230713173212_document.sql
janpetschexain: i would agree that the benefit of faster fetching outweighs the cognitive load coming with multiple tables.

web-api/src/storage/postgres.rs
```diff
@@ -145,6 +140,56 @@ impl Database {
...
+        let mut builder = QueryBuilder::new(
+            "INSERT INTO snippet (
+                        document_id,
+                        snippet_idx,
+                        snippet,
+                        embedding
+                    ) ",
+        );
+
+        while snippets.peek().is_some() {
```
janpetschexain, line 170: ❓/:art: couldn't we also use the new chunk iterator approach for this loop?

web-api/src/storage/postgres.rs
```diff
@@ -202,38 +247,43 @@ impl Database {
...
-    ) -> Result<Vec<InteractedDocument>, Error> {
+        ids: impl IntoIterator<IntoIter = impl ExactSizeIterator<Item = &SnippetId>>,
+    ) -> Result<Vec<SnippetForInteraction>, Error> {
         let mut builder = QueryBuilder::new(
-            "SELECT document_id, tags, embedding
-            FROM document
-            WHERE document_id IN ",
+            "SELECT s.document_id, s.snippet_idx, s.embedding, d.tags
+            FROM snippet s JOIN document d USING (document_id)
+            WHERE (document_id, snippet_idx) IN ",
```
janpetschexain, line 265: ❓ i would have assumed that we need to use the table alias as a prefix here as well, like `WHERE (d.document_id, s.snippet_idx) IN`? or does it work without the alias because that refers to the colums of the joined tables?

web-api/src/storage/postgres.rs
```diff
@@ -307,45 +368,40 @@ impl Database {
...
-                        })
-                    })
-                    .fetch_all(&mut *tx)
-                    .await?,
-            );
+        let mut chunks = IterAsTuple::chunks(Self::BIND_LIMIT, ids);
+        while let Some(ids) = chunks.next() {
+            let mut rows = builder.reset().push_tuple(ids).build().fetch(&mut *tx);
+
+            while let Some(row) = rows.next().await.transpose()? {
```
janpetschexain, line 383: ❓ could't we just do a `fetch_all()` and then iterate over the resulting vec, or is this more efficient in terms of allocations?

web-api/src/storage/postgres.rs
```diff
@@ -382,56 +446,96 @@ impl Database {
...
+    async fn set_is_candidate_and_return_for_ingestion(
+        tx: &mut Transaction<'_, Postgres>,
+        ids: impl ExactSizeIterator<Item = &DocumentId> + Clone,
+    ) -> Result<Vec<DocumentForIngestion>, Error> {
+        let mut builder = QueryBuilder::new(
+            "SELECT document_id, snippet_idx, snippet, embedding
+            FROM snippet
+            WHERE document_id IN ",
+        );
+        let mut snippets = HashMap::<_, BTreeMap<_, _>>::new();
```
janpetschexain, line 476: :art: can you add a comment regarding the btreemap? it took me a moment to figure out that we use this to return the snippets of each document ordered by their index.

web-api/src/storage/postgres.rs
```diff
@@ -382,56 +446,96 @@ impl Database {
...
+            let mut rows = builder
+                .reset()
+                .push_tuple(ids)
+                .push(" RETURNING document_id, preprocessing_step, properties, tags;")
+                .build()
+                .fetch(&mut *tx);
+
+            while let Some(row) = rows.next().await.transpose()? {
+                let document_id = row.try_get("document_id")?;
+                //Hint: We currently assume there are no gaps
```
janpetschexain, line 510: ❓ by "no gaps" do you mean that we assume there is at least one snippet in the snippet table for every document id in the document table?

web-api/src/storage/postgres.rs
```diff
@@ -676,22 +752,24 @@ impl Database {
...
-        let mut builder =
-            QueryBuilder::new("INSERT INTO interaction (doc_id, user_id, time_stamp) ");
+        let mut builder = QueryBuilder::new(
+            "INSERT INTO interaction (document_id, snippet_idx, user_id, time_stamp) ",
+        );
         while interactions.peek().is_some() {
             builder
                 .reset()
                 .push_values(
                     interactions.by_ref().take(Database::BIND_LIMIT / 3),
```
janpetschexain, line 768: 🔧 the bounded arguments increased by one
```suggestion
                    interactions.by_ref().take(Database::BIND_LIMIT / 4),
```

janpetschexain (commented):
i tried to comment only on changes related to this pr, it's a bit difficult since it's not yet rebased

web-api/src/personalization/stateless.rs
```diff
@@ -95,26 +97,38 @@ pub(super) async fn load_history(
...
-            loaded
-                .remove(&id)
-                .map(|(embedding, tags)| LoadedHistoryEntry {
+        .map(
+            |SnippetForInteraction {
+                 id,
+                 embedding,
+                 tags,
+             }| {
+                let timestamp = history[&id];
```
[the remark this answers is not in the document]
rustonaut: As `loaded ⊆ history`, this shouldn't be able to panic. Which is why I iterate over loaded and get from history instead of iterating over history and getting from loaded.

web-api/src/storage.rs
```diff
@@ -232,7 +246,7 @@ pub(crate) trait Interest {
 }
 
 pub(crate) struct InteractionUpdateContext<'s, 'l> {
-    pub(crate) document: &'s InteractedDocument,
+    pub(crate) document: &'s SnippetForInteraction,
```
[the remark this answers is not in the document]
rustonaut: yes but then I would need to replace 50%+ of all occurrences of the word document in the code, which is ... a lot :thinking:

web-api/src/storage/postgres.rs
rustonaut: What I added is a chunk iteration helper over things which are tuples, not a generic one.  But with how I changed it later it's not much different from a generic one, so I guess yes :thinking:

web-api/src/storage/postgres.rs
rustonaut: it is one allocation less and processes results while they come in, but in this context that is an irrelevant micro optimizations we don't have to do,  it's mainly there because the code went through multiple iterations and at some in-between point it was more readable

I want async for :cry:, and generators :old_man_yelling_at_cloud:

web-api/src/storage/postgres.rs
rustonaut: :+1:  just used a `Vec` and `sorted()` which shouldn't be slower

web-api/src/storage/postgres.rs
rustonaut: I mean that if there are 10 snippets for a specific document then their `snippet_idx` are 0..10, clarified it.

web-api/src/models.rs
```diff
@@ -359,39 +462,47 @@ impl AiDocument for PersonalizedDocument {
...
+pub(crate) struct DocumentForIngestion {
     /// Unique identifier of the document.
     pub(crate) id: DocumentId,
 
+    /// The raw document provided by the client.
+    pub(crate) raw_document: String,
+
     /// Snippet used to calculate embeddings for a document.
-    pub(crate) snippet: DocumentSnippet,
+    pub(crate) snippets: Vec<(DocumentSnippet, NormalizedEmbedding)>,
```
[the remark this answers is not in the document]
rustonaut: I added it but for now named it `DocumentContent` I will rename it when things are a bit more clear.

janpetschexain (approved):
lgtm 👍

PR #1057, opened 2023-08-08
ET-4924 multi snippet support in es [7/n]

rustonaut:
Support multi-snippet documents in elastic search.

**References:**

- story: [ET-4755]
- issue: [ET-4756]
- based on #1056 

### DIFF
--- a/web-api-db-ctrl/src/postgres/extern_migrations.rs
+++ b/web-api-db-ctrl/src/postgres/extern_migrations.rs
@@ -1,0 +1,123 @@
+// Copyright 2023 Xayn AG
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, version 3.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+use std::future::Future;
+
+use anyhow::bail;
+use async_trait::async_trait;
+use derive_more::{From, Into};
+use sqlx::{Postgres, Transaction, Type};
+
+use crate::Error;
+
+/// Allows running migration like "once per tenant" tasks.
+#[async_trait(?Send)]
+pub(crate) trait ExternalMigrator {
+    async fn run_migration_if_needed(
+        &mut self,
+        name: &str,
+        exec: impl Future<Output = Result<(), Error>>,
+    ) -> Result<(), Error>;
+}
+
+#[derive(From, Into)]
+pub(crate) struct PgExternalMigrator {
+    tx: Transaction<'static, Postgres>,
+}
+
+impl PgExternalMigrator {
+    fn tx(&mut self) -> &mut Transaction<'static, Postgres> {
+        &mut self.tx
+    }
+}
+
+#[derive(Debug, Copy, Clone, PartialEq, Type)]
+#[sqlx(type_name = "external_migration_state", rename_all = "snake_case")]
+enum ExternalMigrationState {
+    Pending,
+    Succeeded,
+    Failed,
+}
+
+impl ExternalMigrationState {
+    fn from_result(result: &Result<(), Error>) -> (Self, Option<String>) {
+        match result {
+            Ok(()) => (ExternalMigrationState::Succeeded, None),
+            Err(error) => (ExternalMigrationState::Failed, Some(error.to_string())),
+        }
+    }
+}
+
+#[async_trait(?Send)]
+impl ExternalMigrator for PgExternalMigrator {
+    async fn run_migration_if_needed(
+        &mut self,
+        name: &str,
+        exec: impl Future<Output = Result<(), Error>>,
+    ) -> Result<(), Error> {
+        let res = sqlx::query(
+            "INSERT INTO external_migration(name)
+                VALUES ($1)
+                ON CONFLICT DO NOTHING;",
+        )
+        .bind(name)
+        .execute(self.tx())
+        .await?;
+
+        if res.rows_affected() == 0 {
+            let (existing,) = sqlx::query_as::<_, (ExternalMigrationState,)>(
+                "SELECT state
+                    FROM external_migration
+                    WHERE name = $1
+                    FOR UPDATE;",
+            )
+            .bind(name)
+            .fetch_one(self.tx())
+            .await?;
+
+            return match existing {
+                ExternalMigrationState::Succeeded => Ok(()),
+                ExternalMigrationState::Failed => {
+                    bail!("previous external migration failed: {name}")
+                }
+                ExternalMigrationState::Pending => {
+                    unreachable!(/*
+                        'pending' only exist during ongoing transactions as we don't share the transaction we
+                        run on with other tasks this should make it impossible to observe this state
+                    */)
+                }
+            };
+        }
+
+        let result = exec.await;
+
+        let (state, error) = ExternalMigrationState::from_result(&result);
+        sqlx::query(
+            "UPDATE external_migration
+            SET
+                state = $1,
+                end_time = CURRENT_TIMESTAMP,
+                error = $2
+            WHERE
+                name = $3;",
+        )
+        .bind(state)
+        .bind(error)
+        .bind(name)
+        .execute(self.tx())
+        .await?;
+
+        result
+    }
+}
--- a/web-api-db-ctrl/src/postgres/extern_migrations.rs
+++ b/web-api-db-ctrl/src/postgres/extern_migrations.rs
@@ -36,12 +36,6 @@ pub(crate) struct PgExternalMigrator {
     tx: Transaction<'static, Postgres>,
 }
 
-impl PgExternalMigrator {
-    fn tx(&mut self) -> &mut Transaction<'static, Postgres> {
-        &mut self.tx
-    }
-}
-
 #[derive(Debug, Copy, Clone, PartialEq, Type)]
 #[sqlx(type_name = "external_migration_state", rename_all = "snake_case")]
 enum ExternalMigrationState {
@@ -72,7 +66,7 @@ impl ExternalMigrator for PgExternalMigrator {
                 ON CONFLICT DO NOTHING;",
         )
         .bind(name)
-        .execute(self.tx())
+        .execute(&mut self.tx)
         .await?;
 
         if res.rows_affected() == 0 {
@@ -83,7 +77,7 @@ impl ExternalMigrator for PgExternalMigrator {
                     FOR UPDATE;",
             )
             .bind(name)
-            .fetch_one(self.tx())
+            .fetch_one(&mut self.tx)
             .await?;
 
             return match existing {
@@ -93,9 +87,10 @@ impl ExternalMigrator for PgExternalMigrator {
                 }
                 ExternalMigrationState::Pending => {
                     unreachable!(/*
-                        'pending' only exist during ongoing transactions as we don't share the transaction we
-                        run on with other tasks this should make it impossible to observe this state
-                    */)
+                        'pending' only exist during ongoing transactions as we don't share the
+                        transaction we run on with other tasks this should make it impossible to
+                        observe this state
+                    */);
                 }
             };
         }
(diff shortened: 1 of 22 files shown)


web-api/src/storage/elastic.rs
```diff
@@ -145,72 +150,68 @@ impl Client {
...
-        let response = self
-            .bulk_request(ids.map(|id| Ok(BulkInstruction::Delete { id })))
-            .await?;
-        Ok(response.failed_documents("delete", true).into())
+        let response = self.bulk_request(snippets).await?;
+        Ok(response.failed_documents("index", false).into())
     }
 
-    pub(super) async fn retain_documents(
+    pub(super) async fn delete_by_parents(
```
rustonaut, line 201: This is very convenient in many places but there is one gotcha.

:wrench:  But any previously ingested document doesn't has a `parent` field.

So either delete by parents needs to run and addition batch request to delete all cases where parent == ES-Document id or we need to run a ES migration with a script update query which sets the parent field where missing.

web-api/src/storage/elastic.rs
```diff
@@ -404,6 +409,15 @@ impl Client {
     }
 }
 
+pub(super) trait SerializeDocumentIds: Serialize {}
+impl<T> SerializeDocumentIds for &'_ T where T: SerializeDocumentIds {}
```
janpetschexain, line 413: :art: this could be generalized a bit
```suggestion
impl<T> SerializeDocumentIds for &'_ T where T: SerializeDocumentIds + ?Sized {}
```

web-api/src/storage/elastic.rs
janpetschexain: i'd say it's better to run es migration once instead of running additional batch requests every time.

web-api/src/storage/elastic.rs
```diff
@@ -145,72 +150,68 @@ impl Client {
...
+        Ok(response.failed_documents("index", false).into())
     }
 
-    pub(super) async fn retain_documents(
+    pub(super) async fn delete_by_parents(
         &self,
-        ids: impl IntoIterator<IntoIter = impl ExactSizeIterator<Item = &DocumentId>>,
+        parents: impl SerializeDocumentIds,
     ) -> Result<(), Error> {
-        // https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-delete-by-query.html
```
janpetschexain, line 198: :art: we could keep the link to the es endpoint docs for easier access
```suggestion
        // https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-delete-by-query.html
```

web-api/src/storage/elastic.rs
```diff
@@ -145,72 +150,68 @@ impl Client {
...
+                            serde_json::to_value(BulkInstruction::Create { id: id.to_es_id() })
+                                .map_err(Into::into);
+                        let data = serde_json::to_value(EsDocument {
+                            snippet,
+                            properties: &document.properties,
+                            embedding,
+                            tags: &document.tags,
+                            parent: id.document_id(),
+                        })
+                        .map_err(Into::into);
```
janpetschexain, line 187: :art: no need for extra error conversion
```suggestion
                        });
```

web-api/src/storage/elastic.rs
```diff
@@ -145,72 +150,68 @@ impl Client {
...
+                document
+                    .snippets
+                    .iter()
+                    .enumerate()
+                    .flat_map(|(idx, (snippet, embedding))| {
+                        #[allow(clippy::cast_possible_truncation)]
+                        let id = SnippetId::new(document.id.clone(), idx as _);
+                        let header =
+                            serde_json::to_value(BulkInstruction::Create { id: id.to_es_id() })
+                                .map_err(Into::into);
```
janpetschexain, line 179: :art: no need for extra error conversion
```suggestion
                            serde_json::to_value(BulkInstruction::Create { id: id.to_es_id() });
```

web-api/src/storage/elastic.rs
```diff
@@ -442,10 +456,11 @@ pub(crate) enum IndexUpdateMethod {
 }
 
 #[derive(Debug, Serialize)]
-struct IngestedDocument<'a> {
+struct EsDocument<'a> {
```
janpetschexain, line 459: :art: this is the only `Document` type comming from the `elastic` module, so we could just skip the prefix
```suggestion
struct Document<'a> {
```

web-api/src/storage/postgres.rs
```diff
@@ -421,22 +421,32 @@ impl Database {
...
+
+        ingestable.retain(|id| !unchanged.contains(*id));
 
         let mut builder = QueryBuilder::new(
             "UPDATE document
             SET is_candidate = FALSE
             WHERE document_id IN ",
         );
-        let mut chunks = IterAsTuple::chunks(Self::BIND_LIMIT, removed);
+        let mut chunks = IterAsTuple::chunks(Self::BIND_LIMIT, removed.iter());
```
janpetschexain, line 449: :art: we could leave the iteration to `IntoIterator`
```suggestion
        let mut chunks = IterAsTuple::chunks(Self::BIND_LIMIT, &removed);
```

web-api/src/storage/postgres.rs
```diff
@@ -917,19 +927,17 @@ impl storage::Document for Storage {
                     Either::Right(document.id)
                 }
             });
-        let mut failed_documents = self.elastic.insert_documents(&candidates).await?;
-        failed_documents.extend(self.elastic.delete_documents(&noncandidates).await?);
-
+        let failed_documents = self.elastic.upsert_documents(&candidates).await?;
+        self.elastic.delete_by_parents(&noncandidates).await?;
```
janpetschexain, line 931: :art: newline before return
```suggestion
        self.elastic.delete_by_parents(&noncandidates).await?;

```

web-api/src/storage/postgres.rs
```diff
@@ -917,19 +927,17 @@ impl storage::Document for Storage {
...
 
     async fn delete(
         &self,
         ids: impl IntoIterator<IntoIter = impl Clone + ExactSizeIterator<Item = &DocumentId>>,
     ) -> Result<Warning<DocumentId>, Error> {
-        let (candidates, mut failed_documents) = self.postgres.delete_documents(ids).await?;
-        failed_documents.extend(self.elastic.delete_documents(&candidates).await?);
-
+        let (candidates, failed_documents) = self.postgres.delete_documents(ids).await?;
+        self.elastic.delete_by_parents(&candidates).await?;
```
janpetschexain, line 940: :art: newline before return
```suggestion
        self.elastic.delete_by_parents(&candidates).await?;

```

web-api/src/storage/postgres.rs
```diff
@@ -968,9 +976,8 @@ impl storage::DocumentCandidate for Storage {
         &self,
         ids: impl IntoIterator<Item = &DocumentId>,
     ) -> Result<Warning<DocumentId>, Error> {
-        let (removed, mut failed) = self.postgres.remove_candidates(ids).await?;
-        failed.extend(self.elastic.delete_documents(&removed).await?);
-
+        let (removed, failed) = self.postgres.remove_candidates(ids).await?;
+        self.elastic.delete_by_parents(&removed).await?;
```
janpetschexain, line 980: :art: newline before return
```suggestion
        self.elastic.delete_by_parents(&removed).await?;

```

web-api/src/storage/elastic.rs
rustonaut: `T: Serialize` is implied which also implies `Sized`

web-api/src/storage/elastic.rs
rustonaut: okay added the functionality :+1:

rustonaut: side not it doesn't migrate the ES "mappings" only the missing parent field, but can be used for other things later one too

web-api-shared/src/net.rs
```diff
@@ -67,6 +71,15 @@ impl<F> ExponentialJitterRetryPolicy<F> {
...
 
+    pub fn retry<MakeFut, Fut>(self, futures: MakeFut) -> RetryFuture<Self, MakeFut, Fut>
+    where
+        Self: RetryPolicy<Fut::Output>,
+        MakeFut: FnMut() -> Fut,
+        Fut: Future,
+    {
+        retry(self, futures)
+    }
+
```
rustonaut, line 82: At some point I thought I needed to use a additional retry policy and added this method to allow more convenient usage. As it's still nice I kept that change.

web-api-db-ctrl/src/postgres/extern_migrations.rs
```diff
@@ -0,0 +1,123 @@
...
+#[derive(From, Into)]
+pub(crate) struct PgExternalMigrator {
+    tx: Transaction<'static, Postgres>,
+}
+
+impl PgExternalMigrator {
+    fn tx(&mut self) -> &mut Transaction<'static, Postgres> {
+        &mut self.tx
+    }
+}
```
janpetschexain, line 43: :art: we could just inline `&mut self.tx` at the three places where this function is called.
```suggestion
```

web-api-db-ctrl/src/postgres/extern_migrations.rs
```diff
@@ -0,0 +1,123 @@
...
+        name: &str,
+        exec: impl Future<Output = Result<(), Error>>,
+    ) -> Result<(), Error> {
+        let res = sqlx::query(
+            "INSERT INTO external_migration(name)
+                VALUES ($1)
+                ON CONFLICT DO NOTHING;",
+        )
+        .bind(name)
+        .execute(self.tx())
```
janpetschexain, line 75: :art:
```suggestion
        .execute(&mut self.tx)
```

web-api-db-ctrl/src/postgres/extern_migrations.rs
```diff
@@ -0,0 +1,123 @@
...
+
+        if res.rows_affected() == 0 {
+            let (existing,) = sqlx::query_as::<_, (ExternalMigrationState,)>(
+                "SELECT state
+                    FROM external_migration
+                    WHERE name = $1
+                    FOR UPDATE;",
+            )
+            .bind(name)
+            .fetch_one(self.tx())
```
janpetschexain, line 86:
```suggestion
            .fetch_one(&mut self.tx)
```

web-api-db-ctrl/src/postgres/extern_migrations.rs
```diff
@@ -0,0 +1,123 @@
...
+            return match existing {
+                ExternalMigrationState::Succeeded => Ok(()),
+                ExternalMigrationState::Failed => {
+                    bail!("previous external migration failed: {name}")
+                }
+                ExternalMigrationState::Pending => {
+                    unreachable!(/*
+                        'pending' only exist during ongoing transactions as we don't share the transaction we
+                        run on with other tasks this should make it impossible to observe this state
+                    */)
```
janpetschexain, line 98: :art: terminating commas and 100 character limit formatting
```suggestion
                    bail!("previous external migration failed: {name}");
                }
                ExternalMigrationState::Pending => {
                    unreachable!(/*
                        'pending' only exist during ongoing transactions as we don't share the
                        transaction we run on with other tasks this should make it impossible to
                        observe this state
                    */);
```

web-api-db-ctrl/src/postgres/extern_migrations.rs
```diff
@@ -0,0 +1,123 @@
...
+use sqlx::{Postgres, Transaction, Type};
+
+use crate::Error;
+
+/// Allows running migration like "once per tenant" tasks.
+#[async_trait(?Send)]
+pub(crate) trait ExternalMigrator {
+    async fn run_migration_if_needed(
+        &mut self,
+        name: &str,
```
janpetschexain, line 29: ❓ should we make this a `&'static str`? they name will be fixed anyways for each migration which we add.

janpetschexain (approved):
lgtm

web-api-db-ctrl/src/postgres/extern_migrations.rs
rustonaut: okay, through it was convenient to have as during development the type of `tx` was changed multiple times

web-api-db-ctrl/src/postgres/extern_migrations.rs
rustonaut: I don't think this will stay static.

E.g. we currently don't have a proper old schema => new schema ES migration (e.g. when adding new fields like `tags`)

And one thing I considered doing to fix that and additionally make some other code more robust is to not just check the embedding size for compatibility but check for each property in the expected if it exists and  if not create on the fly a `add-property-<name>` migration which adds that.

We can't hard code that for properties we added in the past, as AFIK our existing tenants might have a mixed state of weather they have an explicit `tags` mapping or not (it's happens to be either-way fine for tags and parents, but the "happens to" part is a bit risky long term)

PR #1058, opened 2023-08-08
ET-4925 support document splitting using cutters [8/n]

rustonaut:
- add support for preprecessing documents using cutters
- readded tests for split documents

**References:**

- story: [ET-4755]
- issue: [ET-4756]
- based on #1057 

### DIFF
--- a/web-api/src/error/common.rs
+++ b/web-api/src/error/common.rs
@@ -146,6 +146,7 @@ pub(crate) struct InvalidDocumentTags {
 impl_application_error!(InvalidDocumentTags => BAD_REQUEST, INFO);
 
 #[derive(Debug, Error, Display, Serialize)]
+#[serde(rename_all = "snake_case")]
 pub(crate) enum InvalidDocumentSnippet {
     /// Malformed document snippet: {0}
     InvalidString(InvalidString),
--- a/web-api/src/error/common.rs
+++ b/web-api/src/error/common.rs
@@ -145,10 +145,13 @@ pub(crate) struct InvalidDocumentTags {
 
 impl_application_error!(InvalidDocumentTags => BAD_REQUEST, INFO);
 
-/// Malformed document snippet: {0}
 #[derive(Debug, Error, Display, Serialize)]
-#[serde(transparent)]
-pub(crate) struct InvalidDocumentSnippet(pub(crate) InvalidString);
+pub(crate) enum InvalidDocumentSnippet {
+    /// Malformed document snippet: {0}
+    InvalidString(InvalidString),
+    /// Input document didn't yield any snippets
+    NoSnippets {},
+}
 
 impl_application_error!(InvalidDocumentSnippet => BAD_REQUEST, INFO);
 
--- a/web-api/tests/split_documents.rs
+++ b/web-api/tests/split_documents.rs
@@ -1,0 +1,502 @@
+// Copyright 2023 Xayn AG
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, version 3.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+use std::collections::HashSet;
+
+use anyhow::Error;
+use itertools::Itertools;
+use reqwest::{Client, StatusCode};
+use serde::Deserialize;
+use serde_json::{json, Map, Value};
+use url::Url;
+use xayn_integration_tests::{
+    send_assert,
+    send_assert_json,
+    test_two_apps,
+    with_dev_options,
+    UNCHANGED_CONFIG,
+};
+use xayn_web_api::{Ingestion, Personalization};
+
+#[derive(Debug, Deserialize)]
+struct PersonalizedDocumentData {
+    id: String,
+    snippet_id: SnippetId,
+    score: f32,
+    // included by test only dev option
+    #[serde(default)]
+    snippet: Option<String>,
+}
+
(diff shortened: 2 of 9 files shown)


web-api/src/ingestion/preprocess_document.rs
```diff
@@ -53,3 +55,21 @@ fn embed_with_summarizer(
...
+
+fn embed_with_cutters(
+    embedder: &Embedder,
+    snippet: &DocumentSnippet,
+) -> Result<DocumentContent, Error> {
+    let mut snippets = Vec::new();
+    for split in cutters::cut(snippet, cutters::Language::English) {
+        let str = split.str.trim();
+        let snippet = DocumentSnippet::new(str, str.len())?;
+        let embedding = embedder.run(split.str)?;
```
janpetschexain, line 67: 🔧 trimming is already done by the snippet constructor. and we should compute the embedding on the `snippet` instead of the `str`, otherwise this could introduce bugs if the snippet constructor gets changed.

❓/:wrench: i'm not sure why the `str.len()` is passed to the snippet constructor? since we enforce the max byte size on the input snippet and the splitter must yield shorter snippets, we can pass the original byte size requirement.

```suggestion
        let snippet = DocumentSnippet::new(split.str, snippet.len())?;
        let embedding = embedder.run(&snippet)?;
```

web-api/src/ingestion/preprocess_document.rs
```diff
@@ -53,3 +55,21 @@ fn embed_with_summarizer(
...
+        snippets.push((snippet, embedding));
+    }
```
janpetschexain, line 69: :art: instead of a push-loop we could just iterate on the cutter values and collect afterwards. (if we stick to the push loop, then we should at least allocate snippets with the capacity to hold all splits from `cutters::cut()` since this would be the happy path of the loop.)
```suggestion
    let snippets = cutters::cut(snippet, cutters::Language::English)
        .into_iter()
        .map(|split| {
            let snippet = DocumentSnippet::new(split.str, snippet.len())?;
            let embedding = embedder.run(&snippet)?;
            Ok::<_, Error>((snippet, embedding))
        })
        .try_collect::<_, Vec<_>, _>()?;
```

web-api/src/error/common.rs
```diff
@@ -137,10 +137,14 @@ pub(crate) struct InvalidDocumentTags {
...
-/// Malformed document snippet: {0}
 #[derive(Debug, Error, Display, Serialize)]
-#[serde(transparent)]
-pub(crate) struct InvalidDocumentSnippet(pub(crate) InvalidString);
+#[serde(rename_all = "snake_case")]
+pub(crate) enum InvalidDocumentSnippet {
+    /// Malformed document snippet: {0}
+    InvalidString(InvalidString),
+    /// Input document didn't yield any snippets
+    NoSnippets {},
```
janpetschexain, line 146: ❓ with the current error impl this would be a bad request on info log level. if the cutter fails, wouldn't that be more like an internal error, or at least warn log level?

web-api/tests/split_documents.rs
```diff
@@ -0,0 +1,502 @@
...
+                )
+            })
+            .collect_vec()
+    }
+
+    fn sanity_checks(&self) {
+        let ordered = self
+            .documents
+            .iter()
+            .zip(self.documents.iter().skip(1))
```
janpetschexain, line 70: :art: instead of zipping two iterators we could slide a tuple window over one iterator (with `itertools`)
```suggestion
            .tuple_windows()
```

web-api/tests/split_documents.rs
```diff
@@ -0,0 +1,502 @@
...
+    UNCHANGED_CONFIG,
+};
+use xayn_web_api::{Ingestion, Personalization};
+
+#[derive(Debug, Deserialize)]
+struct PersonalizedDocumentData {
+    id: String,
+    snippet_id: SnippetId,
+    score: f32,
+    // included by test only dev option
```
janpetschexain, line 37: :art: this is reachable by everyone with dev access, not just the tests
```suggestion
    // included by dev option
```

web-api/src/error/common.rs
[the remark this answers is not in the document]
rustonaut: It's mainly for consistency even if we don't add fields later one,

e.g. with `NoSnippets` we have 

```json
{
  "details": "no_snippets"
}
```

but with `NoSnippets {}` we have

```json
{
  "details": {
    "no_snippets": {}
  }
}
```

which is in line with e.g. 

```json
{
  "details": {
    "invalid_string": {
      "size": {
        ...
       }
    }
  }
}
```

rustonaut: no it's not cutter failing but there not being any semantic sentences (which could be because of a subtle failure, but likely is because of different problems).

E.g. `"--------"` isn't a single valid sentence but a non-empty input. 

Note that I didn't check if `cutters` would just return that verbatim, because it doesn't matter because we do replace it very soon anyway.

rustonaut: Stuff like that is why I last week asked in the channel if we could change terminology to not use splits, because at least long term we don't just split the document. We extract sentences from it. Which might happen to be splits but could very well be something else, and they definitely need all formatting stripped when we get PDFs or similar etc.

web-api/src/ingestion/preprocess_document.rs
rustonaut: I pass len because in this part of the pipeline it's not supposed to fail, neither is the config of the input length limits  supposed to be wired through to here.

But we have to rethink this whole snippet length validation (and payment) anyway (in a different PR).

1. when we accept document which are split we probably want to allow them to be much larger then any specific snippet
2. but at that point  it's _our_ responsibility to keep snippets reasonably small
3.  and if we can't do so for a "small" subset of snippets of the same document we still probably don't want to fail ingesting the whole document I would guess, that would be horrible ux ("hy we reject this document due to arbitrary internal reasons"), no idea what we would do but probably something on the line of if we can't get it smaller summarize it and if that doesn't work truncate it and warn. Anyway outside of this PR.
4. through either-way "small" here is now more dependent on what the tokenizer can handle then storage constraints we have now

janpetschexain (approved):
lgtm

PR #1060, opened 2023-08-09
ET-4938 allow different embedder

Robert-Steiner:
Summary:
- allow different embedder

### DIFF
--- a/web-api/src/embedding.rs
+++ b/web-api/src/embedding.rs
@@ -18,7 +18,7 @@ use xayn_ai_bert::{AvgEmbedder, Config as EmbedderConfig, NormalizedEmbedding};
 use crate::{app::SetupError, error::common::InternalError, utils::RelativePathBuf};
 
 #[derive(Debug, Deserialize, Serialize)]
-#[serde(rename_all = "lowercase")]
+#[serde(rename_all = "lowercase", deny_unknown_fields)]
 pub enum Config {
     Pipeline(Pipeline),
 }
@@ -30,7 +30,7 @@ impl Default for Config {
 }
 
 #[derive(Debug, Deserialize, Serialize)]
-#[serde(default)]
+#[serde(default, deny_unknown_fields)]
 pub struct Pipeline {
     pub(crate) directory: RelativePathBuf,
     pub(crate) token_size: usize,
--- a/web-api/src/embedding.rs
+++ b/web-api/src/embedding.rs
@@ -18,7 +18,7 @@ use xayn_ai_bert::{AvgEmbedder, Config as EmbedderConfig, NormalizedEmbedding};
 use crate::{app::SetupError, error::common::InternalError, utils::RelativePathBuf};
 
 #[derive(Debug, Deserialize, Serialize)]
-#[serde(tag = "type", rename_all = "lowercase", deny_unknown_fields)]
+#[serde(tag = "type", rename_all = "snake_case", deny_unknown_fields)]
 pub enum Config {
     Pipeline(Pipeline),
 }
--- a/web-api/src/embedding.rs
+++ b/web-api/src/embedding.rs
@@ -18,13 +18,25 @@ use xayn_ai_bert::{AvgEmbedder, Config as EmbedderConfig, NormalizedEmbedding};
 use crate::{app::SetupError, error::common::InternalError, utils::RelativePathBuf};
 
 #[derive(Debug, Deserialize, Serialize)]
-#[serde(default, deny_unknown_fields)]
-pub struct Config {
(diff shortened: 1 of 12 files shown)


web-api/src/embedding.rs
```diff
@@ -17,14 +17,26 @@ use xayn_ai_bert::{AvgEmbedder, Config as EmbedderConfig, NormalizedEmbedding};
 
 use crate::{app::SetupError, error::common::InternalError, utils::RelativePathBuf};
 
+#[derive(Debug, Deserialize, Serialize)]
+#[serde(rename_all = "lowercase")]
+pub enum Config {
```
janpetschexain, line 22: :art: we could use an untagged enum for serde, then the current serialized format would stay the same and it wouldn't introduce another nested table in the toml. (this of course only works if `Pipeline` and the new sagemaker config object don't have the exact same fields, which they probably don't have.)
```suggestion
#[serde(rename_all = "lowercase", untagged)]
pub enum Config {
```

web-api/src/embedding.rs
```diff
@@ -17,14 +17,26 @@ use xayn_ai_bert::{AvgEmbedder, Config as EmbedderConfig, NormalizedEmbedding};
...
+    Pipeline(Pipeline),
```
janpetschexain, line 23: :art: we could also make this a structured variant like
```suggestion
    Pipeline {
        directory: RelativePathBuf,
        token_size: usize,
    },
```

web-api/src/embedding.rs
acrrd: untagged is a problem because the user can put the field of both and then the behaviour depends on the order of the enum. I don't know if untagged + deny_unknow can help here, but I prefer internally tagged but there actually the problem of backward compatibility, I don't know if we can specify a default for the tag if not present.

web-api/src/embedding.rs
Robert-Steiner: I just tried to use a structured variant and i ran into the problem that cannot use #[serde(default)] on the enum. this becomes an issue when running the integration tests 
```
Error: invalid type: found string "/Users/debug/projects/xayn_discovery_engine/assets/xaynia_v0002", expected struct RelativePathBuf for key "default.embedding" in TOML source string
```
`cargo test -p xayn-web-api@0.2.0 --test document_candidates -- --nocapture`

Robert-Steiner: I also tried to implement the alternative solution described [here](https://github.com/serde-rs/serde/issues/1799#issuecomment-624978919) to define a default but it don't not work in combination with having a default on the config enum itself.

web-api/src/embedding.rs
janpetschexain: that's true. with the current change to use the internally tagged enum the nesting is gone, so that should be fine.

web-api/src/embedding.rs
janpetschexain: yeah, unfortunately `#[serde(default)]` doesn't extend to structured enum variants. we could work around that in some way, but the current solution should be ok.

web-api/src/embedding.rs
```diff
@@ -17,14 +17,26 @@ use xayn_ai_bert::{AvgEmbedder, Config as EmbedderConfig, NormalizedEmbedding};
 
 use crate::{app::SetupError, error::common::InternalError, utils::RelativePathBuf};
 
+#[derive(Debug, Deserialize, Serialize)]
+#[serde(tag = "type", rename_all = "lowercase", deny_unknown_fields)]
```
janpetschexain, line 21: 🔧 for correct formatting if we add variants with more than one word in the future
```suggestion
#[serde(tag = "type", rename_all = "snake_case", deny_unknown_fields)]
```

web-api/src/embedding.rs
```diff
@@ -17,14 +17,26 @@ use xayn_ai_bert::{AvgEmbedder, Config as EmbedderConfig, NormalizedEmbedding};
...
+        Self::Pipeline(Pipeline::default())
+    }
+}
+
 #[derive(Debug, Deserialize, Serialize)]
 #[serde(default, deny_unknown_fields)]
-pub struct Config {
-    pub(crate) directory: RelativePathBuf,
+pub struct Pipeline {
+    pub(crate) directory: String,
```
janpetschexain, line 35: :wrench: we should keep the `RelativePathBuf` here. this just requires to remove the conversion in l.58 and add it in l.98

```suggestion
    pub(crate) directory: RelativePathBuf,
```

janpetschexain (approved):
lgtm

PR #1065, opened 2023-08-10
ET-4750 sagemaker

Robert-Steiner:
Summary:
- add sagemaker embedder
- add the possibility to add a target model (required for multi model mode)
- add the possibility to retry a sagemaker request

### DIFF
--- a/web-api/src/embedding.rs
+++ b/web-api/src/embedding.rs
@@ -163,7 +163,7 @@ impl Embedder {
             request = request.target_model(target_model);
         };
 
-        let res = request.send().await.map_err(|e| {
+        let response = request.send().await.map_err(|e| {
             InternalError::from_message(format!("Failed to request sagemaker endpoint. Error: {e}"))
         })?;
 
--- a/web-api/src/embedding.rs
+++ b/web-api/src/embedding.rs
@@ -173,13 +173,20 @@ impl Embedder {
             ));
         };
 
-        serde_json::from_slice::<SagemakerResponse>(body.as_ref())
+        let mut embeddings = serde_json::from_slice::<SagemakerResponse>(body.as_ref())
             .map_err(InternalError::from_std)?
-            .embeddings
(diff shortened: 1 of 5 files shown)


web-api/src/embedding.rs
```diff
@@ -54,45 +58,136 @@ impl Default for Pipeline {
...
+        let body = res.body().ok_or(InternalError::from_message(
+            "Received sagemaker response without body.",
+        ))?;
+
+        serde_json::from_slice::<SagemakerResponse>(body.as_ref())
+            .map_err(|e| {
+                InternalError::from_message(format!(
+                    "Failed to deserialize sagemaker response body. Error: {e}"
+                ))
+            })?
```
janpetschexain, line 179: :art: this error would already state that it is a deserialization error for sagemaker, so we can save some message bloat.
```suggestion
            .map_err(InternalError::from_std)?
```

web-api/src/embedding.rs
```diff
@@ -54,45 +58,136 @@ impl Default for Pipeline {
...
+            .embeddings
```
janpetschexain, line 180: ❓ shouldn't we also check that `.embeddings.len() == 1` and otherwise return an error?

web-api/src/embedding.rs
```diff
@@ -54,45 +58,136 @@ impl Default for Pipeline {
...
+                target_model,
+                ..
+            } => Self::run_sagemaker(client, endpoint, target_model, sequence).await,
         }
     }
 
+    async fn run_sagemaker(
+        client: &Client,
+        endpoint: &str,
+        target_model: &Option<String>,
```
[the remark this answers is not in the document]
Robert-Steiner: nice!

web-api/src/embedding.rs
Robert-Steiner: done

PR #1054, opened 2023-08-03
ET-4300 Add ort

acrrd:
**Reference**

- [ET-4300]
- [ET-4912]
- [ET-4301]
- followed by #1066

**Summary**

Initial support for ort to compute the embeddings.
It can be selected with an option.
Some data structure it is converted to the tract one to avoid changing too much.

To deploy it, some changes can be needed in a follow-up.

Once we have ort working and deployed, we can remove tract.

- currently the onnxruntime lib is added to the assets bucket, probably will be added to the ci image later (but no gpu lib yet)
- integration tests already run with onnxruntime instead of tract
- engine defaults run with onnxruntime
- webservice build also copies onnxruntime files
- update the benchmark, the results are in [ET-4301]
- image builds successfully: https://github.com/xaynetwork/xayn_discovery_engine/actions/runs/5857592118

[ET-4300]: https://xainag.atlassian.net/browse/ET-4300?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[ET-4912]: https://xainag.atlassian.net/browse/ET-4912?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[ET-4301]: https://xainag.atlassian.net/browse/ET-4301?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ET-4301]: https://xainag.atlassian.net/browse/ET-4301?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1066, opened 2023-08-14
ET-4943 remove tract

janpetschexain:
**Reference**

- [ET-4943]
- requires #1054
- followed by #1067

**Summary**

- remove `tract` dependency
- use `ort` native types for model inputs and outputs

[ET-4943]: https://xainag.atlassian.net/browse/ET-4943?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bert/src/model.rs
+++ b/bert/src/model.rs
@@ -12,11 +12,10 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-use std::{env, fs::File, io::BufReader};
+use std::env;
 
-use anyhow::Error;
-use derive_more::{Deref, From};
-use ndarray::CowArray;
+use anyhow::{bail, Error};
+use ndarray::{Array, CowArray, IxDyn};
 use ort::{
     environment::Environment,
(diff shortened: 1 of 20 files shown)


bert/src/model.rs
```diff
@@ -144,209 +68,121 @@ impl Runtime {
...
+            vec![0; token_size],
+            vec![0; token_size],
+            Vec::new(),
+            Vec::new(),
+            Vec::new(),
+            Vec::new(),
+            vec![1; token_size],
+            Vec::new(),
+            HashMap::new(),
+        );
```
rustonaut, line 180: :art:  This constructor falls cleanly into the "never write code like that" area, but it's also outside of our code, so do we keep it like that?

janpetschexain: i know it's not nice 😅 it's the shortest workaround i found to set up the test, and since we know that the unset parts are not used, it should be ok to keep it. wdyt?

rustonaut (approved):
lgtm

PR #1067, opened 2023-08-15
ET-4949 dynamic token size

janpetschexain:
**Reference**

- [ET-4949]
- requires #1066

**Summary**

- remove fixed token size:
  - padding is set to longest of the batch, currently that means no padding at all, if we chose to go with batch sizes > 1 then this will actually take effect
  - truncation is still applied, the token size keeps its meaning in that regard that it acts as a max token size, this also means that the embedding computation is not breaking for existing tenants (removed padding is also not breaking because we average-pool the embeddings only over the active tokens which doesn't apply to padding tokens)
- removed unused model config values & simplify/consistify remaining config values, all used models have been updated accordingly:
  - `smbert`: `v0004` -> `v0005`
  - `smbert_mocked`: `v0004` -> `v0005`
  - `e5_mocked`: `v0000` -> `v0001` (this wasn't a mocked model but a small-ish e5 model before, now it's really mocked)
  - `xaynia`: `v0002` -> `v0201`
  - `e5_small_v2`: `v0000` -> `v0001`
  - `multilingual_e5_small`: `v0000` -> `v0001`
- update the python model mock tool

[ET-4949]: https://xainag.atlassian.net/browse/ET-4949?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/scripts/download_assets.sh
+++ b/.github/scripts/download_assets.sh
@@ -50,9 +50,9 @@ if [ $# -gt 0 ]; then
     done
 else
     download smbert v0003
-    download smbert v0004
-    download smbert_mocked v0004
-    download e5_mocked v0000
-    download xaynia v0002
+    download smbert v0005
+    download smbert_mocked v0005
+    download e5_mocked v0001
+    download xaynia v0201
     download ort v1.15.1
 fi
--- a/bert/src/tokenizer.rs
+++ b/bert/src/tokenizer.rs
@@ -40,9 +40,9 @@ impl Tokenizer {
             );
         }
         let mut tokenizer = HfTokenizer::from_file(tokenizer)?;
-        let padding_token = config.extract::<String>("tokenizer.tokens.padding")?;
+        let padding_token = config.extract::<String>("tokenizer.padding")?;
         let padding = PaddingParams {
-            strategy: PaddingStrategy::Fixed(config.token_size),
+            strategy: PaddingStrategy::BatchLongest,
             direction: PaddingDirection::Right,
             pad_to_multiple_of: None,
             pad_id: tokenizer
@@ -59,7 +59,7 @@ impl Tokenizer {
         };
         tokenizer.with_padding(Some(padding));
         tokenizer.with_truncation(Some(truncation));
-        let add_special_tokens = config.extract::<bool>("tokenizer.add-special-tokens")?;
+        let add_special_tokens = config.extract::<bool>("tokenizer.add_special_tokens")?;
 
         Ok(Tokenizer {
             tokenizer,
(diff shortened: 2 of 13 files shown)


.github/scripts/download_assets.sh
```diff
@@ -50,9 +50,9 @@ if [ $# -gt 0 ]; then
...
 else
     download smbert v0003
-    download smbert v0004
-    download smbert_mocked v0004
-    download e5_mocked v0000
-    download xaynia v0002
+    download smbert v0005
+    download smbert_mocked v0005
+    download e5_mocked v0001
+    download xaynia v0201
```
rustonaut, line 56: :question: 201 not 2

bert/src/tokenizer.rs
```diff
@@ -130,18 +135,12 @@ mod tests {
...
-        assert_eq!(
-            encoding.get_attention_mask()[..15],
-            [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 0, 0],
-        );
+        assert_eq!(encoding.get_ids().len(), 12);
         assert_eq!(
-            encoding.get_type_ids()[..15],
-            [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
+            encoding.get_ids(),
+            [101, 2122, 2024, 3671, 1010, 2691, 7861, 8270, 4667, 2015, 1012, 102],
```
rustonaut, line 141: The ids changed here, is that an issue?

.github/scripts/download_assets.sh
janpetschexain: i discussed some time ago with @acrrd how we could version `xaynia` to indicate that it still refers to `xaynia 2` even if we have to do small updates to the archive. the easiest thing we came up with was to use the first two digits for the major and the last two digits for the minor version bump.

but since we for example now also have `ort v1.15.1` as an archive, we could do the same for `xaynia v2.1`. the downside would be that we lose the lexicographical ordering of `v0201`, `v0002`, etc.

bert/src/tokenizer.rs
janpetschexain: before the mocked e5 model wasn't actually mocked but a small-ish e5 model. i have run our python tool to create a real mocked e5 model which i based on the `e5_small_v2` vocabulary. the old mocked e5 must have had a different base model/vocab, hence the changed values. in the end it doesn't matter for those tests as we mainly keep them for input/output size checks. but i checked that the new ids correspond to the correct words.

rustonaut (approved):
lgtm

PR #1068, opened 2023-08-15
ET-4896 allow using python nlp libraries

rustonaut:
- add `snippet-extractor` crate which internally uses python
- add support for using venv in test
- initialize python on startup 

As the official python documentation mainly points to pipenv under "managing packages" I used that for managing packages (it's internally using the python build-in pip and venv  modules).

**References:**

- story: [ET-4866]
- issue: [ET-4896]

[ET-4866]: https://xainag.atlassian.net/browse/ET-4866?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ET-4896]: https://xainag.atlassian.net/browse/ET-4896?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


test-utils/src/python.rs
```diff
@@ -0,0 +1,113 @@
...
+use anyhow::{bail, Error};
+use pyo3::{types::PyDict, Python};
+
+/// Initializes pyo3 and inject venv.
+///
+/// # Panics
+///
+/// As this is a test util it will panic on failure.
+pub fn initialize_python() {
+    pyo3::prepare_freethreaded_python();
```
rustonaut, line 32: This can be called any number of times.

test-utils/src/python.rs
```diff
@@ -0,0 +1,113 @@
...
+
+    static ONCE: Once = Once::new();
+    ONCE.call_once(|| {
+        Python::with_gil(inject_snippet_extractor_venv_activation).unwrap();
+    })
+}
+
+/// Inject the snippet-extractor venv into the given python instance.
+pub fn inject_snippet_extractor_venv_activation(py: Python<'_>) -> Result<(), Error> {
```
rustonaut, line 41: pyo3 naming convention is to name the `Python<'_>` handle `py`.

integration-tests/src/lib.rs
```diff
@@ -639,6 +641,11 @@ where
             toml! {
                 [ingestion.index_update]
                 method = "danger_wait_for_completion"
+
+                [snippet_extractor]
+                language = "english"
+                chunk_size = 10
+                hard_chunk_size_limit = 10
```
rustonaut, line 648: Very low _token_ length so that we can have short dump testing texts. Testing the exact algorithm can be done somewhere else.

web-api/src/embedding.rs
```diff
@@ -27,6 +29,15 @@ pub enum Config {
     Sagemaker(Sagemaker),
 }
 
+impl Config {
+    pub(crate) fn tokenizer_file(&self) -> PathBuf {
+        match self {
+            Config::Pipeline(pipeline) => pipeline.directory.relative().join("tokenizer.json"),
+            Config::Sagemaker(_) => unimplemented!(),
```
rustonaut, line 36: :wrench:  we must decide how to handle this

rustonaut:
Side note: The test fails because ET-4927 isn't done yet, which includes updating the docker image.

web-api/src/ingestion/routes.rs
```diff
@@ -224,7 +224,7 @@ impl UnvalidatedDocumentForIngestion {
             )
                 .into())
             }
-            (true, false) => PreprocessingStep::CuttersSplit,
+            (true, false) => PreprocessingStep::NltkSplitV1,
```
acrrd, line 227: Maybe we can have `PreprocessingStep::default_split()` to keep the information of which the default is near the preprocessing module.

snippet-extractor/Pipfile
```diff
@@ -0,0 +1,14 @@
+[[source]]
+url = "https://pypi.org/simple"
+verify_ssl = true
+name = "pypi"
+
+[packages]
+langchain = "*"
+transformers = "*"
+nltk = "*"
```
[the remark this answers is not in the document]
rustonaut: The versions are fixed in the lock file.

Also because python was very bad at versioning until very recently it was somewhat rare to ever do a new major release after 1.0. 

You also can't have two different versions of the same library which makes depending on concrete version much more tricky.

So we basically want the newest compatible version of the combinations of above's library.

Through I guess we should set the minimal version to the currently locked version.

snippet-extractor/tests/basic_functionality.rs
```diff
@@ -0,0 +1,113 @@
...
+    Ok(())
+}
+
+#[test]
+fn test_multi_threaded_snippet_extraction_works() -> Result<(), Error> {
+    initialize_python();
+
+    let extractor = SnippetExtractor::initialize(
+        &Default::default(),
+        Path::new("/home/user/projects/de2/assets/xaynia_v0002/tokenizer.json"),
```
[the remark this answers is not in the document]
rustonaut: Doesn't, I just forgot to update it to detect the workspace/asset dir.

PR #1070, opened 2023-08-21
change migration to not calculate hash

rustonaut:
This only works if the migration (and any follow up migrations) haven't been applied already due to it changing an existing migration. Outside of one demo tenant that seems to be the case.

Run `cargo clean -p xayn-web-api-db-ctrl` can be a good idea to preclude any cargo cache related issues.

### DIFF
--- a/web-api-db-ctrl/postgres/tenant/20230713173212_document.sql
+++ b/web-api-db-ctrl/postgres/tenant/20230713173212_document.sql
@@ -25,7 +25,9 @@ CREATE TABLE snippet (
 INSERT INTO snippet(document_id, sub_id, snippet, embedding)
     SELECT document_id, 0, snippet, embedding FROM document;
 
-ALTER TABLE document ALTER COLUMN snippet TYPE BYTEA USING sha256(snippet::bytea);
-ALTER TABLE document RENAME COLUMN snippet TO original_sha256;
-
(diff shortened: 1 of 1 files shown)


web-api-db-ctrl/postgres/tenant/20230713173212_document.sql
```diff
@@ -25,7 +25,9 @@ CREATE TABLE snippet (
...
     SELECT document_id, 0, snippet, embedding FROM document;
 
-ALTER TABLE document ALTER COLUMN snippet TYPE BYTEA USING sha256(snippet::bytea);
-ALTER TABLE document RENAME COLUMN snippet TO original_sha256;
-
-ALTER TABLE document DROP COLUMN embedding;
+ALTER TABLE document
+    -- 32 byte sha256 hash used to determine if the original changed since last ingestion
+    ADD COLUMN original_sha256 BYTEA NOT NULL
+        DEFAULT '\x0000000000000000000000000000000000000000000000000000000000000000'::bytea,
```
rustonaut, line 31: Since pg11 this doesn't write to every row, i.e. it's very cheap to do. By setting the default to the zero hash we don't need to change any other code, it will be as if any re-ingested document was changed. This works as we do not ever compare hashes between different stored documents, only stored and new documents.

acrrd (approved):

PR #1072, opened 2023-08-24
Upload file

acrrd:
Allow to upload a binary. We pass the binary through tika to extract the text.
By default, this is disabled at the system level. If used when disabled, an error that says that it is disabled is returned.
The text extractor configuration already provides a way to specify which extractor we are using to make it future-proof. At the moment, we don´t plan to use something else but it will make it easier to switch to something else. 

Some changes involve allowing the preprocess to return a bad request kind of error.

Also with updated spec.

### DIFF
--- a/integration-tests/src/lib.rs
+++ b/integration-tests/src/lib.rs
@@ -611,12 +611,12 @@ pub fn with_dev_options() -> Option<Table> {
     })
 }
 
-pub fn with_text_extractor_options(allowed_content_type: Vec<String>) -> Option<Table> {
+pub fn with_text_extractor_options(allowed_media_type: Vec<String>) -> Option<Table> {
     Some(toml! {
         [text_extractor]
         enabled = true
         extractor = "tika"
-        allowed_content_type = allowed_content_type
+        allowed_media_type = allowed_media_type
     })
 }
 
--- a/integration-tests/src/lib.rs
+++ b/integration-tests/src/lib.rs
@@ -611,6 +611,15 @@ pub fn with_dev_options() -> Option<Table> {
     })
 }
 
+pub fn with_text_extractor_options(allowed_content_type: Vec<String>) -> Option<Table> {
+    Some(toml! {
+        [text_extractor]
+        enabled = true
+        extractor = "tika"
+        allowed_content_type = allowed_content_type
+    })
+}
+
 pub fn extend_config(current: &mut Table, extension: Table) {
     for (key, value) in extension {
         if let Some(current) = current.get_mut(&key) {
@@ -715,6 +724,16 @@ pub fn build_test_config_from_parts_and_names(
 
                 [snippet_extractor.tokenizers]
                 default = tokenizer
+            },
+        );
+    }
+
+    if *RUNS_IN_CONTAINER {
+        extend_config(
+            &mut config,
+            toml! {
+                [text_extractor]
+                url = "http://tika:9998/"
             },
         );
     }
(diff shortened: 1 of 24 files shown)


web-api/src/error/common.rs
```diff
@@ -153,10 +153,31 @@ pub(crate) enum InvalidDocumentSnippet {
...
+
+impl_application_error!(FileUploadNotEnabled => BAD_REQUEST, INFO);
+
+/// Content-type of the uploaded file is not supported.
+#[derive(Debug, Error, Display, Serialize)]
+pub(crate) enum InvalidBinary {
+    /// Unrecognized content-type.
+    Unrecognized,
+    /// Unsupported content-type. Found {found}
+    ContentType { found: String },
```
rustonaut, line 174: :wrench:  As far as I can tell this is what "formerly was known" to be a mime-type. In which case naming it media type would be more appropriate.  

In http it's set with the `Content-Type` header because it's the media type of the content, instead of the media type of the body (which could have a content transfer encoding which has it's one media type).

And I guess http is  why tikas field is also called `Content-Type`.

<details>
<summary>Random fun fact about media types</summary>: The only reason media types where named mime types was because there was a different XML related protocol which also had media types and the authors of the MIME set of RFCs wanted to avoid confusion. But the other media type has lost any relevance so now people refer to it again as media type. But to top that off is that http didn't use the MIME specs it used their own spec which is _mosty but not exactly_ the MIME specs so if you want to nitpick http never had mime types :upside_down_face:  One of the few places where MIME and http differ is in how exactly they define media types :rofl:  Through there is somewhat of a "media type" spec by now which works forth both use cases and media types in files etc. 
</details>

web-api/compose.db.yml
```diff
@@ -34,15 +34,31 @@ services:
...
+  tika:
+    image: apache/tika:2.8.0.0-full
+    mem_limit: 256m
+    restart: always
+    ports:
+      - "9998:9998"
+    networks:
+      - internal
+    healthcheck:
+      test: wget --no-verbose --tries=1 --spider --fail 'http://localhost:9998/'
```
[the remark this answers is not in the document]
acrrd: Exactly. that's also why at the beginning the CI was silently failing.

web-api/src/extractor.rs
```diff
@@ -0,0 +1,239 @@
...
+            });
+        }
+
+        let inner = match &config.config {
+            ExtractorConfig::Tika {
+                url,
+                allowed_content_type,
+            } => {
+                let url = Url::parse(url)?;
+                if url.cannot_be_a_base() {
```
rustonaut, line 137: In `web-api-shared` there is `SegmentableUrl`  which is a `Url` newtype you could use here, through weather you want to use this is fully your choice I won't request a change for it.

integration-tests/src/lib.rs
```diff
@@ -719,6 +728,16 @@ pub fn build_test_config_from_parts_and_names(
...
 
+    if *RUNS_IN_CONTAINER {
+        extend_config(
+            &mut config,
+            toml! {
+                [text_extractor]
+                url = "http://tika:9998/"
+            },
+        );
+    }
```
rustonaut, line 739: This is probably what is causing the problems on CI.

For tests which do not use `with_text_extractor_options` this lead to there being a:

```
                [text_extractor]
                url = "http://tika:9998/"
```

config without any other fields

which fails to deserialize due to missing the enum tag

I'm not sure why deserialization errors in the config aren't displayed properly anymore (or maybe never where :thinking: ).

acrrd: Was that, thanks.

rustonaut (approved):
lgmt

PR #1073, opened 2023-08-30
ET-4896 take 2 use external python to extract snippets

rustonaut:
As we had problems with pyo3 and cross compilation and anyway can't statically embed python we instead now use an approach where we spawn a python child process and then use `stdin`/`stdout` to  communicate. 

For parallelism it was considered to use ptyhons `multiprocess` put this comes with additional complications, as we anyway have to have the ability to spawn and re-spawn a child we also can use a pool of child processes instead. 

There are 3 additional things we could do:

- use config relative path, problem we would need to make web-api/utils a crate
- allow configuring the pool.get timeout, currently it's fixed to a high arbitrary value of 15
- use a managed pool which would allow us to have timeouts on starting the child and
  the health check ping, through it also adds a bit of additional complexity

**References:**

- story: [ET-4866]
- issue: [ET-4896], [ET-4927]

[ET-4866]: https://xainag.atlassian.net/browse/ET-4866?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ET-4896]: https://xainag.atlassian.net/browse/ET-4896?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ET-4927]: https://xainag.atlassian.net/browse/ET-4927?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


snippet-extractor/Pipfile
```diff
@@ -0,0 +1,15 @@
+[[source]]
+url = "https://pypi.org/simple"
+verify_ssl = true
+name = "pypi"
+
+[packages]
+langchain = "*"
+transformers = ">=4.31.0, ==4.*"
+nltk = ">= 3.8.1, == 3.*"
+msgpack = ">= 1.0.1, == 1.*"
```
rustonaut, line 10: Hint: This doesn't use `~=` so that it has the same behavior compared to what we are used to from rust.

I.e. `~= 3.8.1` means `>= 3.8.1, == 3.8.*` which isn't quite what we want/

rustonaut: Also langchain currently doesn't do any versioning, i.e. with every update they bump a `0.0.xxx` version.

snippet-extractor/python_src/snippet_extractor.py
```diff
@@ -0,0 +1,148 @@
...
+    unpacker = Unpacker(stdin, max_buffer_size = 0)
+    packer = Packer()
+
+    stdout.write(packer.pack("ready"));
+    stdout.flush()
+
+    for msg in unpacker:
+        result = None
+        try:
+            tag = msg['tag']
```
rustonaut, line 126: We can make this more fancy getting better error messages if the format doesn't match, but then this is an internal interface with just 2 messages, so probably not worth it.

snippet-extractor/python_src/snippet_extractor.py
```diff
@@ -0,0 +1,148 @@
...
+            cmd = msg['cmd']
+            if tag == 'initialize_tokenizer':
+                tokenizers[cmd['name']] = PreTrainedTokenizerFast(tokenizer_file=cmd['path'])
+                result = ok(True)
+            elif tag == 'extract':
+                snippets = SnippetExtractor(
```
acrrd, line 132: This will read the model from the disk every time. On the other script we have, if I initialize it 100times it takes more than 1s.
We can use a cache like we are doing for the tokenizers so we take an hit only the first time.

snippet-extractor/src/python_child.rs
```diff
@@ -0,0 +1,125 @@
...
+}
+
+impl PythonChild {
+    pub(crate) fn spawn(
+        workspace: impl AsRef<Path>,
+        python_file: impl AsRef<Path>,
+        use_pipenv: bool,
+    ) -> Result<Self, io::Error> {
+        let mut cmd = if use_pipenv {
+            let mut cmd = Command::new("pipenv");
```
acrrd, line 43: Why not always use the env?

snippet-extractor/src/pool.rs
```diff
@@ -0,0 +1,69 @@
...
+}
+
+impl SnippetExtractorPool {
+    #[allow(clippy::missing_panics_doc)]
+    pub fn new(config: &Config) -> Result<Self, Error> {
+        let num_cpus = num_cpus::get();
+        let pool = Pool::from_config(&PoolConfig {
+            max_size: num_cpus,
+            // TODO[pmk/now] decide value based on whole request timeout and make configurable
+            timeout: Some(Duration::from_secs(15)),
```
acrrd, line 40: Can't we move this already to `Config`?

snippet-extractor/src/pool.rs
```diff
@@ -0,0 +1,69 @@
...
+#[derive(Clone)]
+pub struct SnippetExtractorPool {
+    // Hint: Internally `Arc`ed
+    pool: Pool<SnippetExtractor>,
+}
+
+impl SnippetExtractorPool {
+    #[allow(clippy::missing_panics_doc)]
+    pub fn new(config: &Config) -> Result<Self, Error> {
+        let num_cpus = num_cpus::get();
```
acrrd, line 36: This should also be part of the configuration.

web-api/src/error/common.rs
```diff
@@ -338,4 +339,32 @@ impl_from_std_error!(
...
+    fn status_code(&self) -> StatusCode {
+        match self {
+            PoolError::Timeout => StatusCode::SERVICE_UNAVAILABLE,
+            PoolError::Closed | PoolError::NoRuntimeSpecified => StatusCode::INTERNAL_SERVER_ERROR,
+        }
+    }
+
+    fn kind(&self) -> &str {
+        match self {
+            PoolError::Timeout => "ServiceOverloaded",
```
acrrd, line 355: I am not convinced about exposing "ServiceOverloaded"

snippet-extractor/python_src/snippet_extractor.py
rustonaut: it doesn't only the tokenizer reads any model from disk, which we initialize beforhand

snippet-extractor/src/python_child.rs
rustonaut: it's using the env

web-api/src/error/common.rs
rustonaut: can you explain why, an overload of the service is the only way you can normally run into this timeout  (it's a timeout  on retrieving a extractor)

rustonaut: As far as I remember we also did discuss exposing errors cause by elasticsearch being overloaded due to too many parallel ingestion as `503 Service Unavailable` in the future, which is a similar case.

snippet-extractor/python_src/snippet_extractor.py
rustonaut: I just realized you might not be speaking about the `punkt` model from the tokenizer.

But then which model?

The NLTK  split only creates a lambda, which might internally load some model, but if so does cache it internally AFIK as we don't create any state for it.

And the other splitters don't have any model.

snippet-extractor/src/pool.rs
rustonaut: Because I wanted to get feedback about weather we use the managed pool which will add additional timeouts, and without that (and the new `threads_per_cpu` config) there was only one pool config so I probably would have inlined it into `super::Config`.

It's now there as `config.pool.acquisition_config`.

snippet-extractor/src/python_child.rs
rustonaut: Through if you want we can change it to use `pipenv install --deploy` instead and then always use `pipenv run`.

rustonaut: I did change it that way.

snippet-extractor/src/pool.rs
acrrd: ok we can make it a factor later.

snippet-extractor/python_src/snippet_extractor.py
acrrd: I was talking about punkt. I got a super high time because I was partially retraining the model to better understand legal abbreviations. That is super slow. but we should put that in a pickle file and use that as an asset.

acrrd (approved):

snippet-extractor/python_src/snippet_extractor.py
rustonaut: Currently we do load the punk tokenizer only once per process and keep it around as PreTrainedTokenizer, the `NLTK` send_tokens function also seems to load something from punk but also caches it.

So AFIK there is nothing which will do any disk access on every request.

snippet-extractor/src/pool.rs
rustonaut: I already did so when adding the timeout pool config.

PR #1075, opened 2023-09-04
ET-4957 quick fix for es deletion bug

rustonaut:
filter get_personalized by "WHERE is_candidate"

this means that even if we fail to delete documents in elastic search (or there is some race condition) we filter out documents which are no longer candidates when enriching search hits with additional data  

this doesn't fully fix the bug, but reduces the negative consequences of it

### DIFF
--- a/web-api/src/storage/postgres.rs
+++ b/web-api/src/storage/postgres.rs
@@ -303,6 +303,14 @@ impl Database {
     ) -> Result<Vec<PersonalizedDocument>, Error> {
         let mut documents = Vec::with_capacity(scores.len());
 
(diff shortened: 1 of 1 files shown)


acrrd (approved):
Can you please add a comment in the code so we remember why it is there and to remove it later?

rustonaut:
I don't think we should remove it later:

- With ES being eventual consistent we always can have situations where ES returned "just" deleted documents.
- In some use-cases we have far more non candidates then candidates, in this case this can yield a performance boost due to the is_candidate partial index we have for other endpoints.

PR #1076, opened 2023-09-05
ET-4957 more test

rustonaut:
Replicate behavior which lead to some deletion bugs in test.


acrrd (approved):
Can you please add a comment with the issue number and say that we are trying to replicate that bug, but atm it is not possible.

PR #1077, opened 2023-09-05
ET-4891 use rrf everywhere

rustonaut:
**References:**

- issue: [ET-4891]

[ET-4891]: https://xainag.atlassian.net/browse/ET-4891?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/web-api/src/personalization/knn.rs
+++ b/web-api/src/personalization/knn.rs
@@ -24,6 +24,7 @@ use crate::{
     error::common::InternalError,
     models::{PersonalizedDocument, SnippetId},
     personalization::filter::Filter,
+    rank_merge::{rrf_score, DEFAULT_RRF_K},
     storage::{self, Exclusions, KnnSearchParams, SearchStrategy},
     Error,
 };
@@ -118,13 +119,12 @@ async fn merge_knn_searchs(
             Ok(documents) => {
                 // the same document can be returned with different elastic scores, hence the
                 // documents are deduplicated and only the highest score is retained for each
-                for document in documents {
+                for (idx, mut document) in documents.into_iter().enumerate() {
+                    document.score = rrf_score(DEFAULT_RRF_K, idx, 1.0);
                     all_documents
                         .entry(document.id.clone())
                         .and_modify(|PersonalizedDocument { score, .. }| {
-                            if *score < document.score {
-                                *score = document.score;
-                            }
+                            *score += document.score;
                         })
                         .or_insert(document);
                 }
--- a/web-api/src/personalization/knn.rs
+++ b/web-api/src/personalization/knn.rs
@@ -57,7 +57,11 @@ where
         let coi_weights = compute_coi_weights(interests.clone(), self.horizon, self.time);
         let cois = interests
             .zip(coi_weights)
-            .sorted_by(|(_, w1), (_, w2)| w1.total_cmp(w2).reverse())
+            .sorted_by(|(coi1, w1), (coi2, w2)| {
+                w1.total_cmp(w2)
+                    .then_with(|| coi1.id.cmp(&coi2.id))
+                    .reverse()
+            })
             .take(self.max_cois)
             .collect_vec();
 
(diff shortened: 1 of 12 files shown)


web-api/src/rank_merge.rs
```diff
@@ -0,0 +1,164 @@
+// Copyright 2023 Xayn AG
```
rustonaut, line 1: just moved

web-api/src/personalization/rerank.rs
```diff
@@ -73,35 +74,32 @@ fn rerank_by_tag_weight<'a>(
...
-        })
-        .collect_vec();
-    for (document, score) in documents.iter_mut().zip(scores) {
-        document.score = score;
+    let scores = rrf(
+        DEFAULT_RRF_K,
+        [
+            (score_weights[0], interest_scores),
+            (score_weights[1], tag_weight_scores),
+            (score_weights[2], search_scores),
```
rustonaut, line 94: :question:  I'm not sure we can use that weights just like that.

The default weights have been largely determined by the difference in the distributions of the different metricts, which with rrf doesn't matter.

rustonaut:
:wrench:  There is an issue with a test being flacky in the order of  personalized documents.

web-api/src/personalization/rerank.rs
acrrd: We should set everything to one. The weights here weren't thought to be used with rrf. We can remove them completely. Maybe in the future, we could enable another _dev option in case we want to experiment.

rustonaut: Be aware that we currently set the weight for tags for personalization to `0` and for semantic search to 1/2 the weight of the other weights.

Which was the main reason I didn't strip it our (or replace it).

rustonaut: I now set it to `1, 1, 0` and `1, 1, 0.5` which produces more similar results then what we had before and in turn requires less changes to tests.

web-api/src/rank_merge.rs
```diff
@@ -0,0 +1,170 @@
...
+pub fn rrf<K>(k: f32, scores: impl IntoIterator<Item = (f32, ScoreMap<K>)>) -> ScoreMap<K>
+where
+    K: Eq + Hash + Ord,
+{
+    let rrf_scores = scores.into_iter().flat_map(|(weight, scores)| {
+        scores
+            .into_iter()
+            // For testing we want to make sure that in case of s1 == s2 we still get a
+            // deterministic result, for this we use the key ordering for equal scores
+            .sorted_by(|(k1, s1), (k2, s2)| s1.total_cmp(s2).then_with(|| k1.cmp(k2)).reverse())
```
rustonaut, line 100: To get more deterministic results in tests I made sorting a bit more stable where appropriate.

web-api/src/personalization/knn.rs
```diff
@@ -118,13 +123,12 @@ async fn merge_knn_searchs(
             Ok(documents) => {
                 // the same document can be returned with different elastic scores, hence the
                 // documents are deduplicated and only the highest score is retained for each
-                for document in documents {
+                for (idx, mut document) in documents.into_iter().enumerate() {
+                    document.score = rrf_score(DEFAULT_RRF_K, idx, 1.0);
```
rustonaut, line 127: @acrrd  I realized this can also use rrf. But I'm not sure if I should switch it back to select the max score instead.

In general I realized that if we had time could could make sense to move the multiple cois search + merge part into the elastic code and apply it before fetching additional information from postgres.

web-api/src/personalization/rerank.rs
acrrd: ok let's go with that

acrrd (approved):

PR #1078, opened 2023-09-06
ET-4971 doc for split documents

rustonaut:
Update getting started documentation to reflect changes regarding split documents and documents with multiple snippets.

### DIFF
--- a/docs/source/getting_started/getting_started.md
+++ b/docs/source/getting_started/getting_started.md
@@ -4,7 +4,7 @@ Here we take a high-level look at how the system works. The API is divided into 
 
 The back office system can be used to ingest your content into the system. Every piece of content you ingest is stored as one document with one or more snippets.
 
-A snippet is a bit of text which the user might search for. In the simplest case you only have one snippet per document in which case the snippet is roughly equivalent to the document. If you have such a use case you can mentally replace snippet with document in most of the remaining documentation. Through as there are limits on
+A document is comprised of one or more snippets. Each snippet represents a piece of the content of the document. In the simplest case, you only have one snippet per document in such a case you can mentally replace snippet with document in most of the remaining documentation. Through as there are limits on
 how long a snippet can be, we provide functionality to either summarize the content or split the content of a document into multiple snippets. In many places both a document id and the id of a specific snippet can be used.
 
 During ingestion, the system creates a mathematical representation of each snippet which is used to match the snippets to the user's interests and searches.
@@ -128,7 +128,7 @@ curl -X PATCH "$URL/users/u1234/interactions" \
 ```
 
 ```{note}
-Please note that if an interaction between a user and a document is added, all snippets of a document will **not** be part of the documents returned for future calls to the personalised endpoint.
+Please note that if an interaction between a user and a document is added, the document will **not** be part of the documents returned for future calls to the personalised endpoint. This includes all snippets associated to that document```
 ```
 
 Let's ask for personalised documents again now:
@@ -167,7 +167,7 @@ In the request, we ask the system to include the properties of the returned docu
 We also have a `score` field which represents how well the documents match the user's interests. The higher the number, the better the documents match. It should be noted that the scores only have meaning in relation to other
 scores from the same requests.
 
-`snippet_id` is the id of the snippet which was found. The `document_id` in the `snippet_id` is the id of the document associated with the snippet. If you do not use ingestion options like `split` and in turn only have
+The field `snippet_id` identifies a specific snippet. The `document_id` in the `snippet_id` is the id of the document associated with the snippet. If you do not use ingestion options like `split` and in turn only have
 one snippet per document then you can always use `snippet_id.document_id` and ignore the rest. In many places
 both a the full snippet id object or a document id string can be used. If documents which have multiple snippets are ingested it's highly recommended to always use the full snippet id.
 
@@ -177,7 +177,7 @@ Depending on the use-case searching for documents can be achieved as a search fo
 
 ## Similar documents
 
-In this search variant either a _document id_ string or a _snippet id_ object must be provided to the [`/semantic_search`](https://docs.xayn.com/front_office.html#tag/front-office/operation/getSimilarDocuments) endpoint.
+In this search variant either a _document id_ or a _snippet id_ must be provided to the [`/semantic_search`](https://docs.xayn.com/front_office.html#tag/front-office/operation/getSimilarDocuments) endpoint.
 
 ```bash
 curl -X POST "$URL/semantic_search" \
@@ -189,13 +189,13 @@ curl -X POST "$URL/semantic_search" \
     }'
 ```
 
-The result contains a list of snippets that are similar to the identified snippet/document.
+The result contains a list of snippets that are similar to the identified snippet.
 
 If only documents with a single snippet are used you can provide only the document id
 like this: `"document": { "id": "xayn_cd5604c" },`.
 
-Be aware that until our implementation is further improved this is equivalent to doing
-a semantic search on the first snippet of a document.
+Be aware that at this moment, this is equivalent to doing a semantic search on the first snippet of a document.
+In the future, this behaviour could change to better represent the intent to search for something similar to the whole document. If you want to explicitly search for the first snippet use a _snippet id_.
 
 ## Free Text search
 
@@ -410,12 +410,7 @@ The candidates can facilitate fast transitions between different sets of documen
 
 We provide a functionality to extract multiple snippets from the provided content of a document.
 
-This is roughly equivalent to splitting the document into multiple parts.
+The system uses Natural language processing (NLP) algorithms to split the document into multiple parts.
 
-We try to have natural language boundaries between snippets as far as possible. For example while the
-algorithm tries to split at sentence boundaries it doesn't do so by naively splitting based on dots as
-they are also e.g. used with abbreviations. Instead it uses a natural language aware approach.
-
-Currently the splitting will only work reliable well with text of the language configured by Xayn for
-a specific tenant (i.e. a specific front office/back office instance pair). We are currently working on integrating language detection to make this more convenient usable and allow having documents of different
-languages in the same system.
+Currently, automatic splitting works only with one language set when the system is set up; by default, it is English. If you need another one, please contact us.
+We are working to add support for multiple languages to our text-splitting algorithms.
--- a/docs/source/getting_started/getting_started.md
+++ b/docs/source/getting_started/getting_started.md
@@ -2,19 +2,22 @@
 
 Here we take a high-level look at how the system works. The API is divided into two parts: the back office and the front office.
 
-The back office system can be used to ingest your content into the system. Every piece of content you ingest is managed as one document within the Xayn system.
+The back office system can be used to ingest your content into the system. Every piece of content you ingest is stored as one document with one or more snippets.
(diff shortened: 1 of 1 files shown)


docs/source/getting_started/getting_started.md
```diff
@@ -111,7 +128,7 @@ curl -X PATCH "$URL/users/u1234/interactions" \
 ```
 
 ```{note}
-Please note that if an interaction between a user and a document is added, the document will **not** be part of the documents returned for future calls to the personalised endpoint.
+Please note that if an interaction between a user and a document is added, all snippets of a document will **not** be part of the documents returned for future calls to the personalised endpoint.
```
rustonaut, line 131: :pushpin:  When updating this line I realized that semantic search is currently quite hard to use for anything which has documents with multiple snippets (except for a use case of refining existing search).

The problem is if you don't do semantic search based on previous search results you have no idea which snippet ids you want to use as the customer only knows the document ids and that there might be more then one snippet but not how many and which snippets there are :thinking:

rustonaut:
Like I have mentioned before I'm not fully happy with the terms snippet / `Personalized Document`.

Writing this documentation I think I now know what would be a better choice of language:

- use search targets for thinks which can be found
- use hits for things which have been found

Through idk. if we have time to change that.

It also would be a braking change, somewhat, depending on how much this feature is already used.

docs/source/getting_started/getting_started.md
```diff
@@ -2,19 +2,22 @@
 
 Here we take a high-level look at how the system works. The API is divided into two parts: the back office and the front office.
 
-The back office system can be used to ingest your content into the system. Every piece of content you ingest is managed as one document within the Xayn system.
+The back office system can be used to ingest your content into the system. Every piece of content you ingest is stored as one document with one or more snippets.
 
-During ingestion, the system creates a mathematical representation of the document which is used to match the document to the user's interests and searches.
+A snippet is a bit of text which the user might search for. In the simplest case you only have one snippet per document in which case the snippet is roughly equivalent to the document. If you have such a use case you can mentally replace snippet with document in most of the remaining documentation. Through as there are limits on
```
acrrd, line 7: > A snippet is a bit of text which the user might search for.
I think it's more clear to present snippets as pieces or the whole content of a document.

```suggestion
A document is comprised of one or more snippets. Each snippet represents a piece of the content of the document. In the simplest case, you only have one snippet per document in such a case you can mentally replace snippet with document in most of the remaining documentation. Through as there are limits on
```

docs/source/getting_started/getting_started.md
acrrd, line 131: I would keep the document in focus and then specify about the snippets.

```suggestion
Please note that if an interaction between a user and a document is added, the document will **not** be part of the documents returned for future calls to the personalised endpoint. This includes all snippets associated to that document```

File: docs/source/getting_started/getting_started.md [hunk 3]
```diff
@@ -131,7 +148,7 @@ As a result, we will get something like:
 {
   "documents": [
   {
-      "id": "xayn_5283ef3",
+      "snippet_id": { "document_id": "xayn_5283ef3", "sub_id": 0 },
```
acrrd, line 151: Does this happen even if you never use the split option? I think that for backward compatibility if split is never used this example should stay the same.
If this is the case would be better to specify this and show it with two different examples.

docs/source/getting_started/getting_started.md
```diff
@@ -147,27 +164,38 @@ As a result, we will get something like:
 
 In the request, we ask the system to include the properties of the returned documents. We can use this data to implement a 'more like this' section.
 
-We also have a 'score' field which represents how well the documents match the user's interests. The higher the number, the better the documents match.
+We also have a `score` field which represents how well the documents match the user's interests. The higher the number, the better the documents match. It should be noted that the scores only have meaning in relation to other
+scores from the same requests.
+
+`snippet_id` is the id of the snippet which was found. The `document_id` in the `snippet_id` is the id of the document associated with the snippet. If you do not use ingestion options like `split` and in turn only have
```
acrrd, line 170:
```suggestion
The field `snippet_id` identifies a specific snippet. The `document_id` in the `snippet_id` is the id of the document associated with the snippet. If you do not use ingestion options like `split` and in turn only have
```

acrrd: I would make `If you do not use ingestion options like split` the start of a new paragraph. This part is very important and it is easier that will be noted.

docs/source/getting_started/getting_started.md
acrrd: Ok, is specified underneath, but maybe we can find a way to specify both? 
In the ingestion example we never used "split = true".

docs/source/getting_started/getting_started.md
```diff
@@ -147,27 +164,38 @@ As a result, we will get something like:
...
 
 # Search
 
-Depending on the use-case searching for documents can be achieved as a search for documents _similar_ to a given document or as a _free-text search_. Both variants can then be run as a anonymous search or a search that is personalized. Personalization comes in two fashions, with a _user-id_ or by providing a interaction _history_.
+Depending on the use-case searching for documents can be achieved as a search for documents _similar_ to a given snippet/document or as a _free-text search_. Both variants can then be run as a anonymous search or a search that is personalized. Personalization comes in two fashions, with a _user-id_ or by providing a interaction _history_.
 
 ## Similar documents
 
-In this search variant only a _document id_ must be provided to the [`/semantic_search`](https://docs.xayn.com/front_office.html#tag/front-office/operation/getSimilarDocuments) endpoint.
+In this search variant either a _document id_ string or a _snippet id_ object must be provided to the [`/semantic_search`](https://docs.xayn.com/front_office.html#tag/front-office/operation/getSimilarDocuments) endpoint.
```
acrrd, line 180: We already showed what a document and snippet ids are. I wouldn´t specify that they are string and objects.
```suggestion
In this search variant either a _document id_ or a _snippet id_ must be provided to the [`/semantic_search`](https://docs.xayn.com/front_office.html#tag/front-office/operation/getSimilarDocuments) endpoint.
```

docs/source/getting_started/getting_started.md
```diff
@@ -147,27 +164,38 @@ As a result, we will get something like:
...
     --header "Content-Type: application/json" \
     --data '{
-        "document": { "id": "xayn_cd5604c" },
+        "document": { "id": { "document_id": "xayn_cd5604c", "sub_id": 2 } },
         "include_properties": true
     }'
 ```
 
-The result contains a list of documents that are similar to the provided document (identified by its _document id_).
+The result contains a list of snippets that are similar to the identified snippet/document.
```
acrrd, line 192: We pass a snippet id
```suggestion
The result contains a list of snippets that are similar to the identified snippet.
```

docs/source/getting_started/getting_started.md
```diff
@@ -147,27 +164,38 @@ As a result, we will get something like:
...
+
+If only documents with a single snippet are used you can provide only the document id
+like this: `"document": { "id": "xayn_cd5604c" },`.
+
+Be aware that until our implementation is further improved this is equivalent to doing
+a semantic search on the first snippet of a document.
```
acrrd, line 198:
```suggestion
Be aware that at this moment, this is equivalent to doing a semantic search on the first snippet of a document.
In the future, this behaviour could change to better represent the intent to search for something similar to the whole document. If you want to explicitly search for the first snippet use a _snippet id_.
```

docs/source/getting_started/getting_started.md
```diff
@@ -371,7 +401,21 @@ curl -X PUT "$URL/documents/candidates" \
 ```
 
 ```{note}
-Please note, that setting candidates can only be undone by sending the complete list of all ingested document-ids again. 
+Please note, that setting candidates can only be undone by sending the complete list of all ingested document-ids again.
 ```
 
 The candidates can facilitate fast transitions between different sets of documents without compromising the users' centers of interest (COIs) with which they were engaging. One practical scenario is handling outdated news articles that should not reappear in the recommendations. However, the past user interactions with those outdated articles should still influence the suggested documents.
+
+# Split Documents
```
acrrd, line 409: I would move this to a subsection of the ingestion at the beginning.

docs/source/getting_started/getting_started.md
```diff
@@ -371,7 +401,21 @@ curl -X PUT "$URL/documents/candidates" \
...
+
+We provide a functionality to extract multiple snippets from the provided content of a document.
+
+This is roughly equivalent to splitting the document into multiple parts.
+
+We try to have natural language boundaries between snippets as far as possible. For example while the
+algorithm tries to split at sentence boundaries it doesn't do so by naively splitting based on dots as
+they are also e.g. used with abbreviations. Instead it uses a natural language aware approach.
```
acrrd, line 417: Too many implementation details.

```suggestion
The system uses Natural language processing (NLP) algorithms to split the document into multiple parts.
```

docs/source/getting_started/getting_started.md
```diff
@@ -371,7 +401,21 @@ curl -X PUT "$URL/documents/candidates" \
...
+
+Currently the splitting will only work reliable well with text of the language configured by Xayn for
+a specific tenant (i.e. a specific front office/back office instance pair). We are currently working on integrating language detection to make this more convenient usable and allow having documents of different
+languages in the same system.
```
acrrd, line 421:
```suggestion
Currently, automatic splitting works only with one language set when the system is set up; by default, it is English. If you need another one, please contact us.
We are working to add support for multiple languages to our text-splitting algorithms.
```

docs/source/getting_started/getting_started.md
rustonaut: The `snippet_id` is  always returned the system doesn't know (or care) if `split` was used or not.

For backward compatibility we still return the previous `id` but it's deprecated in the OpenApi spec, so I didn't include it in the example.

acrrd: ok thanks.

acrrd:
> Like I have mentioned before I'm not fully happy with the terms snippet / `Personalized Document`.
I also think we should find a way to avoid to say every time "snippet/document". We can do it now or in another round.

> It also would be a braking change, somewhat, depending on how many `snippet_id` fields are already used.
I think `snippet_id` is fine. it represents what it is.

acrrd (approved):

docs/source/getting_started/getting_started.md
```diff
@@ -63,14 +66,24 @@ Each document has a unique identifier that can be used to refer to it in the sys
...
 The data that can be included in the properties is limited in terms of type and size. We support numbers, strings, boolean, date and list of strings, none of which are nullable. Please see [createDocuments](https://docs.xayn.com/back_office.html#tag/documents/operation/createDocuments) for more information on properties.
 
 This example assumes that we will eventually display the returned documents as a 'for-you' section, where we want to display an article's image, title, text preview, and a link (for click-through), so we have included these specific properties during ingestion.
 
+## Split Documents
+
+We provide a functionality to extract multiple snippets from the provided content of a document.
+
+The system uses Natural language processing (NLP) algorithms to split the document into multiple parts.
+
```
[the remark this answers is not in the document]
rustonaut: :pushpin: https://github.com/xaynetwork/xayn_discovery_engine/pull/1080

placed it into follow up PR

PR #1083, opened 2023-09-12
ET-4999 make search query size configurable

rustonaut:
- updated `string_wrapper!` to allow not having a hard coded len bound 
- now we can use it again for `DocumentSnippet`, so we do so
- then change `DocumentQuery` to not use a hard coded bound

**References:**

- issue: [ET-4999]

[ET-4999]: https://xainag.atlassian.net/browse/ET-4999?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/web-api/src/error/common.rs
+++ b/web-api/src/error/common.rs
@@ -15,12 +15,13 @@
 use std::{
     borrow::Cow,
     fmt::{Debug, Display},
+    ops::{Bound, RangeBounds},
 };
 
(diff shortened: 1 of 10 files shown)


Robert-Steiner (approved):
lgtm

PR #1086, opened 2023-09-18
parallel polling in hybrid search

rustonaut:
We just realized that we never fixed the "FIXME parallelize polling".

This will poll knn/bm25 in hybrid search.

Through this could affect load characteristics on ES in unexpected ways.

### DIFF
--- a/web-api/src/storage/elastic.rs
+++ b/web-api/src/storage/elastic.rs
@@ -145,7 +145,6 @@ impl Client {
             }),
             generic_parameters,
         ]);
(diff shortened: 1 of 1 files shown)


acrrd (approved):

acrrd:
@Robert-Steiner Can we deploy this and check if the latency of hybrid search decreases?

Robert-Steiner:
> @Robert-Steiner Can we deploy this and check if the latency of hybrid search decreases?

Yes we can do a test

PR #1087, opened 2023-09-19
Incorporate elser hybrid search

krajewskiML:
Creates additional search option which utilizes ELSER as query expansion instead of BM25


rustonaut:
stale

PR #1096, opened 2023-10-02
Use project local tooling/dependencies where possible.

rustonaut:
- switch doc generation to use `pipenv`
- set `PIPENV_VENV_IN_PROJECT` to have in project storage of the venv (without this it would store them out of band but that has the drawback that removing a project from the disk won't remove it's venv)
- install  tooling to `/.local` instead of system/user wide
  - for npm this sets the  `npm_config_prefix` env variable which is the same mechanism  you also use to setup per-user installations
  - currently we don't do so for cargo-sort and flamegraph as cargo doesn't support it
- use `npx` to call `npm` installed tools to avoid any potential problems
- provide proxy commands for cargo, npm,npx, pipenv which call the given programs through just to make sure we call them with exactly the same environment other just commands would


Robert-Steiner (approved):
LGTM

PR #1097, opened 2023-10-05
ET-5059 rename personalize to recommendations

rustonaut:
Fully rename personalized_documents to recommendations in documentation (rust code still uses `PersonalizedDocuments` ans similar  in some cases internally).

This now clearly separates the feature of finding recommendation and personalizing search. It also avoids issues like `PersonalizedDocument` not being a personalized document as we do not personalize the document. It's just part of a group of documents which order/choice of inclusion has been personalized. (It's also strictly speaking not a document but a snippet id and the scoring of the search for that snippet and optionally the properties of the document the snippet is from).

`/users/{user_id}/personalized_documents` => `/users/{user_id}/recommendations`

### DIFF
--- a/web-api/openapi/front_office.yaml
+++ b/web-api/openapi/front_office.yaml
@@ -58,7 +58,7 @@ security:
   - ApiKeyAuth: []
 
 paths:
-  /users/{user_id}/personalized_documents:
+  /users/{user_id}/recommendations:
     parameters:
       - $ref: './parameters/path/id.yml#/UserId'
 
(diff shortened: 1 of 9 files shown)


web-api/openapi/front_office.yaml
```diff
@@ -48,32 +48,83 @@ tags:
...
 
     post:
       tags:
         - front office
-        - search
+        - recommendation
+      deprecated: true
       summary: Personalize documents for the user
       description: |-
+        Use `/users/{user_id}/recommendations` instead.
```
acrrd, line 126:
```suggestion
        This endpoint has been renamed as `/users/{user_id}/recommendations`, use it instead.
```
I would delete everything else in the description

acrrd (approved):

PR #1099, opened 2023-10-11
Bubbling raw scores from elastic up to the response

acrrd:
Show raw scores from Elastic if a dev option is passed.

Passing the dev option is at the moment missing.

### DIFF
--- a/web-api/src/models.rs
+++ b/web-api/src/models.rs
@@ -452,6 +452,23 @@ pub(crate) struct PersonalizedDocument {
 
     /// The tags associated to the document.
     pub(crate) tags: DocumentTags,
+
+    /// Additional data about the document that can be helpful while tuning or debugging the system.
+    pub(crate) dev: Option<DocumentDevData>,
+}
+
+#[derive(Clone, Debug, Default, Serialize)]
+pub(crate) struct DocumentDevData {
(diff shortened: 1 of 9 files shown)


web-api/src/models.rs
```diff
@@ -452,6 +452,23 @@ pub(crate) struct PersonalizedDocument {
 
     /// The tags associated to the document.
     pub(crate) tags: DocumentTags,
+
+    /// Additional data about the document that can be helpful while tuning or debugging the system.
+    pub(crate) dev: Option<DocumentDevData>,
```
rustonaut, line 457: I think we can remove this indirection.

If we ever decide to make it a stable non dev option this would add additional friction.

On the other hand a "omitted if not set" `raw_scores` field should always be fine.

rustonaut (approved):
lgtm, through there is one small interface thing I would change

web-api/src/models.rs
rustonaut: through then if we make it stable we probably add it to the document representation directly :thinking: 

so lets keep it this way

rustonaut: through then even for dev options having this on the returned documents would be nice :thinking:

PR #1102, opened 2023-10-19
Update single machine deploy

acrrd:
Fix some ports.
Move back end front office to different ports.
Enable tika.
Update image

### DIFF
--- a/web-api/compose.single-machine-deploy.yml
+++ b/web-api/compose.single-machine-deploy.yml
@@ -10,8 +10,6 @@ services:
       POSTGRES_DB: "xayn"
     volumes:
       - postgres_data:/var/lib/postgresql/data
-    ports:
-      - "5432"
     networks:
       - internal
     healthcheck:
@@ -30,8 +28,6 @@ services:
       - xpack.security.enabled=false
     volumes:
       - elasticsearch_data:/usr/share/elasticsearch/data
-    ports:
-      - "9200"
     networks:
       - internal
     healthcheck:
@@ -95,8 +91,6 @@ services:
     image: apache/tika:2.8.0.0-full 
     mem_limit: 512m 
     restart: always 
-    ports: 
-      - "9998" 
     networks: 
       - internal 
     healthcheck: 
@@ -105,16 +99,6 @@ services:
       timeout: 5s
       retries: 5
 
-  web:
-    image: nginx:latest
-    mem_limit: 256m
-    restart: always
-    volumes:
-      - ./demo:/usr/share/nginx/html/ 
-      # The web interface is optional, but if it is needed,
-      # make sure the demo folder contains the compiled demo web app
-    ports:
-      - "8000:8000"
 
   db_ready:
     image: debian:bookworm-slim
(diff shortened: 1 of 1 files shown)


web-api/compose.single-machine-deploy.yml
```diff
@@ -45,47 +41,50 @@ services:
...
     environment:
       XAYN_WEB_API__STORAGE__POSTGRES__BASE_URL:  "postgres://${POSTGRES_USER:-user}:${POSTGRES_PASSWORD:-pw}@postgres:5432/xayn"
       XAYN_WEB_API__STORAGE__ELASTIC__URL: "http://elasticsearch:9200/"
-      XAYN_WEB_API__NET__BIND_TO: "0.0.0.0:8000"
+      XAYN_WEB_API__NET__BIND_TO: "0.0.0.0:8081"
       XAYN_WEB_API__STORAGE__POSTGRES__USER: "${POSTGRES_USER:-user}"
       XAYN_WEB_API__STORAGE__POSTGRES__PASSWORD: "${POSTGRES_PASSWORD:-pw}"
       XAYN_WEB_API__STORAGE__POSTGRES__APPLICATION_NAME: "xayn"
-      XAYN_WEB_API__EMBEDDING__TOKEN_SIZE: "384"
+      XAYN_WEB_API__EMBEDDING__TOKEN_SIZE: "512"
```
acrrd, line 55: Token size is the maximum amount of tokens that we want to use to compute the embedding.

web-api/compose.single-machine-deploy.yml
```diff
@@ -21,10 +21,6 @@ services:
       retries: 5
 
   elasticsearch:
-    # Don't forget to apply mapping.json to elasticsearch
-    # See example mapping in web-api-db-ctrl
-    # Make sure vector dimensions match model and run something like
-    # curl -XPUT 'http://localhost:3092/xayn/_mapping' -H 'Content-Type: application/json' -d @mapping.json
```
acrrd, line 27: The code already does this.

web-api/compose.single-machine-deploy.yml
```diff
@@ -45,47 +41,50 @@ services:
...
-    image: xaynetci/xayn_discovery_ingestion_service:main-230831084031-f320e8c-multilingual_e5_small_v0001-amd64
+    image: xaynetci/xayn_discovery_ingestion_service:main-231019104503-d06d92c-e5_small_v2_v0001-amd64
     mem_limit: 1.5g
     cpus: 2
     restart: always
     environment:
       XAYN_WEB_API__STORAGE__POSTGRES__BASE_URL:  "postgres://${POSTGRES_USER:-user}:${POSTGRES_PASSWORD:-pw}@postgres:5432/xayn"
       XAYN_WEB_API__STORAGE__ELASTIC__URL: "http://elasticsearch:9200/"
-      XAYN_WEB_API__NET__BIND_TO: "0.0.0.0:8000"
+      XAYN_WEB_API__NET__BIND_TO: "0.0.0.0:8081"
```
acrrd, line 51: We could avoid to change the port and expose it to another mapping but I think it's easier to read and less to misuse if they are the same

rustonaut: The docker network internal port is somewhat irrelevant so we can use whatever is the most convenient.

rustonaut (approved):
I have a few considerations about ports:

- should we even expose ports of internal only services (postgres, tika)? 

- by removing `HOST_PORT_SCOPE` we remove the ability to avoid port binding collisions (which was an issue I had ran into before hence why it exists), if we do so maybe we shouldn't use very common ports like `808x` and similar

acrrd:
> * should we even expose ports of internal only services (postgres, tika)?
removed

> * by removing `HOST_PORT_SCOPE` we remove the ability to avoid port binding collisions (which was an issue I had ran into before hence why it exists), if we do so maybe we shouldn't use very common ports like `808x` and similar

It is expected that this is the only thing that runs on that machine. If ports need to be changed, it can be done on a per-case basis locally.

HOST_PORT_SCOPE is a pain in the ass, honestly. Once you run `just web-dev-up` I expect to be able to seamlessly run an instance that I build locally but I always need to configure the ports. That's a huge waste of time. Collision for postgres and elastic almost never happens. The only case is when we connect to a remote system. It is very rare, the authorization will forbid us to mess with it. The fix is easy and is not worth complicating our day-to-day workflow.

petricevich (approved):

PR #1103, opened 2023-10-19
Add stateless recommendations

joecks:


acrrd (commented):
I would avoid using stateless in the names. Also, it is already possible to use the user id instead in this implementation.

web-api/src/personalization/routes.rs
```diff
@@ -820,6 +874,103 @@ async fn semantic_search(
...
+    rerank(
+        &state.coi,
+        &mut documents,
+        &interests,
+        &tag_weights,
+        state.config.semantic_search.score_weights,
+        time,
+    );
+
+    #[cfg_attr(not(test), allow(irrefutable_let_patterns))]
```
rustonaut, line 960: This shouldn't be needed.

rustonaut, line 960: This shouldn't be needed.

```suggestion
```

joecks: done

web-api/compose.db.yml
```diff
@@ -36,7 +36,7 @@ services:
 
   tika:
     image: apache/tika:2.8.0.0-full
-    mem_limit: 256m
+    mem_limit: 512m
```
rustonaut, line 39: This seems somewhat unrelated to this PR, I'm not sure if this mem limit is just for convenience when locally developing or could have some importance in some context :thinking: 

I guess it should be fine.

web-api/openapi/front_office.yaml
```diff
@@ -280,6 +280,49 @@ paths:
...
           $ref: './responses/generic.yml#/BadRequest'
 
+  /recommendations:
+    post:
+      tags:
+        - front office
+        - recommendation
+      summary: Recommendations for a given user_id or a history
+      description: |-
+        Finds a number of recommendations for the given user_id or a history.
```
rustonaut, line 290:
```suggestion
        Finds a number of recommendations for the given `user_id` or a history.
```

web-api/openapi/front_office.yaml
```diff
@@ -280,6 +280,49 @@ paths:
...
+
+        Recommendations are based on snippets and each recommendation contains the snippet id it is based on as well as a score. A higher score means that the document matches the preferences of the user or the history better. Scores can be compared only with other scores that belong to the same request; comparing scores of documents that have been obtained through different requests can lead to unexpected results.
+
+        Depending on the request parameters the recommendation can also include additional fields.
+
+        The recommendations also contain the properties of the snippets parent document if this is requested and the properties are not empty.
+
+        By default documents that have been interacted with by the user or are listed in the history are filtered out from the result. This behavior can be changed with `exclude_seen`.
+
+        Note that you can request personalized documents for a specific user_id, only after that same user_id has made enough interactions via our system.
```
rustonaut, line 300:
```suggestion
        Note that you can request personalized documents for a specific `user_id`, only after that same `user_id` has made enough interactions via our system.
```

web-api/src/personalization/routes.rs
```diff
@@ -268,35 +273,7 @@ async fn recommendations(
...
-        Either::Right((
-            Json(PersonalizedDocumentsError::NotEnoughInteractions),
-            StatusCode::CONFLICT,
-        ))
-    };
-
-    Ok(deprecate!(if is_deprecated {
-        documents
-    }))
+    _recommendations(state, request, storage).await
```
rustonaut, line 276: :art: Leading `_` has a special meaning in rust it means: "don't warn if this thing is not used".

Note that this isn't the same as dead code. I.e. it's mainly used for variables where you assign a value which has some side-effect when being dropped but doesn't need to be accessed anymore (e.g. release lock on end of scope). 

So it's better to rename this to e.g.  `recommendations_inner`, `recommendations_` or similar.

web-api/src/personalization/routes.rs
```diff
@@ -707,6 +744,17 @@ struct SemanticSearchRequest {
     is_deprecated: bool,
 }
 
+#[allow(clippy::too_many_arguments)]
+#[allow(clippy::struct_excessive_bools)]
```
rustonaut, line 748:
```suggestion
```

Both of this seem to not necessary (when I removed them on a local checkout, both with the tooling we use and nightly clippy).

web-api/tests/recommendations.rs
```diff
@@ -122,311 +84,103 @@ macro_rules! assert_order {
         );
         for documents in $documents.windows(2) {
             let [d1, d2] = documents else { unreachable!() };
-            assert!(d1.score >= d2.score, $($arg)*);
-            assert!(d2.score >= 0., $($arg)*);
+            assert!(1. >= d1.score, $($arg)*);
+            assert!(d1.score > d2.score, $($arg)*);
+            assert!(d2.score >= -1., $($arg)*);
```
rustonaut, line 89: :wrench:  This is wrong.

We don't give any guarantees about the range scores are in.

Through we currently should only have scores in range [0;1],
but that could change but we kinda don't want to return scores < 0.

(Similarities tend to be in range [-1;1] but we return scores not similarities.)

web-api/tests/recommendations.rs
```diff
@@ -89,29 +61,19 @@ async fn interact(client: &Client, personalization_url: &Url) -> Result<(), Erro
...
 }
 
-#[derive(Debug, Deserialize, PartialEq)]
-enum RecommendationsResponse {
-    #[serde(rename = "documents")]
-    Documents(Vec<PersonalizedDocumentData>),
-    #[serde(rename = "kind")]
-    Error(PersonalizedDocumentsError),
+#[derive(Debug, Deserialize)]
+struct SemanticSearchResponse {
```
rustonaut, line 71: :wrench:  We get recommendation responses here not semantic search responses.

(Which have the same format, but the name is misleading.)

web-api/tests/recommendations.rs
```diff
@@ -122,311 +84,103 @@ macro_rules! assert_order {
...
+fn test_full_personalization_with_inline_history() {
     test_two_apps::<Ingestion, Personalization, _>(
         UNCHANGED_CONFIG,
-        UNCHANGED_CONFIG,
-        |client, ingestion_url, personalization_url, _| async move {
-            ingest_with_dates(&client, &ingestion_url).await?;
-            personalize(
+        Some(toml! {
+            [semantic_search]
+            score_weights = [0.5, 0.5, 0.]
```
rustonaut, line 100: Out of interest why did you set the weight of the search score to 0?

:wrench:  Additionally it should be:

```suggestion
            [personalization]
            score_weights = [0.5, 0.5, 0.]
```

web-api/src/personalization/routes.rs
```diff
@@ -820,6 +868,102 @@ async fn semantic_search(
...
+    }
+    .run_on(&storage)
+    .await?;
+
+    rerank(
+        &state.coi,
+        &mut documents,
+        &interests,
+        &tag_weights,
+        state.config.semantic_search.score_weights,
```
rustonaut, line 950: :wrench:  This is using the wrong config option.

```suggestion
        state.config.personalization.score_weights,
```

rustonaut (commented):
mostly just minor things

web-api/compose.db.yml
joecks: Tika wouldn't start up on a mac, so you have to run the tests at least twice. Increasing the limit helped.

web-api/openapi/front_office.yaml
joecks: done

web-api/openapi/front_office.yaml
joecks: done

web-api/src/personalization/routes.rs
joecks: done

web-api/src/personalization/routes.rs
joecks: done

web-api/tests/recommendations.rs
joecks: I simply copied the code from the previous recommendations test, that now is called user_recommendations

joecks: I removed that check

web-api/tests/recommendations.rs
joecks: renamed

web-api/tests/recommendations.rs
joecks: again I just copied this from the other test, didn't think too much about it.

web-api/src/personalization/routes.rs
joecks: done

web-api/tests/recommendations.rs
rustonaut: The diff says you replaced:

```rust
 assert!(d1.score >= d2.score, $($arg)*);
 assert!(d2.score >= 0., $($arg)*);
```

with

```rust
assert!(1. >= d1.score, $($arg)*);
assert!(d1.score > d2.score, $($arg)*);
assert!(d2.score >= -1, $($arg)*);
```

but the diff seems also to be _very_ confused :thinking: 

I think you might have copied from a outdated pice of code which was already fixed upstream.


```rust
            // While score can be any arbitrary value it _currently_ should happen to be in [0;1]
            assert!(1. >= d1.score, $($arg)*);
            assert!(d1.score >= d2.score, $($arg)*);
            assert!(d2.score >= 0., $($arg)*);
```

(hit: this doesn't just change the bound from -1 to 0 but also the `>` to a `>=` as both scores could very well be the same)

rustonaut (approved):
lgtm (through there is one change left which I think would make sense as part of this PR)

PR #1104, opened 2023-10-24
Fix stuck tests

acrrd:
During the drop of `Session` the `Environment` is dropped before the session, this causes the random block of the test.

This has been fixed upstream https://github.com/pykeio/ort/pull/105 but it has not been ported to the `1.16` branch nor it has been released yet.


rustonaut (approved):
lgtm

PR #1105, opened 2023-10-25
justfile misc improvements

rustonaut:


justfile
```diff
@@ -250,16 +253,32 @@ install-openapi-doc-generator:
     npm install -g \
       @redocly/cli@${REDOCLY_CLI_VERSION}
 
+install-doc-generator:
+    #!/usr/bin/env -S bash -eux -o pipefail
+    cd docs
+    pipenv install --deploy
```
rustonaut, line 259: Deploy isn't needed but we also `cargo ... --locked` in some places and that's roughly the same.

joecks (approved):

PR #1106, opened 2023-10-31
Fix service name in docker compose

Robert-Steiner:
In #1072 the service name `db_ready` has been renamed to `services_ready`. This pr fixes the references to the old service name.


rustonaut (commented):
Lgtm. maybe we should rename it `compose.single-machine-deploy.yml` to fit.

rustonaut (approved):

rustonaut (approved):
lgtm

PR #1107, opened 2023-11-14
ET-4885 allow to prefix snippets

rustonaut:
Allows specifying two prefixes:

- `embedding.prefix.query` prefix all queries with this before computing their embedding
- `embedding.prefix.snippet` prefix all snippets with this before computing their embedding

works for both local and sagemaker setups

Env variables:

- `XAYN_WEB_API__EMBEDDING__PREFIX__QUERY`
- `XAYN_WEB_API__EMBEDDING__PREFIX__SNIPPET`

### DIFF
--- a/web-api/src/embedding.rs
+++ b/web-api/src/embedding.rs
@@ -12,8 +12,6 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-use std::borrow::Cow;
-
 use aws_config::retry::RetryConfig;
 use aws_sdk_sagemakerruntime::{config::Region, primitives::Blob, Client};
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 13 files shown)


web-api/src/embedding.rs
```diff
@@ -128,19 +161,39 @@ impl Embedder {
...
+            (
+                EmbeddingKind::Content,
+                Prefix {
+                    snippet: content, ..
+                },
+            ) => content,
+        };
+
+        let sequence = if let Some(prefix) = prefix {
+            Cow::Owned(format!("{prefix} {sequence}"))
```
rustonaut, line 180: Is the concatenation using a space okay?

web-api/src/embedding.rs
```diff
@@ -33,6 +35,16 @@ impl Default for Config {
     }
 }
 
+#[derive(Debug, Clone, Default, Deserialize, Serialize)]
+#[serde(default)]
+#[cfg_attr(test, serde(deny_unknown_fields))]
+pub(crate) struct Prefix {
+    /// Prefix prepended to search queries when embedding them.
+    pub(crate) query: Option<String>,
```
acrrd, line 43: I am not sure if we care to distinguish between prefix not given and an empty prefix. The end result will be the same.

acrrd (approved):

web-api/src/embedding.rs
rustonaut: removed

web-api/src/embedding.rs
rustonaut: If you by default separate the prefix with a white-space from the rest it makes a difference.

Through we removed the default space separator so now it's fine.

PR #1109, opened 2023-11-20
Add support for openai

acrrd:
Add support to use OpenAi embedding models. This allows us to use openai and azure services.

This also fixes the prefix for sagemaker.


joecks (approved):
Honestly, can not see any issue.

PR #1113, opened 2023-11-30
remove cyclic crate dependencies

rustonaut:
- originally integration tests where in  `integrations-tests/tests` to avoid cyclic dependencies between crates

- but you can have such cyclic dependencies as the build graph was `web-api(lib) -> integration-tests(lib) -> web-api(dev/tests)`, so we moved them to `web-api/tests`

- but while it works in general it did repeatedly cause issues, like triggering bugs in cargo or leading to degraded functionality with rust-analyzer so I moved it back to `integration-tests`


rustonaut:
I cam to the conclusion this should only be merged after the refactoring changes.

acrrd:
Sorry, I saw the diff and I thought it was only about the dependency. I manged to ignore all the the files that are moved :see_no_evil: 
I don't see much conflicts with the work in the other PR. @x3ro what do you think about the order of merging this?

x3ro:
Let's merge this one first 🚀

acrrd (approved):

PR #1115, opened 2023-11-30
ET-5228 Unify Rust API applications in code, to make one binary

x3ro:


web-api/src/ingestion.rs
```diff
@@ -138,7 +64,7 @@ impl IngestionConfig {
 }
 
 #[derive(AsRef)]
-pub struct Extension {
+pub(crate) struct Extension {
     snippet_extractor: SnippetExtractorPool,
 }
```
acrrd, line 69: Shouldn't we be able to remove this?

acrrd (commented):
How do we know that the current configurations will work with these changes?

acrrd:
> > How do we know that the current configurations will work with these changes?
> 
> I don't think they do at the moment. This is something I wanted to tackle once I got the docker image working and can test it locally.

Ok but then I would wait to merge this till we know that it works. You can base the other PRs on this one.

acrrd (commented):
I think we can have few more cleaning up in the test for the unification.

PR #1116, opened 2023-12-04
ET-5231 Update image generation for unified WebApi

x3ro:


web-api/compose.single-machine-deploy.yml
```diff
@@ -36,8 +36,8 @@ services:
       timeout: 5s
       retries: 5
 
-  backoffice:
-    image: xaynetci/xayn_discovery_ingestion_service:main-231019104503-d06d92c-e5_small_v2_v0001-amd64
+  webapi:
+    image: xaynetci/xayn_discovery_web_service:ET-5231-update-image-generation-231204120506-693e23c-xaynia_v0201-arm64
```
x3ro, line 40: This needs to be updated once the branch is merged, because otherwise the image name will contain the branch name and not `main`

PR #1122, opened 2023-12-07
fix config test

rustonaut:
The config test had been broken:

- they are ignored/skipped if the binary they run against doesn't exist anymore, which is the case since we merged the front and back office).
- they where out of sync
- if they are placed under `integration-tests/test` `cargo test` on the workspace might decide it doesn't need to build the `web-api` binaries and in that case you might end up running against an older binary, potentially from a different branch, so I moved them
-  tests didn't validate the config, nor did we do so when calling `--print-config` which could lead to confusing results

this PR fixes all of that

While doing so it removes some now unneeded generic args and both enforces that `.validate` is called and makes sure that logging can be setup before validate is called so that we can log warning in config validations.


web-api/src/bin/web-api.rs
```diff
@@ -13,16 +13,15 @@
...
 
 #[tokio::main]
 #[instrument(err)]
 async fn main() -> Result<(), anyhow::Error> {
-    let config: Config = config::load(application_names!());
-    logging::initialize_global(config.as_ref())?;
-    config.validate()?;
-
+    let config = Config::load(application_names!());
+    logging::initialize_global(config.logging_config())?;
```
rustonaut, line 24: x3ro: We can't call `logging::initialize_global` in the config loading code as it must _not_ be called when we run integration tests as logging is in that case scoped to the test and it already setup before calling the config through other means

web-api/src/config.rs
```diff
@@ -55,50 +55,67 @@ pub struct WebApiConfig {
...
+            config,
+            print_config,
+        } = self;
+        config.ingestion.validate()?;
+        config.personalization.validate()?;
+        config.semantic_search.validate()?;
+
+        if print_config {
+            println!("{}", serde_json::to_string_pretty(&config)?);
+            if exit_on_print {
```
rustonaut, line 107: Being able make tests print the config is nice for debugging, through we don't want to exit in that case.

x3ro (approved):

PR #1121, opened 2023-12-06
ET-5261 make es index name a teant config

rustonaut:
- added functionality to have mutable per-tenant configurations
- made the tenants es index name such a changeable configuration
- added support for changing the es index name to `web-api-db-ctrl` and `/_ops/silo_management`
- tested if it works

**References:**

- task: [ET-5261]
- story: [ET-5260]
- based on #1122

[ET-5261]: https://xainag.atlassian.net/browse/ET-5261?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ET-5260]: https://xainag.atlassian.net/browse/ET-5260?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1123, opened 2023-12-07
ET-5285 handle multiple models in same deployment

rustonaut:
- allow specific multiple models/embedders in the same deployment as a name=>embedder map
- inject the `embedding` config if given into that map as `"default"=>...`
- have a runtime changable per-tenant config in `management.tenant` which specify which model/embedder to use by name, if not given fall back to `"default"`
- update `TenantState(storage)` to `TenantState(storage, embedder)` to make the right embedder available on each request

**References:**

- story: [ET-5260]
- task: [ET-5285]
- based on: #1121, #1122

[ET-5260]: https://xainag.atlassian.net/browse/ET-5260?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ET-5285]: https://xainag.atlassian.net/browse/ET-5285?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rustonaut:
This doesn't contain code for switchign models, because switching models only works if you at the same time also switch the index and computed embeddings in postgres.

x3ro (approved):

PR #1127, opened 2024-01-17
Fix release CI

acrrd:
We have some duplication in the Dockerfile and the release_artifact action, but I don´t think it is worth addressing it now.
It's not ideal because of some duplication, but it's the most straightforward approach.
To have a multi-platform build in docker, we also need to configure it on each computer on which it runs.


acrrd:
https://github.com/xaynetwork/xayn_discovery_engine/actions/runs/7555733583

janpetschexain (approved):

